PR #354, opened 2014-12-18
Overlord2/mabdeljalil Version 1

msabdeljalil:


overlord/features/step_definitions/activate_bomb_steps.rb
```diff
@@ -0,0 +1,62 @@
+# =============== OUTLINE =========================
+When(/^I activate a bomb with (\d+)$/) do |arg1|
```
sshao: the regex on line 7 will catch strings and fixnums (`I activate a bomb with "7890"` and `I activate a bomb with "abcd"` will both match). 

also note that within the block, `arg1` always comes in as a string anyways -- even if you use `\d` to match:

```shell
From: steps.rb @ line 7 :

     6: Then(/^I see the number (\d+)$/) do |arg|
 =>  7:   require "pry"; binding.pry
     8: end

[1] pry(#<Cucumber::Rails::World>)> arg
=> "678"
[2] pry(#<Cucumber::Rails::World>)> arg.class
=> String
```

so you can't really test strings and fixnums in your cukes, anyway. if you really wanted to, you should do that in a bomb model spec, not a cuke.

overlord/lib/models/bomb.rb
```diff
@@ -0,0 +1,68 @@
+class Bomb
+  attr_reader :activation_code, :deactivation_code, :timer
```
sshao: just use the ivars directly (`@activation_code/@deactivation_code`)

overlord/lib/models/bomb.rb
```diff
@@ -0,0 +1,68 @@
...
+  attr_accessor :status, :deactivation_attempts, :messages
+
+  ACTIVE = "ACTIVE"
+  INACTIVE = "Inactive"
```
sshao: nah, don't use constants at all. you'd delete lines 5 and 6 and instead of `@status = INACTIVE`, you'd do `@status = :inactive`. 

strings like "ACTIVE" and "Inactive" are mutable; symbols aren't. a status identifier should be immutable, ie, we should never be able to modify the object representing an "active" state. so symbols are better suited for the purpose of representing bomb state.

overlord/lib/overlord.rb
```diff
@@ -0,0 +1,66 @@
+require 'sinatra/base'
+require "pry"
+require_relative "models/bomb"
+
+class Overlord < Sinatra::Base
+  enable :sessions
+
+  before '/bomb/*' do
+    @bomb = bomb(params[:bomb])
+    redirect "/exploded" if bomb.bomb_exploded?
```
sshao: i don't follow how that'd be triggered by the user -- if a bomb hasn't been created, then the ivar assignment (`@bomb = bomb(params[:bomb])`) that precedes rendering any page other than `'/'` would fail, likely showing up as an exception when rendering the view, right?

overlord/lib/views/_status.erb
```diff
@@ -0,0 +1,3 @@
+<h1 id="bomb_status">Bomb Status: <%= @bomb.status %></h1>
```
sshao: oops, yeah that's right -- my bad :)

overlord/lib/models/bomb.rb
```diff
@@ -0,0 +1,68 @@
...
+  BOMB_TIME = 7
+
+  DEFAULT_ACTIVATE = 1234
+  DEFAULT_DEACTIVATE = 1234
+
+  def initialize(args)
+    @activation_code = activation_check(args[:create_activation_code])
+    @deactivation_code = deactivation_check(args[:create_deactivation_code])
```
sshao: oh no, i meant just changing the symbol name -- so you'd do `activation_check(args[:activation_code])` instead of `activation_check(args[:create_activation_code])`. 

`create_(de)activation_code` is just a bad name for what it represents -- since it's got that `create` verb in there, it sounds like it represents a method instead of an string/number code.

PR #386, opened 2014-12-30
Pmail - Incomplete

djkotowski:
Looking for some feedback with this draft before I advance too much further in the exercise.


djkotowski:
Received hard copy of notes from Dan V

PR #388, opened 2014-12-31
Overlord2/mabdeljalil Version 2

msabdeljalil:


johnmcconnell:
The controller code is really nice. You achieved that by making your model big, which is good! (I'd argue your model could probably being broken up more IMO).

One thing I have more strong feelings about is:
https://github.com/msabdeljalil/level_up_exercises/blob/overlord2/mabdeljalil/overlord/app/models/bomb.rb#L12

```
  def initialize(args)
    @activation_code = activation_check(args[:create_activation_code])
    @deactivation_code = deactivation_check(args[:create_deactivation_code])
```

You can rename `******_check(arg)` to something more intuitive. Maybe `default_*****_code(arg)` I'm not positive.

PR #389, opened 2014-12-31
Shopping cart/mabdeljalil v2

msabdeljalil:


shopping_cart/features/add_item.feature
```diff
@@ -0,0 +1,29 @@
...
+  Background: 
+    Given an empty cart
+
+  @happy
+  Scenario: Adding 1 valid item
+    When I add "item1" for $10.00
+    Then I expect to see "item1" in my cart
+    Then the cart quantity should be 1
+    And have a subtotal of $10.00
+
```
zsyed1: You might want to consider breaking this test into two (maybe three) separate tests.

```Scenario: Adding an item and seeing it in my cart
  When I add "item1" to my cart
  Then I expect to see "item1" in my cart```

```Scenario: Adding an item updates the quantity
  When I add a new item to my cart
  Then the quantity for the item is 1```

```Scenario: Adding an item updates the total
  When I add an item that costs $10
  Then the total  increases by $10```

Comment (zsyed1) [inline shopping_cart/features/add_item.feature]: Sorry, multiline comments don't seem to work for me :disappointed:

File: shopping_cart/features/add_item.feature [hunk 2]
```diff
@@ -0,0 +1,29 @@
...
+  @happy
+  Scenario: Adding multiple valid items
+    When I add 1 "item1" for $10.00
+    And  I add 2 "item2" for $15.00
+    Then I should "item1", "item2", and "item3" in my cart
+    And the cart quantity should be 6
+    And have a subtotal of $40.00
+
```
zsyed1: You shouldn't need to test this. The user will most likely add items one at a time. They would add item1 and then subsequently add item2. You can write a scenario for quantity though. `Scenario: Adding a quantity larger than 1`.

shopping_cart/features/add_item.feature
```diff
@@ -0,0 +1,29 @@
...
+  @happy
+  Scenario: Adding multiple items to a cart with existing items
+    Given a cart with 2 Cars for $10.00 each and 3 Trucks for $15.00 each
+    When I add 1 plane for $5.00 and 1 boat for $1.00
+    And have a subtotal of $71.00
+    And the cart quantity should be 6
```
zsyed1: You can replace the tests concerning quantity with two tests:
Scenario: I increase the quantity of an item
Scenario: I decrease the quantity of an item

Within those two you can check that the total is correct, or you can make two separate tests to check the total.
If those tests pass, then all these scenarios will be covered.

The benefit of writing tests atomically like this, is that you help isolate the feature. If this is unclear please hit me up on slack. I'd be happy to explain.

shopping_cart/features/coupons.feature
```diff
@@ -0,0 +1,65 @@
+Feature: Apply coupon/promotion
+	As a shopper
+	I want to be able to claim coupons and promotions on items I have in my cart
+	So that I can save money on the items I want to purchase
+
+	Scenario: Cannot enter promotion when cart is empty
+		Given a cart with no items
+		When I am on the cart checkout page
+		Then I should not be able to enter a promotion/coupon code
+
```
zsyed1: Good test!

shopping_cart/features/coupons.feature
```diff
@@ -0,0 +1,65 @@
...
+	Scenario: Entering a coupon/promotion code
+		Given a non-empty cart
+		When I am on the cart checkout page
+		Then I should be able to enter a code for a coupon or promotion
+
```
zsyed1: `Then I should be able to enter a promotion/coupon code` to keep the language consistent with the previous scenario.

shopping_cart/features/coupons.feature
```diff
@@ -0,0 +1,65 @@
...
+	Scenario: Applying a valid coupon/promotion code
+		Given a non-empty cart
+		When I am on the cart checkout page
+		And I enter a valid code for a coupon or promotion
+		Then I should see the relevant discout applied to my subtotal
+
```
zsyed1: Spelling: `discount`

shopping_cart/features/coupons.feature
```diff
@@ -0,0 +1,65 @@
...
+Feature: Apply coupon/promotion
+	As a shopper
+	I want to be able to claim coupons and promotions on items I have in my cart
+	So that I can save money on the items I want to purchase
+    
+    Background:
+    	Given I have a cart with "item_1" at priced at "price_1"
+		And I add "item_2" at priced at "price_2"
+		And I have a subotal of "(price_1 + price_2)" # Can u do arithmetic here?
+
```
zsyed1: Extra word `at` in line 44.

zsyed1: You shouldn't need line 46.

shopping_cart/features/coupons.feature
```diff
@@ -0,0 +1,65 @@
...
+    Scenario Outline:
+        When I apply coupon #"<coupon_code>" for "<amount>" off
+        Then I should see the "<message>"
+        And the subtotal should be "<subtotal>"
+
```
zsyed1: This is probably too much detail. You can just do:

When I apply coupon #"<coupon_code>"
Then the subtotal should be "<subtotal>"
And I should see "<message>"

A customer will simply enter a coupon code. The application will determine the discount from that code.

zsyed1: Sorry, the test in the quotes got chopped off.

shopping_cart/features/coupons.feature
```diff
@@ -0,0 +1,65 @@
...
+        | thisiscoupon1   | 5       | (price_1 + price_2) - 5   |         |
+        | thisiscoupon2   | 10      | (price_1 + price_2) - 10  |         |
+        | thisiscoupon3   | 15      | (price_1 + price_2) - 15  |         |
+
+    @bad
+    Examples:
+        | coupon_code   | amount    | subtotal            | message                                                  |
+        | notrealcoupon |           | (price_1 + price_2) | Invalid coupon code |
+        | expiredCoupon |           | (price_1 + price_2) | Expired coupon      |
+        |               |           | (price_1 + price_2) | Invalid coupon code |
```
zsyed1: I would suggest splitting happy and unhappy into different scenarios rather than keeping `message` blank for happy scenarios. You can remove `And I should see the <message>` from the happy cases and in the sad cases put `And I should see the <message>`.

shopping_cart/features/remove_items.feature
```diff
@@ -0,0 +1,27 @@
...
+		Then I should not see "item_1" or "item_2" in the cart
+		And the total quantity should be 0
+		And the subtotal should $0.00
+
+	Scenario: I can remove, and re-add a product
+		When I remove "item_1"
+		And I add "item_1"
+		Then I should see "item_1" in the cart
+		And the total quantity should be 2
+		And the subtotal should ("price_1" + "price_2")
```
zsyed1: I don't think you need to test this with two different items in the cart. Perhaps one test with multiple items would suffice. When all the tests have multiple items, it makes it harder for the reader to understand what each scenario is testing.

shopping_cart/features/shipping_estimates.feature
```diff
@@ -0,0 +1,46 @@
+Feature: Get estimated shipping costs based on address
+    As a shopper
+    I want to receive the estimated cost of shipping for the the items in my cart
+    to an address I enter
+
```
zsyed1: Should have a `So I can...`

shopping_cart/features/shipping_estimates.feature
```diff
@@ -0,0 +1,46 @@
...
+    Scenario Outline: Receive estimate of shipping costs for items
+        Given I have multiple items in my cart
+        And a subtotal of $10
+        And I am on the cart checkout page
+        When I enter "<address>", "<address2>", "<city>", "<state>" and "<zipcode>"
+        And I click "Estimate shipping" button
+        Then I should see "<message>"
+        And the shipping cost has value "<shipping_cost>"
+        And the total cost is "<total_cost>"
+
```
zsyed1: Split this into happy and sad. Remove line 22 from happy.

Also, I would move the testing of `total_cost` into a separate (single scenario). That will help reduce the number of things you are testing in a scenario. Also, line 17 and 18 will no longer be relevant. So your test would reduce to lines 17, 19, 20, 21, 23 for the happy path. For sad, you would have line 22 instead of 23.

shopping_cart/features/view_item_detail.feature
```diff
@@ -0,0 +1,11 @@
...
+	As a shopper
+	I want to be able to view an items product page from the cart
+	So that I can review an items details before carrying out a purchase
+
+
+	Scenario: View product detail page by clicking on item in cart
+		Given I am on the cart index page 
+		And I have a cart with multiple items
+		When I click on an item in the item list
+		Then I should be taken to the item detail page
```
zsyed1: You probably do not need multiple items in the cart to test this.

zsyed1: Your tests are pretty thorough! Some of these comments might be better explained over dialogue than single-line notes. Let me know if you want to discuss. I would recommend splitting the bigger ones up and removing irrelevant elements, then resubmitting for a final round.

shopping_cart/features/add_item.feature
djkotowski: Fixed 'em

djkotowski: I think this is combination works fine as-is.  It's testing each scenario individually, but there are a couple values to check for each.

shopping_cart/features/add_item.feature
```diff
@@ -0,0 +1,29 @@
...
+    When I add "item1" for $10.00
+    Then I expect to see "item1" in my cart
+    Then the cart quantity should be 1
+    And have a subtotal of $10.00
+
+  @happy
+  Scenario: Adding multiple valid items
+    When I add 1 "item1" for $10.00
+    And  I add 2 "item2" for $15.00
+    Then I should "item1", "item2", and "item3" in my cart
```
djkotowski: Quick fix:  "I should have"

Longer fix:  I would split this into three separate tests.  If you test for two or five items, then this step is useless.

shopping_cart/features/add_item.feature
djkotowski: I think it's fine checking multiple valid items.  You'd want to ensure that you have all items in your cart, and it doesn't show just the latest one (I've seen this issue happen in other applications).

shopping_cart/features/coupons.feature
```diff
@@ -0,0 +1,65 @@
...
+
+Feature: Apply coupon/promotion
+	As a shopper
+	I want to be able to claim coupons and promotions on items I have in my cart
+	So that I can save money on the items I want to purchase
+    
+    Background:
+    	Given I have a cart with "item_1" at priced at "price_1"
+		And I add "item_2" at priced at "price_2"
+		And I have a subotal of "(price_1 + price_2)" # Can u do arithmetic here?
```
djkotowski: No arithmetic.  It's just a regex.

shopping_cart/features/coupons.feature
djkotowski: To expand, you should not have tests in your `Given` statements.

shopping_cart/features/coupons.feature
djkotowski: To expand on this, you could have a Given statement in this test that specifies the valid coupons and their amounts.  Then this test would just have the expected results for the given codes.  On a side note, remove the arithmetic and manually compute the expected subtotals.  When you calculate expected values, you've added another point of failure in your testing, as you're relying on the computed values being correct.

shopping_cart/features/remove_items.feature
djkotowski: Remove the calculation.

shopping_cart/features/shipping_estimates.feature
djkotowski: "In order to" seems to be the preferred syntax

shopping_cart/features/coupons.feature
djkotowski: Also, spelling.

djkotowski:
It's looking nice.  Take a look at the comments and refactor a few things.  Remove those calculations.

djkotowski:
Reopened for additional review.

PR #390, opened 2015-01-01
Supportive

tharumax:


supportive/supportive.rb
```diff
@@ -9,96 +9,70 @@ class BlagPost
...
-        DISALLOWED_CATEGORIES.include? category
-      end
-    else
-      @categories = []
-    end
-
-    @comments = args[:comments] || []
-    @body = args[:body].gsub(/\s{2,}|\n/, ' ').gsub(/^\s+/, '')
-    @publish_date = (args[:publish_date] && Date.parse(args[:publish_date])) || Date.today
+    args.to_options!
```
ashtonthomas: I think you are only really using with_indifferent_access here, but `.symbolize_keys!` may be more appropriate

supportive/supportive.rb
```diff
@@ -9,96 +9,70 @@ class BlagPost
...
-    return '' unless comments.length > 0
-
-    ordinal = case comments.length % 10
-      when 1 then "st"
-      when 2 then "nd"
-      when 3 then "rd"
-      else "th"
+  def assign_categories(categories)
+    @categories = Array.wrap(categories).reject do |c|
+      c.in? DISALLOWED_CATEGORIES
```
ashtonthomas: FWIW see: https://github.com/jmmastey/level_up_exercises/pull/383#discussion_r22241746

supportive/supportive.rb
```diff
@@ -9,96 +9,70 @@ class BlagPost
...
     end
-    "You will be the #{comments.length}#{ordinal} commenter"
   end
 
-  def comments_allowed?
-    publish_date + (365 * 3) > Date.today
+  def assign_comments(comments)
+    @comments = Array.wrap(comments)
```
ashtonthomas: also FWIW see: https://github.com/jmmastey/level_up_exercises/pull/282#discussion_r20746015

ashtonthomas:
See comments but this looks really good @tharumax 

:+1:

supportive/supportive.rb
tharumax: According to http://apidock.com/rails/Hash/to_options `#to_options!` is an alias for `#symbolize_keys!`. However I get that `#symbolize_keys!` is more meaningful in this scenario.

ashtonthomas: Sweet! Thanks for the info on this one @tharumax! Yeah, `symbolize_keys!` is very intention revealing/appropriate

PR #391, opened 2015-01-02
CYOA - Iteration 1

paulh16:
I have deployed this app on Heroku at the following URL:

https://artjourney.herokuapp.com

I wanted you to be able to see the actual app online. Since Heroku requires the Gemfile to be in the root directory, I had to rearrange the level_up_exercises directory. For some reason, some of the other exercise files still got pulled into this pull request. I can provide you with a clean directory of my app if that would be easier.

To sign in, I've created an account for you:

Email: joe@enova.com
Password: mastey

The gist of the app is that admin users can go to the Artists#Index page and add an artist from the Artsy API: 

[Artsy AP!](https://developers.artsy.net)

This will search for an artist. For example, try "Edward Hopper". This will add the artist and their artworks if it doesn't already exist. 

All users have a My Profile page, where they can display their favorite artists. The actual favoriting and unfavoriting is handled on the Artist#Show page. This will add the artist to their profile page. In addition, a set of images, one random image from each followed artist, is shown on the User Profile.

My app is fairly conventional. A couple things to note are:

1. I have an `artsy_api_wrapper.rb` file in the lib directory. This file handles the connection and requests to the external API.

2. I have two files in `app/services`, `retrieve_artist.rb` and `retrieve_artworks.rb`, that handle calling the API Wrapper and creating artist and artwork records.

Please let me know if you have any questions. Thank you for reviewing!


djkotowski:
Resubmitted.

PR #393, opened 2015-01-02
Data science

jfinzel:
it "hopes to be super duper code"

### DIFF
--- a/data_science/compare_samples.rb
+++ b/data_science/compare_samples.rb
@@ -1,0 +1,26 @@
+require 'abanalyzer'
+
+class CompareSamples
+  def initialize(sample_a, sample_b)
+    @tester = ABAnalyzer::ABTest.new(samples(sample_a, sample_b))
+  end
+
+  def different?
+    tester.different?
+  end
+
+  def chisquare_p
+    tester.chisquare_p
+  end
+
+  private
+
+  attr_reader :tester
+
+  def samples(sample_a, sample_b)
+    { agroup: { conversions: sample_a.conversions,
+                non_conversions: sample_a.non_conversions },
+      bgroup: { conversions: sample_b.conversions,
+                non_conversions: sample_b.non_conversions } }
+  end
+end
--- a/data_science/load_conversion_data.rb
+++ b/data_science/load_conversion_data.rb
@@ -1,0 +1,19 @@
+require 'json'
+
+class LoadConversionData
+  attr_accessor :data, :sample_key
+  SAMPLE_KEY_DEFAULT = "cohort"
+
+  def initialize(json_file, field_info = {})
+    self.data = JSON.parse(File.read(json_file))
+    self.sample_key = field_info.fetch(:sample_key, SAMPLE_KEY_DEFAULT)
+  end
+
+  def sample_data(sample)
+    data.select { |user| user[sample_key] == sample }
+  end
+
+  def to_s
+    data
+  end
+end
--- a/data_science/sample.rb
+++ b/data_science/sample.rb
@@ -1,0 +1,30 @@
+require 'abanalyzer'
+
+class Sample
+  attr_accessor :sample, :result_key, :valid_result
+  attr_reader :size, :conversions
+  RESULT_KEY_DEFAULT = "result"
+  VALID_RESULT_DEFAULT = 1
+
+  def initialize(results, field_info = {})
+    self.sample = results
+    self.result_key = field_info.fetch(:result_key, RESULT_KEY_DEFAULT)
+    self.valid_result = field_info.fetch(:valid_result, VALID_RESULT_DEFAULT)
+  end
+
+  def size
+    @size = sample.length
+  end
+
+  def conversions
+    @conversions = sample.count { |user| user[result_key] == valid_result }
+  end
+
+  def non_conversions
+    @non_conversions = size - conversions
+  end
+
+  def confidence_interval(confidence)
+    ABAnalyzer.confidence_interval(conversions, size, confidence)
+  end
+end
--- a/data_science/test_samples.rb
+++ b/data_science/test_samples.rb
@@ -1,0 +1,68 @@
+require_relative 'load_conversion_data'
+require_relative 'sample'
+require_relative 'compare_samples'
+
+def conversion_data
+  LoadConversionData.new("source_data.json")
+end
+
+def conversion_data_a
+  conversion_data.sample_data("A")
+end
+
+def conversion_data_b
+  conversion_data.sample_data("B")
+end
+
+describe LoadConversionData do
+  it "should parse a JSON file" do
+    expect(conversion_data.data).to be_an(Array)
+  end
+
+  it "should understand how to filter by sample" do
+    expect(conversion_data_a).to be_an(Array)
+  end
+end
+
+describe Sample do
+  it "should have a size equal to the array length" do
+    sample = Sample.new(conversion_data.data)
+    expect(sample.size).to eq(conversion_data.data.length)
+  end
+
+  it "should show conversions" do
+    sample = Sample.new(conversion_data_a)
+    expect(sample.conversions).to be_an(Integer)
+    puts "Conversions: #{sample.conversions}"
+  end
+
+  it "should have a size equal to conversions plus non-conversions" do
+    sample = Sample.new(conversion_data_a)
+    expect(sample.size).to eq(sample.conversions + sample.non_conversions)
+  end
+
+  it "can give a confidence interval" do
+    sample = Sample.new(conversion_data_a)
+    confidence = sample.confidence_interval(0.95)
+    expect(confidence).to be_an(Array)
+  end
+end
+
+describe CompareSamples do
+  it "says an identical data set is not different" do
+    sample = Sample.new(conversion_data_a)
+    different = CompareSamples.new(sample, sample).different?
+    expect(different).to be(false)
+  end
+
+  it "says a different data set is different" do
+    sample_a = Sample.new(conversion_data_a)
+    sample_b = Sample.new(conversion_data_b)
+    compare = CompareSamples.new(sample_a, sample_b)
+    if compare.chisquare_p < 0.95
+      expect(compare.different?).to be(true)
+    else
+      raise "Test data sets are not different"
+    end
+  end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+*.iml


data_science/load_conversion_data.rb
```diff
@@ -0,0 +1,19 @@
+require 'json'
+
+class LoadConversionData
```
sshao: `LoadConversionData` sounds more like a method name (verb-y) than a class name (noun). `ConversionDataLoader` instead?

data_science/test_samples.rb
```diff
@@ -0,0 +1,68 @@
...
+  conversion_data.sample_data("A")
+end
+
+def conversion_data_b
+  conversion_data.sample_data("B")
+end
+
+describe LoadConversionData do
+  it "should parse a JSON file" do
+    expect(conversion_data.data).to be_an(Array)
```
sshao: might want to check the contents of `conversion_data.data` here, too. what if the format of the array is incorrect? what if it corrupts the data during the parsing? what if the array is somehow empty?

data_science/test_samples.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  it "should understand how to filter by sample" do
+    expect(conversion_data_a).to be_an(Array)
```
sshao: make some assertions on `conversion_data_a`'s contents, same as above

data_science/load_conversion_data.rb
```diff
@@ -0,0 +1,19 @@
...
+  attr_accessor :data, :sample_key
+  SAMPLE_KEY_DEFAULT = "cohort"
+
+  def initialize(json_file, field_info = {})
+    self.data = JSON.parse(File.read(json_file))
+    self.sample_key = field_info.fetch(:sample_key, SAMPLE_KEY_DEFAULT)
```
sshao: `sample_key` is a confusing name here. it suggests that it maps to a single sample, when it actually maps to a *group* of samples. try `cohort_key` instead

data_science/load_conversion_data.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def sample_data(sample)
+    data.select { |user| user[sample_key] == sample }
```
sshao: instead of having to call `select` every time you want to grab one sample group, you can use something like `data.group_by { |d| d[sample_key] }` in `initialize`, and save the result of that to an instance variable (say `sample_data`). 

then you can get a sample group by calling `sample_data["A"]`, and you can delete this method entirely.

sshao: to take the above a little further -- instead of saving it as a separate variable `sample_data`, can you just save it as `data` instead? you're going to have to split `data` into cohort groups later on down the line anyway, so it doesn't seem like you'll need the raw `data` result for anything else.

data_science/test_samples.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+end
+
+describe Sample do
+  it "should have a size equal to the array length" do
```
sshao: "array" means nothing here -- what array? where did it come from? what does it contain? rewording to "should have a size equal to the dataset size" will clear up the purpose of this test.

data_science/test_samples.rb
```diff
@@ -0,0 +1,68 @@
...
+    sample = Sample.new(conversion_data.data)
```
sshao: curious as to why you used the full `conversion_data.data` set here, as it seems that in practice, the Sample class should only used on separated data groups (ie `Sample.new(conversion_data_a)`, as on lines 59-60 below). it's not necessarily wrong and the test does fulfill its purpose -- it just seems weird because it doesn't reflect how `Sample.new` would be used in a 'real-world' application.

data_science/test_samples.rb
```diff
@@ -0,0 +1,68 @@
...
+    expect(sample.size).to eq(conversion_data.data.length)
+  end
+
+  it "should show conversions" do
+    sample = Sample.new(conversion_data_a)
+    expect(sample.conversions).to be_an(Integer)
+    puts "Conversions: #{sample.conversions}"
```
sshao: whoa, why is there a `puts` in your test? this should be removed, as it doesn't have a purpose. the assertion on line 35 should be the only assertion you need to assure that `Sample#conversions` works as expected.

data_science/test_samples.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+end
+
+describe Sample do
+  it "should have a size equal to the array length" do
+    sample = Sample.new(conversion_data.data)
+    expect(sample.size).to eq(conversion_data.data.length)
+  end
+
+  it "should show conversions" do
```
sshao: this description suggests that you want to test that *some* unspecified `Sample` method *outputs* correctly (as implied by the verb "show"), but you're making assertions on the `Sample#conversions` method instead... so now I'm actually unclear on what exactly you're trying to test here. 

if you actually want to test that it "shows [conversions/number of conversions]", what you really want is something like `expect(sample.to_s).to be expected_number_of_conversions1.

if you want to test that the `Sample#conversion` method is correct, then you want `it "returns the correct number of conversions"` what you have + the suggestions i made in the comment below.

data_science/load_conversion_data.rb
jfinzel: I think it works with the usage sample_key just indicates the name of the field in the file, which in this case is cohort.  But what cohort represents is the same thing as "sample", which I think is a more general word for what this represents.  Yes, there are several samples within a file but this is just the field name.  Thought?

data_science/test_samples.rb
```diff
@@ -0,0 +1,68 @@
...
+    expect(sample.conversions).to be_an(Integer)
+    puts "Conversions: #{sample.conversions}"
+  end
+
+  it "should have a size equal to conversions plus non-conversions" do
+    sample = Sample.new(conversion_data_a)
+    expect(sample.size).to eq(sample.conversions + sample.non_conversions)
+  end
+
+  it "can give a confidence interval" do
```
sshao: you want stronger wording and stronger assertions here: `it "gives the correct confidence interval"`, and tests to check that. again, giving *a* confidence interval means nothing if it's not the *correct* confidence interval :)

data_science/test_samples.rb
```diff
@@ -0,0 +1,68 @@
...
+    sample = Sample.new(conversion_data_a)
+    confidence = sample.confidence_interval(0.95)
+    expect(confidence).to be_an(Array)
+  end
+end
+
+describe CompareSamples do
+  it "says an identical data set is not different" do
```
sshao: why do you need the `different?` method? it doesn't seem like you use it anywhere except in these tests.

data_science/test_samples.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  it "says a different data set is different" do
+    sample_a = Sample.new(conversion_data_a)
+    sample_b = Sample.new(conversion_data_b)
+    compare = CompareSamples.new(sample_a, sample_b)
+    if compare.chisquare_p < 0.95
+      expect(compare.different?).to be(true)
+    else
+      raise "Test data sets are not different"
+    end
```
sshao: this is a weird control flow for a test. 

if a test fails, you want it to fail in `expect` -- which will then raise the failure for you. you shouldn't have to explicitly `raise` yourself. i see that the reason you do this is because you're not sure that the expectation on line 63 will be reached but you should structure your tests such that you *know* it will -- otherwise it's not a very good test. so in this case, you need to tailor your test data such that you *know* `compare.chisquare_p` will be less than 0.95.

sshao: also, it doesn't look like you have any explicit tests on the `chisquare_p` method. given that it's a public method, you should :)

data_science/load_conversion_data.rb
jfinzel: If I were to do the group by solution, that would make sense to eliminate the data variable as the whole data set, but I need it if I am filtering instead of grouping by.  I guess the question is: is it much better to do the group by vs. a filtering like I am doing?  I think I do like it better and am glad to know about that option!

data_science/test_samples.rb
jfinzel: It does mean something - it means that it was able to parse the data into an Array, which is the result of the JSON.parse method.  But what you are getting at I think is more detail about what kind of information this contains, as you said above?

data_science/compare_samples.rb
```diff
@@ -0,0 +1,26 @@
+require 'abanalyzer'
+
+class CompareSamples
```
sshao: again, noun classes instead of verb classes. `SampleComparer` -- or in this case, i'd rename the entire thing to `Tester`, perhaps?

data_science/test_samples.rb
jfinzel: True, but here I am testing only this class, and that it is able to simply import the file initially before grouping or filtering.  Isn't that worth testing separately?

data_science/compare_samples.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def chisquare_p
+    tester.chisquare_p
+  end
+
+  private
+
+  attr_reader :tester
+
+  def samples(sample_a, sample_b)
```
sshao: using `samples` as a method name is getting confusing here, given that you've got the `Sample` class, and then the `sample_a/sample_b` variables. i'd rename it to `format` instead.

data_science/test_samples.rb
jfinzel: It's because I have never written a test before in my life :).  I am doing this to learn how to test.  But more specifically, I was just wanting to see the numbers that were coming out of the file to feed out to my command prompt.  I actually think I intended to delete this line.

Again, I disagree that it means nothing - it does mean something that it's an integer, not a string or nil.  Yes, I need it to be the right value, but how do I test that it's the right integer value without using the very same code to count the number of conversions?  I believe there's an answer to the latter, but I didn't do that because I don't really know how to do that, other than hardcoding "47" which is not helpful.

data_science/sample.rb
```diff
@@ -0,0 +1,30 @@
+require 'abanalyzer'
+
+class Sample
```
sshao: again, this class name suggests that it maps to a single sample rather than a group of samples. `SampleGroup`?

data_science/sample.rb
```diff
@@ -0,0 +1,30 @@
...
+  RESULT_KEY_DEFAULT = "result"
+  VALID_RESULT_DEFAULT = 1
+
+  def initialize(results, field_info = {})
+    self.sample = results
+    self.result_key = field_info.fetch(:result_key, RESULT_KEY_DEFAULT)
+    self.valid_result = field_info.fetch(:valid_result, VALID_RESULT_DEFAULT)
+  end
+
+  def size
```
sshao: you have both a method `size` and an attr_reader `size`. given that `size` should be constant, i'd delete the method, keep the attr_reader, and assign `@size = sample.length` in `initialize`.

data_science/sample.rb
```diff
@@ -0,0 +1,30 @@
...
+    @size = sample.length
+  end
+
+  def conversions
+    @conversions = sample.count { |user| user[result_key] == valid_result }
+  end
```
sshao: same as above. another thing you could do here is `@conversions ||= sample.count { ... }`, so it'd only ever make the `#count` call once.

data_science/sample.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def non_conversions
+    @non_conversions = size - conversions
```
sshao: `@non_conversions ||= size - conversions` or just assign it in `initialize`, same as above.

data_science/test_samples.rb
jfinzel: I think this is related to the same thing above.  I literally wanted to test that Sample shows conversions as an attribute, but maybe that's too simplistic to test.  But I don't know how to test the number of conversions without using the same code I am already using the count the number of conversions.  Am I supposed to use a different code snippet inside my test?  And if so, don't I have to then test my code snippet inside my test also?  I'm just not sure how this kind of particular test should be conducted: I am writing a piece of code to make a calculation - I want to test that it's correct.  How do I test that?  Do I try to write the calculation another way and then make it that way?  But then it seems like I need to test that calculation as well!  I would appreciate understanding this in particular.

data_science/load_conversion_data.rb
sshao: ah, i think you're right -- it looks like the word "sample" does refer to a set of data, and not just a singular piece. i'm not very good at statistics....

data_science/test_samples.rb
jfinzel: I have the same question about this as above on how do I independently verify a calculation?

data_science/test_samples.rb
jfinzel: That was the purpose of the exercise - to tell whether or not the samples are different with significance.

data_science/load_conversion_data.rb
sshao: i'm unclear as to what you mean by "filtering" -- do you have an example of what that'd look like in code?

data_science/test_samples.rb
jfinzel: I simply have no idea how to do this.  Again, I have the same issue that keeps repeating itself.  I don't know how I would independently test that the samples are different - that's the very purpose of my class.  The independent test is the chisquare_p.  But then I have no idea how chisquare_p works - how would I test this?

data_science/compare_samples.rb
jfinzel: Yes I see.

data_science/sample.rb
jfinzel: I think this works.  It is only a single sample in the way I've written my project.

data_science/sample.rb
jfinzel: I don't see how that's really different?  It's not a method, it's an override of the getter for attr_reader.  I have been told by several that even if you are overriding either getter or setter logic, it is still good to use attr_reader/accessor, and then override the logic in an explicitly defined getter or setter.

data_science/sample.rb
jfinzel: I do see your point here.  I think of the logic of a field as its definition relative to the class which I think is why I often do it this way.  I like the idea of doing ||= on it.

data_science/load_conversion_data.rb
jfinzel: I mean that what I'm doing is "filtering" the data down to the "A" group by doing a select, and likewise with the "B" group, so I have an array of only the "A" group, or the "B" group, respectively.

data_science/compare_samples.rb
jfinzel: Yes thanks for helping me think through these class names.

data_science/test_samples.rb
sshao: imagine that you're some other user who hasn't read your code. when you run this test, you'll see its description as: `"Sample should have a size equal to the array length."` you don't know what "the array" is. you don't know that `Sample` takes an array in its `initialize` method. you literally don't know what "the array" is referring to -- that's what i meant by "it means nothing". a complete stranger looking at this test has no context for where "the array" comes from.

actually, to make the intentions of this test even clearer, i'd restructure it is:

```ruby
describe Sample do
  describe "#new" do
    it "creates a Sample with all the data passed to it" do
      sample = Sample.new(conversion_data.data)
      expect(sample.size) to eq(conversion_data.data.size)
    end
  end
end
```

now the description of this test is "Sample#new creates a Sample with all the data passed to it", and there's no ambiguities about what it's trying to assert.

data_science/test_samples.rb
sshao: coming back to this after your comments below -- ignore suggestions 2 and 3. crafting a smaller test dataset is really the only solution here.

sshao:
take another shot at this and resubmit for another reviewer to look at afterwards :+1:

data_science/test_samples.rb
jfinzel: Thanks that explanation makes a lot of sense.  I'm trying to get a big
picture on what kind of tests I need to write, the depth and the extent of
the tests I need to write.

data_science/test_samples.rb
jfinzel: Sorry, I misunderstood what you were commenting on above as referring to the JSON parse part!  Reading your initial response you are exactly right - I should be using one or the other sample, not the whole data set.

data_science/test_samples.rb
jfinzel: This is very helpful!  I have always had the mindset of not hardcoding values in programming, but I see that you have to think about testing differently.

data_science/test_samples.rb
jfinzel: Yes but having chisquare_p public is how I test that my own code is actually how I verify that's it's different in my tests - it just runs the method from the gem on my data.  Isn't testing chisquare_p outside the scope of this exercise?  I am honestly trying to do as much as I can to learn about testing here without being nitpicky but I feel like testing chisquare_p is nitpicky :)

sshao: oh yeah, when you use a third-party library you're usually supposed to assume that *they've* done thorough testing. i think i commented on your tests before i went up and commented on your source code, and missed that it's a call to the abanalyzer gem. so you're right that it's not in scope to test chisquare_p :)

but if i understand you correctly, and you're using `ComareSamples#chisquare_p` to *verify* that your code is working correctly: then it sounds like you've implemented the method solely for the purposes of testing, which is frowned upon. unless you expect a user to actually use that method, you shouldn't implement it.

jfinzel: Well it does make sense to run chisquare_p in this context, even though I
am really only including it for testing.  It does make sense to just run
chisquare_p on a set of samples like I'm doing.

data_science/load_conversion_data.rb
jfinzel: Thanks for the explanation here.  I haven't used group by and it looks really useful.  In this case, I was trying to just say that my code still accomplishes the same purpose a different way.

PR #395, opened 2015-01-05
refactred the code for robot_name

sriking1783:
Please look at the refactored code for robot name

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -3,23 +3,37 @@ class NameCollisionError < RuntimeError; end
 class Robot
   attr_accessor :name
 
-  @@registry
-
   def initialize(args = {})
     @@registry ||= []
     @name_generator = args[:name_generator]
+  end
 
+  def name
     if @name_generator
-      @name = @name_generator.call
+      robot_name = @name_generator.call
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+      robot_name = generate_characters + generate_numbers
     end
+    raise NameCollisionError, "There was a problem generating the robot name!" unless name_valid?(robot_name)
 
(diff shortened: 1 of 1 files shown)


robot_name/robot_name.rb
```diff
@@ -3,23 +3,37 @@ class NameCollisionError < RuntimeError; end
...
+
+  def generate_characters
+    characters = ""
+    2.times { characters << ('A'..'Z').to_a.sample }
+    characters
+  end
+
+  def generate_numbers
+    numbers = ""
+    3.times { numbers << rand(10).to_s }
```
sriking1783: I guess its personal choice, but I personally I feel, why are we multiplying a string? what do you think?

sriking1783:
There would be name collisions and it would raise exception
@dulyanov I added this
```1000.times.each do |n|
  puts " My pet robot's name is #{robot.name}, but we usually call him sparky."
end```

whenever I got exception, it raises exception and exits

**Comment by dulyanov** [comment]
What I meant is that your instance of a robot should get a name assigned once, it shouldn't get a new name each time you ask what it's name is.

File: robot_name/robot_name.rb (hunk 1 above)
Comment (dulyanov) [inline robot_name/robot_name.rb]: I didn't think about the fact you're generating 3 unique numbers, never mind. But you could simplify it using inject.

File: robot_name/robot_name.rb [hunk 2]
```diff
@@ -3,23 +3,37 @@ class NameCollisionError < RuntimeError; end
...
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+      robot_name = generate_characters + generate_numbers
     end
+    raise NameCollisionError, "There was a problem generating the robot name!" unless name_valid?(robot_name)
+
+    @@registry << robot_name
+    robot_name
```
djkotowski: Each time `#name` is called, a new Robot name is generated and the registry grows.

djkotowski:
This is looking good.  Take a look at the comments and refactor, especially regarding `#name` generating new names each time.

PR #396, opened 2015-01-05
Supportive

paulh16:
Thank you for reviewing this exercise. You can see the ActiveSupport methods I used in my commit messages. I also did a general refactoring of the class.

Thanks!

### DIFF
--- a/supportive/supportive.rb
+++ b/supportive/supportive.rb
@@ -28,7 +28,7 @@ class BlagPost
 
     @comments = args[:comments] || []
     @body = args[:body].squish
-    @publish_date = args[:publish_date].try { |date| Date.parse(date) } || Date.today
+    @publish_date = args[:publish_date].try { |publish_date| publish_date.to_date } || Date.today
   end
 
   def to_s
--- a/supportive/supportive.rb
+++ b/supportive/supportive.rb
@@ -28,7 +28,7 @@ class BlagPost
 
     @comments = args[:comments] || []
     @body = args[:body].gsub(/\s{2,}|\n/, ' ').gsub(/^\s+/, '')
-    @publish_date = (args[:publish_date] && Date.parse(args[:publish_date])) || Date.today
+    @publish_date = args[:publish_date].try { |date| Date.parse(date) } || Date.today
   end
 
   def to_s
@@ -119,4 +119,4 @@ blag = BlagPost.new("author"        => "Foo Bar",
                         ARTICLE
                    )
 puts blag.to_s
-#puts blag.inspect
+puts blag.inspect
--- a/supportive/supportive.rb
+++ b/supportive/supportive.rb
@@ -14,28 +14,28 @@ class BlagPost
       hash
     end
 
-    if args[:author].present? && args[:author_url].present?
-      @author = Author.new(args[:author], args[:author_url])
-    end
-
-    if args[:categories]
-      @categories = args[:categories].reject do |category|
-        category.in? DISALLOWED_CATEGORIES
-      end
-    else
-      @categories = []
-    end
-
+    @author = create_author(args)
+    @categories = validate_category(args[:categories]) || []
     @comments = Array.wrap(args[:comments])
     @body = args[:body].squish
-    @publish_date = args[:publish_date].try { |publish_date| publish_date.to_date } || Date.today
+    @publish_date = create_date(args) || Date.today
   end
 
   def to_s
-    [ category_list, byline, abstract, commenters ].join("\n")
+    [category_list, byline, abstract, commenters].split(/\n/)
   end
 
   private
+
+  def create_author(args)
+    if args[:author] && args[:author_url]
+      @author = Author.new(args[:author], args[:author_url])
+    end
+  end
+
+  def create_date(args)
+    args[:publish_date].try(&:to_date)
+  end
 
   def byline
     if author.blank?
@@ -45,20 +45,26 @@ class BlagPost
     end
   end
 
+  def validate_category(categories)
+    categories.reject do |category|
+      category.in? DISALLOWED_CATEGORIES
+    end
+  end
+
   def category_list
     if categories.empty?
       ""
     else
       category_titles = categories.map { |category| as_title(category) }
-      "#{pluralizer(categories.length, 'Category')}: #{category_titles.to_sentence}"
+      "#{pluralizer(categories, 'Category')}: #{category_titles.to_sentence}"
     end
   end
 
-  def pluralizer(size, label)
-    if size == 1
-      label.singularize
-    else
+  def pluralizer(object, label)
+    if object.many?
       label.pluralize
+    else
+      label.singularize
     end
   end
 
@@ -67,10 +73,11 @@ class BlagPost
   end
 
   def commenters
-    return '' unless comments_allowed?
-    return '' unless comments.length > 0
-
-    "You will be the #{comments.length.ordinalize} commenter"
+    if comments_allowed? && comments.length > 0
+      "You will be the #{comments.length.ordinalize} commenter"
+    else
+      ''
+    end
   end
 
   def comments_allowed?
@@ -80,13 +87,12 @@ class BlagPost
   def abstract
     body.truncate(204)
   end
-
 end
 
 blag = BlagPost.new("author"        => "Foo Bar",
                     "author_url"    => "http://www.google.com",
                     "categories"    => [:theory_of_computation, :languages, :gossip],
-                    "comments"      => [ ["hello"], ["hello"], ["hello"] ], # because comments are meaningless, get it?
+                    "comments"      => [["hello"], ["hello"], ["hello"]],
                     "publish_date"  => "2013-02-10",
                     "body"          => <<-ARTICLE
                         Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec a diam lectus.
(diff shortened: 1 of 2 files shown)


supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
...
-
-    @comments = args[:comments] || []
-    @body = args[:body].gsub(/\s{2,}|\n/, ' ').gsub(/^\s+/, '')
-    @publish_date = (args[:publish_date] && Date.parse(args[:publish_date])) || Date.today
+    @author = create_author(args)
+    @categories = validate_category(args[:categories]) || []
+    @comments = Array.wrap(args[:comments])
+    @body = args[:body].squish
+    @publish_date = create_date(args) || Date.today
   end
```
zsyed1: In line 21, move the `|| Date.today` into the method `create_date`

supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
...
 
   def to_s
-    [ category_list, byline, abstract, commenters ].join("\n")
+    [category_list, byline, abstract, commenters].split(/\n/)
   end
```
zsyed1: Why are you splitting instead of joining?

supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
...
-      last_category = categories.pop
-      suffix = " and #{as_title(last_category)}"
+  def pluralizer(object, label)
+    if object.many?
+      label.pluralize
     else
-      suffix = ""
+      label.singularize
     end
-
```
zsyed1: A shorter way to do this is `label.pluralize(object.size)`. After all, you are passing the singular version `Category` into this method.
Also, `object` is the most generic name for something in ruby. Better to use something more descriptive like `collection`.

supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
...
 
-    if categories.length == 1
-      label = "Category"
+  def category_list
+    if categories.empty?
+      ""
     else
-      label = "Categories"
+      category_titles = categories.map { |category| as_title(category) }
+      "#{pluralizer(categories, 'Category')}: #{category_titles.to_sentence}"
```
zsyed1: Another way to do line 58 is `categories.map(&:to_s).map(&:humanize).map(&:titleize)`. You could then move this into a method called `def category_titles` and line 59 would stay the same.

supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
...
   end
 
   def as_title(string)
-    string = String(string)
-    words = string.gsub('_', ' ').split(' ')
-
-    words.map!(&:capitalize)
-    words.join(' ')
+    String(string).humanize.titleize
   end
```
zsyed1: I prefer not to name variables by their type but rather by description. You could make a case that this method was generically intended for any string, but still you are only passing `category`. Something like `def as_title(category)` would be better.

zsyed1: On line 72 I think `string.to_s` is better than `String(string)`

supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
...
-      when 1 then "st"
-      when 2 then "nd"
-      when 3 then "rd"
-      else "th"
+    if comments_allowed? && comments.length > 0
+      "You will be the #{comments.length.ordinalize} commenter"
+    else
+      ''
     end
-    "You will be the #{comments.length}#{ordinal} commenter"
```
zsyed1: You don't really need the `else` block. If you remove it, the method will return `nil` and then up above in line 25 in the method `to_s` just add `.reject(&:blank?)` to the end of your array. If this is unclear, please hit me up on slack.

supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
...
 end
 
 blag = BlagPost.new("author"        => "Foo Bar",
                     "author_url"    => "http://www.google.com",
                     "categories"    => [:theory_of_computation, :languages, :gossip],
-                    "comments"      => [ [], [], [] ], # because comments are meaningless, get it?
+                    "comments"      => [["hello"], ["hello"], ["hello"]],
                     "publish_date"  => "2013-02-10",
                     "body"          => <<-ARTICLE
                         Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec a diam lectus.
```
zsyed1: Ask Dan Kotowski what line 98 means. He's a Latin expert.

supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
...
+  end
+
   def byline
-    if author.nil?
+    if author.blank?
       ""
     else
       "By #{author.name}, at #{author.url}"
     end
   end
```
zsyed1: For `if/else` I prefer the two-line version instead of the five-line version:
`return '' unless author.present?`
`"By #{author.name}, at #{author.url}"`

supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
       hash
     end
```
zsyed1: The variable name `attributes` is better than `args`. You don't have to change it though - just an fyi.

supportive/supportive.rb
zsyed1: @djkotowski I think this can be closed.

supportive/supportive.rb
djkotowski: Don't reference variable types in their names.  There's also an ActiveSupport extension that will simplify lines 12-15.

supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
...
-      end
-    else
-      @categories = []
-    end
-
-    @comments = args[:comments] || []
-    @body = args[:body].gsub(/\s{2,}|\n/, ' ').gsub(/^\s+/, '')
-    @publish_date = (args[:publish_date] && Date.parse(args[:publish_date])) || Date.today
+    @author = create_author(args)
+    @categories = validate_category(args[:categories]) || []
```
djkotowski: Why `|| []`?   When would `validate_category` return `nil`?

supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
...
   private
 
+  def create_author(args)
+    if args[:author] && args[:author_url]
+      @author = Author.new(args[:author], args[:author_url])
+    end
+  end
+
+  def create_date(args)
+    args[:publish_date].try(&:to_date)
```
djkotowski: No need to use a block here.  Just pass `try` the symbolized method name.

supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
...
     else
       "By #{author.name}, at #{author.url}"
     end
   end
 
-  def category_list
-    return "" if categories.empty?
+  def validate_category(categories)
+    categories.reject do |category|
+      category.in? DISALLOWED_CATEGORIES
```
djkotowski: Use parens for method arguments.

supportive/supportive.rb
djkotowski: I like sticking with the original structure, to some extent.  Using `as_title` is fine.  That said, I think you're doing a bit much in the interpolated string by calling `pluralizer`.  Take a look at @zsyed1's comments about `pluralizer` below.  By passing the size into pluralize, you'll shorten that method enough where you can eliminate `pluralizer` entirely.

supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
...
-      label = "Categories"
+      category_titles = categories.map { |category| as_title(category) }
+      "#{pluralizer(categories, 'Category')}: #{category_titles.to_sentence}"
     end
+  end
 
-    if categories.length > 1
-      last_category = categories.pop
-      suffix = " and #{as_title(last_category)}"
+  def pluralizer(object, label)
```
djkotowski: This method name implies that it returns a pluralizer, not that it pluralizes.

supportive/supportive.rb
djkotowski: Normally, I would agree with choosing a more descriptive name.  However, while this method is used to "titleize" categories, it is generic enough to be used for anything.  In this case, I'm not opposed to using `string` or `text` as the parameter name.

supportive/supportive.rb
djkotowski: Alternative would be
```
return '' unless comments.allowed? && comments.length > 0
"You will be the #{comments.length.ordinalize} commenter"
```

supportive/supportive.rb
```diff
@@ -14,95 +14,85 @@ def initialize(args)
...
+      "You will be the #{comments.length.ordinalize} commenter"
+    else
+      ''
     end
-    "You will be the #{comments.length}#{ordinal} commenter"
   end
 
   def comments_allowed?
-    publish_date + (365 * 3) > Date.today
+    publish_date > Date.today.years_ago(3)
```
djkotowski: `Date.today.years_ago(3)` could be substituted with `3.years.ago`.

supportive/supportive.rb
djkotowski: I memorized the first 11 lines of the Aeneid for class and had to know roughly what it meant.  I now remember the first 4 lines, knowing very little of the English translation.  Not a Latin expert anymore.

djkotowski:
Looking good.  No need for a resubmission, but take a look at the comments anyway.

paulh16:
@djkotowski @zsyed1 Thanks for the review. I have updated my solution in light of these comments, but won't submit a pull request. Thanks!

PR #398, opened 2015-01-06
Data science refactor 1

jfinzel:
Thanks!


data_science/data_conversion_loader.rb
```diff
@@ -0,0 +1,19 @@
+require 'json'
+
+class DataConversionLoader
+  attr_accessor :data, :sample_key
```
jmmastey: Are these accessors necessary? I can't find any place where you use the sample_key outside the class.

data_science/data_conversion_loader.rb
```diff
@@ -0,0 +1,19 @@
...
+  SAMPLE_KEY_DEFAULT = "cohort"
```
jmmastey: Probably `DEFAULT_SAMPLE_KEY`. Meh.

data_science/data_conversion_loader.rb
```diff
@@ -0,0 +1,19 @@
...
+    self.data = JSON.parse(File.read(json_file))
+    self.sample_key = field_info.fetch(:sample_key, SAMPLE_KEY_DEFAULT)
+  end
+
+  def sample_data(sample)
+    data.select { |user| user[sample_key] == sample }
+  end
+
+  def to_s
+    data
```
jmmastey: Isn't data a parsed JSON array of hashes? `to_s` should always return a string.

data_science/data_conversion_loader.rb
jfinzel: My thought with sample key is that if you had a json set of data with a different field name for "cohort" you could pass that into the field_info hash.  But actually there's not any reason I need to know about it outside the class if that were passed in.  I do need data - that is the full result of the parse, which I then work on using sample_data to get a specific cohort within the data.

data_science/data_conversion_loader.rb
jfinzel: No I understand and sympathize with your naming convention there which reads like plain english, rather than like a suffix, as I had it originally.

data_science/data_conversion_loader.rb
jfinzel: I don't recall why I left that in there but I would just remove it.  I am not using it.  Point well taken.

data_science/sample.rb
```diff
@@ -0,0 +1,30 @@
...
+  VALID_RESULT_DEFAULT = 1
+
+  def initialize(results, field_info = {})
+    self.sample = results
+    self.result_key = field_info.fetch(:result_key, RESULT_KEY_DEFAULT)
+    self.valid_result = field_info.fetch(:valid_result, VALID_RESULT_DEFAULT)
+  end
+
+  def size
+    @size ||= sample.length
```
jmmastey: Making `sample` an accessor means you'll have staleness problems with this. If you move them to attr_reader, that problem goes away.

data_science/spec/data_conversion_loader_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+describe DataConversionLoader do
+  let(:data) { conversion_data }
+  let(:data_a) { conversion_data_a }
+
+  it "should parse a JSON file" do
+    expect(data).to be_an(Array)
+  end
+
+  it "should be able to parse the correct fields" do
+    DATA_FIELDS.each { |field| expect(data[0]).to have_key(field) }
```
jmmastey: Maybe `expect(data[0].keys.sort).to eq(DATA_FIELDS)`

data_science/spec/data_conversion_loader_spec.rb
```diff
@@ -0,0 +1,27 @@
+require_relative 'test_helper'
+
+describe DataConversionLoader do
+  let(:data) { conversion_data }
+  let(:data_a) { conversion_data_a }
```
jmmastey: This is a little hard to understand, with conversion data coming from some mysterious source.

data_science/spec/data_conversion_loader_spec.rb
jmmastey: Your form might have better error messages, though. Not convinced.

data_science/spec/sample_comparer_spec.rb
```diff
@@ -0,0 +1,20 @@
+require_relative 'test_helper'
+
+describe SampleComparer do
+  let(:sample_a) { Sample.new(conversion_data_a) }
+  let(:sample_b) { Sample.new(conversion_data_b) }
+  let(:comparer_same) { SampleComparer.new(sample_a, sample_a) }
+  let(:comparer_different) { SampleComparer.new(sample_a, sample_b) }
+
+  it "says an identical data set is not different" do
+    expect(comparer_same.different?).to be(false)
```
jmmastey: `expect(comparer_same).not_to be_different`

data_science/spec/sample_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    expect(sample.non_conversions).to eq(DATA_NON_CONVERSIONS_A)
+  end
+
+  it "should have a size equal to conversions plus non-conversions" do
+    expect(sample.size).to eq(sample.conversions + sample.non_conversions)
+  end
+
+  it "gives the correct confidence interval" do
+    class_confidence = sample.confidence_interval(CONFIDENCE)
+    check_confidence = ABAnalyzer.confidence_interval(DATA_CONVERSIONS_A, DATA_SIZE, CONFIDENCE)
```
jmmastey: Ehhh, this is pretty tautological. I'm not sure this is a high quality test.

data_science/spec/test_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+  DataConversionLoader.new("source_data.json").data
+end
+
+def conversion_data_a
+  DataConversionLoader.new("source_data.json").sample_data("A")
+end
+
+def conversion_data_b
+  DataConversionLoader.new("source_data.json").sample_data("B")
+end
```
jmmastey: No reason not to keep this stuff in the test file itself. I will say that you've now pegged your tests to the production data, which is an anti-pattern.

jmmastey:
Solid work. Left a few notes. Make some corrections and move on, no resubmission needed.

data_science/spec/test_helper.rb
jfinzel: Joe, could you explain this one to me?  One thing I don't understand is how
you are supposed to independently verify that the data is correct without
explicitly hardcoding the values, or just picking some other way to check
it which itself should have to be checked?  Does my question make sense?

*Jeremy Finzel*

*Senior Database Developer*
*M* 630-777-4520

PR #402, opened 2015-01-07
Data science initial implementation

arrubin:


data_science/lib/abtest/analyzer.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    def confidence_interval_by_cohort
+      cohorts.each_with_object({}) do |cohort, confidence_intervals|
+        confidence_intervals[cohort.name] = cohort.confidence_interval
+      end
+    end
+
+    def chi_squared
+      cohorts.inject(0) do |chi_squared, cohort|
```
sshao: just `cohorts.inject do |...|` should work here

data_science/spec/abtest/analyzer_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+        analyzer = Analyzer.new(test_data)
+        expect(analyzer.chi_squared).to eq(1.82)
+      end
+    end
+
+    describe "#significant?" do
+      it "returns false when difference is not significant" do
+        analyzer = Analyzer.new(test_data)
+        expect(analyzer.significant?).to be false
+      end
```
sshao: might want to check that it returns true when difference *is* significant, too

data_science/lib/abtest/analyzer.rb
```diff
@@ -0,0 +1,81 @@
...
+      cohorts.each_with_object({}) do |cohort, standard_errors|
+        standard_errors[cohort.name] = cohort.standard_error
+      end
+    end
+
+    def confidence_interval_by_cohort
+      cohorts.each_with_object({}) do |cohort, confidence_intervals|
+        confidence_intervals[cohort.name] = cohort.confidence_interval
+      end
+    end
```
sshao: do you need all of these `*_by_cohort` methods, or are you just providing them as helpers to the user? in the case of the latter, it just seems like it'd be easier to provide `cohorts` as an attr_reader or something, and let the user manipulate `cohorts` themself to get these values.

arrubin: Originally most of the logic for the calculations was in these methods. After I extracted the Cohort class they make less sense. I am reconsidering what the API for the analyzer should be.

sshao:
This is solid. Consider my couple of comments, especially the analyzer's public API, but you don't need to resubmit.

PR #403, opened 2015-01-07
Yadda v1

djkotowski:


jmmastey:
Can you submit a pull for the relevant hound and gitignore stuff so we can include it upstream?

yadda/populate/.bundle/config
```diff
@@ -0,0 +1,2 @@
+---
+BUNDLE_JOBS: 8
```
jmmastey: Oh yeah?

djkotowski:
I'll take a look at the hound file.  It looks like there's some duplication from a merge.

yadda/populate/.bundle/config
djkotowski: I'll ensure `.bundle/` is excluded from Git in the future, given these configs are supposed to be machine-specific.

jmmastey:
This is all good :)

PR #404, opened 2015-01-08
Happenings (Local Events Feed) Resubmission

zsyed1:


zsyed1:
Was using wrong hound file. Am fixing hound voilations now, will resubmit shortly...

PR #405, opened 2015-01-08
Add data science

sriking1783:
Please Cr datascience


data_science/calculator.rb
```diff
@@ -0,0 +1,97 @@
...
+    (observed - expected)**2 / expected
+  end
+
+  def chi_square
+    chi_squared_for_experiments.inject(0) do |sum, (_k, v)|
+      sum + v
+    end
+  end
+
+  def experiment
```
dulyanov: I'd put this at the top to make it easy to read from top to bottom

data_science/experiment.rb
```diff
@@ -0,0 +1,43 @@
...
+    @data = JsonParser.parse(File.open(file_name, "r"))
+    @group_stats ||= {}
+    @data.keys.each do |cohort|
+      @group_stats[cohort] ||= {}
+    end
+  end
+
+  def observed_conversion_rate(cohort)
+    return @group_stats[cohort]["conversion_rate"] if
+                              @group_stats[cohort].key? "conversion_rate"
```
dulyanov: Remember the comment in one of your other pull requests on hash[key] returning nil if key doesn't exist? You can simplify this to `if @group_stats["conversion_rate"]` (same below)

data_science/experiment.rb
```diff
@@ -0,0 +1,43 @@
...
+    cohort_stats = @data[cohort]
+    conversion_rate = ((cohort_stats["conversions"].to_f / cohort_stats["total_visits"])).round(4)
+    @group_stats[cohort]["conversion_rate"] = conversion_rate
```
dulyanov: The entire method could be further simplified to
` @group_stats[cohort]["conversion_rate"] ||= ((@data[cohort]["conversions"].to_f / @data[cohort]["total_visits"])).round(4)

data_science/experiment.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def standard_error(cohort)
+    return @group_stats[cohort]["standard_error"] if
+                             @group_stats[cohort].key? "standard_error"
+    conversion_rate = @group_stats[cohort]["conversion_rate"] || observed_conversion_rate(cohort)
```
dulyanov: What's the point of this `||`? `observed_conversion_rate` already handles that

data_science/experiment.rb
```diff
@@ -0,0 +1,44 @@
...
+    trials          = @data[cohort]["total_visits"]
+    error           = Math.sqrt((conversion_rate * (1 - conversion_rate)) / trials).round(4)
+    @group_stats[cohort]["standard_error"] = error
+  end
+
+  def expected_conversion_rate(cohort)
+    return @group_stats[cohort]["expected_conversion_rate"] if
+                              @group_stats[cohort].key? "expected_conversion_rate"
+    conversion_rate = @group_stats[cohort]["conversion_rate"] || observed_conversion_rate(cohort)
```
dulyanov: same comment as above regarding these `||`

data_science/calculator.rb
```diff
@@ -0,0 +1,97 @@
...
+
+class Calculator
+  def initialize(file_name = "source_data.json")
+    @file = file_name
+    build_experiments
+  end
+
+  def build_experiments
+    experiment.data.each_key do |cohort|
+      experiment.observed_conversion_rate(cohort)
```
dulyanov: Since this is the only place you're referencing `observed_conversion_rate` and the other two methods outside of Experiment and you throw away the return values, why not just initialize everything in the experiment itself?

data_science/calculator.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def conversions_for_experiment
+    experiment.data.each_with_object({}) do |(cohort, values), conversions|
+      conversions[cohort] = values["conversions"]
+    end
+  end
+
+  def total_visits
+    visits_for_experiment.inject(0) do |sum, (_k, v)|
```
dulyanov: `values.inject(:+)`?

johnmcconnell:
Why so many hound errors?

data_science/experiment.rb
sriking1783: didn't understand sorry. Will talk to you later.

data_science/calculator.rb
sriking1783: didn't understand sorry. Will talk to you later.

johnmcconnell:
Okay so there are some things that you have done that are pretty cool like:
https://github.com/sriking1783/level_up_exercises/blob/add_data_science/data_science/spec/json_parser_spec.rb#L6
You avoided using method stubbing which is very nice.

Somethings that I do not like (IMO):
https://github.com/sriking1783/level_up_exercises/blob/add_data_science/data_science/calculator.rb#L4
in your initializer you call `build_experiments`. Executing logic in an initializer makes it very had for someone to reuse your code. If someone wants to extend your class they know have to be worried about whether their data cooperates with `build_experiments` and if they override `experiments` the method `build_experiments` is still called on initialization.

Just have `experiment` run `build_experiments` on initializing Experiment
Like so
```ruby
def experiment
    @experiment ||= 
       begin
          Experiment.new(@file)
          build_experiment
        end
end
```
Even that can be improved upon. You are explicitly saying that this class must use the `Experiment` not anything that acts as the `Experiment` class. The experiment should probably be passed into the initializer.

https://github.com/sriking1783/level_up_exercises/blob/add_data_science/data_science/experiment.rb
These methods could use some refactoring:
Stuff like:
```ruby
Math.sqrt((conversion_rate * (1 - conversion_rate)) / trials).round(4)
```
IMO should be in its own method **not** inline

All in all, I am pretty opinionated about this stuff. So if you disagree on my comments let me know.
I would recommend more refactoring .

data_science/experiment.rb
```diff
@@ -0,0 +1,43 @@
...
+    @group_stats[cohort]["conversion_rate"] =  ((@data[cohort]["conversions"] /
+      @data[cohort]["total_visits"].to_f)).round(4)
+  end
+
+  def standard_error(cohort)
+    return @group_stats[cohort]["standard_error"] if
+                             @group_stats[cohort]["standard_error"]
+    conversion_rate = observed_conversion_rate(cohort)
+    trials          = @data[cohort]["total_visits"]
+    error           = Math.sqrt(
```
johnmcconnell: Stuff like:
```ruby
Math.sqrt((conversion_rate * (1 - conversion_rate)) / trials).round(4)
```
IMO should be in its own method **not** inline

data_science/experiment.rb
```diff
@@ -0,0 +1,43 @@
...
+      (conversion_rate * (1 - conversion_rate)) / trials).round(4)
+    @group_stats[cohort]["standard_error"] = error
+  end
```
johnmcconnell: Stuff like:
```ruby
Math.sqrt((conversion_rate * (1 - conversion_rate)) / trials).round(4)
```
IMO should be in its own method **not** inline

data_science/calculator.rb
```diff
@@ -0,0 +1,102 @@
...
+class Calculator
+  def initialize(file_name = "source_data.json")
+    @file = file_name
+    build_experiments
+  end
+
+  def experiment
+    @experiment ||= Experiment.new(@file)
+  end
+
```
johnmcconnell: Executing logic in an initializer makes it very had for someone to reuse your code. If someone wants to extend your class they know have to be worried about whether their data cooperates with `build_experiments` and if they override `experiments` the method `build_experiments` is still called on initialization.

Just have `experiment` run `build_experiments` on initializing Experiment
Like so
```ruby
def experiment
    @experiment ||= 
       begin
          Experiment.new(@file).tap do |experiment|
            build_experiment(experiment)
          end
        end
end
```
Even that can be improved upon. You are explicitly saying that this class must use the `Experiment` not anything that acts as the `Experiment` class. The experiment should probably be passed into the initializer.

data_science/calculator.rb
johnmcconnell: Executing logic in an initializer makes it very had for someone to reuse your code. If someone wants to extend your class they know have to be worried about whether their data cooperates with `build_experiments` and if they override `experiments` the method `build_experiments` is still called on initialization.

Just have `experiment` run `build_experiments` on initializing Experiment
Like so
```ruby
def experiment
    @experiment ||= 
       begin
          Experiment.new(@file).tap do |experiment|
            build_experiment(experiment)
          end
        end
end
```
Even that can be improved upon. You are explicitly saying that this class must use the `Experiment` not anything that acts as the `Experiment` class. The experiment should probably be passed into the initializer.

data_science/spec/calculator_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  it "calculates total conversions experiment" do
+    total = described_class.new("test_data.json").total_visits
+    expect(total).to eq(8)
+  end
+  it "calculates total conversions experiment" do
+    total = described_class.new("test_data.json").chi_square
+    expect(total).to eq(0.53)
+  end
+end
```
johnmcconnell: Okay so there are some things that you have done that are pretty cool like:
https://github.com/sriking1783/level_up_exercises/blob/add_data_science/data_science/spec/json_parser_spec.rb#L6
You avoided using method stubbing which is very nice.

PR #406, opened 2015-01-09
Local Events Feed (Happenings) - Revision

zsyed1:


zsyed1:
@jmmastey I made the Ruby changes you suggested to the application views, but did not spend much time on fixing the CSS. I plan to use pmail for that purpose. Ashton has given me similar comments on my CSS there.

djkotowski:
I recognize many of these wallpapers.

local_events_feed/app/helpers/application_helper.rb
```diff
@@ -0,0 +1,51 @@
+module ApplicationHelper
+
+  include ShowingsHelper
```
jmmastey: It's a little weird for ApplicationHelper to include another Helper module (unless it's another truly global helper, I suppose). Not clear here.

local_events_feed/app/helpers/application_helper.rb
```diff
@@ -0,0 +1,51 @@
...
+    calendar.add_event(showing.to_ics)
+    calendar.publish
+    calendar.to_ical
+  end
+
+  def get_local_time_view(time)
+    LocalTimeView.new(time)
+  end
+
+  class LocalTimeView
```
jmmastey: I actually like this internal class here.

local_events_feed/app/helpers/showings_helper.rb
```diff
@@ -0,0 +1,53 @@
+require 'theatre_in_chicago/event'
```
jmmastey: This shouldn't be necessary.

local_events_feed/app/views/application/_user.html.erb
```diff
@@ -0,0 +1,39 @@
...
+	  <td>
+		<div class="add-to-cal">
+		  <%= link_to image_tag("calendar.png", :class => 'add-image tooltip', :title => 'Add Calendar'),
+		  showing_path(showing, format: :ical), :id => "add_cal-#{showing.id}" %>
+		</div>
+	  </td>
+	  <td class="attributes">
+		<ul>
+		  <li>
+			<% local_time_view = ApplicationHelper::LocalTimeView.new(showing.time) %>
```
jmmastey: I keep thinking you may be able to access this without the ApplicationHelper namespace. I believe the ApplicationHelper should be included in this scope, which would shorten these a bit.

jmmastey:
Hey, sorry this took so long. Added a few notes, but this looks good! Thanks.

PR #410, opened 2015-01-11
Pmail

jmarks1enova:
Done with plain CSS, plain HTML, and basic JQuery. (All style and UI logic done by hand.) A subset of interface features are implemented (had to stop somewhere!). I'm interested in advise regarding technique and organization.


pmail/inbox.html
```diff
@@ -0,0 +1,635 @@
...
+<meta charset="utf8" />
+<link rel="stylesheet" href="inbox.css" />
+<link rel="shortcut icon" href="FIXME.jpg" />
+<script src="jquery-1.11.2.js"></script>
+<script src="inbox.js"></script>
+</head>
+<body>
+  <header id="PageTop">
+
+    <hgroup id="TitleHeader" class="pageheader-container titleheader sidebar sidebar-indented">
```
ashtonthomas: The `hgroup` tag has been removed from the spec, so we should probably avoid it:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup

pmail/inbox.html
```diff
@@ -0,0 +1,635 @@
...
+<html lang="en">
+<head>
+<meta charset="utf8" />
+<link rel="stylesheet" href="inbox.css" />
+<link rel="shortcut icon" href="FIXME.jpg" />
+<script src="jquery-1.11.2.js"></script>
+<script src="inbox.js"></script>
+</head>
+<body>
+  <header id="PageTop">
```
ashtonthomas: This is entirely up to you, but I most often see a convention of naming id's and classes in all lower case with -'s

#foo-bar
.foo-bar

page-top
title-header
etc.

pmail/inbox.html
```diff
@@ -0,0 +1,635 @@
...
+
+      <div>
+
+        <table id="UnreadMessages" class="mailbox-list">
+
+          <tr class="mailbox-list-row mailbox-message-unread">
+            <td class="mailbox-list-handle"><a href="#"></a></td>
+            <td class="mailbox-list-select">
+              <div class="checkbox">
+                <img src="graphics/checkmark.png">
```
ashtonthomas: just note that sometimes it is good to use icon fonts:
http://fortawesome.github.io/Font-Awesome/icons/

This is probably more efficient than sending graphics over the wire

ashtonthomas:
Good use of semantic elements

pmail/inbox.html
```diff
@@ -0,0 +1,635 @@
...
+        </div>
+
+        <span class="to-right message-count" title="Show more messages">
+          <span class="bold">1-10</span> of <span class="bold">10</span>
+        </span>
+      </div>
+
+      <div>
+
+        <table id="UnreadMessages" class="mailbox-list">
```
ashtonthomas: I'm not a big fan of the use of a table here. I'm not saying you need to do another PR, but if you did I would suggest focusing on removing the table as a way of controlling the layout of messages

pmail/inbox.html
```diff
@@ -0,0 +1,635 @@
...
+
+          <tr class="mailbox-list-row mailbox-message-unread">
```
ashtonthomas: You could try adding a :hover state for `cursor:pointer` to give the user the finger :)

ashtonthomas:
The site isn't fully responsive, which is fine, but you may want to think about adding a min-width to the main containing element so the layout doesn't break if you are on small screens. You can argue that if the user want's to shrink the screen then it's their problem. However, I think this would only apply to smaller than mobile screens. A horizontal scroll bar on small screens isn't the worse thing (well you just need to evaluate horizontal scroll over a breaking layout - if the layout break is really that bad and it may not be)

pmail/inbox.html
ashtonthomas: You could use the same `cursor:pointer` styling to anything that is clickable and maybe include a very subtle style change

pmail/inbox.css
```diff
@@ -0,0 +1,735 @@
...
+body {
+  background-color: white;
+  margin: 0px;
+  font-family: sans-serif;
+  font-size: 13px;
+}
+
+/* LAYOUT */
+#PageTop {
+  position: fixed;
```
ashtonthomas: A good exercise would be to avoid using fixed

pmail/inbox.css
```diff
@@ -0,0 +1,735 @@
...
+
+#AccountControl {
+  position: relative;
+  display: inline-block;
+  text-align: right;
+  font-size: 13px;
+}
+
+#AppControls {
+  z-index: -1;
```
ashtonthomas: z-index makes me thing something isn't done quite right (and you have fixed position in here as well). I'm thinking there is a better way (feel free to talk to me if you want to dig into it)

pmail/inbox.css
```diff
@@ -0,0 +1,735 @@
...
+  width: 100%;
+}
+
+.appcontrol-container {
+  height: 40px;
+  padding-top: 5px;
+}
+
+#AppMain {
+  position: absolute;
```
ashtonthomas: here is another indicator that we aren't taking advantage of a fluid layout. The problem would be that as things change, this layout would be very brittle and may be hard to understand why things change/don't change when you want them to

ashtonthomas:
The main stuff is here. I would say that the use of absolute/fixed positioning and z-index along with the use of a table to the content layout are the biggest offenses, but I don't think it warrants another submission. Please feel free to talk to me if you want to dig into any of these comments

PR #413, opened 2015-01-12
Update supportive solution

zachary-huang:


supportive/supportive.rb
```diff
@@ -9,94 +9,50 @@ class BlagPost
...
   def initialize(args)
-    args = args.inject({}) do |hash, (key, value)|
-      hash[key.to_sym] = value
-      hash
-    end
-
-    if args[:author] != '' && args[:author_url] != ''
-      @author = Author.new(args[:author], args[:author_url])
-    end
+    args.symbolize_keys
```
incazteca: Using `symbolize_keys` like this will require you to store the results in another hash. If you want the args hash to have its keys symbolized use the method `symbolize_keys!`

supportive/supportive.rb
```diff
@@ -9,94 +9,50 @@ class BlagPost
...
+    cat = cat.presence || []
+    cat -= DISALLOWED_CATEGORIES unless cat.empty?
+    @categories = cat
+  end
 
-    @comments = args[:comments] || []
-    @body = args[:body].gsub(/\s{2,}|\n/, ' ').gsub(/^\s+/, '')
-    @publish_date = (args[:publish_date] && Date.parse(args[:publish_date])) || Date.today
+  def publish_date=(date)
+    date = (date.presence && Date.parse(date)) || Date.today
```
incazteca: Could just have this line assign to `@publish_date`. Also since this method just becomes a single line it kind of makes sense to just remove it and have publish_date assigned in the `initialize` method.

supportive/supportive.rb
```diff
@@ -9,94 +9,50 @@ class BlagPost
...
-      when 3 then "rd"
-      else "th"
-    end
-    "You will be the #{comments.length}#{ordinal} commenter"
+    return '' unless comments_allowed? || comments.length > 0
+    "You will be the #{comments.length.ordinalize} commenter"
   end
 
   def comments_allowed?
     publish_date + (365 * 3) > Date.today
```
incazteca: This method can be refactored using an ActiveSupport method

supportive/supportive.rb
```diff
@@ -9,94 +9,50 @@ class BlagPost
...
-  end
-
-  def as_title(string)
-    string = String(string)
-    words = string.gsub('_', ' ').split(' ')
-
-    words.map!(&:capitalize)
-    words.join(' ')
+    label = 'Category'.pluralize if categories.length > 1
+    "#{label} : #{categories.map! { |cat| String(cat).titleize }.to_sentence}"
```
incazteca: This line can also be simplified.

incazteca:
@zachary-huang Would like to see another pull request with fixes.

djkotowski:
Nice work.  Take a look at the comments and refactor with the necessary ActiveSupport extensions.

PR #414, opened 2015-01-12
Overlord

jfinzel:
Be careful


overlord/bomb.rb
```diff
@@ -0,0 +1,57 @@
...
+    if code == DEACTIVATE_CODE
+      deactivate_confirm
+    else
+      deactivate_deny
+    end
+  end
+
+  private
+
+  ACTIVATE_CODE = "1234"
```
incazteca: Please place constants at the top.

jfinzel: Even if private?

incazteca:
features directory should be in the project root path. Not within the spec directory.

incazteca:
I'm also unable to have the bomb work on my machine. Seems that Sinatra is unable to find the path.

jfinzel:
I had it under root, but then I was told specifically that features is to go under the spec folder by someone.  Also, the Capybara documentation says to put features under the spec folder ('for a Rails app', even though this isn't one).

How are you trying to run the bomb?  You have to run it with rackup command.
rackup

incazteca:
Using rackup command worked. Thanks

incazteca:
Missing tests for ensuring that input is numeric.

overlord/spec/features/bomb_control.feature
```diff
@@ -0,0 +1,17 @@
...
+
+  Background:
+    Given I have booted the bomb
+
+  Scenario: Activate the bomb
+    When I enter the right activation code
+    Then the status indicator shows as activated
+
+  Scenario: Deactivate the bomb
+    Given I enter the right activation code
```
incazteca: You can create a step or background to just give you an activated bomb. It will save you from having to rewrite the activation step and checking if activated.

overlord/spec/features/boot_steps_spec.rb
```diff
@@ -0,0 +1,13 @@
+require './spec/spec_helper.rb'
+include OverlordTest
+
+Given /^I have not booted the bomb$/ do
```
incazteca: Is this step needed?

overlord/spec/features/bomb_control.feature
jfinzel: So you mean I would add "I enter the right activation code" to background, then remove the redundancy for the two scenarios?

overlord/spec/features/control_steps_spec.rb
```diff
@@ -0,0 +1,16 @@
+require_relative '../spec_helper.rb'
+include OverlordTest
+
+When(/^I enter the right activation code$/) do
```
incazteca: Include tests where you do not enter a right activation code/deactivation code as well as tests for ensuring you're accepting numeric input.

jfinzel: I see what you mean about entering the wrong codes but no - the input is actually text.  0000 is the deactivation code.

overlord/spec/features/bomb_control.feature
incazteca: Yeah you could do that in say a feature file that tests for deactivation and have a background step that already has the bomb activated. Or you could make a statement like `Given I have an activated bomb`.

overlord/spec/features/boot_steps_spec.rb
jfinzel: I wasn't sure how to write this scenario - there really isn't a background because all I want to check is that it doesn't activate on "boot".  I want to have something that checks when I "boot" the bomb that it does not activate.

See boot_bomb.feature.  I am interested in a better way to write this.

incazteca:
@jfinzel I would like to see another request made. Also I'll try to set up a time for us to work on this together.

overlord/bomb.rb
djkotowski: `private` does not affect constants.  They can always be accessed via `Class::CONSTANT`

overlord/spec/spec_helper.rb
```diff
@@ -0,0 +1,93 @@
+require 'rspec'
+require 'capybara'
+require 'cucumber'
+
+# This file was generated by the `rspec --init` command. Conventionally, all
+# specs live under a `spec` directory, which RSpec adds to the `$LOAD_PATH`.
+# The generated `.rspec` file contains `--require spec_helper` which will cause this
```
djkotowski: Feel free to remove generated comments in the future.  It will reduce the number of hound errors you receive.

djkotowski:
Good work.  Take a look at the comments and refactor.

PR #415, opened 2015-01-13
Traffic control 2

paulh16:
For this first iteration, I have only created routes for the API endpoints defined in the instructions. Once these are all correct, I will establish additional API endpoints.

Thank you for reviewing!

### DIFF
--- a/traffic_control/routes.txt
+++ b/traffic_control/routes.txt
@@ -1,0 +1,46 @@
+# Item 1
+flight_bookings       GET  /flights/:flight_number/bookings                     bookings#index
+
+# Item 2
+flight                GET  /flights/:flight_number                              flights#show
+
+# Item 3
+plane_trips           GET  /planes/:plane_number/trips                          trips#index
+
+# Item 4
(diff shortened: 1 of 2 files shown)


traffic_control/routes.txt
```diff
@@ -0,0 +1,46 @@
...
+flight_bookings       GET  /flights/:flight_number/bookings                     bookings#index
+
+# Item 2
+flight                GET  /flights/:flight_number                              flights#show
+
+# Item 3
+plane_trips           GET  /planes/:plane_number/trips                          trips#index
+
+# Item 4
+plane_trips           GET  /planes/:plane_number/trips/:trip_number/refuel      trips#refuel
```
ashtonthomas: this requirement is new to me. 

To me, I would remove the trip aspect and just have a concept of 'stops' so a plane has stops and a stop may or may not require a refuel. `/planes/#/stops/DATE/refuel` 

no idea, though.

ashtonthomas:
LGTM

PR #416, opened 2015-01-13
add dinosaur that breaks everyone's period.split(' ').last == value code

jmmastey:
I like to be mean. More importantly, the "all observed period values are two words, the latter of which is the actual period" is a really brittle approach. Add a dinosaur that makes clear that this approach doesn't work well on real datasets.


ashtonthomas:
I like it.

Are we looking for a `.in?` solution?

jmmastey:
`.in?`, `.include?`, `=~`. There are still lots of options, but they all recognize the limitations of the previous approach.

ashtonthomas:
:+1:

PR #418, opened 2015-01-13
Add data science

sriking1783:
Refactored after 1st PR


data_science/calculator.rb
```diff
@@ -0,0 +1,37 @@
+require_relative 'experiment'
+
+class Calculator
+  def initialize(file_name = "source_data.json")
+    @file = file_name
+  end
+
+  def experiment
+    experiment ||= Experiment.new(@file)
```
dulyanov: `@experiment`?

data_science/experiment.rb
```diff
@@ -0,0 +1,50 @@
+require_relative 'json_parser'
+require_relative 'cohort'
+
+class Experiment
+  attr_accessor :cohorts
+
+  def initialize(file_name)
+    data = JsonParser.parse(File.open(file_name, "r"))
+    @cohorts = []
+    data.each do |dat|
```
dulyanov: you can just use map here `@cohorts = data.map {|key_value| ... }`

sriking1783: fixed in 099075801a1611d1a622d984e8e78be34a12ac15

data_science/calculator.rb
sriking1783: Ya thought if it, but again thought why, whats the point of instance variable. Does that serve any purpose? Or am I missing anything?

dulyanov: You are missing something, memorization. Otherwise the local experiment variable is getting thrown away each time.

sriking1783: dang, I should have gotten that clue atleast from hound complaint.
fixed abacd6f9562ec70003223de59aa6b246e00c9be6

ashtonthomas:
Hey, @sriking1783. I just took a look at this. I think there is some really great ruby code in here, and some good tests. However, I'm not sure I am following your `Calculator#chi_square` method. More specifically, can you add a test that illustrates testing 3 different data sets: 
- Group A is significantly better
- Group B is significantly better
- The difference is not significant and there is no clear winner

95% of the code here is a real pleasure to read. but I think this last example will give me what I need to follow the rest of it. I'm just not sure I'm smart enough to understand how you are using `chi_squared_for_experiments`

- A

ashtonthomas:
It would have also been pretty nice if we had a main.rb/runner.rb or some easy way to run the code all the way through. Your specs are a great way of doing this, but I found myself wanting a simple way to just run the code instead of reading the unit tests and thinking through how everything would work together

PR #420, opened 2015-01-13
Overlord

jperkins:
Looking for feedback on the configuration of guard, etc. as well as the Sinatra app layout. Thanks!


ashtonthomas:
@jperkins, @aeberlin is probably your guy for Guard configuration, and he can probably help with Sinatra layout.

djkotowski:
@jperkins, @aeberlin paired.

PR #422, opened 2015-01-13
Data science

ddzz:
My first submission for the data science exercise.  I have created three classes: one for parsing and tallying json data (ABDataSummary), one for cohort data/calculations (Cohort), and one for confidence level calculation using the ABAnalyzer gem (ABCalculator).


data_science/lib/cohort.rb
```diff
@@ -0,0 +1,37 @@
...
+    (@conversions.to_f / total_trials).round(6)
+  end
+
+  def standard_error
+    (sqrt(conversion_rate * (1 - conversion_rate) / total_trials)).round(6)
+  end
+
+  def conversion_range
+    Range.new(
+      (conversion_rate - 1.96 * standard_error).round(6),
```
ashtonthomas: `1.96` might be a good candidate for a constant

data_science/lib/cohort.rb
```diff
@@ -0,0 +1,37 @@
...
+      (conversion_rate + 1.96 * standard_error).round(6),
+    )
+  end
+
+  private
+
+  def validate_params(conversions, nonconvs)
```
ashtonthomas: You have this `valdiate_*` pattern in these files. I would like to suggest a video by Avdi: https://www.youtube.com/watch?v=T8J0j2xJFgQ

In more than one place, you are checking the format of things. How about if you extracted all of the areas where you need to validate things into a different place, and then you acted with confidence everywhere else?

Let's have confidence in your code. Your unit tests should be the area where you can do sanity checks.

data_science/lib/a_b_data_summary.rb
```diff
@@ -0,0 +1,37 @@
...
+    results = Hash.new { |h, k| h[k] = 0 }
+    data.each do |item|
+      results[(item[:cohort].downcase + "_" + item[:result].to_s).to_sym] += 1
+    end
+    validate_results(results)
+    results
+  end
+
+  def validate_results(results)
+    raise ArgumentError unless (results.keys - VALID_RESULTS).empty?
```
ashtonthomas: This is a unique way of validating something. See my comment about confident coding, and we should be able to make a lot of this go away

ashtonthomas:
Hey, @ddzz. Any chance I could get a runner.rb/main.rb that runs through a full example? I know the specs give me a lot of this, but it would be nice to a full example for:

- A is significant
- B is significant
- No clear winner

ashtonthomas:
@ddzz, thanks for the added file. This is exactly what I was looking for. In adding this file, I think you have added more logic that probably should be in a different place. 

What do you think about the following logic? Should this logic live in your app or within the user of your app (here runner.rb acts as the user I think)

```
if a_b_test.different?
  puts "Samples are different at 0.05 significance"
  if a_cohort.conversion_rate > b_cohort.conversion_rate
    puts "A is winner"
  else
    puts "B is winner"
  end
end
```

I'm thinking there is something to gain by moving this logic into an appropriate place in your main code and trying to do away with all of the valid_results, validate_results stuff (having more confidence - and if there is an area where you don't have confidence, move it out of the core logic and isolate it so you can be confident in the areas that matter) - no idea if that makes any sense so .. this is all FWIW..

PR #423, opened 2015-01-13
Data science feature

jillalynch:


data_science/experiment.rb
```diff
@@ -0,0 +1,52 @@
...
+  PROBABILITY_THRESHOLD = 0.05 # why
+
+  def initialize(results)
+    create_cohorts(pageviews(results))
+    @test = ABAnalyzer::ABTest.new(format_for_abanalyzer)
+  end
+
+  # prints out the winner cohort (make a winner method)
+  # if there's no winner you print out another message
+  def report
```
ashtonthomas: Is this method supposed to print or return a string? It looks like it does print sometimes but not always

data_science/page_view.rb
```diff
@@ -0,0 +1,16 @@
+class PageView
```
ashtonthomas: FWIW, I would ask myself if `PageView` is more specific if than was you are able to do with a more generic ABtest?

data_science/experiment.rb
```diff
@@ -0,0 +1,52 @@
+require 'abanalyzer'
+require_relative 'page_view.rb'
+require_relative 'cohort.rb'
+
+class Experiment
+  CONFIDENCE_LEVEL      = 1.96 # why
```
ashtonthomas: "why"? haha

data_science/experiment.rb
```diff
@@ -0,0 +1,52 @@
...
+    if winner
+      p winner.to_s
+      winner.name
+    else
+      "There is no clear winner!"
+    end
+  end
+
+  def winner
+    @cohorts.sort_by(&:conversion_percentage).to_a.last if difference?
```
ashtonthomas: Maybe you can ignore the `difference?` in `#winner` and instead use that in the `#report` method?

Cause maybe you do want the winner even if it isn't significant

ashtonthomas: Also, do you need to do the `.to_a` before .last?

ashtonthomas:
Look'in good. Enjoyed listening to ruby rogues while reviewing this!

See my comments. Up to you if you want to PR. I don't think it would be necessary

PR #425, opened 2015-01-13
Rc02 dino catalog

magicalbanana:


magicalbanana:
@jmmastey please review sir.

ashtonthomas:
@magicalbanana, I'm thinking you want your /coverage folder in the .gitignore that is the root directory not just the one in /dino_catalog

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,78 @@
...
+class DinoDex
+
+  attr_accessor :dinosaurs
+
+  def initialize
+    parse_csv
+    print_to_console(@dinosaurs)
+  end
+
+  def dinosaurs
```
ashtonthomas: I think you are overriding your attr_accessor (dynamic method), and I'm not sure about having a noun-like method that prints instead of returning. I think this could lead users(other programmers who are using your code) astray

ashtonthomas:
Hey, @magicalbanana, can you add some kind of main.rb or runner.rb that runs the app? I know you have specs that do the unit tests, but it would be helpful for us to see a full run through and be able to offer feedback on how everything comes together

ashtonthomas:
Talked in person. Will let another Bucketeer do the next PR.

Thanks @magicalbanana (awesome handle btw)

PR #428, opened 2015-01-14
Rc03 triangle facts

magicalbanana:
@jmmastey please review sir.

Please forgive me for not adding the coverage off the bat.


ashtonthomas:
Yo, let's update your gitignore :-)

triangle_facts/triangle_facts.rb
```diff
@@ -1,49 +1,45 @@
 # Killer facts about triangles AWW YEAH
 class Triangle
-	attr_accessor :side1,:side2,:side3
+	attr_accessor :side_a,:side_b,:side_c
```
ashtonthomas: So I'm not sure this is the biggest gain.. a, b, c - 1, 2, 3 - not sure we can give you credit for this one ;-)

I think we are going for the lack of spaces in between these guys and in the method params below

triangle_facts/triangle_facts.rb
```diff
@@ -1,49 +1,45 @@
...
-
 		puts 'This triangle is also a right triangle!' if angles.include? 90
 		puts ''
 	end
 
-	def calculate_angles(a,b,c)
-		angleA = radians_to_degrees(Math.acos((b**2 + c**2 - a**2) / (2.0 * b * c)))
-		angleB = radians_to_degrees(Math.acos((a**2 + c**2 - b**2) / (2.0 * a * c)))
-		angleC = radians_to_degrees(Math.acos((a**2 + b**2 - c**2) / (2.0 * a * b)))
+	def calculate_angles(side_a,side_b,side_c)
```
ashtonthomas: Again, not sure the name/variable changes are all that significant. One thing to think about here is that there seems to be no reason for this to be an instance method. Maybe it should be, and if so, maybe we can get rid of the parameters and even utilize memoization. Just a thought.

ashtonthomas:
@magicalbanana your tests are something to be admired, plz add coverage to gitignore. I think you got a lot of the desired changes and left a few out and maybe did some unnecessary name changing.

Another thing to do is get your editor setup to convert tabs into spaces hint, hint :smile: 

Do another PR when you get a chance, and I'm sure it will be a simple glance. Shouldn't take long to knock out

PR #429, opened 2015-01-16
CYOA: Opposite of progress

tharumax:
Fully functional version of this app can be found in the following location.

https://congressional-good-deeds.herokuapp.com/

JSON API can be accessed at following paths. Every api path returns paginated data. `api/*/favorites` does not work currently as app still does not support authentication via HTTP.

```
api/good_deeds
api/good_deeds/:id
api/good_deeds/favorites # show good deeds from favorited legislators and bills
api/bills
api/bills/:id
api/bills/favorites # show good deeds from favorited bills
api/legislators
api/legislators/:id
api/legislators/favorites # show good deeds from favorited legislators
```


tharumax:
Due to the large number of images files (legislators images) present in the source tree, diff is not visible in this PR. Going to close this PR and resubmit without the images.

PR #435, opened 2015-01-16
Shopping cart

jfinzel:


shopping_cart/spec/features/browse_items.feature
```diff
@@ -0,0 +1,19 @@
...
+  As a customer
+  I want to be able to browse the items
+
+  Scenario Outline: Browsing Scenarios
+    Given I am <logged_in>
+    And <number_of_items> items exist
+    And I am on the <page> page
+    And the "<item_name>" item exists
+    When I click on <something>
+    Then <something_happens>
```
jmmastey: This one is pretty weird as a step. Usually you'd use the table technique to make sure that some message appears depending on the input. This really seems to be several test cases masquerading as one.

shopping_cart/spec/features/enter_address.feature
```diff
@@ -0,0 +1,29 @@
...
+
+  @happy
+  Examples: Go to Address Page
+    | logged_in     | saved_address      | number_of_items | page_from     | address_prompt                   |
+    | logged in     | a saved address    | 1               | items         | address confirmation content     |
+    | logged in     | no saved addresses | 1               | items         | add new address for user content |
+    | logged in     | a saved address    | 1               | shopping cart | address confirmation content     |
+    | logged in     | no saved addresses | 1               | shopping cart | add new address for user content |
+    | not logged in | no saved addresses | 1               | items         | enter address for order content  |
+    | not logged in | no saved addresses | 1               | shopping cart | enter address for order content  |
```
jmmastey: One thing you may want to consider is removing the `number_of_items` entry altogether, since it's always one. Then, perhaps split the test cases into logged-in vs. not. It looks like the page_from is really not a variant in the test, which would drop half the remaining cases and leave you with a few remaining cases.

shopping_cart/spec/features/enter_address.feature
```diff
@@ -0,0 +1,29 @@
...
+
+  @bad
+  Examples: Do Not Proceed
+    | logged in     | a saved address    | 0               | items         | no items message                 |
+    | logged in     | a saved address    | 0               | shopping cart | no items message                 |
+    | not logged in | no saved addresses | 0               | items         | no items message                 |
+    | not logged in | no saved addresses | 0               | shopping cart | no items message                 |
```
jmmastey: This is another good example. Really, all you're trying to say is "if there aren't any items in my cart, and I click on checkout, then it tells me I have no items". That doesn't need four cases to enumerate. Your logged in status, addresses and source page are all irrelevant.

shopping_cart/spec/features/login.feature
```diff
@@ -0,0 +1,33 @@
+Feature: Login
+	In order to save my shopping history
```
jmmastey: TAB!

shopping_cart/spec/features/login.feature
```diff
@@ -0,0 +1,33 @@
...
+  @happy
+  Examples: Successful logins
+    | user | password | exists | user_entry | password_entry | result                                 |
+    | john | secret   | exists | john       | secret         | I see the successful login result page |
+
+  @sad
+  Examples: Unsuccessful logins
+    | user | password | exists         | user_entry | password_entry | result                                 |
+    | john | secret   | exists         | john       | wrong          | I see an unsuccessful login message    |
+    | john | secret   | exists         | john       | wrong          | I see an unsuccessful login message    |
```
jmmastey: The table made it so hard to read that you ended up with a duplicate test case :)

shopping_cart/spec/features/login.feature
```diff
@@ -0,0 +1,33 @@
...
+    | john | secret   | does not exist | john       | secret         | I see an unsuccessful login message    |
```
jmmastey: Not sure how a non-existent user could possibly authenticate? This is probably already covered reasonably well by the "wrong password" case. Not sure what a failure would look like here, either.

shopping_cart/spec/features/login.feature
```diff
@@ -0,0 +1,33 @@
...
+
+  @bad
+  Examples: Invalid logins
+    | user | password | exists         | user_entry | password_entry | result                                        |
+    | john | secret   | exists         |            | wrong          | I see a missing username message              |
+    | john | secret   | exists         | john       |                | I see a missing password message              |
+    | john | secret   | exists         |            |                | I see a missing username and password message |
```
jmmastey: Only useful insofar as you have different messages for errors. If there weren't separate messages these cases would be superfluous.

shopping_cart/spec/features/register.feature
```diff
@@ -0,0 +1,38 @@
...
+
+  Background:
+    Given I navigate to the registration page
+
+  Scenario Outline: Try registrations
+    Given user <user> with password <password> <exists>
+    When I enter username <user_entry>
+    And I enter password <password_entry>
+    And I submit the registration information
+    Then <result>
```
jmmastey: Your unit tests should verify that the user model is properly validated. These integration tests are better for the interactions between parts of the system, and for routing.

If there were 10 items on the registration page, this test would explode to infinite combinations. Don't create infinite combinations!

shopping_cart/spec/features/register.feature
```diff
@@ -0,0 +1,38 @@
...
+    | john | secret   | exists         | john       | wrong          | I see a user already exists message |
+
+  @bad
+  Examples: Invalid registrations
+    | user | password | exists         | user_entry | password_entry | result                                         |
+    | john | secret   | does not exist |            | wrong          | I see a missing username message               |
+    | john | secret   | does not exist | john       |                | I see a missing password message               |
+    | john | secret   | does not exist |            |                | I see a missing username and password message  |
+    | john | secret   | exists         |            | wrong          | I see a missing username message               |
+    | john | secret   | exists         | john       |                | I see a user already exists message            |
```
jmmastey: This one, though, might be worth a test. Not sure.

shopping_cart/spec/features/update_shopping_cart.feature
```diff
@@ -0,0 +1,62 @@
...
+    Then I am on the <result_page> page
+    And I have <result_number_of_item> "<item_name>" items in my cart
+
+  @happy
+  Examples:
+    | logged_in     | number_of_item | item_name | result_page   | result_number_of_item |
+    | logged in     | 0              | freezer   | shopping cart | 1                     |
+    | logged in     | 1              | freezer   | shopping cart | 2                     |
+    | not logged in | 0              | freezer   | shopping cart | 1                     |
+    | not logged in | 1              | freezer   | shopping cart | 2                     |
```
jmmastey: Not sure what your test is getting at here? Maybe that if you add an item it starts to appear in your cart, and if it was already there it gets incremented? In that case, two small tests are more comprehensible than this arrangement.

jmmastey: What if there aren't enough items left?

shopping_cart/spec/features/update_shopping_cart.feature
```diff
@@ -0,0 +1,62 @@
...
+
+  Scenario Outline: Remove Items
+    Given I am <logged_in>
+    And I have <number_of_item> "<item_name>" items in my cart
+    And I am on the shopping cart page
+    When I click to remove the "<item_name>" from my cart
+    Then I am on the shopping cart page
+    And I have <result_number_of_item> "<item_name>" items in my cart
```
jmmastey: Again, logged_in is not a real variant here. One thing you could do is just to always start with 3 freezer items in your cart. Then, by removing one, you test for two left. By removing three, you have none left, etc etc.

shopping_cart/spec/features/use_coupons.feature
```diff
@@ -0,0 +1,28 @@
...
+    When I enter coupon code <enter_coupon_code>
+    Then I see <message>
+    And my total is <total>
+
+  @happy
+  Examples: Successfully applied coupons
+    | logged_in     | subtotal | coupon_code | valid   | coupon_value | enter_coupon_code | message                                              | total |
+    | logged in     | 10.00    | 1234        | valid   | 5.00         | 1234              | a coupon successfully applied message                | 5.00  |
+    | logged in     | 20.00    | 1111        | valid   | 20.00        | 1111              | a coupon successfully applied message                | 0.00  |
+    | logged in     | 20.00    | 1111        | valid   | 30.00        | 1111              | a coupon successfully applied with remainder message | 0.00  |
```
jmmastey: Set up the coupons in a `Background` block, including codes and value, along with the logged in bit (which is probably not relevant here). Ditto the shopping cart.

Then you can concentrate on the real purpose of the test: the coupon.

shopping_cart/spec/features/use_coupons.feature
```diff
@@ -0,0 +1,28 @@
...
+    
+  @sad
+  Examples: Unsuccessfully applied coupons
+    | logged_in     | subtotal | coupon_code | valid   | coupon_value | enter_coupon_code | message                                              | total |
+    | not logged in | 10.00    | 1234        | valid   | 5.00         | 1234              | a message that only registered users can use coupons | 10.00 |
+    | logged in     | 20.00    | 1111        | invalid | 0.00         | 1111              | a coupon does not exist message                      | 20.00 |
+    | logged in     | 10.00    | 1234        | valid   | 0.00         | 1234              | a coupon balance is exhaused message                 | 10.00 |
```
jmmastey: Not sure if that's how coupons work...

shopping_cart/spec/features/view_item_from_shopping_cart.feature
```diff
@@ -0,0 +1,16 @@
...
+    Given I am on the shopping cart page
+    And I have <number_of_item> "<item_name>" items in my cart
+    When I click on the "<item_name>" item
+    Then <something_happens>
+
+  @happy
+  Examples:
+    | number_of_item | item_name  | something_happens                  |
+    | 1              | fishbowl   | I go to the "fishbowl" item page   |
+    | 2              | monkey hat | I go to the "monkey hat" item page |
```
jmmastey: Aww c'mon, this one clearly doesn't deserve a table. It's just a single test.

jmmastey:
Too many tables, overall. They're clouding your structure. I like the brainstorming, and you hit most of the good cases, but it's hard to understand what the software does when there are so many cases.

Try to pare them down. Ask me if you have questions.

PR #436, opened 2015-01-16
refactor dinodex for concision

katieschilling:
dinodex.search is longer than 7 lines, but my pair agreed.


dino_catalog/lib/dino.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  def initialize(args = {})
+    super
+    @continent = "Africa"
+    @name      = args[:genus]
+    @carnivore = args[:carnivore] == 'Yes'
+    @weight    = args[:weight]
+    @diet      = 'Carnivore' if @carnivore
+  end
+end
```
jmmastey: This is the only thing so far I'm not totally excited about. It happens to work for this dataset, but I think this approach would become pretty untenable over time.

dino_catalog/lib/dinodex.rb
```diff
@@ -0,0 +1,25 @@
...
+  def method_missing(method, *args)
+    return all.send(method) if all.respond_to?(method)
+    super
+  end
+
+  private
+
+  def read_csv
+    CSV.table(@csv_file).each_with_object([]) do |i, arr|
+      arr << Hash[i.to_h.map { |k, v| [k.downcase.to_sym, v] }]
```
jmmastey: This construction is really hard to read; I'd hate to have to debug it. Find something better.

dino_catalog/lib/dinoset.rb
```diff
@@ -0,0 +1,60 @@
+require 'json'
+
+class DinoSet
+  include Enumerable
```
jmmastey: Be careful including all of Enumerable. There are _lots_ of methods on it, and you may not want them all.

dino_catalog/lib/dinoset.rb
```diff
@@ -0,0 +1,60 @@
...
+  def search(args)
+    args.each_with_object(self.dup) do |(key, value), results|
+      key = key.to_s.downcase
+      value.downcase! if value.respond_to?(:downcase!)
+
+      results.pluck! do |dino|
+        result = dino.send(key)
+        result.downcase! if result.respond_to?(:downcase!)
+        result == value
+      end
```
jmmastey: The variable names make this is a bit hard to read. That said, I think you might be better off with `@dinos.each_with_object({})` then using `criteria.all?` to see if a particular dino matches. The dup-and-remove strategy here is tough to keep track of. Clever, though :)

jmmastey:
I added a few minor notes, but overall this works just fine. Consider making the revisions and move on.

jmmastey:
Let me know if it's unclear why the `AfricanDino` solution is not a good one for dealing with different CSVs.

PR #437, opened 2015-01-16
Triangle facts in OOD

katieschilling:


jmmastey:
Dirty commits to master boooo

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,55 @@
...
+      when :isosceles
+        'This triangle is isosceles! Also, that word is hard to type.'
+      else 'This triangle is scalene and mathematically boring.'
+    end
+  end
+
+  def angle_facts
+    angles = calculate_angles
+    "The angles of this triangle are #{angles.join(',')}" \
+      "#{"\nThis triangle is also a right triangle!" if angles.include? 90}"
```
jmmastey: Ehh, this is only a styleguide assignment, but avoid this style of string concat. It's really hard to tell that it's conditional (and you're at 80 chars anyway).

On the topic of 80 chars: remember that fake linebreaks don't count as satisfying the rule. Reduce the complexity of the line instead.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,55 @@
...
+  end
+
+  def triangle_type
+    return :equilateral if side1 == side2 && side2 == side3
+    return :isosceles   if [side1, side2, side3].uniq.length == 2
+    :scalene
+  end
```
jmmastey: I was more looking for you to know that `def scalene()` should be `def scalene?`. This works, though.

Similarly to the above, be careful about methods with many short-circuit returns. This is basically the same as a case statement, or an if/else, but harder to read.

jmmastey:
Added a few notes, but you're good to go.

PR #438, opened 2015-01-16
Arrowhead

katieschilling:


jmmastey:
Ack! The dirty commits tricked me into reviewing arrowhead on #439 instead of #438! Assume I did them both on the other one.

PR #439, opened 2015-01-16
Robot

katieschilling:


arrowhead/arrowhead.rb
```diff
@@ -17,17 +17,19 @@ class Arrowhead
...
-    end
+    return identify_shape(region, shape) if CLASSIFICATIONS.include? region
+    raise "Unknown region, please provide a valid region."
+  end
+
+  def self.identify_shape(region, shape)
+    shapes = CLASSIFICATIONS[region]
+    shape_error unless shapes.include? shape
+    "You have a(n) '#{shapes[shape]}' arrowhead. Probably priceless."
+  end
```
jmmastey: This meets the letter, but the levels of abstraction are lopsided. For instance, `classify` only cares if `CLASSIFICATIONS.include? region`, and raises if it doesn't understand region. `identify_shape` is at a different level of abstraction (since it gets called by identify shape), duplicates the knowledge about the classification/region relationship, then delegates even further for throwing an error.

Take a look at the `Hash#fetch` method for a cleaner way to accomplish it.

jmmastey:
It got shorter, but also harder to read. Try once more with the `Hash#fetch` thing. Also, please make sure your repo is clean of dinosaur changes :)

jmmastey:
Ack! The dirty commits tricked me into reviewing arrowhead on #439 instead of #438! Assume I did them both here.

robot_name/robot_name.rb
```diff
@@ -1,25 +1,29 @@
...
-    @@registry ||= []
-    @name_generator = args[:name_generator]
+  def generate
+    letters = ('A'..'Z').to_a + ('a'..'z').to_a
+    name = letters.sample(2).join + (0..9).to_a.sample(3).join
 
-    if @name_generator
-      @name = @name_generator.call
+    if used?(name)
+      generate
```
jmmastey: There's a `retry` in ruby that might help you here. It runs the risk of an infinite loop, though.

More importantly, you removed my ability to use my own `args[:name_generator]`. Try adding that back and still fighting off the complexity.

PR #441, opened 2015-01-20
Add overlord

sriking1783:
PR for overlord part 1


overlord/app/helpers/bomb_helpers.rb
```diff
@@ -0,0 +1,31 @@
...
+  def self.match_activation_code?(bomb, code)
+    bomb.activation_code == code["activation_code"] &&
+      !BombHelpers.active?(bomb)
+  end
+
+  def self.match_deactivation_code?(bomb, code)
+    bomb.deactivation_code == code["deactivation_code"] &&
+      BombHelpers.active?(bomb)
+  end
+end
```
djkotowski: These methods all take in a bomb.  Why aren't they in `Bomb`?

overlord/app/helpers/bomb_helpers.rb
```diff
@@ -0,0 +1,31 @@
...
+  def self.bomb_status(bomb)
+    return { bomb_status: nil }.to_json if bomb.nil?
+    if self.bomb_exploded? bomb
+      bomb.status = :explode
+      bomb.save!
+    end
+
+    { bomb_status: bomb.status, detonation_time: bomb.detonation_time,
+      wires: bomb.wires }.to_json
+  end
```
djkotowski: 1.  The method returns a JSON object rather than a hash.  This responsibility seems like it belongs elsewhere.
2.  `bomb_status` returns a hash including `bomb_status`, `detonation_time`, and `wires`.  This seems like three methods in one.

overlord/app/helpers/bomb_helpers.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def self.bomb_exploded?(bomb)
```
djkotowski: Should this be private, since it's just used in `bomb_status`?  Also, the method implies it is returning if the bomb is exploded or not, when it is used in `bomb_status` to set the status.  This is confusing.

overlord/app/helpers/bomb_helpers.rb
```diff
@@ -0,0 +1,31 @@
...
+    !(bomb.activated_time.nil?) &&
+      (Time.now >= bomb.activated_time + bomb.detonation_time.seconds)
+  end
+
+  def self.match_activation_code?(bomb, code)
+    bomb.activation_code == code["activation_code"] &&
+      !BombHelpers.active?(bomb)
```
djkotowski: I'd use `self` instead of `BombHelpers`.  Otherwise, if you change the module name, you'll have to change the reference throughout the module.

overlord/Gemfile
```diff
@@ -1,6 +1,31 @@
...
 
 gem "sinatra"
+gem 'capistrano', '2.14.2'
+gem 'haml', '3.1.7'
+gem 'sass', '3.2.1'
+gem 'shotgun', '0.9'
+gem 'dm-sqlite-adapter'
+gem 'datamapper', require: 'data_mapper'
+gem "sinatra-activerecord"
+gem "sqlite3"
```
djkotowski: In the future, I'd get used to using Postgres.  This doesn't need to be changed in this project, though.

overlord/app/overlord.rb
```diff
@@ -0,0 +1,45 @@
...
+
+register Sinatra::Sprockets
+
+class Hash
+  def self.to_ostructs(obj, memo = {})
+    return obj unless obj.is_a? Hash
+    os = memo[obj] = OpenStruct.new
+    obj.each { |k, v| os.send("#{k}=", memo[v] || to_ostructs(v, memo)) }
+    os
+  end
```
djkotowski: Please don't modify base classes.

overlord/app/overlord.rb
```diff
@@ -0,0 +1,45 @@
...
+set :database, adapter: "sqlite3", database: "detonation_device"
+set :views, "#{File.dirname(__FILE__)}/views"
+
+class Overlord < Sinatra::Application
+  include OverlordHelpers
+  configure :development do
+    DataMapper::Logger.new($stdout, :debug)
+  end
+  get '/' do
+    Bomb.destroy_all
```
djkotowski: Here your DB doesn't support multiple users.  Another user navigates to `/` and the first user's bomb is destroyed.

overlord/app/overlord.rb
```diff
@@ -0,0 +1,45 @@
...
+    haml :index
+  end
+
+  use Rack::Session::Cookie, key: 'rack.session',
+                           domain: 'http://localhost:9292/',
```
djkotowski: If the port changes, this is no longer accurate.

overlord/app/overlord_helpers.rb
```diff
@@ -0,0 +1,5 @@
+module OverlordHelpers
+  def start_time
```
djkotowski: Just move this into Overlord.

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,91 @@
+require 'sinatra'
+require 'dm-sqlite-adapter'
+require 'json'
+
+require_relative '../helpers/bomb_helpers'
+class Overlord < Sinatra::Application
+  include BombHelpers
+
+  get '/bomb.json' do
+    bomb = Bomb.last
```
djkotowski: Again, multiple users?

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,91 @@
...
+    haml :bomb
```
djkotowski: Why render Haml in the JSON route?

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,91 @@
...
+  get '/bomb' do
+    bomb = Bomb.last
+    BombHelpers.bomb_status(bomb)
+    haml :bomb
+  end
+
+  get '/bomb/diffuse' do
+    bomb = Bomb.last
+    wire = Wire.where(color: params["color"]).first
+    if wire.diffuse?
```
djkotowski: This logic should be in the bomb.

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,91 @@
...
+      bomb.status = "inactive"
+    else
+      bomb.status = "explode"
+    end
+    bomb.save!
+    haml :bomb
+  end
+
+  post '/bomb' do
+    request_json = {}
```
djkotowski: This isn't JSON yet.  It's a hash.
Also, don't list data types in variable names.

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,91 @@
...
+      request_json = JSON.parse(request.body.read)
+    end
+
+    bomb = Bomb.create(
+      activation_code: request_json["activation_code"],
+      deactivation_code: request_json["deactivation_code"],
+      detonation_time: request_json["detonation_time"])
+
+    request_json["wires"] ||= [{ color: "red", diffuse: true },
+                                { color: "green", diffuse: true }]
```
djkotowski: Why are wire defaults defined here?

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    bomb.save!
+
+    request_json["wires"].each do |wire|
+      wire_string = bomb.wires.build(wire)
```
djkotowski: Why not `wire = bomb.wires.build(wire_options)`?  As it is, `wire` is a hash and `wire_string` is a wire.

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  post '/bomb/activate' do
+    request.body
+    request_json = JSON.parse(request.body.read)
+    bomb = Bomb.last
+    if BombHelpers.match_activation_code?(bomb, request_json)
+      bomb.status = :active
+      bomb.activated_time = Time.now
+      bomb.failed_attempts = 0
+      bomb.save!
```
djkotowski: This logic should be in your bomb.

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,91 @@
...
+      if bomb.failed_attempts == 3
+        bomb.status = :explode
+        bomb.activated_time = nil
+        bomb.save!
+      end
+    end
+    BombHelpers.bomb_status(bomb)
+  end
+end
```
djkotowski: This is a *lot* of logic for a routes/controller file.  Try to extract some of this logic to other classes.

overlord/app/views/bomb.haml
```diff
@@ -0,0 +1,191 @@
+%link{:href => "/style.css", :rel => "stylesheet"}/
```
djkotowski: If you're going to put script references and HTML body in the same file, split it into `<head>` and `<body>` tags.  As it is now, it relies heavily on the forgiveness of browsers for poorly-structured HTML.

overlord/app/views/bomb.haml
```diff
@@ -0,0 +1,191 @@
...
+  %p
+    %span.ui-icon.ui-icon-alert{:style => "float:left; margin:0 7px 20px 0;"}>
+    These items will be permanently deleted and cannot be recovered. Are you sure?
+.timer
+
+.label
+  %label Status
+
+#wires.bomb_wires
+  %a{:href => "http://localhost:9292/bomb/diffuse?color=green"}
```
djkotowski: Don't hardcode hostnames and protocols in the anchors.

overlord/app/views/bomb.haml
```diff
@@ -0,0 +1,191 @@
...
+    %font{:color => "green"} Green
```
djkotowski: Use CSS classes for styling.

overlord/app/views/bomb.haml
```diff
@@ -0,0 +1,191 @@
...
+%textarea{ readonly: true, id: "bomb_status",class: "bomb_inactive" }
+  ="Inactive"
+#bomb
+  / Screen and clear key
+  .top
+    %span.clear C
+    .screen
+  .keys
+    / operators and other keys
+    %span.7 7
```
djkotowski: UI is not the focus of this exercise, but I would pick a class more detailed than `<number>` for these.  Maybe `keypad-7`?

overlord/app/views/bomb.haml
```diff
@@ -0,0 +1,191 @@
...
+    %span.6 6
+    %span.inactive_operator.deactivate Deactivate
+    %span.1 1
+    %span.2 2
+    %span.3 3
+    %span.operator Status
+    %span.0 0
+/ PrefixFree
+
+:javascript
```
djkotowski: Separate JS into a separate file.

overlord/app/views/bomb.haml
```diff
@@ -0,0 +1,191 @@
...
+
+
+
+
+
+
+
+
+
+
```
djkotowski: Excess newlines.

overlord/db/schema.rb
```diff
@@ -0,0 +1,31 @@
...
+# of editing this file, please use the migrations feature of Active Record to
+# incrementally modify your database, and then regenerate this schema definition.
+#
+# Note that this schema.rb definition is the authoritative source for your
+# database schema. If you need to create the application database on another
+# system, you should be using db:schema:load, not running all the migrations
+# from scratch. The latter is a flawed and unsustainable approach (the more migrations
+# you'll amass, the slower it'll run and the greater likelihood for issues).
+#
+# It's strongly recommended that you check this file into your version control system.
```
djkotowski: Remove generated comments if you don't need them.  Hound loves to report millions of these, and send you one email for each :)

overlord/features/bomb.feature
```diff
@@ -0,0 +1,21 @@
+Feature: Bomb
+  In order to activate bomb
+  As a super villain
+  I want to enter an activation code and deactivation code
+
+  Scenario: Bomb is booted
+    Given I am on the bomb page
+    Then I should see "Inactive" with id of textarea "#bomb_status"
```
djkotowski: This knows too much about the implementation.  Does it really matter what the ID is?  Does the person reviewing these tests care about the implementation?

overlord/features/bomb.feature
```diff
@@ -0,0 +1,21 @@
...
+
+  Scenario: Bomb is booted with no activation code configured
+    Given I am on the bomb page
+    And "#bomb_status" says "Inactive"
```
djkotowski: Ditto above.

overlord/features/bomb.feature
```diff
@@ -0,0 +1,21 @@
...
+    And "activation_code" is not configured
```
djkotowski: `And the activation code is not configured` would read better in this case.  Too much knowledge about the implementation.  Let your step definitions handle this.

overlord/features/bomb.feature
```diff
@@ -0,0 +1,21 @@
+Feature: Bomb
```
djkotowski: This feature doesn't encapsulate the whole bomb.  It is only concerned with the _configuration_ of the bomb.

overlord/features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,228 @@
+require 'uri'
```
djkotowski: Get rid of `web_steps`.  They encourage poor testing practices.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,37 @@
...
+Given(/^"(.*?)" is not configured$/) do |arg1|
+  Bomb.create(activation_code: "1234")
+  visit "/bomb"
+  %w(1 2 3 4).each do |num|
+    find(:xpath, "//span[@class='"+num+"']").click
+  end
+end
+
+Given(/^"(.*?)" is configured to "(.*?)"$/) do |arg1, arg2|
+  visit "/bomb"
```
djkotowski: `arg1` is never used.  Make it a non-capturing group.

djkotowski: Also, give `arg2` a better name.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,37 @@
+Given(/^"(.*?)" says "(.*?)"$/) do |selector_id, status_text|
+  selector_id.should have_content(status_text)
+end
+
+Given(/^"(.*?)" is not configured$/) do |arg1|
```
djkotowski: Make `arg1` a non-capturing group (eliminating the need for a parameter list).

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,37 @@
...
+end
+
+Given(/^"(.*?)" is configured to "(.*?)"$/) do |arg1, arg2|
+  visit "/bomb"
+  arg2.each_char do |num|
+    find(:xpath, "//span[@class='"+num+"']").click
+  end
+end
+
+When(/^I enter right "(.*?)"$/) do |arg1|
```
djkotowski: Ditto arg naming.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,37 @@
...
+  visit "/bomb"
+  %w(1 2 3 4).each do |num|
+    find(:xpath, "//span[@class='"+num+"']").click
+  end
+end
+
+Then(/^"(.*?)" should contain "(.*?)"$/) do |arg1, arg2|
```
djkotowski: Ditto arg naming.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,37 @@
...
+  visit "/bomb"
+  arg2.each_char do |num|
+    find(:xpath, "//span[@class='"+num+"']").click
+  end
+end
+
+Then /^(?:|I )should see "([^\"]*)"(?: with id of textarea "([^\"]*)")?$/ \
```
djkotowski: This is too complex of a regex.  Simplify.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,37 @@
...
+do |text, selector|
+  selector.should have_content("Inactive")
```
djkotowski: Don't use "should" syntax.  Use the newer "expect".

overlord/public/js/bomb.js
```diff
@@ -0,0 +1,82 @@
+// Get all the keys from document
```
djkotowski: You have several large JS files and in-document scripts that go through each of the keys and operate on them.  I would look at a way to consolidate this logic.

overlord/public/js/bomb.js
```diff
@@ -0,0 +1,82 @@
...
+var keys = document.querySelectorAll('#bomb span');
+var operators = ['+', '-', 'x', '÷'];
+var decimalAdded = false;
+
+// Add onclick event to all the keys and perform operations
+for(var i = 0; i < keys.length; i++) {
+  keys[i].onclick = function(e) {
+    alert("hi");
```
djkotowski: Debugging code?

overlord/public/js/bomb.js
djkotowski: What is the difference between this and `public/bomb.js`?

overlord/spec/models/bomb_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  it "should have default deactivation code" do
+    expect(bomb.deactivation_code).to eq("0000")
+  end
+
+  it "should have activation code configurable" do
+    expect(valid_bomb.activation_code).to eq("12342")
+  end
+
+  it "should have default deactivation code" do
```
djkotowski: This is duplicate text from 15.  Do you mean "has deactivation code configurable"?

overlord/spec/models/bomb_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+describe 'bomb' do
+  let(:bomb) { Bomb.create }
+
+  let (:valid_bomb) { Bomb.create(activation_code: '12342',
+                        deactivation_code: '10220',
+                        detonation_time: '55') }
+
+
+  it "should have default activation code" do
```
djkotowski: Remove "should" from your `it` statements.

overlord/spec/overlord_spec.rb
```diff
@@ -0,0 +1,8 @@
+require_relative 'spec_helper'
+
+describe 'overlord' do
+  it 'should allow accessing the home page' do
```
djkotowski: How useful is this test?

overlord/spec/routes/bomb_spec.rb
```diff
@@ -0,0 +1,140 @@
+require_relative '../spec_helper'
+require 'timecop'
+
+describe 'bomb' do
```
djkotowski: This belongs with the integration tests, not the unit tests.

djkotowski:
I didn't see any specs for `BombHelpers`.

djkotowski:
This is looking like a good start.  It looks like you basically tried to get Rails in Sinatra.  Lot of work!

Work on moving logic to the correct classes/locations.  Take a look at our comments and refactor.

overlord/app/models/wire.rb
```diff
@@ -0,0 +1,6 @@
+require 'dm-types'
+require 'dm-validations'
+
+class Wire < ActiveRecord::Base
+  belongs_to :bomb
+end
```
jmmastey: This is very generalized, but it's probably too much abstraction for such a simple problem. In a real app, I'd try to limit the number of records I had to create for one entry.

overlord/app/overlord.rb
```diff
@@ -0,0 +1,37 @@
...
+set :database, adapter: "sqlite3", database: "detonation_device"
+set :views, "#{File.dirname(__FILE__)}/views"
+
+class Overlord < Sinatra::Application
+  include OverlordHelpers
+  configure :development do
+    DataMapper::Logger.new($stdout, :debug)
+  end
+  get '/' do
+    @bombs = Bomb.all
```
jmmastey: This would be overkill if there were multiple people using the app, fwiw. Why load multiple bombs?

overlord/app/overlord.rb
jmmastey: This is true. Production apps need to be runnable on any port.

overlord/app/overlord_helpers.rb
jmmastey: The reasoning here is that the abstraction is only saving you from two methods in the main class. The extra complexity isn't particularly worth it.

overlord/app/overlord_helpers.rb
```diff
@@ -0,0 +1,9 @@
...
+    session[:start_time] ||= (Time.now).to_s
+  end
+
+  def base_url
+    @base_url ||= "#{request.env['rack.url_scheme']}://#{request.env['HTTP_HOST']}"
```
jmmastey: Why do this? Also why memoize it? Usually, you should be able to use relative paths to ignore all of this. If not those, then you probably still don't need to memoize this value. A single string concatenation isn't going to save you much, especially relative to retrieving objects from the database.

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,103 @@
...
+    BombHelpers.explode_bomb(@bomb)
+
+    respond_to do |format|
+      format.json { @bomb.to_json }
+      format.html { haml :bomb }
+    end
+  end
+
+  post '/bomb/diffuse' do
+    params = {}
```
jmmastey: Yep, delete this. You're assigning them on the next line anyway. Also, Sinatra has facilities for parsing parameters. Any reason you're doing it this way?

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,103 @@
...
+    params = JSON.parse(request.body.read)
+    @bomb = Bomb.where("id = ?", params["bomb_id"]).first
```
jmmastey: Since you have ActiveRecord: `Bomb.find(params["bomb_id"])`

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,103 @@
...
+    wire = Wire.where(color: params["color"]).first
```
jmmastey: Does this mean there are only a few wires in the entire database? This is a confusing way to model the domain.

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,103 @@
...
+    @bomb.save!
+
+    haml :bomb
+  end
+
+  post '/bomb' do
+    request_json = {}
+    if request.body.try("string") != ""
+      request_json = JSON.parse(request.body.read)
+    end
```
jmmastey: dan's correct. also, this is a good candidate for a helper method, since it's not central to the action of this method.

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,103 @@
...
+    request_json["wires"].each do |wire_options|
+      wire = bomb.wires.build(wire_options)
+      wire.save
+    end
+
+    bomb.to_json
+  end
+
+  post '/bomb/activate' do
+    request.body
```
jmmastey: What's the purpose here?

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    request_json = JSON.parse(request.body.read)
+    bomb = Bomb.where("id = ?", request_json["bomb_id"]).first
+    if bomb.match_activation_code?(request_json)
+      bomb.active!
+      bomb.activated_time = Time.now
+      bomb.failed_attempts = 0
+      bomb.save!
+    end
+    BombHelpers.explode_bomb(bomb)
```
jmmastey: This is deceptive. It looks like the bomb should explode no matter what?

overlord/app/views/bomb.haml
```diff
@@ -0,0 +1,48 @@
...
+%script{:src => "/jquery-1.11.1.min.js"}
+%script{:src => "/jquery-ui.min.js"}
+%link{:href => "https://code.jquery.com/ui/1.11.1/themes/smoothness/jquery-ui.css", :rel => "stylesheet"}/
+
+%script{:src => "/bomb.js", :type => "text/javascript"}
+%h4 This is the interface
+
+%input{type: "hidden", value: "#{@bomb.id}", id: "bomb_id"}
+
+#dialog-confirm{:style => "display: none;", :title => "Activate the bomb?"}
```
jmmastey: Make sure to do CSS in a separate file, not inline. It's hard to maintain or debug inline CSS, so it's a big antipattern.

overlord/db/schema.rb
jmmastey: I think it'll probably regenerate the comments whenever it updates the schema. You can edit the rubocop config to ignore this file safely, since you never manually update it.

overlord/features/bomb.feature
jmmastey: E.g. the feature descriptions should be narrowly scoped so that people know what to expect when reading the tests

overlord/features/bomb.feature
```diff
@@ -0,0 +1,24 @@
...
+  In order to activate bomb
+  As a super villain
+  I want to enter an activation code and deactivation code
+
+  Background: A bomb exists
+    Given a bomb
+
+  Scenario: Bomb is booted
+    Given I am looking at the page bomb
+    Then I should see "Inactive" with id of textarea "#bomb_status"
```
jmmastey: Specifying a textarea is somewhat brittle. I'd rather see `The bomb status is "Inactive"`. (the status can still be checked via `#bomb_status` in the `step_definitions.rb` file.

overlord/features/bomb.feature
```diff
@@ -0,0 +1,24 @@
...
+    And "#bomb_status" says "Inactive"
+    And "activation_code" is not configured
+    When I enter right "activation_code"
+    Then I should see "Active" within "#bomb_status"
+
+  Scenario: Bomb is booted with activation code configured
+    Given I am looking at the page bomb
+    And "activation_code" is configured to "2345"
+    Then "activation_code" should contain "2345"
+    Then I should see "Inactive" within "#bomb_status"
```
jmmastey: What happens if you enter an erroneous activation code?

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,47 @@
...
+
+Given(/^I am looking at the page bomb$/) do
+  visit (current_url + "/" + @bomb.id.to_s)
+end
+
+Given(/^"(.*?)" is not configured$/) do |arg1|
+  Bomb.create(activation_code: "1234")
+  visit "/bomb"+ "/" + @bomb.id.to_s
+  %w(1 2 3 4).each do |num|
+    find(:xpath, "//span[@class='"+num+"']").click
```
jmmastey: This is fine for here, but you should try to use CSS selectors and not XPath. The latter is way harder to understand. Also, just throw a `def click_number` up top and this will be much easier to read.

overlord/public/bomb.js
```diff
@@ -0,0 +1,220 @@
...
+      input.innerHTML = '';
+      decimalAdded = false;
+    }
+
+    // If eval key is pressed, calculate and display the result
+    else if(btnVal == '=') {
+      var equation = inputVal;
+      var lastChar = equation[equation.length - 1];
+
+      // Replace all instances of x and ÷ with * and / respectively. This can be done easily using regex and the 'g' tag which will replace all instances of the matched character/substring
```
jmmastey: this is a good example of unnecessary comments. though I _am_ confused as to why we're doing this equation, you've basically reiterated the line of code below in English terms, which is not helpful for understanding what the code is actually doing.

overlord/public/bomb.js
```diff
@@ -0,0 +1,220 @@
...
+        input.innerHTML += btnVal;
+
+      // Replace the last operator (if exists) with the newly pressed operator
+      if(operators.indexOf(lastChar) > -1 && inputVal.length > 1) {
+        // Here, '.' matches any character while $ denotes the end of string, so anything (will be an operator in this case) at the end of string will get replaced by new operator
+        input.innerHTML = inputVal.replace(/.$/, btnVal);
+      }
+
+      decimalAdded =false;
+    }
```
jmmastey: I'm still quite confused as to what we're trying to accomplish here?

Also, you may want to use jQuery to simplify most of this code in a cross-browser way.

overlord/public/bomb.js
```diff
@@ -0,0 +1,220 @@
...
+
+      // Allow minus if the string is empty
+      else if(inputVal == '' && btnVal == '-')
+        input.innerHTML += btnVal;
+
+      // Replace the last operator (if exists) with the newly pressed operator
+      if(operators.indexOf(lastChar) > -1 && inputVal.length > 1) {
+        // Here, '.' matches any character while $ denotes the end of string, so anything (will be an operator in this case) at the end of string will get replaced by new operator
+        input.innerHTML = inputVal.replace(/.$/, btnVal);
+      }
```
jmmastey: This is a triple-nested conditional, within (I think) the third elsif. Think back to arrowhead and find a way to organize this better.

overlord/public/bomb.js
```diff
@@ -0,0 +1,220 @@
...
+            configureBomb();
+          }
+          else if(btnVal == 'C') {
+            input.innerHTML = '';
+          }
+          else {
+            input.innerHTML += btnVal;
+          }
+        }
+      }
```
jmmastey: Same as above, this is a tangle of javascript. Abstract this stuff into methods. There's even a confusing indentation problem on #110.

overlord/public/bomb.js
```diff
@@ -0,0 +1,220 @@
...
+      var start = 0;
+      var refreshId = window.setInterval(function(){
+        $.ajax({
+          type: "GET",
+          url: "http://localhost:9292/bomb/"+$("#bomb_id").val(),
```
jmmastey: Use relative paths so that you don't have to dictate hostname and port.

overlord/public/bomb.js
```diff
@@ -0,0 +1,220 @@
...
+              $('.timer').css("display", "block");
+              $(".bomb_wires").css("display", "block");
+              $("#bomb_status").removeClass("bomb_inactive");
+              $("#bomb_status").removeClass("bomb_explode");
+              $("#bomb_status").addClass("bomb_active");
+              $("#bomb_status").val("Active");
+              $('.timer').html(((returnObject.detonation_time) - start) + " Seconds ");
+              var html = ""
+              var str = ""
+              var btn_str = ""
```
jmmastey: Why initialize three locals, make them empty, and then ignore them?

overlord/public/js/bomb.js
jmmastey: Keep in mind that with jQuery, you can do `$("#bomb span").click(whatever)` and it will tag each element.

overlord/spec/routes/bomb_spec.rb
jmmastey: This is technically actually a route test, which is definitely not the right one. The integration tests are definitely the right place for this stuff.

jmmastey:
So there are definite fixes to make to the backend, but as a testing assignment, the biggest thing you should do is flesh out the integration tests. Right now they cover very little of the app.

Please take another stab.

overlord/db/schema.rb
djkotowski: This is a better idea.

PR #447, opened 2015-01-21
Add shopping cart

sriking1783:
2nd PR for shopping cart


ashtonthomas:
@sriking1783, this is real nice. I looked at the previous reviews and comments and it seems like a really nice upgrade.

LGTM

PR #449, opened 2015-01-21
Add data science

sriking1783:
Data science 2nd revision


data_science/calculator.rb
```diff
@@ -0,0 +1,110 @@
...
+                              18.48, 20.09, 21.67, 23.21],
+                    0.05 =>  [7.88, 10.6, 12.84, 14.86, 16.75, 13.55,
+                              20.28, 21.95, 23.59, 25.19],
+                    0.025 => [9.14, 11.98, 14.32, 16.42, 18.39, 20.25,
+                              22.04, 23.77, 25.46, 27.11],
+                    0.01 =>  [10.83, 13.82, 16.27, 18.47, 20.51,
+                              22.46, 24.32, 26.12, 27.83, 29.59],
+                    0.005 => [12.12, 15.2, 17.73, 20, 22.11, 24.1,
+                              26.02, 27.87, 29.67, 31.42] }
+
```
zsyed1: You can get this table from the gem `statistics2` using the method `Statistics2.chi2dist(deg_of_freedom, value)`

data_science/calculator.rb
```diff
@@ -0,0 +1,110 @@
...
+  SIGNIFICANT_PERCENTAGE = 5
+
+  def initialize(file_name = "source_data.json")
+    @file = file_name
+  end
+
+  def experiment
+    @experiment ||= Experiment.new(@file)
+  end
```
zsyed1: You only use `@file` to create an experiment. And the other code in this class uses experiment. So it probably makes sense to replace line 32 with `@experiment = Experiment.new(file_name)` in the constructor itself and lose the @file. You can then add `attr_reader experiment` and remove lines 35-37.

data_science/calculator.rb
```diff
@@ -0,0 +1,110 @@
...
+      conversions_squared  = chi_squared_numerator(cohort.conversions, e_conv)
+      failures_squared     = chi_squared_numerator(cohort.failures, e_failed)
+      chi_squares[cohort.name]  = (conversions_squared + failures_squared).round(2)
+    end
+  end
+
+  def chi_squared_numerator(observed, expected)
+    (observed - expected)**2 / expected
+  end
+
```
zsyed1: The method `chi_squared_numerator` should be private.

data_science/calculator.rb
```diff
@@ -0,0 +1,110 @@
...
+  def next_higher_value
+    percentages[previous_percentages_index]
+  end
+
+  def p_value_matched?(dof)
+    index = degrees_of_freedom - 1
+    dof[index] >= chi_square
+  end
+end
```
zsyed1: The class `Experiment` has many public methods. This is fine if you really want to expose these methods, but then you need to have tests for each method. So you have two options:

1) Add a test for every method in `Calculator`.
2) Make more of the methods in `Calculator` private.

For example, the task only asks that you compute `chi_square` and it's `p-value`. So you don't really need to display `chi_squared_for_experiments`. Please hit me up on Slack if you want to discuss. I recommend option 2.

data_science/cohort.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def initialize(data)
+    @name = data.keys.first
+    @visits = data[@name]["total_visits"]
+    @conversions = data[@name]["conversions"]
+  end
+
+  def conversion_rate
+    (conversions.to_f / visits).round(4)
+  end
```
zsyed1: What if `visits == 0`?

data_science/cohort.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def failures
+    visits - conversions
+  end
+
```
zsyed1: Need to have tests for all public methods. So `cohort_spec.rb` should contain a test for the method `failures`.

data_science/calculator.rb
```diff
@@ -0,0 +1,111 @@
...
+  def p_value
+    [p_value_index, next_higher_value]
+  end
+
+  def winner
+    conversions = {}
+    experiment.cohorts.each do |cohort|
+      conversions[cohort.name] = cohort.conversion_rate
+    end
+    conversions.select { |_cohort, conv| conv == conversions.values.max }.keys.first
```
zsyed1: I think you can use the method `find` here instead of `select/first`.

data_science/calculator.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  def degrees_of_freedom
+    experiment.cohorts.count - 1
+  end
+
+  def p_value_index
+    P_VALUE_TABLE.select do |_percent, dof|
+      p_value_matched?(dof)
+    end.keys.first
+  end
```
zsyed1: Please see comment above about using the `statistics2` gem

data_science/calculator.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  private
+
+  def percentages
+    P_VALUE_TABLE.keys
+  end
+
+  def previous_percentages_index
+    return 0 if  percentages.index(p_value_index)
```
zsyed1: Extra space after the `if` in line 98

data_science/calculator.rb
zsyed1: You've written some clever code to get the p-value out of the table. If you use the statistics2 gem you can remove most of this.

data_science/cohort.rb
```diff
@@ -0,0 +1,32 @@
...
+  def standard_error
+    Math.sqrt((conversion_rate * (1 - conversion_rate)) / visits).round(4)
+  end
+
+  def expected_conversion_rate
+    sample_mean = conversion_rate * 100
+    { max: (sample_mean + (STANDARD_DEVIATION * standard_error)).round(2),
+      min: (sample_mean - (STANDARD_DEVIATION * standard_error)).round(2) }
+  end
```
zsyed1: The repetition in lines 29 and 30 looks pretty. You may want to move `STANDARD_DEVIATION * standard_error` into a separate method.

data_science/runner.rb
```diff
@@ -0,0 +1,33 @@
...
+
+print "The total conversions of the experiment is "
+puts Experiment.new("source_data.json").total_conversions.to_s
+
+print "The expected conversions of the experiment is "
+puts Experiment.new("source_data.json").expected_conversions
+
+print "The chi_square for cohorts is "
+puts Calculator.new.chi_squared_for_experiments.to_s
+
```
zsyed1: Probably don't need to display this.

data_science/experiment.rb
```diff
@@ -0,0 +1,47 @@
+require_relative 'json_parser'
+require_relative 'cohort'
+
+class Experiment
+  attr_accessor :cohorts
+
+  def initialize(file_name)
+    data = JsonParser.parse(File.open(file_name, "r"))
+    @cohorts = data.map { |key, value| Cohort.new(key => value) }
+  end
```
zsyed1: Instead of passing in a filename to `Experiment.new`, it probably makes more sense to pass in a data object. So in your application `JsonParser` parses the file and produces a `data` object. You then pass `data` into the constructor as `Experiment.new(data)`. This will more cleanly separate the responsibilities of parsing and processing.

This will also remove the need creating and destroying a file in your unit test below. Instead, in `experiment_spec.rb` you would fashion a data object and feed it directly to your class. Let me know if any of this is unclear.

data_science/spec/calculator_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+                { date: "2015-01-03", cohort: "A", result: 0 },
+                { date: "2015-01-03", cohort: "B", result: 0 },
+                { date: "2015-01-03", cohort: "B", result: 1 },
+                { date: "2015-01-03", cohort: "B", result: 1 },
+                { date: "2015-01-03", cohort: "A", result: 1 },
+                { date: "2015-01-03", cohort: "B", result: 1 },
+                { date: "2015-01-03", cohort: "B", result: 1 }]
+    file.write(raw_data.map { |o| Hash[o.each_pair.to_a] }.to_json)
+    file.rewind
+  end
```
zsyed1: See note above about constructing `Experiment` off of a data object rather than a file name.

data_science/calculator.rb
zsyed1: Have `Calculator` construct off of a data object rather than a file name. Please see other notes about this - sorry for scattering this comment in multiple places.

data_science/spec/calculator_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'spec_helper'
+
+describe Calculator do
+  let(:calculator) { described_class.new("test_data.json") }
+  before do
+    file     = File.open("test_data.json", "w")
+    raw_data = [{ date: "2015-01-01", cohort: "A", result: 1 },
```
zsyed1: No need for extra spaces after the `file` in line 6.

sriking1783:
@zsyed1 : Hey I agree I should have used gem, but already I have invested so much time here, I don't want to go to the gem. I will address the other comments that you mentioned. As long as its doing fine even without using gem, do you think I have to redo it?
Thoughts?

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+                { date: "2015-01-02", cohort: "B", result: 1 },
+                { date: "2015-01-03", cohort: "A", result: 1 },
+                { date: "2015-01-03", cohort: "A", result: 0 },
+                { date: "2015-01-03", cohort: "B", result: 0 },
+                { date: "2015-01-03", cohort: "B", result: 1 },
+                { date: "2015-01-03", cohort: "B", result: 1 },
+                { date: "2015-01-03", cohort: "A", result: 1 },
+                { date: "2015-01-03", cohort: "B", result: 1 },
+                { date: "2015-01-03", cohort: "B", result: 1 }]
+    file.write(raw_data.map { |o| Hash[o.each_pair.to_a] }.to_json)
```
zsyed1: This `raw_data` seems to be repeated in three places. So it makes sense to put this into a method in a helper file say `data_helper.rb`. That way if you have to change it, you only need to change it in one place. For example:

`def raw_data
  [{ date: "2015-01-01", cohort: "A", result: 1 },
...
}]
end`

data_science/spec/json_parser_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      expect { JsonParser.parse("bogus file") }.to raise_error ArgumentError
+    end
+
+    it "errors when json is invalid" do
+      expect { JsonParser.parse(invalid_file) }.to raise_error JSON::ParserError
+    end
+
+    it "parses json and stores results in to hash with date as keys" do
+      data_json = described_class.parse(file)
+      expect(data_json["B"]["total_visits"]).to eq(6)
```
zsyed1: You can put these variables into a `let` statement like: `let(:data_json) { described_class.parse(file) }` and move them up to into the `describe` section.

zsyed1: Overall you've done a good job of implementing the features. One policy we follow in testing is that _every_ public method of a class/module must be tested. So again you have two options. Either make those methods private _or_ add a test. I would also recommend changing `Experiment` and `Calculator` so that they construct off a data object instead of file name.

zsyed1: @sriking1783 I would recommend fixing the test-coverage issue and resubmitting. That is a more serious issue than the other comments. Please see above comment.

data_science/spec/cohort_spec.rb
djkotowski: You could also place it in a secondary JSON file (e.g. `test_data.json`) to be used by your tests.

djkotowski:
This is a good start.  Take a look at the comments and refactor.

PR #451, opened 2015-01-21
refactor arrowhead code

katieschilling:


arrowhead/arrowhead.rb
```diff
@@ -17,17 +17,20 @@ class Arrowhead
...
-      raise "Unknown region, please provide a valid region."
-    end
+    shapes = CLASSIFICATIONS[region]
+    raise "Unknown region, please provide a valid region." unless shapes
+    identify_shape(shapes, shape)
+  end
+
+  def self.identify_shape(shapes, type)
+    shape = shapes[type]
+    shape_error unless shape
```
incazteca: Is there a specific reason you moved the raise into a separate method? Keep the error message close to the actual error.

katieschilling: Line too long error. Moved it at polish dan's suggestion.

incazteca:
@katieschilling Looks good I would probably remove the shape_error method as the error being thrown is not the one you created but a different error ```arrowhead.rb:27:in `identify_shape': undefined local variable or method `shape_error' for Arrowhead:Class (NameError)```

arrowhead/arrowhead.rb
djkotowski: I don't think this is too bad given `shape_error` is near `identify_shape`.  Another option would be to declare a constant with the long error message.

arrowhead/arrowhead.rb
```diff
@@ -17,17 +17,20 @@ class Arrowhead
...
+    "You have a(n) '#{shape}' arrowhead. Probably priceless."
+  end
+
+  def shape_error
```
djkotowski: Make this a class method.

djkotowski:
This looks good.  Take a look at the comments, but no need for a resubmission.  Nice work!

PR #453, opened 2015-01-21
Data

cnorrisjr:
Refactor based on Ashton's suggestions.


data_science/data_science.rb
```diff
@@ -0,0 +1,22 @@
+require_relative 'Data_loader'
+require_relative 'sample'
+
+class DataScience
+  my_data = DataLoader.new
+  my_data.read_in_json
```
zsyed1: Don't use the prefix `my_` in variables. After all, whose else would it belong to? :-)

data_science/data_loader.rb
```diff
@@ -0,0 +1,47 @@
...
+  def message
+    "Your file could not be found"
+  end
+end
+
+require 'json'
+class DataLoader
+  def initialize(file_name = 'source_data.json')
+    @file_name = file_name
+  end
```
zsyed1: Why not call `validate` and `build_data` right away. Then you don't need to store `@file_name` and clients of this class will not have to invoke multiple methods. They just call `DataLoader.new(file_name)` and are done. You can also move `@all_in = []` from line 20 into the constructor at line 16.

data_science/data_loader.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def build_data
+    @all_in = []
+    @parsed.each do |row|
```
zsyed1: I would also suggest renaming `@all_in` to something like `@all_traffic`.

data_science/data_loader.rb
zsyed1: I meant call `read_in_json` right away. That in turn calls `validate`...

data_science/sample.rb
```diff
@@ -0,0 +1,75 @@
...
+    @values[@key] = { num_convs: num_convs, vistors: vistors }
+    { low: con_l, high: con_h }
+  end
+
+  def print(range, confidence, name)
+    puts "Name:                     #{name} "
+    puts "Conversions:              #{range[:num_convs]}"
+    puts "Sample Size:              #{range[:vistors]}"
+    percent = Sample.calc_percent(range[:num_convs], range[:vistors])
+    puts "Percent:                  #{percent}%"
```
zsyed1: In line 27 you can remove the `Sample.` I think.

data_science/sample.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def initialize(group, values)
+    @group = group
+    @values = values
+    @key = @group[0].cohort.to_sym
+  end
+
+  def calc_values
+    num_convs, vistors = Sample.find_conversions(@group)
+    con_l, con_h = Sample.find_confidence(num_convs, vistors, Sample.percent)
```
zsyed1: Calling `visitors` something like `num_visitors` would make this more readable. Just nitpicking.

data_science/sample.rb
```diff
@@ -0,0 +1,75 @@
...
+    puts "Chi Score:                #{chi_percent}%"
+  end
+
+  def self.number_of_conversions(group_in)
+    total = 0
+    group_in.each do |row|
+      total += 1 if row.result == 1
+    end
+    total
+  end
```
zsyed1: An shorter way would be replace lines 52-56 with the single line: `group_in.count{ |row| row.result == 1 }`

data_science/spec/dataloader_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  it "Data Objects are built" do
+    my_data = DataLoader.new("../smalldata.json")
+    my_data.read_in_json
+    expect(my_data.build_data).to eql(
+    [{ "date" => "2014-03-20", "cohort" => "B", "result" => 0 },
+     { "date" => "2014-03-20", "cohort" => "B", "result" => 1 },
+     { "date" => "2014-03-20", "cohort" => "A", "result" => 0 },
+     { "date" => "2014-03-20", "cohort" => "A", "result" => 1 }])
+  end
```
zsyed1: You can move this data structure into a `let` statement so that you don't have to repeat it twice.

data_science/spec/dataloader_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  it "Sample json data is seperated into groups"do
+    my_data = DataLoader.new("../smalldata.json")
+    my_data.read_in_json
+    my_data.build_data
+    group1, group2 = my_data.seperate_groups
+    expect(group1).not_to be_empty
+    expect(group1.size).to be(2)
+    expect(group2).not_to be_empty
```
zsyed1: You can use `expect(group1).to have(2).items`

data_science/spec/sample_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    sample_a = Sample.new(group_a, @values)
+    @confidence_a = sample_a.calc_values
+  end
+
+  it "Calculate_values function is working, using group_a's data" do
+    expect(@confidence_a[:low]).to eql(-19.3)
+    expect(@confidence_a[:high]).to eql(119.3)
+    expect(@values[:A][:num_convs]).to eql(1)
+    expect(@values[:A][:vistors]).to eql(2)
+  end
```
zsyed1: You can use `eq` here instead of `eql`.

data_science/traffic.rb
```diff
@@ -0,0 +1,13 @@
...
+  def initialize(options = {})
+    @date   = options[:date]
+    @cohort = options[:cohort]
+    @result = options[:result]
+  end
+
+  def to_s
+    "#{@date} #{@cohort} #{@result}"
+  end
+end
```
zsyed1: I like your `Traffic` class that encapsulates a data point.

data_science/spec/sample_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  before(:each) do
+    a = DataLoader.new("../smalldata.json")
+    a.read_in_json
+    a.build_data
+    group_a, _groupb = a.seperate_groups
+    @values = {}
+    sample_a = Sample.new(group_a, @values)
+    @confidence_a = sample_a.calc_values
+  end
```
zsyed1: I would split this into two `let` statements: one for `values` and one for `confidence_a`.

data_science/traffic.rb
zsyed1: Nice job overall. I would recommend closing this.

data_science/sample.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def initialize(group, values)
+    @group = group
+    @values = values
+    @key = @group[0].cohort.to_sym
+  end
+
+  def calc_values
+    num_convs, vistors = Sample.find_conversions(@group)
```
djkotowski: Spelling.

data_science/sample.rb
djkotowski: Especially since conversion counts use the `num_` prefix.

data_science/sample.rb
djkotowski: `self.class` would be more robust.

data_science/spec/dataloader_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  it "File is not valid and a NotAJSONFileError comes up" do
+    no_json = "../smalldata.jsn"
+    my_data = DataLoader.new(no_json)
+    expect { my_data.read_in_json }.to raise_error(NotAJSONFileError)
+  end
+
+  it "File is not valid and a FileDoesNotExistError Error comes up" do
+    no_file = "../smalldat.json"
```
djkotowski: I would recommend naming these a bit more different.  It's confusing to glance over these filenames when the differences are so small (e.g. `small_data.jsn` vs `small_data.json`)

data_science/spec/dataloader_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    expect(data_in).to eql(
+    [{ "date" => "2014-03-20", "cohort" => "B", "result" => 0 },
+     { "date" => "2014-03-20", "cohort" => "B", "result" => 1 },
+     { "date" => "2014-03-20", "cohort" => "A", "result" => 0 },
+     { "date" => "2014-03-20", "cohort" => "A", "result" => 1 }])
+  end
+
+  it "File is not valid and a NotAJSONFileError comes up" do
+    no_json = "../smalldata.jsn"
+    my_data = DataLoader.new(no_json)
```
djkotowski: I could see easily merging 18 and 19 (and 24 + 25).  I don't think splitting this into two lines is very helpful in this case.

data_science/spec/dataloader_spec.rb
```diff
@@ -0,0 +1,50 @@
+require_relative '../data_loader'
+require_relative '../traffic'
+
+describe DataLoader do
+
+  it "File should be valid and will be read in" do
```
djkotowski: These should read as English (and describe DataLoader, not File).

data_science/spec/dataloader_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    [{ "date" => "2014-03-20", "cohort" => "B", "result" => 0 },
+     { "date" => "2014-03-20", "cohort" => "B", "result" => 1 },
+     { "date" => "2014-03-20", "cohort" => "A", "result" => 0 },
+     { "date" => "2014-03-20", "cohort" => "A", "result" => 1 }])
+  end
+
+  it "Sample json data is seperated into groups"do
+    my_data = DataLoader.new("../smalldata.json")
+    my_data.read_in_json
+    my_data.build_data
```
djkotowski: This is more of a procedural approach.  Why not read in JSON and build on initialization?

data_science/sample.rb
```diff
@@ -0,0 +1,75 @@
...
+    con_l, con_h = Sample.find_confidence(num_convs, vistors, Sample.percent)
+    @values[@key] = { num_convs: num_convs, vistors: vistors }
+    { low: con_l, high: con_h }
+  end
+
+  def print(range, confidence, name)
```
djkotowski: Does printing to standard output belong in this class?

djkotowski:
This is looking good.  Take a look at the comments and refactor.  I don't think a resubmission is necessary, though.

PR #456, opened 2015-01-22
Rc05 robot name

magicalbanana:
@jmmastey please review.


robot_name/name_generator.rb
```diff
@@ -0,0 +1,32 @@
...
+    letters = ('A'..'Z').to_a
+    alpha_characters = randomizer(letters)
+    numeric_characters = randomizer(numbers)
+    name = alpha_characters.sample(2) + numeric_characters.sample(3)
+    name.join
+  end
+
+  private
+
+  def self.randomizer(characters)
```
djkotowski: Name implies that it returns a randomizer, not that it randomizes.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,32 @@
...
+    length = 10
```
djkotowski: You're generating 10 random characters each time.  This is a magic number that, if you end up needing > 10 characters, will cause this program to fail.  Generate what you need, no more, no less.

robot_name/validate_name.rb
```diff
@@ -0,0 +1,12 @@
+require 'set'
+
+class ValidateName
```
djkotowski: Given names are only for robots, why is this in a separate class?
Also, given that all the methods here are class methods, why isn't this a module?

robot_name/name_generator.rb
```diff
@@ -0,0 +1,32 @@
+require 'securerandom'
+
+class NameGenerator
+  def self.call
+    generate_name
+  end
+
+  private
+
+  # The reason I put this in private is because this type of method should not be accessible to other classes.
```
djkotowski: Class methods are not affected by `private`.

robot_name/name_generator.rb
magicalbanana: Yeah, should just be randomize.

robot_name/validate_name.rb
magicalbanana: Just habit of building a really big app. When building a really big app like rails you often have a seperate class for validations. In this case, there are two types of validations for the name, the regex and the existing of the name on the registry.

robot_name/name_generator.rb
magicalbanana: The length is a fail over when length was initially passable. The original code for generating a name was not like this, it was actually more complex but it could not produce the necessary data required to pass the regex validation. It should be length = characters.length or size

robot_name/name_generator.rb
magicalbanana: Yep. Forgot you need the private_class_method here.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,32 @@
+require 'securerandom'
+
+class NameGenerator
+  def self.call
+    generate_name
+  end
+
+  private
```
jmmastey: It's true. Perhaps unintuitively, class methods are not affected by private declarations. You could use `private_class_method`, or more likely move to using an instance. At that point, `self.call` just does `new.generate_name`.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,32 @@
...
+  def self.generate_name
+    # I am using SecureRandom here just for kicks and giggles.
+    # No really I am using it because it's a good method to generate
+    # random instances of a number. -svajone
+    numbers = ('0'..'9').to_a
+    letters = ('A'..'Z').to_a
+    alpha_characters = randomize(letters)
+    numeric_characters = randomize(numbers)
+    name = alpha_characters.sample(2) + numeric_characters.sample(3)
+    name.join
```
jmmastey: Eh, this is still excessive for simple rand generation. If the intent is short methods, this isn't particularly doing the job.

For instance, with one more level of abstraction:

```
def generate_name
  random_characters('0'..'9', count: 2) + random_characters('A'..'Z', count: 3)
end

def random_characters(sample_set, count: 3)
  sample = Array(sample_set).shuffle.pluck(count)
  sample.join('')
end

File: robot_name/robot_name.rb [hunk 7]
```diff
@@ -1,24 +1,20 @@
+require './name_generator.rb'
+require './validate_name.rb'
+
 class NameCollisionError < RuntimeError; end
+class NameRegexError < RuntimeError; end
```
jmmastey: Don't imply the implementation in the name. You're leaking details unnecessarily.

robot_name/robot_name.rb
```diff
@@ -1,24 +1,20 @@
...
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
-
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
+    name_generator = options[:name_generator] || NameGenerator
+    @name = name_generator.call
```
jmmastey: This I like. Good combination of the two methods of input.

robot_name/robot_name.rb
```diff
@@ -1,24 +1,20 @@
...
+    raise NameRegexError, "The name generated does not pass the Regex" unless ValidateName.match_regex?(@name)
+    raise NameCollisionError, "The name '#{@name}' already exists in the registry!" if ValidateName.name_exists?(@@registry, @name)
```
jmmastey: These length errors are actually a problem, just switch to the normal form of `if` and you'll be fine. Also, these two lines should probably be in a different method. Remember we want each method to be small and to have very few responsibilities.

robot_name/spec/robot_name/name_generator_spec.rb
```diff
@@ -0,0 +1,9 @@
+require './spec/spec_helper.rb'
+
+describe NameGenerator do
+  describe "#self.call" do
+    it "generates a name" do
+      expect(NameGenerator.call).to_not be_empty
```
jmmastey: It's not a brittle test, that much is true. Still, is there nothing else meaningful we can specify here?

robot_name/validate_name.rb
jmmastey: It's overkill in this situation, but it's true that in a large app, you'd start to need to split this stuff out. Just be wary of the tradeoffs of obscuring your implementation with abstractions.

robot_name/validate_name.rb
```diff
@@ -0,0 +1,12 @@
...
+  def self.name_exists?(registry, name)
+    set = Set.new(registry)
+    return true if set.include? name
```
jmmastey: Just let it implicitly return a boolean:

```
def self.name_exists?(registry, name)
  set = Set.new(registry)
  set.include? name
end
```

Seems like the additional creation of a set object might also be a lot of work for something that could be handled efficiently enough with a hash.

robot_name/validate_name.rb
```diff
@@ -0,0 +1,12 @@
...
+  end
+
+  def self.match_regex?(name)
+    return true unless (/[[:alpha:]]{2}[[:digit:]]{3}/ =~ name).nil?
```
jmmastey: Same as above, leverage implicit returns.

spec/spec_helper.rb
```diff
@@ -0,0 +1,82 @@
+require 'simplecov'
+SimpleCov.start
+
+require 'rspec'
+require 'pry'
+require 'faker'
+require './dino_catalog/dinosaur.rb'
```
jmmastey: Dinosaurs, huh? Clean your helpers :)

jmmastey:
Added a few notes inline. Take a look at them, make changes as necessary. No need to resubmit.

PR #457, opened 2015-01-23
Explain how the Five Second Rule for code review works.

magicalbanana:
@jmmastey  Five-Second rule sir. Please review.


jmmastey:
Yep, that works.

PR #458, opened 2015-01-23
Exercise: Knows why comments lie, and how to write self-documenting code

magicalbanana:
@jmmastey Exercise: Knows why comments lie, and how to write self-documenting code, please review.


jmmastey:
Yep, works for me! You're right that big programs are rarely self-documenting in their entirety. Still, look at stuff like http://blog.8thlight.com/uncle-bob/2011/09/30/Screaming-Architecture.html to see examples of how to get much closer to that reality.

PR #459, opened 2015-01-23
Exercise: Knows the cases where comments are really necessary

magicalbanana:
@jmmastey Exercise: Knows the cases where comments are really necessary, please review.


jmmastey:
The only other case I'd add is when you do something that seems non-optimal, but is actually required due to an inconsistency or hack in the libraries you're working with. Someone may attempt to refactor the code and re-introduce a subtle bug if not given sufficient warning.

PR #465, opened 2015-01-27
test datascience.rb and treat cohort.rb and parser.rb as private api

katieschilling:


data_science/cohort.rb
```diff
@@ -0,0 +1,14 @@
+class Cohort
+  def initialize(data, cohort_name)
+    @cohort_name   = cohort_name
+    @events = data.select { |event| event['cohort'] == "#{@cohort_name}" }
```
zsyed1: Remove extra spaces before equal sign (=) in line 3

data_science/parser.rb
```diff
@@ -0,0 +1,21 @@
+require 'json'
+
+class Parser
+  class << self
+    def parse(file)
```
zsyed1: Use `def self.parse(file)` notation for class methods instead of `class << self`.

data_science/datascience.rb
```diff
@@ -0,0 +1,71 @@
...
+  attr_reader :json_file
+
+  def initialize(json_file = 'source_data.json')
+    @json_file = json_file
+  end
+
+  def data
+    @data ||= Parser.parse(json_file)
+  end
+
```
zsyed1: I don't see a need for defining a default `json_file` here. I would suggest removing the default and removing lines 13-15. Instead move line 14 into the constructor as `@data = Parser.parse(json_file)`. You can then remove the `default_json_file` test below in your `data_science_spec.rb`.

data_science/spec/datascience_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+  end
+
+  describe '#data' do
+    let(:json_data) { { name: "test", description: "test" }.to_json }
+
+    before do
+      expect(experiment).to receive(:json_file).and_return('foo.json')
+      allow(File).to receive(:read).with('foo.json').and_return(json_data)
+    end
+
```
zsyed1: By passing parsed data into the `DataScience` constructor, you can avoid mocking this file i/o. Instead if you want to create fake data, you can just construct the data hash and pass it in.

data_science/datascience.rb
zsyed1: Please consider sending parsed data to the constructor instead of a file. You built a nice object to parse files. So you can separate the responsibility of parsing and processing. Parse the data first, then pass the data into `DataScience`. This will reduce your tests as well. Please see comment below in your `data_science_spec.rb`.

data_science/datascience.rb
```diff
@@ -0,0 +1,71 @@
...
+  def chi_sq
+    difference = tester.different?
+    chi = tester.chisquare_p.round(3)
```
zsyed1: In line 18, the name `difference` implies a quantity whereas the right side `different?` implies a boolean. Better to name the left variable `different` or `is_different`.

data_science/parser.rb
```diff
@@ -0,0 +1,21 @@
...
+      raise ArgumentError unless File.exist?(file)
+      JSON.parse(File.read(file))
+    end
+
+    def counts(data)
+      cohort_names(data).each_with_object({}) do |cohort_name, counts|
+        new_cohort = Cohort.new(data, cohort_name).count_events
```
zsyed1: In line 11 the variable `counts` has the same name as the method (line 10 `def counts`). You want to avoid this (called _shadowing_) by renaming the variable name. Here the variable name _shadows_ the method name.

data_science/cohort.rb
```diff
@@ -0,0 +1,14 @@
...
+  end
+
+  def count_events
+    @events.each_with_object(Hash.new { |h, k| h[k] = 0 }) do |event, counts|
+      counts["#{@cohort_name}_1"]     += 1 if event['result'] == 1
+      counts["#{@cohort_name}_0"]     += 1 if event['result'] == 0
+      counts["#{@cohort_name}_total"] += 1
+    end
```
zsyed1: The way you have used strings to distinguish the different values is nice. A better way to do this is to store the data hierarchically. That is:
```
  counts[@cohort_name][:success] += 1 if event['result'] == 1
  counts[@cohort_name][:failure] += 1 if event['result'] == 0
  counts[@cohort_name][:total] += 1
```

data_science/parser.rb
```diff
@@ -0,0 +1,21 @@
...
+        counts.merge!(new_cohort)
+      end
+    end
+
+    def cohort_names(data)
+      data.map { |event| event['cohort'] }.uniq
```
zsyed1: Since you are referring to each data item as `event`, a better name for `data` would be `events`. If you chose to rename it you would have to replace it everywhere.

data_science/spec/datascience_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+      {
+        'date'   => Date.today.to_s,
+        'cohort' => i.even? ? 'A' : 'B',
+        'result' => i % 3 == 0 ? 0 : 1,
+      }
+    end
+  end
+  let(:mock_cohort_names) { %w(A B) }
+  let(:mock_counts) do
+    { "B_0" => 1000, "B_total" => 1500, "B_1" => 500, "A_0" => 1000, "A_total" => 1500, "A_1" => 500 }
```
zsyed1: Please see my comments about passing in parsed-data into `DataScience` instead of filenames. That will remove the need for setting up mocks.

data_science/spec/datascience_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+    let(:json_data) { { name: "test", description: "test" }.to_json }
+
+    before do
+      expect(experiment).to receive(:json_file).and_return('foo.json')
+      allow(File).to receive(:read).with('foo.json').and_return(json_data)
+    end
+
+    context 'with a file specified' do
+      before { expect(File).to receive(:exist?).and_return(true) }
+
```
zsyed1: Did you want this to be an `allow` rather than an `expect`?

data_science/datascience.rb
```diff
@@ -0,0 +1,71 @@
...
+    [difference, chi]
+  end
```
zsyed1: I think it would be better to return a descriptive hash in line 20 instead of an array: `{difference: different, chi_square: chi}`. It makes it easier for clients to know what's being returned. Please see comment in `experiment_spec.rb` below.

data_science/spec/datascience_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+    end
+
+    it 'returns a boolean as the first value' do
+      expect(experiment.chi_sq[0]).to match boolean
+    end
+
+    it 'returns a number as the second value' do
+      expect(experiment.chi_sq[1]).to be_a Float
+    end
+
```
zsyed1: If `Experiment::chi_sq` returned a descriptive hash this would read a little better: `expect experiment.chi_sq[:different]).to match boolean`. In fact you probably wouldn't need this test as much.

data_science/spec/datascience_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+  describe '#conversion_rate' do
+    context 'with counts given for each cohort' do
+      before do
+        allow(experiment).to receive(:cohort_names).and_return(mock_cohort_names)
+        allow(experiment).to receive(:counts).and_return(mock_counts)
+      end
+
+      it 'returns a hash of expected conversion rates' do
+        expect(experiment.conversion_rate).to eq("A" => 0.333, "B" => 0.333)
```
zsyed1: This description should be `it 'returns the correct conversion rates' do`

data_science/parser.rb
```diff
@@ -0,0 +1,21 @@
...
+    def parse(file)
+      raise ArgumentError unless File.exist?(file)
+      JSON.parse(File.read(file))
+    end
+
+    def counts(data)
+      cohort_names(data).each_with_object({}) do |cohort_name, counts|
+        new_cohort = Cohort.new(data, cohort_name).count_events
+        counts.merge!(new_cohort)
+      end
```
zsyed1: Line 13 in conjunction with the way you defined `Cohort::count_events` is an interesting approach. Another way to have done this is to have methods: `Cohort::successes`, `Cohort::failures`, `Cohort::total` and then build up the hash in this method (`def counts`). I think your way is fine though.

data_science/spec/datascience_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+          bgroup: { :"1" => 5,  :"0" => 295 },
+        }
+      end
+
+      before do
+        expect(experiment).to receive(:values).and_return(mock_values)
+      end
+
+      it 'returns an expected chi sq value' do
+        expect(experiment.chi_sq).to eq [true, 0.023]
```
zsyed1: Line 83 `it 'returns the expected chi square value'`. Using the definite article _the_ implies that there is one correct answer. This is real nitpicking - sorry!

data_science/spec/datascience_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+      end
+    end
+
+    context 'with identically distributed data' do
+      before do
+        expect(experiment).to receive(:data).and_return(mock_data)
+      end
+
+      it 'returns max chi sq value' do
+        expect(experiment.chi_sq).to eq [false, 1.0]
```
zsyed1: A more succinct test is `it 'returns no winner' do` and you would only test the fact that `different` is `false`.

data_science/spec/datascience_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+      Array.new(1000) { { 'cohort' => 'B', 'result' => 0 } } +
+        Array.new(500)  { { 'cohort' => 'B', 'result' => 1 } } +
+        Array.new(1000) { { 'cohort' => 'A', 'result' => 0 } } +
+        Array.new(500)  { { 'cohort' => 'A', 'result' => 1 } }
+    end
+
+    before { expect(experiment).to receive(:data).and_return(mock_data) }
+
+    it 'returns a summary of the data' do
+      expect(experiment.counts).to eq("B_0" => 1000, "B_total" => 1500, "B_1" => 500, "A_0" => 1000, "A_total" => 1500, "A_1" => 500)
```
zsyed1: Nice job! I don't think this needs a re-submission, although @djkotowski may disagree. (A little good cop, bad cop)

PR #467, opened 2015-01-27
Add test plan for shopping cart

arrubin:
Based on initial requirements, I have developed an initial test plan for the
shopping cart.

### DIFF
--- a/shopping_cart/features/authentication.feature
+++ b/shopping_cart/features/authentication.feature
@@ -1,0 +1,17 @@
+Feature: authentication
+
+  As a customer
+  I want my cart to remain unchanged when I authenticate
+  So that I do not have to log in to shop
+
+  Scenario: authenticate with items in cart
+    Given my cart contains items
+    And I am not authenticated
+    When I login
+    Then the items appear in my cart
+
+  Scenario: log out with items in cart
+    Given my cart contains items
+    And I am authenticated
+    When I login
+    Then the items appear in my cart
--- a/shopping_cart/features/update_cart.feature
+++ b/shopping_cart/features/update_cart.feature
@@ -1,0 +1,20 @@
+Feature: update shopping cart
+
+  As a customer
+  I want to update my cart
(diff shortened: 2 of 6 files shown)


shopping_cart/features/authentication.feature
```diff
@@ -0,0 +1,17 @@
+Feature: authentication
+
+  As a customer
```
jvanderz22: I think you're missing some scenarios relating to login here. What happens if the username doesn't exist? Or the password is wrong?

shopping_cart/features/update_cart.feature
```diff
@@ -0,0 +1,20 @@
+Feature: update shopping cart
+
+  As a customer
```
jvanderz22: These scenarios look good, but I think there is room for a little wider coverage here. Are we sure that nothing happens to the other items in the cart if we make a change to one of them?

jvanderz22:
There's room for a more expansive test plan to be implemented here, but you do a good job covering the basic scenarios, including happy, sad, and bad paths. The features and step definitions are given descriptive names that make sense in plain english and all of the scenarios follow good cucumber syntax. 

Review the areas where additional tests would provide greater code coverage, but you don't have to resubmit this exercise.

shopping_cart/features/authentication.feature
```diff
@@ -0,0 +1,17 @@
...
+    Given my cart contains items
+    And I am not authenticated
+    When I login
+    Then the items appear in my cart
+
+  Scenario: log out with items in cart
+    Given my cart contains items
+    And I am authenticated
+    When I login
+    Then the items appear in my cart
```
djkotowski: What about merging carts?  (Acct has 1 item X and 1 item Y, anon user has 1 item X in cart; when anon user logs in to acct, cart has 2 item X and 1 item Y).

djkotowski:
Ditto what Jordan said.  Take a look at the comments, but no need for resubmission.

PR #468, opened 2015-01-28
Shopping cart refactor

jfinzel:
I greatly simplified my many-tabled scenarios and also added the shipping estimate feature.


shopping_cart/spec/brainstorm
```diff
@@ -0,0 +1,18 @@
+Happy:
```
jvanderz22: This probably isn't a type of file that should be included in any commits

shopping_cart/spec/features/browse_items.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Browse Items
+  In order to choose what I want to buy
+  As a customer
+  I want to be able to browse the items
+
+  @happy
+  Scenario: View All Items
+    Given 5 items exist
```
jvanderz22: This seems like a self-verifying test or, at least, a non-descriptive one. What would it mean to have 5 items that exist if the 5 items aren't shown on the page? This would be more clear if the given was more like "Given I add 5 items to the shopping cart"

shopping_cart/spec/features/existing_user_address.feature
```diff
@@ -0,0 +1,27 @@
...
+
+  @happy
+  Scenario: Checkout with existing address
+    Given I have a saved address
+    And my cart has 1 item
+    When I click the checkout button
+    Then I am asked to confirm my existing address
+
+  @sad
+  Scenario: Checkout with no items
```
jvanderz22: This is a good test, but it should belong to a different feature. This feature is specific to address, but you're testing whether you can proceed to checkout with a relation to items. A good, similar test here would be check if you can proceed to checkout without an address entered.

shopping_cart/spec/features/login.feature
```diff
@@ -0,0 +1,38 @@
...
+    Then I see an invalid user or password message
+
+  @bad
+  Scenario: Missing Password
+    When I submit with a missing password
+    Then I see a missing password message
+
+  @bad
+  Scenario: Missing Username
+    When I submit with a missing password
```
jvanderz22: Oops, duplicate "when" condition here

shopping_cart/spec/features/update_shopping_cart.feature
```diff
@@ -0,0 +1,42 @@
...
+    And I am on the shopping cart page
+    When I change the quantity to 1
+    Then I have 1 of the item in my cart
+
+  @bad
+  Scenario: Invalid Quantity Update
+    Given I have 2 of the item in my cart
+    And I am on the shopping cart page
+    When I change the quantity to -1
+    Then I see an "invalid quantity" message
```
jvanderz22: What about when you have two of the item and you want to remove all of them from the cart?

shopping_cart/spec/features/browse_items.feature
jfinzel: ​This isn't about the shopping cart.  If the 5 items don't show on the
page, that means something is wrong with the functionality that shows the
items on the page.

I grant that it's not very clear though.  This is what I mean: there are
actually 5 products that exist in the database.  The test is that the items
page actually shows all of them.

shopping_cart/spec/features/existing_user_address.feature
jfinzel: I see.

shopping_cart/spec/features/update_shopping_cart.feature
jfinzel: Yes.  I think I had that test and removed it.  The latter could be tested as a further integration test.

shopping_cart/spec/features/browse_items.feature
jvanderz22: Oh okay, I understand what you mean here now. In that case, I think changing the wording a bit and maybe changing the feature to be "I want to be able to browse the items in the store" would help for clarity, but it's probably fine as is

jvanderz22:
Overall, this seems good. A few of the cases and step definitions could be written to provide more clarity on what's being tested, but in general, the tests are clear, straightforward and cover the important scenarios. Keep the comments in mind, but this exercise doesn't have to be resubmitted.

PR #469, opened 2015-01-28
Bte03 data science

magicalbanana:
@jmmastey please review.


data_science/model/datum.rb
```diff
@@ -0,0 +1,18 @@
...
+    def initialize(cohort, date, result)
+      @cohort = cohort
+      @date = date
+      @result = result
+    end
+
+    def self.process_json(file)
+      json = Service::JsonParser.parse(file)
+      json.map { |j| new(j[:cohort], j[:date], j[:result]) }
+    end
```
jmmastey: I'm not sure it makes sense for the Datum class itself to understand how to parse the incoming data file. What if we start to feed this system from a webservice, or a CSV, or statsd? If this method accepted a hash and generated datum objects from it, that would make more sense.

Keep in mind also, though, that if we needed to feed in 10,000,000 hits, this would probably crank to a halt.

data_science/model/sample.rb
```diff
@@ -0,0 +1,74 @@
...
+
+module Model
+  class Sample
+    #because the plural for datum is data
+    attr_accessor :data
+
+    #CONFIDENCE VALUE > 95
+    CONVERSION_RATE_MULTIPLIER = 1.96
+
+    #DEFAULT FILE
```
jmmastey: Delete the unused lines. We have source control.

data_science/model/sample.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+
+    def standard_error(cohort)
+      factor = ( 1 - conversion_rate(cohort)) / cohort_size(cohort)
+      Math.sqrt(conversion_rate(cohort) * factor)
+    end
+
+    def error_bars(cohort)
+      (standard_error(cohort) * CONVERSION_RATE_MULTIPLIER) * 100
+    end
```
jmmastey: So notice here that the last six methods are all obsessed with `cohort`. This is called Feature Envy, and there's a class trying to get out. Try adding a cohort object and see how it affects your system. Even better, since you test a lot, try to do it without breaking your tests.

data_science/model/sample.rb
```diff
@@ -0,0 +1,74 @@
...
+    def calculate_confidence_level
+      values = Hash.new
+      get_cohorts.each do |cohort|
+        values[cohort.to_sym] = { :non_conversions => cohort_non_conversions(cohort),
+                                  :conversion => cohort_conversions(cohort)
+        }
+      end
+      tester = ABAnalyzer::ABTest.new(values)
+      1 - tester.chisquare_p
+    end
```
jmmastey: When a method returns a value (but doesn't have any side effects) we call it a query (to distinguish it from a command). Queries should have noun-like names (`confidence_level`, not `calculate_confidence_level`). This naming principle means that you could move `confidence_level` to an `attr_reader` later on without your class becoming weird and confusing. It also helps the next person reason about the behavior of your system.

data_science/model/sample.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    def get_cohorts
+      cohorts = []
+      @data.uniq { |datum| datum.cohort }.each { |d| cohorts << d.cohort }
+      cohorts
+    end
```
jmmastey: Same here as above. Also, whenever you see this pattern:

```
thing = []
@other_things.each { |i| thing << i }
thing
```

There is always a better refactor. Usually, that means `inject`, `each_with_object`, or `select`. In this case, I'd probably do:

```
def cohorts
  @data.map(&:cohort).uniq
end
```

data_science/service/json_parser.rb
```diff
@@ -0,0 +1,21 @@
+require 'json'
+
+module Service
+  class JsonParser
+    def self.parse(file)
+      unless file_exists?(file)
+        raise "The file = '#{file}' could not be located"
+      end
```
jmmastey: It's probably semantically just as reasonable (and more concise) to do `raise FileNotFoundError unless file_exists?(file)`

data_science/service/json_parser.rb
```diff
@@ -0,0 +1,21 @@
...
+      JSON.parse(file, :symbolize_names => true)
+    end
+
+    def self.file_exists?(file)
+      File.exists?(file)
+    end
+
+    def self.file_empty?(file)
+      File.zero?(file)
+    end
```
jmmastey: Unused method? Also, I think JSON parsing will probably flag this anyway.

data_science/spec/data_science/model/sample_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+  end
+
+  describe "#cohort_size" do
+    { "A" => 1349, "B" => 1543 }.each do |cohort, size|
+      it "returns the size of cohort: '#{cohort}' == '#{size}'" do
+        expect(sample.cohort_size(cohort)).to eq(size)
+      end
+    end
+  end
```
jmmastey: This is a common problem that I see in people's tests. In this case, the source data file is production data. We don't want to peg our tests to production data, as it's likely to change. For instance, if you pull the master repo, you'll see that the filename has since changed (because of this problem). All your tests will break, and not because your implementation broke.

Instead, it'd be better to have a small test file, or even better to generate test data within this file. That way, the data is in the same geographical area as the test, which makes it easier to grok.

data_science/spec/data_science/model/sample_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+
+  describe "#conversions" do
+    { "A" => 47, "B" => 79 }.each do |cohort, conversion|
+      it "returns the conversion of cohort: '#{cohort}' == '#{conversion}'" do
+        expect(sample.cohort_conversions(cohort)).to eq(conversion)
+      end
+    end
+  end
```
jmmastey: Also as a general note. If the test works for A, it probably works for B. I'd rather see some tests around edge cases. Can I have a cohort with zero conversions or hits? What happens at the boundary of significant / non-significant tests? What if there aren't enough datapoints to tell?

data_science/spec/data_science/model/sample_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+
+  describe "#conversion_rate" do
+    { "A" => 0.034840622683469234, "B" => 0.05119896305897602 }.each do |cohort, rate|
+      it "returns the conversion rate of cohort: '#{cohort}' == '#{rate}" do
+        expect(sample.conversion_rate(cohort)).to eq(rate)
+      end
+    end
+  end
```
jmmastey: Probably misleadingly specific. A common idiomatic way to do this would be `expect(conversion_rate).to be_within(0.001).of(0.03)`.

data_science/spec/data_science/model/sample_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+
+  describe "#error_bars" do
+    { "A" => 0.9785715107120809, "B" => 1.0997431518046341}.each do |cohort, error|
+      it "returns the error bars for cohort: '#{cohort}' == '#{error}" do
+        expect(sample.error_bars(cohort)).to eq(error)
+      end
+    end
+  end
```
jmmastey: I think error bars should have a positive and negative component. Or is this simply the magnitude?

data_science/spec/data_science/service/json_parser_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    let(:json_parser) { Service::JsonParser.parse(file) }
+    context "when file_exists? = true" do
+      let(:file) { './source_data.json' }
+      it "parses and is not empty" do
+        expect(json_parser).to_not be_empty
+      end
+      it "returns an array" do
+        expect(json_parser).to be_a(Array)
+      end
+    end
```
jmmastey: If the prod data file gets big (a few GB), this test is going to be a mess. Take a look at the Sandi Metz unit testing talk for some ideas on how you could isolate these tests better.

data_science/spec/data_science/split_testing_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  let(:split_testing) { SplitTesting.new }
+
+  describe "#initialize" do
+    it "initializes a new object" do
+      expect(split_testing).to be_a(SplitTesting)
+    end
+  end
+
+  describe "#run" do
+    let(:output) { capture_stdout { split_testing } }
```
jmmastey: It'd be better just to return strings from the library itself. That way you could test this more directly. Generally, outside of toy apps, we almost never write to stdout.

data_science/spec/support/matchers/have_attr_accessor.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  failure_message_when_negated do |object_instance|
+    "expected attr_accessor for #{field} not to be defined on #{object_instance}"
+  end
+
+  description do |object_instance|
+    "have attr_accessor = #{field} on #{object_instance.class}"
+  end
+end
```
jmmastey: I like this, but you can generalize slightly further than `attr_accessor`. The real thing that happens here is that you define a setter for the attribute. If I wanted to refactor away from accessor and just define the setter itself, this test does (and should) still pass, but the error messages no longer make sense.

data_science/split_testing.rb
```diff
@@ -0,0 +1,32 @@
...
+  def run
+    puts "Test Results: Total Sample Size = #{@sample.total_size}"
+    @sample.get_cohorts.sort.each do |cohort|
+      puts "The number of conversions for cohort: #{cohort} is #{@sample.cohort_conversions(cohort)}."
+      puts "The number of non conversions for cohort: #{cohort} is #{@sample.cohort_non_conversions(cohort)}."
+      puts "The conversion rate for cohort: #{cohort} is #{format_output(@sample.conversion_rate(cohort))}"
+      puts "With an error bar of #{format_output(@sample.error_bars(cohort), :error_bar => true)}."
+    end
+    puts "The confidence level for this test is #{format_output(@sample.calculate_confidence_level)}."
+  end
```
jmmastey: Technically, the running happens in the `Model::Sample.new` as much as here. :D

jmmastey:
Hey, one big suggestion for refactor (based on the Feature Envy described elsewhere). Other things are suggestions for robustness of implementation / tests, but not a big deal. Do the refactor for cohorts, your call if it needs another review.

PR #470, opened 2015-01-29
Add yadda part 1

sriking1783:
This is part 1.

Some comments that were raised in previous request(updating updated_*  columns), I put as comments here as why I am doing that.

### DIFF
--- a/yadda/create_yadda_schema.sql
+++ b/yadda/create_yadda_schema.sql
@@ -2,6 +2,13 @@ BEGIN;
 
 CREATE SCHEMA IF NOT EXISTS yadda AUTHORIZATION sastaputhra;
 
+DROP TABLE IF EXISTS yadda.addresses CASCADE;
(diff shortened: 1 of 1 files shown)


yadda/create_yadda_schema.sql
```diff
@@ -0,0 +1,175 @@
...
+  updated_by    TEXT         ,
+  updated_on    TIMESTAMPTZ
+);
+
+CREATE OR REPLACE FUNCTION  yadda.tg_inserts()
+RETURNS TRIGGER AS $$
+  BEGIN
+    NEW.created_on := current_timestamp;
+    NEW.created_by := current_user;
+    -- While inserting/creating a record only created_* are populated
```
incazteca: To remain consistent with Rails convention set the updated_* fields when the row is created.

incazteca:
@sriking1783 Looks good. Please initialize the udpate_* fields when the row is created. This change is small though so you don't need to make another pull.

PR #471, opened 2015-01-29
Overlord

ddzz:


overlord/bomb.rb
```diff
@@ -0,0 +1,83 @@
+class Bomb
+  DEFAULT_ACT_CODE = "1234"
+  DEFAULT_DEACT_CODE = "0000"
+  MAX_DEACTIVATION_ATTEMPTS = 3
+
+  attr_reader :attempts_remaining, :bad_activation_code
```
incazteca: I don't see why bad_activation_code has to be an attribute. It'll make more sense as a method named `bad_activation_code?` that would return a boolean value.

overlord/bomb.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  def initialize
+    @state = :unbooted
+    @attempts_remaining = MAX_DEACTIVATION_ATTEMPTS
+    @bad_activation_code = false
+    @already_booted = false
```
incazteca: This can also be a method `already_booted` that would check the state of the bomb.

overlord/bomb.rb
```diff
@@ -0,0 +1,83 @@
+class Bomb
+  DEFAULT_ACT_CODE = "1234"
+  DEFAULT_DEACT_CODE = "0000"
+  MAX_DEACTIVATION_ATTEMPTS = 3
+
+  attr_reader :attempts_remaining, :bad_activation_code
+
+  def initialize
+    @state = :unbooted
+    @attempts_remaining = MAX_DEACTIVATION_ATTEMPTS
```
incazteca: Could change this to a counter method that increments for each attempt and then just compare to `MAX_DEACTIVATION_ATTEMPTS`.

overlord/bomb.rb
```diff
@@ -0,0 +1,83 @@
...
+    @bad_activation_code = false
+    @already_booted = false
+  end
+
+  def boot(act_code = '', deact_code = '')
```
incazteca: Should set a default activation and deactivation code.

overlord/bomb.rb
```diff
@@ -0,0 +1,83 @@
...
+    @activation_code = assign_act_code(act_code)
+    @deactivation_code = assign_deact_code(deact_code)
+    @state = :deactivated
+  end
+
+  def activate(code)
+    raise RuntimeError unless deactivated?
+    if code == @activation_code
+      @state = :activated
+      @bad_activation_code = false
```
incazteca: These `@bad_activation_code` are no longer needed. Once the state is set you can just call a method to check it.

overlord/bomb.rb
```diff
@@ -0,0 +1,83 @@
...
+      @bad_activation_code = true
+    end
+  end
+
+  def deactivate(code)
+    raise RuntimeError unless activated?
+    if code == @deactivation_code
+      deactivate_bomb
+    else
+      incorrect_deactivation_code
```
incazteca: You could just increment the attempt counter and then check if it's more than `MAX_DEACTIVATION_ATTEMPTS`

overlord/bomb.rb
```diff
@@ -0,0 +1,83 @@
...
+    end
+  end
+
+  def unbooted?
```
incazteca: You could just make status an attribute if needed. Though these methods work fine.

overlord/features/activate_bomb.feature
```diff
@@ -0,0 +1,28 @@
...
+
+  Scenario: Reach activation page
+    Given the bomb has been booted with the default codes
+    Then the bomb is not activated yet
+    And I see a field to enter the activation code
+    And I see a button to submit the activation code
+
+  Scenario: Bomb activates with default activation code
+    Given the bomb has been booted with the default codes
+  	When I enter the default activation code
```
incazteca: Please fix indentation

incazteca: I don't think the user needs to enter the default codes. They should just be the default.

overlord/features/deactivate_bomb.feature
```diff
@@ -0,0 +1,22 @@
...
+  Background:
+    Given the bomb has been booted and activated with the default codes
+
+  Scenario: Bomb deactivates with default deactivation code
+  	When I enter the default deactivation code
+  	Then the bomb is deactivated
+
+  Scenario: Bomb does not deactivate after one incorrect deactivation attempt
+  	When I enter an incorrect deactivation code
+    And I try to deactivate the bomb
```
incazteca: Fix indentation

overlord/features/deactivate_bomb.feature
```diff
@@ -0,0 +1,22 @@
...
+
+  Background:
+    Given the bomb has been booted and activated with the default codes
+
+  Scenario: Bomb deactivates with default deactivation code
+  	When I enter the default deactivation code
+  	Then the bomb is deactivated
+
+  Scenario: Bomb does not deactivate after one incorrect deactivation attempt
+  	When I enter an incorrect deactivation code
```
incazteca: Fix indentiation

incazteca:
@ddzz Your tests look great however I don't see the bomb exploding other than inputting the wrong deactivation code. Please implement a detonation feature and include the proper tests.

overlord/bomb.rb
ddzz: I wrote the ```boot``` method this way because I pass the form input (the activation and deactivation codes) directly to this method, and these inputs are strings.  If no codes are entered then the method receives two empty strings for the activation codes, so ```def boot (act_code = DEFAULT_ACT_CODE, deact_code = DEFAULT_DEACT_CODE)``` would not use the default codes in any scenario as it is currently defined.

But now that you've pointed this out, ```def boot (act_code = DEFAULT_ACT_CODE, deact_code = DEFAULT_DEACT_CODE)``` would work if I added some logic to ```post '/boot'``` to check if the codes are empty strings and simply call ```boot``` without any arguments, and this would probably clean up the ```boot``` method as well.

overlord/features/deactivate_bomb.feature
ddzz: Sorry about the indentation issues.  For some reason the idents were not showing up like this in Sublime.  I will check my indentation settings.

ddzz:
@incazteca I have designed the bomb so that it detonates only if an incorrect deactivation code is entered three times.  Is a separate feature to directly detonate it necessary?

overlord/features/activate_bomb.feature
ddzz: I have designed the bomb so that the user does not have to enter any codes only when they boot the bomb.  After it has been booted an activation or deactivation code must be entered, whether the bomb is set to use the defaults or not.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,263 @@
...
+      expect(bomb).not_to be_activated
+      bomb.activate(DEFAULT_ACT_CODE)
+      bomb.deactivate(DEFAULT_DEACT_CODE)
+      expect(bomb).not_to be_activated
+      bomb.activate(DEFAULT_ACT_CODE)
+      3.times do
+        bomb.deactivate("1111")
+      end
+      expect(bomb).not_to be_activated
+    end
```
djkotowski: This test is doing a lot.  With four assertions, it's a bit more difficult to diagnose when the test fails.

djkotowski:
This is looking good.  I like that you're using HTML5 markup in some cases.  Take a look at the comments and refactor.

overlord/features/activate_bomb.feature
```diff
@@ -0,0 +1,28 @@
...
+    And I try to activate the bomb with the default code
```
djkotowski: I would combine 14 and 15 into one step.  Line 14 implies it is doing more than it actually is (implies it uses the default code somehow, not that it clicks "Submit").

djkotowski:
You do not need an additional detonation feature.  The only requirement for detonation is three wrong deactivation attempts.  All others are optional.

djkotowski:
Revision:  No resubmission needed.

PR #473, opened 2015-01-30
Add overlord revision

sriking1783:
Overlord revision 2


overlord/app/helpers/bomb_helpers.rb
```diff
@@ -0,0 +1,14 @@
+module BombHelpers
+  def self.explode_bomb(bomb)
+    return if bomb.nil?
+    bomb.explode! if self.time_to_explode? bomb
+  end
+
+  private
+
+  def self.time_to_explode?(bomb)
+    !(bomb.activated_time.nil?) &&
```
aeberlin: I think this could be made simpler by using guard clauses.. e.g.:
```
def foo
  return true if ...
  false
end
```

overlord/app/helpers/wire_helpers.rb
```diff
@@ -0,0 +1,6 @@
+module WireHelpers
+  def self.default_wires
+    [{ color: "red", diffuse: false },
+     { color: "green", diffuse: true }]
+  end
+end
```
aeberlin: I would consider using constants here instead, instead of class methods on the module.

overlord/app/models/bomb.rb
```diff
@@ -0,0 +1,51 @@
+require 'dm-types'
+require 'dm-validations'
+
+class Bomb < ActiveRecord::Base
+  validates_length_of :activation_code, :deactivation_code, minimum: 4
+  validates :activation_code, :deactivation_code, format: { with: /\A[0-9]*\z/ }
+  enum status: %w(active inactive explode)
+
+  has_many :wires
+  before_validation :initialize_defaults
```
aeberlin: I'm slightly confused by this, as I am with the `initialize_defaults` method. What does this accomplish? I noticed you have your defaults properly set in your migrations, so what is the purpose of redefining them here and reinitializing before each validation attempt?

overlord/app/models/bomb.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def explode
+    explode!
+    self.activated_time = nil
+    save!
+  end
+
+  private
+
+  def initialize_defaults
```
aeberlin: See my comment above.

overlord/app/routes/bomb.rb
```diff
@@ -0,0 +1,124 @@
...
+  get "/bomb_activate" do
+    haml :bomb_activate
+  end
+
+  get "/bomb_deactivate" do
+    haml :bomb_deactivate
+  end
+
+  post '/bomb/diffuse' do
+    params = request.body.read.split('=')
```
aeberlin: I see this line several times. I would suggest extracting it into a `before` filter and running that filter on each request.

http://www.sinatrarb.com/intro.html#Filters

overlord/app/helpers/bomb_helpers.rb
sriking1783: Sorry, I don't understand, how a guard clause can simplify this. There are 3 boolean statements I am checking, I will refactor it a little, but still believe adding a guard clause will complicate it

jmmastey:
The tests on this are awesome, but the implementation is getting more complicated with the revisions. You should sit with Danila and work through these things to find a much less complicated solution.

PR #474, opened 2015-02-02
Add yadda part 2

sriking1783:
This is part 2 of  yadda

### DIFF
--- a/yadda/create_yadda_schema.sql
+++ b/yadda/create_yadda_schema.sql
@@ -2,6 +2,13 @@ BEGIN;
 
 CREATE SCHEMA IF NOT EXISTS yadda AUTHORIZATION sastaputhra;
 
+DROP TABLE IF EXISTS yadda.addresses CASCADE;
+DROP TABLE IF EXISTS yadda.breweries CASCADE;
+DROP TABLE IF EXISTS yadda.beers CASCADE;
+DROP TABLE IF EXISTS yadda.users CASCADE;
+DROP TABLE IF EXISTS yadda.ratings CASCADE;
+DROP TABLE IF EXISTS yadda.beer_styles CASCADE;
+
 CREATE TABLE yadda.addresses(
   id            SERIAL       PRIMARY KEY,
   line_1        TEXT         NOT NULL,
@@ -9,8 +16,8 @@ CREATE TABLE yadda.addresses(
   state         TEXT         NOT NULL,
   country       TEXT         NOT NULL,
   zip_code      TEXT         NOT NULL,
-  created_by   TEXT         ,
-  created_on   TIMESTAMPTZ  ,
+  created_by    TEXT         ,
+  created_on    TIMESTAMPTZ  ,
   updated_by    TEXT         ,
   updated_on    TIMESTAMPTZ
 );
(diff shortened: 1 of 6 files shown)


yadda/create_yadda_schema.sql
```diff
@@ -0,0 +1,182 @@
...
+  id            SERIAL       PRIMARY KEY,
+  style_id      INTEGER      NOT NULL REFERENCES yadda.beer_styles ON DELETE CASCADE,
+  brewing_year  INTEGER      NOT NULL,
+  brewery_id    INTEGER      REFERENCES yadda.breweries ON DELETE CASCADE,
+  created_by    TEXT         ,
+  created_on    TIMESTAMPTZ  ,
+  updated_by    TEXT         ,
+  updated_on    TIMESTAMPTZ  ,
+  description   TEXT
+);
```
jmmastey: I think you're missing `name` :)

yadda/create_yadda_schema.sql
```diff
@@ -0,0 +1,182 @@
...
+
+CREATE TABLE yadda.users(
+  id            SERIAL       PRIMARY KEY,
+  name          TEXT         NOT NULL,
+  email         TEXT,
```
jmmastey: I'm not sure that nullable email is realistic anymore. Any reason for this and password to be nullable?

yadda/create_yadda_schema.sql
```diff
@@ -0,0 +1,182 @@
...
+  created_on    TIMESTAMPTZ  ,
+  updated_by    TEXT         ,
+  updated_on    TIMESTAMPTZ
+);
+
+CREATE OR REPLACE FUNCTION  yadda.tg_inserts()
+RETURNS TRIGGER AS $$
+  BEGIN
+    NEW.created_on := current_timestamp;
+    NEW.created_by := current_user;
```
jmmastey: So, fwiw, this will not work correctly in a production app. The `current_user` will be set to the app credentials (`cnuapp`, or `netcredit`, or what have you)

yadda/create_yadda_schema.sql
```diff
@@ -0,0 +1,182 @@
...
+    RETURN NEW;
+  END;
+$$ LANGUAGE plpgsql;
+
+CREATE OR REPLACE FUNCTION  yadda.tg_updates()
+RETURNS TRIGGER AS $$
+  BEGIN
+    NEW.updated_on := current_timestamp;
+    NEW.updated_by := current_user;
+    -- While updating a record only updated_* are populated
```
jmmastey: Probably not a useful comment.

yadda/create_yadda_view.sql
```diff
@@ -0,0 +1,54 @@
+BEGIN;
+
+CREATE OR REPLACE VIEW yadda.average_rating_view AS
+  SELECT st.id,
+         br.name,
```
jmmastey: How does this execute without the actual `beer.name` field?

yadda/create_yadda_view.sql
```diff
@@ -0,0 +1,54 @@
...
+  GROUP BY name, style, st.id
+  ORDER BY name
+;
+
+CREATE OR REPLACE VIEW yadda.top_beer_from_brewery AS
+ SELECT name,style,rating
+  FROM (
+    SELECT name,rating,style, rank() OVER(PARTITION BY name
+      ORDER BY rating DESC) AS rank FROM  yadda.average_rating_view)ss
+where rank=1;
```
jmmastey: Have you tried this on a realistic dataset? Without an index, this is going to eat your lunch on performance.

yadda/create_yadda_view.sql
```diff
@@ -0,0 +1,54 @@
...
+         br.name,
+         st.style,
+         AVG((r.look+r.smell+r.taste+r.feel+r.overall)) rating
+    FROM yadda.ratings r
+    JOIN yadda.beers b on r.beer_id = b.id
+    JOIN yadda.breweries br on b.brewery_id = br.id
+    JOIN yadda.beer_styles st on b.style_id = st.id
+    WHERE r.created_on > CURRENT_DATE - INTERVAL '6 months'
+    GROUP BY name, style, st.id
+    ORDER BY name;
```
jmmastey: Doesn't this duplicate most of the `average_rating_view`?

yadda/create_yadda_view.sql
```diff
@@ -0,0 +1,54 @@
+BEGIN;
+
+CREATE OR REPLACE VIEW yadda.average_rating_view AS
```
jmmastey: Generally, don't use `_view` in the name, for the same reason you wouldn't call an object instance var `.name_instance_var`. It leaks abstraction details unnecessarily.

yadda/create_yadda_view.sql
```diff
@@ -0,0 +1,54 @@
...
+          SELECT st.style from yadda.beer_styles st
+            JOIN yadda.beer_styles st1 ON st.category_id = st1.category_id
+          AND st1.style =' || quote_literal($1) ||'
+      ) AND style <> '|| quote_literal($1) ||' ORDER BY rating DESC LIMIT 3
+  )tmp ORDER by random()';
+    BEGIN
+      RETURN QUERY EXECUTE sql
+      USING beer;
+    END;
+$$ LANGUAGE plpgsql;
```
jmmastey: This causes a text-based match on `average_rating_view.style`. Also, I'm not sure this works, given the style logic? Have you tested this code?

yadda/create_yadda_schema.sql
```diff
@@ -0,0 +1,182 @@
...
+  line_2        TEXT,
+  city          TEXT,
+  state         TEXT,
+  country       TEXT,
+  zip_code      TEXT,
+  created_by    TEXT         ,
+  created_on    TIMESTAMPTZ  ,
+  updated_by    TEXT         ,
+  updated_on    TIMESTAMPTZ
+);
```
jmmastey: So these aren't normalized, though that might be a pain for this assignment. More particularly, most of the fields in this schema are nullable text fields, which doesn't allow a lot of room for data quality. Any reason not to use a more particular data type?

jmmastey:
A few notes, but overall the schema is reasonable. I'm interested to see where it goes when you get to part three and implement the requirement of "a million drinks over maybe 200,000 drinkers". Some of the existing structures would fall over under that relatively light load. Make changes but go on to part 3 / 4 before submitting anything else.

yadda/create_yadda_view.sql
sriking1783: I tested it by adding 4-5 record in a table. I don't understand how it will not work.

PR #475, opened 2015-02-03
Refactored arrowhead to remove arrowheads

geekaholic:
Don't merge, but do send feedback :)

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -1,4 +1,9 @@
 class Arrowhead
+  # Exception msgs
+  UNKNOWN_REGION_EXCEPTION = "Unknown region, please provide a valid region."
+  UNKNOWN_SHAPE_EXCEPTION =
+    "Unknown shape value. Are you sure you know what you're talking about?"
+
   # This seriously belongs in a database.
   CLASSIFICATIONS = {
     far_west: {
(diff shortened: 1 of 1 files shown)


arrowhead/arrowhead.rb
```diff
@@ -1,4 +1,9 @@
 class Arrowhead
+  # Exception msgs
```
ashtonthomas: errrr.. just because I have nothing else to say, this comment does nothing. Style guides usual say to avoid organizational comments.

ashtonthomas:
@geekaholic, I think this is as simple as you can make it. Definitely removes the arrow-like if/else statements. I would leave out the organizational comment, though.

We could probably make this more complicated by adopting an OOP approach (breaking the classify method up so we don't have a guard clause, operation, guard clause. However, it is definitely not worth it for this exercise. I think this is as simple as the exercise calls for.

LGTM

geekaholic:
Thanks Ashton. Will take note on the feedback on comments.

Bud

On Tuesday, February 3, 2015, Ashton Thomas <notifications@github.com>
wrote:

> Closed #475 <https://github.com/jmmastey/level_up_exercises/pull/475>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jmmastey/level_up_exercises/pull/475#event-229254297>.
>


-- 
Bud

Geek with an Attitude
Blog: http://geekaholic.org
twitter: @geekaholic

PR #479, opened 2015-02-03
Zb dinos

zbright:


dino_catalog/catalog.rb
```diff
@@ -0,0 +1,75 @@
+require 'csv'
+require 'terminal-table'
+require_relative 'dinosaur'
+require_relative 'dino_list'
+
+def reset_arr(original)
+  d_arr = DinoList.new
+  original.each { |d| d_arr << d.dup }
```
zsyed1: In the ruby the generally accepted practice is to name variables as close as possible to English words. So a better name for `d_arr` would be `dinosaurs` or `dinos` for short. You definitely want to avoid putting variable types (like `arr` and `array`) into the name. Since your class is called `DinoList` you could use `dino_list` as a variable name.

dino_catalog/catalog.rb
```diff
@@ -0,0 +1,75 @@
...
+  new_dino.continent = row['CONTINENT']
+  new_dino.diet = row['DIET'] ||
+                  (row['Carnivore'] == 'Yes' ? 'Carnivore' : 'Herbivore')
+  new_dino.weight = row['Weight'] || row['WEIGHT_IN_LBS']
+  new_dino.walking = row['Walking'] || row['WALKING']
+  new_dino.description = row['DESCRIPTION']
+
+  dino_arr << new_dino
+end
+
```
zsyed1: Lines 17 through 31 loads an existing DinoList from a csv-file. Instead of laying this out here you could instead move this into a method on `DinoList` say `def load_from_csv(path)` then invoke it as `dino_arr.load_from_csv(ARGV[0])`. If you want to keep this functionality out of `DinoList` then at the very least you could move this lines into a function within this file say `def load_from_csv(dino_arr, path)`.

dino_catalog/catalog.rb
```diff
@@ -0,0 +1,75 @@
...
+CSV.foreach(ARGV[0], headers: true) do |row|
+  dino_arr.is_pirate_list |= !row.header?('DESCRIPTION')
+
+  new_dino = Dinosaur.new
+  new_dino.name = row['Genus'] || row['NAME']
+  new_dino.period = row['Period'] || row['PERIOD']
+  new_dino.continent = row['CONTINENT']
+  new_dino.diet = row['DIET'] ||
+                  (row['Carnivore'] == 'Yes' ? 'Carnivore' : 'Herbivore')
+  new_dino.weight = row['Weight'] || row['WEIGHT_IN_LBS']
```
zsyed1: In line 24 you can move the conversion logic into a method `def extract_diet(row)` and then replace line 24 with `new_dino.diet = extract_diet(row)`.

dino_catalog/catalog.rb
```diff
@@ -0,0 +1,75 @@
...
+  new_dino.walking = row['Walking'] || row['WALKING']
+  new_dino.description = row['DESCRIPTION']
+
+  dino_arr << new_dino
+end
+
+exit = false
+original_array = reset_arr dino_arr
+
```
zsyed1: In line 34 put parentheses around the arguments: `reset_arr(dino_arr)`

dino_catalog/catalog.rb
```diff
@@ -0,0 +1,75 @@
...
+require_relative 'dino_list'
+
+def reset_arr(original)
+  d_arr = DinoList.new
+  original.each { |d| d_arr << d.dup }
+  d_arr.is_pirate_list = original.is_pirate_list
+
+  d_arr
+end
+
```
zsyed1: The method `reset_arr` should just be a method on `DinoList`:

```
def reset(original)
  clear
  original.each { |d| push(d.dup) }
  is_pirate_list = original.is_pirate_list
end
```

dino_catalog/catalog.rb
```diff
@@ -0,0 +1,75 @@
...
+  new_dino.description = row['DESCRIPTION']
+
+  dino_arr << new_dino
+end
+
+exit = false
+original_array = reset_arr dino_arr
+
+while !exit
+  puts "Note: if entering more than one "
```
zsyed1: I believe `exit` is a keyword. It would be better to use a different name like `done`

dino_catalog/catalog.rb
```diff
@@ -0,0 +1,75 @@
...
+  puts "A) Display all bipeds"
+  puts "B) Display all carnivores"
+  puts "C) Display all from a time period"
+  puts "D) Display all big dinos (> 2 tons)"
+  puts "E) Display all small dinos (<= 2 tons)"
+  puts "F) Select one dino"
+  puts "R) Remove filters"
+  puts "Hit enter to view current list"
+  puts "Type EXIT to quit"
+
```
zsyed1: Another way to do this (so as to avoid writing all the `puts`) is:
```
puts ["Enter a letter (or multiple) for your choice(s):",
      "A) Display all bipeds",
      "B) Display all carnivores",
      ...
      "Type EXIT to quit"].join("\n")
```

dino_catalog/catalog.rb
```diff
@@ -0,0 +1,75 @@
...
+  user_input = $stdin.gets.chomp.downcase
+
+  exit = true and next if user_input == 'exit'
```
zsyed1: I would put this in two separate lines:

```
if user_input == 'exit'
  exit = true
  next
end
```
Though @djkotowski may disagree...

dino_catalog/catalog.rb
```diff
@@ -0,0 +1,75 @@
...
+    puts "Enter the time period you would like to view: "
+    period_input = $stdin.gets.chomp.downcase
+    dino_arr.get_from_period period_input
+  end
+
+  dino_arr.get_big if user_input.include? 'd'
+  dino_arr.get_small if user_input.include? 'e'
+
+  if user_input.include? 'f'
+    puts "Please enter a dinosaur's name/genus: "
```
zsyed1: Put parentheses around function arguments: `user_input.include?('d')`

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,8 @@
+class Dinosaur
+  attr_accessor :name, :period, :continent, :diet, :weight, :walking, :description
+
+  def to_json
+    { 'name' => @name, 'period' => @period, 'continent' => @continent, 'diet' => @diet,
+      'weight' => @weight, 'walking' => @walking, 'description' => @description }
```
zsyed1: Please wrap all function arguments in parentheses - I didn't want to flood the conversation by putting a note at every occurrence :smile:

dino_catalog/dino_list.rb
```diff
@@ -0,0 +1,78 @@
...
+    print_desc = select { |dino| dino.description }.any?
+
+    Terminal::Table.new do |t|
+      t << set_headers(print_weight, print_desc)
+      t << :separator
+      each do |dino|
+        t << prettify_dino(dino, print_weight, print_desc)
+      end
+    end
+  end
```
zsyed1: Nice gem you found!

dino_catalog/dino_list.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  def get_carnivores
+    select! { |dino| dino.diet.downcase != 'herbivore' }
+  end
+
+  def get_by_name(name)
+    select! { |dino| dino.name.downcase == name }
+  end
+
```
zsyed1: All these methods are called `get_<something>` but what they really do is _filter_ the list to a possibly smaller set. I would change the word `get` to `filter` and then put an exclamation mark at the end so as to convey that they may mutate the internal data. So for example, you could rename `get_carnivores` to `filter_carnivores!`. You can use something other than `filter` of course, that was just one option that came to mind.

dino_catalog/catalog.rb
zsyed1: Instead of giving the user options indexed by letters `A, B, C, ...` perhaps you could have them enter the keywords `bipeds, carnivores, big, ...`.

dino_catalog/catalog.rb
zsyed1: You could encapsulate the functionality in `catalog.rb` into a class called `class Catalog` and then separate these blocks into methods. For example:

Lines 17 - 31 would go into a `load` method.
Line 36's while loop could go into a `run` method
Line 37 - 47 would go into a `display_menu` method
Lines 51 - 73 would go into a `process_input` method

Then the catalog would be instantiated and run like `Catalog.new(ARGV[0]).run`

This will help readers understand the various components of your catalog and how they work.

dino_catalog/dinosaur.rb
zsyed1: Nice job implementing the DinoDex features! I would recommend doing the following:

1) Fix the parentheses-around-arguments issues.
2) Consider making your catalog an object (see comment above about this)
3) Make a `reset` and a `load_csv` method on `DinoList`

I think these are the biggest things, but please consider the other comments as well.

dino_catalog/dino_list.rb
jmmastey: Interestingly, since they change the internal data array, `filter_carnivores!` is not a bad call for names (verbs for things that mutate state). If you instead return a new instance (letting someone chain but keep the old instance), a noun name would be better `carnivores` (and also simpler.)

zsyed1: Cool.

PR #481, opened 2015-02-04
Add event system revised

sriking1783:


dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,102 @@
...
+result = dinosaur.search_dinosaur(options1)
+p result.get_name.inspect
+
+options2['WEIGHT_IN_LBS'] = 2000
+options2['compare'] = "greater"
+result = dinosaur.search_dinosaur(options2).search_dinosaur(options1)
+p result.get_name.inspect
+
+result = dinosaur1.search_dinosaur(options2)
+p result.get_name.inspect
```
jmmastey: This file really shouldn't be in the pull. Please make sure to keep your branches well maintained.

event_system/app/assets/javascripts/register.js.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
```
jmmastey: Delete unused files. I think we've covered this one before.

event_system/app/controllers/register_controller.rb
```diff
@@ -0,0 +1,11 @@
+class RegisterController < ApplicationController
+  skip_before_action :verify_authenticity_token
+  def index
+  end
+
+  def create
+    EmailContact.create(email: params["region"],
+      region_id: Region.select("region_id").where(city: params["region"]))
```
jmmastey: Try not to leak database details via your controllers. This should be a scope or a method on `Region`. Also, take a look at `.pluck`, which unlike `.select` does not create an object.

event_system/app/controllers/weather_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+class WeatherController < ApplicationController
+  include WeatherHelper
+  def index
+    if params["region"] && params["region"]["name"] != ""
+      @high_low_temperatures = WeatherHelper.parse_temperatures
+      @detail_forecasts = WeatherHelper.parse_details
+      region_id = Region.where(:city => "Chicago").first.region_id
+
+      today_forecast = WEATHER_WEAR.keys.select do |word|
+        @detail_forecasts.last.last.values.first.downcase.match(word)
```
jmmastey: This line is super confusing. Not sure what you're trying to do here.

event_system/app/controllers/weather_controller.rb
```diff
@@ -0,0 +1,21 @@
+
+class WeatherController < ApplicationController
+  include WeatherHelper
+  def index
+    if params["region"] && params["region"]["name"] != ""
+      @high_low_temperatures = WeatherHelper.parse_temperatures
+      @detail_forecasts = WeatherHelper.parse_details
```
jmmastey: When you include helpers you don't need to reference their name to use their methods.

event_system/app/helpers/weather_helper.rb
```diff
@@ -0,0 +1,57 @@
+require 'nokogiri'
+require 'open-uri'
+module WeatherHelper
+  BASE_URL = "http://forecast.weather.gov/MapClick.php?textField1=41.8500262820005&textField2=-87.65004892899964"
+  def self.parse_temperatures
+    if WeatherForecast.where("created_at::date = current_date").count <= 0
```
jmmastey: `WeatherForecast.for_today.any?`

event_system/app/helpers/weather_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+      document = fetch_document
+      scraping =  WebScraper.scrape_temperatures document
+      scraping.each do |key, hash_temp|
+        wf = WeatherForecast.new
+        wf.weather_day = key
+        wf.high        = hash_temp["high"] if hash_temp["high"]
+        wf.low         = hash_temp["low"]  if hash_temp["low"]
+        wf.save!
```
jmmastey: If the `hash_temp["high"]` etc values are nonexistent keys, these modifier-`ifs` have no effect. Instead:

```
WeatherForecast.create!(
  weather_day: key,
  high: hash_temp["high"],
  low: hash_temp["low"]
)
```

event_system/app/helpers/weather_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+    else
+      weather_forecast = WeatherForecast.select("weather_day,high,low").
+                         where("created_at::date = current_date")
```
jmmastey: Again, leaking DB details out of the model.

event_system/app/helpers/weather_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+      scraping = {}
+      weather_forecast.each do |rec|
+        scraping[rec.weather_day] = {}
+        scraping[rec.weather_day] = { "high" => rec.high, "low" => rec.low }
+      end
+    end
+    scraping
+  end
```
jmmastey: In general, these two lemmas of the `if` should be separate methods, which would also allow you to return values directly from them. That, in turn would eliminate the `scraping` variable.

event_system/app/helpers/weather_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+        scraping[rec.weather_day] = { "detail_afternoon" => rec.detail_afternoon, "detail_night" => rec.detail_night }
+      end
+    end
+    scraping.first(1)
+  end
+
+  def self.fetch_document
+    # return Nokogiri::HTML(open(BASE_URL)) if !Rails.env.test?
+    Nokogiri::HTML(File.open("temperatures.html", "r"))
+  end
```
jmmastey: Since this is a query, not a command (it does not modify any data), it should have a noun-like name (`document` would work). That'd render #30 irrelevant.

event_system/app/helpers/weather_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+      weather_forecast.each do |rec|
+        scraping[rec.weather_day] = {}
+        scraping[rec.weather_day] = { "high" => rec.high, "low" => rec.low }
+      end
+    end
+    scraping
+  end
+
+  def self.parse_details
+    if WeatherForecastDetail.where("created_at::date = current_date").count <= 0
```
jmmastey: Same as the other one, this is too much complexity in one method. It also appears to mostly duplicate the previous method, in which case it probably makes sense to de-duplicate.

event_system/app/helpers/weather_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+    scraping.first(1)
+  end
+
+  def self.fetch_document
+    # return Nokogiri::HTML(open(BASE_URL)) if !Rails.env.test?
+    Nokogiri::HTML(File.open("temperatures.html", "r"))
+  end
+end
+
+WeatherHelper.parse_details
```
jmmastey: This line means that when you include the file it will do a bunch of work for no reason and then discard the result. Why is this call here?

event_system/app/mailers/weather_alert.rb
```diff
@@ -0,0 +1,10 @@
+class WeatherAlert < ActionMailer::Base
+  default from: "from@example.com"
+
+  def weather_alert(contact, weather)
+    Rails.logger.info " BEFORE SENDING #{contact.inspect}"
```
jmmastey: Remove debugging statements.

event_system/app/views/register/index.html.haml
```diff
@@ -0,0 +1,6 @@
+%h1 Register#email
+%form{:name => "city", :id => "city", :action => "/register/create", :method => "post", :enctype => "text/plain"}
```
jmmastey: Why the enctype? Shouldn't be necessary here. Also, for ID, just do `#city%form`

event_system/app/views/weather/index.html.haml
```diff
@@ -0,0 +1,26 @@
...
+    %li
+      %input{:type => "text", :name => "region[name]", :id => "city_name", :class => "text"}
+    %input{:type => "submit", :value => "Send", :id => "send", :class => "button"}
+-if params["region"] && params["region"]["name"] != ""
+  %table{id: "temperatue_forecast_table"}
+    - @high_low_temperatures.each do |key,hash_temp|
+      %tr
+        %td= key
+        %td= hash_temp["low"].to_s  if hash_temp["low"]
+        %td= hash_temp["high"].to_s if hash_temp["high"]
```
jmmastey: This will probably not work the way you want it to. If the key were missing, you'd have a ragged table row, which would look really weird. Why are you worried about rendering a nil here?

event_system/app/views/weather/show.html.haml
```diff
@@ -0,0 +1,2 @@
+%h1 Weather#show
+%p Find me in app/views/weather/show.html.haml
```
jmmastey: ???

event_system/app/helpers/weather_helper.rb
```diff
@@ -0,0 +1,57 @@
+require 'nokogiri'
+require 'open-uri'
+module WeatherHelper
```
jmmastey: So this isn't actually a helper. Helpers are for removing logic from views. This is more accurately something that belongs in `lib`.

event_system/config/application.yml
```diff
@@ -0,0 +1,2 @@
+gmail_username: 'sriking@gmail.com'
+gmail_password: 'Sairam123'
```
jmmastey: Errr... you should change your gmail password now.

event_system/app/controllers/register_controller.rb
```diff
@@ -0,0 +1,11 @@
+class RegisterController < ApplicationController
+  skip_before_action :verify_authenticity_token
```
jmmastey: Why? This causes the app to be vulnerable to CSRF attacks.

event_system/app/controllers/weather_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      region_id = Region.where(:city => "Chicago").first.region_id
+
+      today_forecast = WEATHER_WEAR.keys.select do |word|
+        @detail_forecasts.last.last.values.first.downcase.match(word)
+      end
+      @user = EmailContact.where(region_id: region_id)
```
jmmastey: So, if I understand the flow correctly, I register my email and ask for a particular region. Then, I have to visit `/weather/index`, which sends me an email? This is a really weird workflow...

event_system/app/controllers/register_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+  def index
+  end
+
+  def create
+    EmailContact.create(email: params["region"],
```
jmmastey: I'm pretty sure this is broken, isn't it? `email: params["region"]`?

event_system/config/routes.rb
```diff
@@ -0,0 +1,62 @@
+Rails.application.routes.draw do
+  get 'weather/index'
+  get 'weather/show'
+
+  get 'register/index'
+  post "register/create" => "register#create"
```
jmmastey: There's some confusing verbiage here. The `register` controller creates an `EmailContact`, which is often referred to as a `user`. Typically you should try to keep one name for a domain concept.

event_system/features/index.feature
```diff
@@ -0,0 +1,11 @@
...
+
+  Scenario: Should not see high low temperatures
+    Given I visit "weather/index" page
+    And I did not submit the city name
+    Then I should not see temperature forecast
+
+  Scenario: Should see high low temperatures
+    Given I visit "weather/index" page
+    And I submit the city name
+    Then I should see temperature forecast
```
jmmastey: What about EmailContact creation? Or the weather gear? This is not a bad test, but it's missing a ton.

event_system/features/step_definitions/weather_steps.rb
```diff
@@ -0,0 +1,20 @@
+Given(/^I visit "(.*?)" page$/) do |arg1|
+  visit arg1
+end
+
+Given(/^I submit the city name$/) do
+  find(:css, "#city_name").set("Chicago")
```
jmmastey: `fill '#city_name', with: 'Chicago'`

event_system/lib/web_scraper.rb
```diff
@@ -0,0 +1,47 @@
...
+
+class WebScraper
+
+  def self.scrape_temperatures(doc)
+    document = doc
+
+    low_high_temp = {}
+    d1 = Date.today
+    document.css("div.one-ninth-first").each do |para|
+      if para.css("p.txt-ctr-caps").text[-5..-1].downcase == "night"
```
jmmastey: What's a `para`?

event_system/lib/web_scraper.rb
```diff
@@ -0,0 +1,47 @@
...
+        temp_hash_description[d1]["detail_night"] = li.text.strip
+        d1 += 1.day
+      else
+        temp_hash_description[d1] ||= {}
+        temp_hash_description[d1]["detail_afternoon"] = li.text.strip
+      end
+    end
+
+    temp_hash_description
+  end
```
jmmastey: There's a ton of logic in this method. It needs to be decomposed into smaller methods.

event_system/lib/web_scraper.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+    end
+    low_high_temp
+  end
+
+  def self.detailed_scrape(doc)
+    document = doc
+    temp_hash_description = {}
+    top_level = document.search('div.point-forecast-7-day > ul > li')
+    d1 = Date.today
```
jmmastey: `d1` is a really unhelpful name.

event_system/spec/controllers/weather_controller_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  context "GET index" do
+
+  end
+
+  context "GET show" do
+
+  end
+
+end
```
jmmastey: ????

event_system/spec/helpers/weather_helper_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+     before do
+      WeatherForecast.destroy_all
+
+      [Date.today, Date.today + 1.days, Date.today + 2.days].each do |date|
+        wf = WeatherForecast.new
+        wf.weather_day = date
+        wf.high        = "34"
+        wf.low         = "21"
+        wf.save!
+      end
```
jmmastey: Use factories for stuff like this.

event_system/spec/helpers/weather_helper_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    it "saves the data to database" do
+      WeatherHelper.parse_temperatures
+      expect(WeatherForecast.count).to eq(5)
+    end
+  end
+
+  context "trying to pull data again on same day " do
+     before do
+      WeatherForecast.destroy_all
```
jmmastey: Why is the `destroy_all` here? What good does it do?

jmmastey:
Hey Srikanth. This is a good start, but it's got some real issues that need to get resolved. 

1. Several required features are missing. Registration, sending weather forecasts on a schedule, JSON, and historical data.
2. Basic refactory stuff, mentioned in the comments. Specifically, a lot of the code organization needs help.
3. Tests.

Keep at it.

PR #482, opened 2015-02-04
Zb dinos

zbright:
@zsyed1 - Thanks for all the feedback! I made your suggested changes, if there is anything else you think could use some work just let me know


dino_catalog/catalog.rb
```diff
@@ -0,0 +1,59 @@
...
+            "Bipeds - Display all bipeds",
+            "Carnivores - Display all carnivores",
+            "Time Period - Display all from a specified time period",
+            "Big - Display all big dinos",
+            "Small - Display all small dinos",
+            "Select - Select one dion to view",
+            "Reset - Remove all filter",
+            "Exit - quit program",
+            "You can also hit enter to view current list again"].join("\n")
+
```
zsyed1: I would put this string (lines17 - 26) into a method `def menu` but this is fine too.

dino_catalog/catalog.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    dino_list.get_big if user_input.include?('big')
+    dino_list.get_small if user_input.include?('small')
+
+    if user_input.include?('select')
+      puts "Please enter a dinosaur's name/genus: "
+      name_input = $stdin.gets.chomp.downcase
+
+      dino_list.get_by_name(name_input)
```
zsyed1: No need for an empty line (line 55)

dino_catalog/dino_list.rb
```diff
@@ -0,0 +1,113 @@
...
+      arr << dino.walking
+      arr << dino.description if print_desc
+      arr
+    end
+  end
+
+  def to_json
+    dinos = []
+    each { |dino| dinos << dino.to_json }
+
```
zsyed1: You can do this with a `map` call: `puts map { |dino| dino.to_json }`

dino_catalog/main.rb
```diff
@@ -0,0 +1,5 @@
+require 'csv'
+require 'terminal-table'
+require_relative 'catalog'
+
+Catalog.new(ARGV[0]).run
```
zsyed1: Nice work.

dino_catalog/catalog.rb
```diff
@@ -0,0 +1,59 @@
...
+            "Time Period - Display all from a specified time period",
+            "Big - Display all big dinos",
+            "Small - Display all small dinos",
+            "Select - Select one dion to view",
+            "Reset - Remove all filter",
+            "Exit - quit program",
+            "You can also hit enter to view current list again"].join("\n")
+
+      user_input = $stdin.gets.chomp.downcase
+      done = true and next if user_input == 'exit'
```
jmmastey: Or just break out of the loop. `while true; break if user_input == exit`

PR #485, opened 2015-02-04
removed the arrowhead shenanigans

zbright:


jmmastey:
The arrowhead is gone, but this is probably still too much going on in one method. Try splitting the two checks into different methods. No need to resubmit, though.

PR #486, opened 2015-02-05
Add supportive

sriking1783:
This is supportive

### DIFF
--- a/supportive/supportive.rb
+++ b/supportive/supportive.rb
@@ -9,58 +9,56 @@ class BlagPost
   DISALLOWED_CATEGORIES = [:selfposts, :gossip, :bildungsromane]
 
   def initialize(args)
-    args = args.each_with_object({}) do |(key, value), hash|
-      hash[key.to_sym] = value
-    end
+    args.symbolize_keys!
+
     @categories = []
     if args[:author].present? && args[:author_url].present?
       @author = Author.new(args[:author], args[:author_url])
     end
 
-    if args[:categories]
-      @categories = args[:categories].reject do |category|
-        DISALLOWED_CATEGORIES.include? category
-      end
-    end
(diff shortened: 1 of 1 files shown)


supportive/supportive.rb
```diff
@@ -9,92 +9,57 @@ class BlagPost
...
 
   private
 
   def byline
-    if author.nil?
-      ""
-    else
-      "By #{author.name}, at #{author.url}"
-    end
+    author.nil? ? "" : "By #{author.name}, at #{author.url}"
```
jmmastey: Take a look at `.try`

supportive/supportive.rb
```diff
@@ -9,92 +9,57 @@ class BlagPost
...
-      when 3 then "rd"
-      else "th"
-    end
-    "You will be the #{comments.length}#{ordinal} commenter"
+    "You will be the #{commenter_number} commenter"
   end
 
   def comments_allowed?
-    publish_date + (365 * 3) > Date.today
+    publish_date > 3.years.ago.to_date
```
jmmastey: the `to_date` is superfluous, but otherwise good

supportive/supportive.rb
```diff
@@ -9,92 +9,57 @@ class BlagPost
...
   end
 
   def abstract
-    if body.length < 200
-      body
-    else
-      body[0..200] + "..."
-    end
+    return body.truncate(204) if body.length >= 200
+    body
```
jmmastey: Truncate will do the right thing if the body is too long. You can turn the entire block into `body.truncate(200)`

jmmastey:
Added a note or two. Looks good!

supportive/supportive.rb
sriking1783: I did 204, to get the effect of the original one. If I do 200, it get 197 characters and adds 3 dots at the end.
So, I did 204, :)

PR #487, opened 2015-02-05
Yadda Part 2 - Iteration 2

paulh16:
This is the second iteration of Part 2. I made the recommendation of @incazteca.

Thanks!

### DIFF
--- a/yadda/database_connector.rb
+++ b/yadda/database_connector.rb
@@ -1,56 +1,80 @@
 require 'pg'
 require 'securerandom'
+require 'faker'
 
 # Connect to database
 db_connection = PG.connect(dbname: 'yadda')
 
+# Helper Methods
+
+user_id_array = Array((1..200000))
+beer_id_array = Array((1..50))
+rating_array = Array((1..5))
+style_id_array = Array((1..7))
+
+def beer_id(id_array)
+  id_array.sample
+end
+
+def rating(id_array)
+  id_array.sample
+end
+
+def user_id(id_array)
+  id_array.sample
+end
+
+def style_id(id_array)
+  id_array.sample
+end
+
 # Create User Records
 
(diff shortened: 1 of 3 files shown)


yadda/database_connector.rb
```diff
@@ -0,0 +1,80 @@
...
+
+# Connect to database
+db_connection = PG.connect(dbname: 'yadda')
+
+# Helper Methods
+
+user_id_array = Array((1..200000))
+beer_id_array = Array((1..50))
+rating_array = Array((1..5))
+style_id_array = Array((1..7))
```
jmmastey: (1..7).to_a :)

Also, I'd probably just put this into the default on the method. `rating(id_array = (1..5).to_a)`

yadda/yadda-pg.sql
```diff
@@ -0,0 +1,120 @@
...
+CREATE VIEW you_may_enjoy AS
+  SELECT beer_styles_lookup.name AS "beer_style", beers.name AS "Beer Name",
+    AVG(((ratings.look + ratings.smell + ratings.taste + ratings.feel) / 4))
+    AS "Overall Rating"
+  FROM beers
+  INNER JOIN beer_styles_lookup on beers.style_id = beer_styles_lookup.id
+  INNER JOIN ratings on beers.id = ratings.beer_id
+  GROUP BY beer_styles_lookup.name, beers.name
+  HAVING AVG(((ratings.look + ratings.smell + ratings.taste + ratings.feel) / 4)) > 2.5
+  ORDER BY RANDOM();
```
jmmastey: For views like this, you should expose the ID of the beers and other such useful columns for filtering results. This is because you could potentially have thousands of beer styles and matching the style (vs. the ID) is nonperformant.

jmmastey:
Looks good, keep going.

PR #488, opened 2015-02-05
data science exercise with tests

zbright:


data_science/calculator.rb
```diff
@@ -0,0 +1,66 @@
...
+    @test_data_list = TestDataList.new
+  end
+
+  def setup_data(filepath)
+    test_data_list.parse(filepath)
+  end
+
+  def conversions
+    keys = get_cohorts(test_data_list)
+    Hash[keys.map { |key| [key.to_sym, cohort_conversions(key)] }]
```
jmmastey: These hash formulations are really hard to understand in general, mainly because there's so much going on on this line. There are other comments that should make it irrelevant, though :)

data_science/calculator.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def sample_size
+    keys = get_cohorts(test_data_list)
+    Hash[keys.map { |key| [key.to_sym, cohort_sample_size(key)] }]
+  end
+
+  def cohort_conversions(cohort)
+    converted_results = test_data_list.select { |data| data.result == 1 }
+    converted_results.select { |result| result.cohort == cohort }.count
```
jmmastey: This method does an unnecessary amount of iteration. To get real results, you'll iterate through the data several times. Consider that if you had a more realistic workload (millions of records), this would be totally nonperformant.

data_science/calculator.rb
```diff
@@ -0,0 +1,66 @@
...
+    rates.map { |x| (x * 100).round(2).to_s + '%' }
+  end
+
+  def cohort_sample_size(cohort)
+    test_data_list.select { |result| result.cohort == cohort }.count
+  end
+
+  def get_cohorts(results_set)
+    results_set.map(&:cohort).uniq { |x| x }
+  end
```
jmmastey: You shouldn't need to pass a block to uniq here.

data_science/calculator.rb
```diff
@@ -0,0 +1,66 @@
...
+    ABAnalyzer::ABTest.new(values).different?
+  end
+
+  def setup_confidence_hash(cohort)
+    {
+      visitors: cohort_sample_size(cohort),
+      converted: cohort_conversions(cohort),
+    }
+  end
+end
```
jmmastey: Overall, this class spends too much time iterating back and forth over the data to get cohorts. Nearly every method spends time messing around with the data in this way. This also leads to wasted computation, which would be a problem for a case like this. Given you have tests, try to refactor into adding a Cohort abstraction while keeping your tests green.

jmmastey: Also, most of these methods should be private.

data_science/main.rb
```diff
@@ -0,0 +1,26 @@
+require_relative 'calculator'
```
jmmastey: For future assignments, assume that the code should be a library, as opposed to a script like `main`. That should help you organize into better OO.

data_science/test_data_list.rb
```diff
@@ -0,0 +1,14 @@
+require 'json'
+require_relative 'test_result'
+class TestDataList < Array
```
jmmastey: Inheriting from array has some really nasty side effects. Better off not doing it.

data_science/test_data_list.rb
```diff
@@ -0,0 +1,14 @@
...
+  def parse(filepath)
+    results_hash = {}
+    File.open(filepath, "r") do |f|
+      results_hash = JSON.load(f)
+    end
+
+    results_hash.map do |result|
```
jmmastey: Don't use datatypes in variable names.

If you're using the `results_hash` return value, skip the `push`. If you're not using it, go for `each` rather than `map`.

Also, find a better way to use the File API than the one above.

data_science/rspec/calculator_spec.rb
```diff
@@ -0,0 +1,38 @@
+require_relative '../calculator'
+describe Calculator do
+  filepath = 'test_data.json'
+  calculator = Calculator.new
+  calculator.setup_data(filepath)
```
jmmastey: This belongs inside some other construct. Most likely `let`. If you let your class get initialized a little differently, this would work:

```
let(:filepath) { 'test_data.json' }
subject(:calculator) { Calculator.new(filepath) }

File: data_science/rspec/calculator_spec.rb [hunk 9]
```diff
@@ -0,0 +1,38 @@
...
+
+  it "should have a test data list" do
+    expect(calculator.test_data_list.count).to be > 0
```
jmmastey: Use idiomatic rspec matchers to get better error output and make tests more readable:

```
expect(calculator.test_data_list).not_to be_empty
```

Also, the more I see it, `test_data_list` seems too verbose. Is there a better name you can find for it?

data_science/rspec/calculator_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    expect(sample_size).not_to include(:C, :D)
+  end
+
+  it "should calculate the number of conversions" do
+    conversions = calculator.conversions
+    expect(conversions).to include(:A, :B)
+    expect(conversions).to include(A: 22)
+    expect(conversions).to include(B: 41)
+    expect(conversions).not_to include(:C, :D)
+  end
```
jmmastey: I'd be interested to see what happens to the code if you get zero conversions for A, etc. Have you tried this?

data_science/rspec/calculator_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    conversion_rates = calculator.conversion_rates
+    expect(conversion_rates).to include(:A, :B)
+    expect(conversion_rates).to include(A: ["2.33%", "5.56%"])
+    expect(conversion_rates).to include(B: ["4.5%", "8.29%"])
+    expect(conversion_rates).not_to include(:C, :D)
+  end
+
+  it "should calculate the confidence level" do
+    expect(calculator.confident?).to eq(false)
+  end
```
jmmastey: `expect(calculator).not_to be_confident`. Mostly any time you see `eq(false)` there's a better matcher to use.

jmmastey: Can you test a case where the test IS different? The biggest thing missing from the tests is sad-path verification.

data_science/rspec/test_data_list_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  it "should not have any results without a cohort" do
+    expect(test_data_list.select { |result| result.cohort.nil? }.count).to eq(0)
+  end
+
+  it "should not have any results without a date" do
+    expect(test_data_list.select { |result| result.date.nil? }.count).to eq(0)
+  end
+
+  it "should not have any results without a cohort" do
+    expect(test_data_list.select { |result| result.result.nil? }.count).to eq(0)
```
jmmastey: See above about matchers. On this, what happens if there _is_ a nil result? This test is asserting the test dataset moreso than it is anything about the parser.

data_science/rspec/test_result_spec.rb
```diff
@@ -0,0 +1,23 @@
+require_relative '../test_result'
+
+describe TestResult  do
+  context "initialize" do
+    cohort = "A"
+    date = "2014-05-19"
+    result = 1
+
+    test_result = TestResult.new(cohort, date, result)
```
jmmastey: Use `let` and `subject`

data_science/test_result.rb
```diff
@@ -0,0 +1,8 @@
+class TestResult
+  attr_accessor :cohort, :date, :result
+  def initialize(cohort, date, result)
+    @cohort = cohort
+    @date = date
+    @result = result
+  end
+end
```
jmmastey: Honestly, with no behavior, this may be better off as a `Struct`, in which case there's nothing for you to test either. If it gets behavior, then yeah it would need to be an object.

Consider, though, that if you have to build a few tens of millions of these, that's a lot of objects to track. Based on the refactor I asked about above (building a cohort object), can you drop this object entirely?

jmmastey:
Added a few notes. The tests are pretty solid. See if you can refactor into a `Cohort` object (which should simplify your code significantly), while keeping your tests green. Except for the ones you delete because you make methods private ;)

PR #490, opened 2015-02-06
Adding my version of dinodex, revised after ruby style guide exercise

plasticwrapper:
Cuz Joe Said So

### DIFF
--- a/dino_catalog/dinodex.rb
+++ b/dino_catalog/dinodex.rb
@@ -17,52 +17,6 @@ class DinoDex
     prune_period_values(merge_data(dino_data1, dino_data2))
   end
 
-  def display_all_bipeds
-    display_header "All Bipeds"
-    display_names grab_all_bipeds
-  end
-
-  def display_all_carnivores
-    display_header "All Carnivores"
-    display_names grab_all_carnivores
-  end
-
-  def display_all_jurassics
-    display_header "All Jurassics"
-    display_names grab_all_jurassics
-  end
-
-  def display_all_big_dinos
-    display_header "All Big Dino's"
-    display_names grab_all_big_dinos
-  end
-
-  def display_all_small_dinos
-    display_header "All Small Dino's"
-    display_names grab_all_small_dinos
-  end
-
-  def display_all_facts_about(dino_name)
-    display_header "All About #{dino_name}"
-    display_dino grab_dino_by_name(dino_name)
-  end
-
-  def display_dino(csv_table)
-    csv_table.each do |row|
-      row.each do |header, field|
-        if field
-          display header + ": " + field
-        end
-      end
-      display_line_break
-    end
-  end
-
-  def display_dinos(dino_list)
-    display_header "List of Dino's!"
-    display_dino dino_list
-  end
-
   def grab_all_bipeds
     filter_by_column("WALKING", "Biped")
   end
@@ -145,6 +99,63 @@ class DinoDex
     return "Carnivore" if carnivore_bool == "Yes"
     "Herbivore"
   end
+end
+
+class DinoDexPresenter
+
+  attr_accessor :dinodex
+
+  def initialize
+    @dinodex = DinoDex.new
+  end
+
+  def display_all_bipeds
+    display_header "All Bipeds"
+    display_names dinodex.grab_all_bipeds
+  end
+
+  def display_all_carnivores
+    display_header "All Carnivores"
+    display_names dinodex.grab_all_carnivores
+  end
+
+  def display_all_jurassics
+    display_header "All Jurassics"
+    display_names dinodex.grab_all_jurassics
+  end
+
+  def display_all_big_dinos
+    display_header "All Big Dino's"
+    display_names dinodex.grab_all_big_dinos
+  end
+
+  def display_all_small_dinos
+    display_header "All Small Dino's"
+    display_names dinodex.grab_all_small_dinos
+  end
+
+  def display_all_facts_about(dino_name)
+    display_header "All About #{dino_name}"
+    display_dino dinodex.grab_dino_by_name(dino_name)
+  end
+
+  def display_dino(csv_table)
+    csv_table.each do |row|
+      row.each do |header, field|
+        if field
+          display header + ": " + field
+        end
+      end
+      display_line_break
+    end
+  end
+
+  def display_dinos(dino_list)
+    display_header "List of Dino's!"
+    display_dino dino_list
+  end
+
+  private
 
   def display(message)
     puts message
@@ -172,7 +183,7 @@ class DinoDex
 end
 
 # 1. parse both csv files
-dino = DinoDex.new
+dino = DinoDexPresenter.new
 
 # 2. answer these questions
 ## display all dinosaurs that were bipeds
@@ -194,4 +205,4 @@ dino.display_all_small_dinos
 dino.display_all_facts_about "Albertosaurus"
 
 # 4. display all the dinosaurs in a given collection
-dino.display_dinos dino.grab_all_small_dinos
+dino.display_dinos dino.dinodex.grab_all_small_dinos
--- a/dino_catalog/dinodex.rb
+++ b/dino_catalog/dinodex.rb
@@ -1,0 +1,201 @@
+require 'csv' 
+
+class DinoDex
+
+  DEBUG = false
+  DINO_FILENAME1 = "dinodex.csv"
+  DINO_FILENAME2 = "african_dinosaur_export.csv"
+
+  attr_accessor :dino_data
+
+  def initialize
+    @dino_data = parse_dino_files
+  end
+
+  def parse_dino_files
+    dino_data1 = CSV.read(DINO_FILENAME1, {:headers => true})
+    dino_data2 = CSV.read(DINO_FILENAME2, {:headers => true})
+
+    prune_period_values(merge_data(dino_data1, dino_data2))
+  end
+  
+  def display_all_bipeds
+    display_header "All Bipeds"
+    display_names grab_all_bipeds
+  end
+
+  def display_all_carnivores
+    display_header "All Carnivores"
+    display_names grab_all_carnivores
+  end
+
+  def display_all_jurassics
+    display_header "All Jurassics"
+    display_names grab_all_jurassics
+  end
+
+  def display_all_big_dinos
+    display_header "All Big Dino's"
+    display_names grab_all_big_dinos
+  end
+
+  def display_all_small_dinos
+    display_header "All Small Dino's"
+    display_names grab_all_small_dinos
+  end
+
+  def display_all_facts_about(dino_name)
+    display_header "All About #{dino_name}"
+    display_dino grab_dino_by_name(dino_name)
+  end
+
+  def display_dino(csv_table)
+    csv_table.each do |row|
+      row.each do |header, field|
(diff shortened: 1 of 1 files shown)


plasticwrapper:
houndci is soooo annoying in my inbox.  need to change my notification settings.

plasticwrapper:
split up dinodex class into two classes to adhere to Sandi Metz 100-line class rule.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,207 @@
...
+dino.display_all_big_dinos
+
+## display only small dinosaurs (< 2 tons)
+dino.display_all_small_dinos
+
+# 3. display all the facts for a given dino
+dino.display_all_facts_about "Albertosaurus"
+
+# 4. display all the dinosaurs in a given collection
+dino.display_dinos dino.dinodex.grab_all_small_dinos
```
jmmastey: Boo, now I can't use this library as the inclusion of the file will cause a bunch of puts-spam.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,207 @@
+require 'csv'
+
+class DinoDex
+  DINO_FILENAME1 = "dinodex.csv"
+  DINO_FILENAME2 = "african_dinosaur_export.csv"
```
jmmastey: This works for the test dataset, but makes it really hard to use this library with any other files or data sources.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,207 @@
...
+    csv_table
+  end
+
+  def convert_carnivore(carnivore_bool)
+    return "Carnivore" if carnivore_bool == "Yes"
+    "Herbivore"
+  end
+end
+
+class DinoDexPresenter
```
jmmastey: Ah! A wild presenter appears! This belongs in a `dino_dex_presenter.rb` file.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,207 @@
...
+
+  def display_all_facts_about(dino_name)
+    display_header "All About #{dino_name}"
+    display_dino dinodex.grab_dino_by_name(dino_name)
+  end
+
+  def display_dino(csv_table)
+    csv_table.each do |row|
+      row.each do |header, field|
+        if field
```
jmmastey: This is a bit arrowhead-y. I also think it will result in a ragged table if you are missing columns, which makes me sad.

dino_catalog/dinodex.rb
plasticwrapper: you're darn right it makes it hard.  I've got quite a few lessons to go through, so I'm not making everything super extensible.  It would have been fairly trivial to add in arguments so I could load in files via command line or whatever, but I decided cut scope so I could move on to the next lesson.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,207 @@
...
+    display_names dinodex.grab_all_small_dinos
+  end
+
+  def display_all_facts_about(dino_name)
+    display_header "All About #{dino_name}"
+    display_dino dinodex.grab_dino_by_name(dino_name)
+  end
+
+  def display_dino(csv_table)
+    csv_table.each do |row|
```
jmmastey: I don't think the value _is_ a csv table here. It appears to be a hash. Any reason why you're pegging this implementation to CSV?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,207 @@
...
+        if field
+          display header + ": " + field
+        end
+      end
+      display_line_break
+    end
+  end
+
+  def display_dinos(dino_list)
+    display_header "List of Dino's!"
```
jmmastey: It's not a possessive. Dinos ;)

dino_catalog/dinodex.rb
plasticwrapper: Maybe I don't want to make a new file.  ever thought about that?

dino_catalog/dinodex.rb
plasticwrapper: display_dino takes in a CSV::Table object.  naming it anything else would make people assume they're dealing with a hash when it's definitely not.  Reason for CSV shibbidy... attempting to use as much CSV code from ruby core as possible... and finding out just how limiting the CSV library really is...

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,207 @@
...
+    end
+
+    csv_table
+  end
+
+  def convert_carnivore(carnivore_bool)
+    return "Carnivore" if carnivore_bool == "Yes"
+    "Herbivore"
+  end
+end
```
jmmastey: So overall, this class does too many things. Parsing, standardization, filtering. If you break out the CSV file details (and probably the regularization) into another file, this class gets more comprehensible and shorter.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,207 @@
...
+  end
+
+  def prune_period_values(csv_table)
+    csv_table.each do |row|
+      row["PERIOD"].gsub!("Late ", "")
+      row["PERIOD"].gsub!("Early ", "")
+    end
+
+    csv_table
+  end
```
jmmastey: You're truncating my data. BOO.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,207 @@
...
+    end
+  end
+
+  # This data transformation is brittle, be careful.
+  def merge_data(csv_table1, csv_table2)
+    csv_table2.each do |row|
+      csv_table1 << [
+        row["Genus"],                         # NAME
+        row["Period"],                        # PERIOD
+        nil,                                  # CONTINENT
```
jmmastey: Continent is definitely Africa. It says so in the filename.

dino_catalog/dinodex.rb
plasticwrapper: It's very ragged.  There's no table output at all here.. it's line by line.  one space only.  makes me sad too, but I'm going for barebones legibility rather than spend too much time on aesthetics.

dino_catalog/dinodex.rb
plasticwrapper: touché.

dino_catalog/dinodex.rb
plasticwrapper: maybe I don't want you to include this anywhere.

very well.. the execution should probably get split up into a separate file from the library itself.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,207 @@
...
+    csv_table1
+  end
+
+  def prune_period_values(csv_table)
+    csv_table.each do |row|
+      row["PERIOD"].gsub!("Late ", "")
+      row["PERIOD"].gsub!("Early ", "")
+    end
+
+    csv_table
```
jmmastey: Weird final return value. You're probably better off using `map` and omitting this part.

plasticwrapper: Hmmm I think I might not even need to do that here... I think the each enumerator would still return the csv_table object just as well as map based on the CSV::Table object's behavior so far as I use it... lines 95 and 94 can probably just be deleted...

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,207 @@
...
+
+  def filter_by_weight_less_than(weight)
+    duplicate_csv_table(@dino_data).delete_if do |row|
+      (row["WEIGHT_IN_LBS"].to_i > weight.to_i) ||
+        (row["WEIGHT_IN_LBS"].to_i == 0)
+    end
+  end
+
+  def filter_by_column(column, filter)
+    duplicate_csv_table(@dino_data).delete_if do |row|
```
jmmastey: Lots of duplication in these filter methods. You could probably just coalesce these into one match method.

jmmastey: Also, since you're just returning some data, use `.select`, rather that dup + delete.

dino_catalog/dinodex.rb
plasticwrapper: Yep.  I took advantage of your ambiguous requirements to make things easier for myself.

dino_catalog/dinodex.rb
plasticwrapper: finally.  a comment worth replying to... up until now, you've been low-balling me.  agreed.

dino_catalog/dinodex.rb
jmmastey: Ehhh, fine, but just add it as an arg? That wouldn't have lost you any time.

dino_catalog/dinodex.rb
plasticwrapper: damnit.  missed that.

dino_catalog/dinodex.rb
jmmastey: Don't inflict your laziness on other people. Maybe you don't want to write good code at all. Still should, 'eh?

dino_catalog/dinodex.rb
jmmastey: Yeah, getting it out of CSV an into a hash would be even better. This is almost-a-hash, though, and for your purposes you could still ducktype it and pretend. If I passed a hash in here, it would work.

jmmastey:
Looks fine overall. Sounds like you don't really want to split the parsing class out, but it'd make a big difference. Don't skip forward and make the same mistakes at least?

dino_catalog/dinodex.rb
plasticwrapper: You show me a select method fully functioning on CSV::Table, and I'll use it.

http://ruby-doc.org/stdlib-1.9.3/libdoc/csv/rdoc/CSV/Table.html

plasticwrapper: already working on drying up the filter methods... but the limitations of CSV::Table are getting annoying.  What was at first a quick "hey! I can use CSV library and save time!" has turned into "crap.. because i'm using CSV library I have to do all this extra stuff."

dino_catalog/dinodex.rb
plasticwrapper: no such thing as good code.  we're just bugs on a rock in a void.  why does anybody do anything?  new file it is.

dino_catalog/dinodex.rb
plasticwrapper: I'm trying to read 4 books and get through every single engineering baseline exercise by the end of February while providing you with feedback when I feel it's valuable.  adding args takes more time than I'd like to spare.  Plus, you were all like "The more senior the person, the longer it takes them to show code" so I'm deliberately putting stuff up here that I know is not super awesome.  

Then again.. it would have taken less time to add the args than the 4 comments in this thread... c'est la vie.  But, keep adding comments like this, assuming this is the type of stuff you add to all PR's for all engineers.

plasticwrapper:
So... looking for some guidance from you at this point.  What action can I take that will add the most value for you and the training program?

The majority of the things you outlined were things I strategically decided to leave as is due to "time constraints".  

I can try to barrel forward, and note your comments (all of them I actually agree with), or I can make the changes and resubmit.  

While I want to simulate what our engineers would go through as much as possible, I also want to try to cover as much material as possible in the next few weeks.  What do you suggest?  Make the changes and resubmit?  Or move on?

PR #492, opened 2015-02-06
Robot name exercise

plasticwrapper:
here we go...

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -3,28 +3,52 @@ class NameCollisionError < RuntimeError; end
 class Robot
   attr_accessor :name
 
+  NAME_ERROR_MSG = "There was a problem generating the robot name!"
+
   @@registry
 
   def initialize(args = {})
     @@registry ||= []
     @name_generator = args[:name_generator]
+    @name = pick_a_name
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      @name = generate_name
-    end
-
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
+    raise NameCollisionError, NAME_ERROR_MSG if name_issues?(@name)
 
     @@registry << @name
   end
 
-  def generate_name
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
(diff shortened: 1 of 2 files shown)


robot_name/robot_name_test.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def test_robot_name_conflict
+    assert_raises NameCollisionError do
+      generator = -> { 'AA111' }
+      Robot.new(name_generator: generator)
+      Robot.new(name_generator: generator)
+    end
+  end
+end
```
jmmastey: If you're going to add tests, use rspec :)

robot_name/robot_name.rb
```diff
@@ -3,23 +3,52 @@ class NameCollisionError < RuntimeError; end
...
     @@registry ||= []
     @name_generator = args[:name_generator]
+    @name = pick_a_name
+
+    raise NameCollisionError, NAME_ERROR_MSG if name_issues?(@name)
+
+    @@registry << @name
+  end
 
+  def pick_a_name
```
jmmastey: Most if not all of the methods under this should be private.

robot_name/robot_name.rb
```diff
@@ -3,23 +3,52 @@ class NameCollisionError < RuntimeError; end
...
     if @name_generator
-      @name = @name_generator.call
+      @name_generator.call
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+      generate_name
     end
```
jmmastey: You can use the `Kernel#method` method to turn `generate_name` into a proc, which would eliminate a conditional here.

jmmastey:
The stabby lambdas are not awesome where they are, but otherwise all is well.

robot_name/robot_name.rb
plasticwrapper: agreed.

robot_name/robot_name_test.rb
plasticwrapper: whateva whateva!  I do what I want!

Fine.

plasticwrapper:
agreed on the stabby lambdas... I wanted to just get rid of them, but decided to leave them in and move on.

PR #494, opened 2015-02-09
Overlord

katieschilling:


overlord/overlord.rb
```diff
@@ -1,14 +1,86 @@
...
 
-get '/' do
-  "Time to build an app around here. Start time: " + start_time
-end
+    after do
+      redirect to('/') unless request.path_info == '/'
+    end
+
+    get '/' do
+      session[:state] = :exploded if bomb_exploded?
```
danvayn: smart choice to use symbols instead of strings

overlord/public/style.css
```diff
@@ -0,0 +1,101 @@
+body {
+  margin:0px;
```
danvayn: if you wanna do this just copy paste http://meyerweb.com/eric/tools/css/reset/ instead

overlord/public/style.css
```diff
@@ -0,0 +1,101 @@
...
+}
+a.startanother {
```
danvayn: poor choice of class name and you don't need the hyperlink at the start -- it's a bad way to use selectors.

overlord/public/style.css
```diff
@@ -0,0 +1,101 @@
...
+  font-family: 'Lato', sans-serif;
+  font-size: 250%;
```
danvayn: use ems for text instead: http://alistapart.com/article/howtosizetextincss

overlord/public/style.css
```diff
@@ -0,0 +1,101 @@
...
+  color: #ecf0f1;
+  text-decoration: none;
+  display: block;
+  text-align: center;
+  text-transform: uppercase;
+}
+a:hover {
```
danvayn: use sass (http://sass-lang.com/) so you can nest selectors

overlord/public/style.css
```diff
@@ -0,0 +1,101 @@
...
+  color: #e74c3c;
+  margin: 30 auto;
+}
+input {
+  border-radius: 5px;
+}
+#deactivated {
+  background-image: url('http://s3.amazonaws.com/nicolascage/niccageclapshort.gif')
+}
+img {
```
danvayn: don't do this!

overlord/public/style.css
```diff
@@ -0,0 +1,101 @@
...
+  width: 60%;
+  display: block;
+  margin: 50 auto 30 auto;
+}
+#index, #started, #active {
+  background-color: #2c3e50;
+}
+.submit {
```
danvayn: target submits using input[type=submit]

overlord/overlord.rb
```diff
@@ -1,14 +1,86 @@
...
+      template = session[:state] || :index
```
danvayn: your index page doesnt seem correct. it has an activation field and a deactivation field and neither do anything for me -- it just redirects you to the "arm the bomb" page.  is there a flow problem somewhere?

overlord/overlord.rb
```diff
@@ -1,14 +1,86 @@
...
+
+      set_time_left if template == :active
+
+      erb template
+    end
+
+    post '/start' do
+      redirect_with_notice("activation")   unless validate_code(:activation_code)
```
danvayn: you dont need these fancy notice/flashes when you can just use javascript and a session key having to do with alerts. just have it so that in your js "notice.value = session param" and then if notice.value is nil or empty, hide the div containing the message (in other words, targeting the parent and then doing display: none)

overlord/overlord.rb
```diff
@@ -1,14 +1,86 @@
...
+      session[:remaining_tries] ||= 3
+      session[:state]             = :active
+    end
+
+    before '/deactivate' do
+      unless params[:deactivation_code] == session[:deactivation_code]
+        flash[:notice] = 'Wrong deactivation code'
+        session[:remaining_tries] -= 1
+
+        session[:state] = :exploded if session[:remaining_tries] < 1
```
danvayn: session[:remaining_tries].zero?

overlord/overlord.rb
```diff
@@ -1,14 +1,86 @@
...
+        redirect to('/')
+      end
+    end
+
+    post '/deactivate' do
```
danvayn: good choice with the before '/deactivate' handling the logic

overlord/overlord.rb
```diff
@@ -1,14 +1,86 @@
...
+    private
+
+    def redirect_with_notice(type)
+      flash[:notice] = "Invalid #{type} code"
+      redirect to('/')
+    end
+
+    def validate_code(type)
+      code = Code.new(params[type], type)
+      return session[type] = code.to_s unless code.to_s.nil?
```
danvayn: id just use an if statement for this...something like if code.matches?(session[type])

overlord/overlord.rb
```diff
@@ -1,14 +1,86 @@
...
+    def bomb_exploded?
+      session[:state] == :active && session[:explosion_time] - Time.now <= 0
+    end
 
-# we can shove stuff into the session cookie YAY!
-def start_time
-  session[:start_time] ||= (Time.now).to_s
+    def set_time_left
+      @seconds_left = (session[:explosion_time] - Time.now).to_i
+      @time_left = Time.at(@seconds_left).utc.strftime("%H:%M:%S")
```
danvayn: stating this here isn't very necessary, right?

overlord/public/countdown.js
```diff
@@ -0,0 +1,26 @@
+countdown = function(seconds_left){
+  var countdown = document.getElementById('countdown');
```
danvayn: oh gosh use jquery

overlord/public/countdown.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+  }, 1000);
+};
+
+Number.prototype.toHHMMSS = function(){
+  var hours = Math.floor(this / 3600);
+  var minutes = Math.floor((this - (hours * 3600)) / 60);
+  var seconds = this - (hours * 3600) - (minutes *60);
+
+  if (hours   < 10) { hours   = "0"+hours; }
```
danvayn: i think there are better ways to set up this function, but i think you can figure that out.

overlord/public/jquery.tubular.1.0.js
```diff
@@ -0,0 +1,148 @@
...
+|* http://www.seanmccambridge.com/tubular
+|* version: 1.0
+|* updated: October 1, 2012
+|* since 2010
+|* licensed under the MIT License
+|* Enjoy.
+|* 
+|* Thanks,
+|* Sean */
+
```
danvayn: btw you dont need this if you want good video handling -- everything is in here https://developers.google.com/youtube/js_api_reference

overlord/public/style.css
```diff
@@ -0,0 +1,101 @@
...
+.submit:active {
+  top: 1px;
+  outline: none;
+  -webkit-box-shadow: none;
+  box-shadow: none;
+}
+.submit:hover {
+  cursor: pointer;
+}
+#activated_submit {
```
danvayn: delete rules you dont use, and dont use ids randomly

overlord/public/style.css
```diff
@@ -0,0 +1,101 @@
...
+  background: #27ae60;
+}
+.right {
+  float: right;
+}
+.code_inputs {
+  width:60%;
```
danvayn: you dont have to worry about this junk if you just use a barebones grid system

overlord/spec/code_spec.rb
```diff
@@ -0,0 +1,56 @@
+require_relative '../lib/code'
+
+describe Code do
```
danvayn: YAY TESTS

overlord/spec/code_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  describe '#initialize' do
+    context 'without any arguments' do
+      it 'raises an error' do
+        expect { described_class.new }.to raise_error
```
danvayn: for proper test coverage include the error message

overlord/spec/code_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  let(:code)    {}
+  let(:type)    {}
+
+  describe '#valid' do
+    let(:subject) { described_class.new(code, type) }
+
+    context 'when input is numeric' do
+      let(:code) { number }
+
+      it { should be_valid }
```
danvayn: is this correct rspec syntax?

overlord/spec/code_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  let(:number)  { '1234' }
+  let(:letters) { 'abcd' }
+  let(:code)    {}
+  let(:type)    {}
+
+  describe '#valid' do
+    let(:subject) { described_class.new(code, type) }
+
+    context 'when input is numeric' do
```
danvayn: dont describe a method, describe a scenario where the method will have use.

eg describe a submitted code --> context "is the correct (numeric format)", context "is empty", context "contains letters" (and then get the message returned by the error)

overlord/spec/code_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    context 'when input is empty' do
+      let(:code) { "" }
+
+      it { should_not be_valid }
+    end
+  end
+
+  describe '#to_s' do
+    it 'returns input if self is numeric and not empty' do
+      expect(described_class.new(number, "").to_s).to eq number
```
danvayn: man i dont like described class but im definitely no test expert so its cool

overlord/views/active.erb
```diff
@@ -0,0 +1,24 @@
+<%= erb :_header, :layout => false %>
+<body id="active">
+  <%= erb :_flash, :layout => false %>
+  <img id="malkovich_bunny" src="http://s3.amazonaws.com/nicolascage/malkovich_bunny.jpg">
```
danvayn: dont give your images ids pleaseeeeeeeee you can target an image with class .hero or something since theres no difference between one image and the next in this app

overlord/views/active.erb
```diff
@@ -0,0 +1,24 @@
...
+  <span id="countdown"><%= @time_left %></span>
+  <div class="code_inputs">
+    <form action="/deactivate" method="POST">
+      <span><%= session[:remaining_tries] %> attempts left</span>
+      <br/>
```
danvayn: dont use br tags, theres plenty of solutions you can do instead

overlord/views/active.erb
```diff
@@ -0,0 +1,24 @@
...
+      <span>
```
danvayn: dont use empty spans, use a div with class "row" or something lol

overlord/views/active.erb
```diff
@@ -0,0 +1,24 @@
...
+        <label for="deactivation_code">Deactivation Code</label>
+        <input id="deactivation_code" class="right" type="number" name="deactivation_code">
+      </span>
+      <br/>
```
danvayn: you know you're not doing html right when you have to write an extra tag to compliment the tag setup you just used

overlord/views/active.erb
```diff
@@ -0,0 +1,24 @@
...
+      <input id="activated_submit" class="submit" type="submit" value="Submit">
+    </form>
+  </div>
+</body>
+
+<script type="text/javascript">
+  window.onload = function(){ countdown(<%= @seconds_left %>); };
```
danvayn: you can do this externally and id say thats probably better to have

overlord/lib/code.rb
```diff
@@ -0,0 +1,47 @@
+class Code
```
danvayn: i like this class

overlord/lib/code.rb
```diff
@@ -0,0 +1,47 @@
...
+  def initialize(input, type)
+    @input = input
+    @type  = type
+  end
+
+  def to_s
+    if valid?
+      input
+    elsif present?
+      nil
```
danvayn: just pass "incorrect" or something, returning nil never really seems like a great idea

overlord/views/deactivated.erb
```diff
@@ -0,0 +1,7 @@
+<%= erb :_header, :layout => false %>
+
+<body id="deactivated">
```
danvayn: you should just redirect them to the home page like the rest of this app at this point. like check in the "before '/'" that if their bomb exploded, reset then

overlord/views/exploded.erb
```diff
@@ -0,0 +1,14 @@
...
+
+<body id="exploded">
+  <div id="wrapper" class="black-50">
+    <a class="startanother" href="/reset">Start another bomb</a>
+  </div>
+</body>
+
+<script type="text/javascript">
+  $().ready(function() {
+    $('#wrapper').tubular({videoId: 'oUltheRpJNw', mute: false, start: 26});
```
danvayn: if you use jquery here you coulda used it earlier! aughhgwuhg

overlord/views/started.erb
```diff
@@ -0,0 +1,23 @@
+<%= erb :_header, :layout => false %>
+<body id="started">
+  <%= erb :_flash, :layout => false %>
+  <div>
+    <img id="bomb" src="bomb.jpg">
```
danvayn: no id plz..

overlord/views/started.erb
```diff
@@ -0,0 +1,23 @@
+<%= erb :_header, :layout => false %>
```
danvayn: i think you used this wrong, id check how other overlords did partials

overlord/views/started.erb
```diff
@@ -0,0 +1,23 @@
...
+    <div class="code_inputs">
+      <form action="/activate" method="POST">
+        <span>
+          <label for="activation_code">Activation Code</label>
+          <input id="activation_code" class="right" type="number" name="activation_code">
+        </span>
+        <br/>
+        <span>
+          <label for="countdown">Countdown Seconds</label>
+          <input id="countdown" class="right" type="number" name="countdown">
```
danvayn: dont have a class called "right" please, if your inputs are always on the right of your label you can just give it class "bomb-input" and have the bomb input float right

overlord/views/started.erb
```diff
@@ -0,0 +1,23 @@
+<%= erb :_header, :layout => false %>
+<body id="started">
+  <%= erb :_flash, :layout => false %>
+  <div>
+    <img id="bomb" src="bomb.jpg">
+    <span class="title">Activate Bomb</span>
+    <div class="code_inputs">
```
danvayn: this div should be in place of that empty span below

overlord/features/reset_bomb.feature
```diff
@@ -0,0 +1,24 @@
...
+    And I click "Submit"
+
+  Scenario: after deactivation
+    Given I fill in "Deactivation Code" with "1111"
+    And I click "Submit"
+    When I click link "Start another bomb"
+    Then I should be on the "index" page
+
+  Scenario: after explosion
+    Given I fill in "Deactivation Code" with "2222"
```
danvayn: i think you can dry up this code

overlord/features/reset_bomb.feature
```diff
@@ -0,0 +1,24 @@
+Feature: Resetting the bomb
+  Background: Bomb has been activated
+    Given I start the app
+    And I fill in "Activation Code" with "5678"
```
danvayn: i know your index has some probs but writing around it dont help much

overlord/overlord.rb
katieschilling: this way is simpler.

overlord/features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,12 @@
+
```
djkotowski: Avoid using the prebuilt web steps.  Cucumber no longer includes them by default as they are considered too implementation-based and can lead to fragile tests.

overlord/features/activate_bomb.feature
```diff
@@ -0,0 +1,28 @@
+Feature: Activating a bomb
+  Background: Bomb has already been started
+    Given I start the app
+    And I fill in "Activation Code" with "5678"
+    And I fill in "Deactivation Code" with "1111"
+    And I click "Submit"
+    
+  Scenario: Valid activation code but no countdown time
+    When I fill in "Activation Code" with "5678"
```
djkotowski: While explicitly stating the codes will work, if you change the codes in the background, you will have to change these codes too.  It generally is preferred to replace these web steps with something like

```cucumber
When I activate the bomb
```

and let your step defintions concern themselves with filling in fields and clicking on buttons.

djkotowski: Doing it this way also lets you get rid of the repeated

```cucumber
And I click "Submit"
```

overlord/features/reset_bomb.feature
```diff
@@ -0,0 +1,24 @@
...
+    And I fill in "Deactivation Code" with "1111"
+    And I click "Submit"
+    And I fill in "Activation Code" with "5678"
+    And I click "Submit"
```
djkotowski: These steps are confusing.  Something like "Given I have booted and activated the bomb" would be clearer.

overlord/features/step_definitions/start_bomb_steps.rb
```diff
@@ -0,0 +1,14 @@
...
+When /^I start the app$/ do
+  visit "/"
+end
+
+Then /^I should see "([^\"]*)"$/ do |content|
+  expect(page).to have_content(/#{content}/i)
+end
+
+Then /^I should be on the "([^\"]*)" page$/ do |p|
+  expect(page).to have_css("##{p}")
```
djkotowski: This step requires the feature to have more knowledge about the implementation than it should.  I'd at least split this step into multiple (one for 'started', one for 'exploded') rather than search for the CSS using the value provided in the feature step.

overlord/lib/code.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def default
+    case type
+      when :activation_code
+        '1234'
+      when :deactivation_code
+        '0000'
+      when :countdown
+        '30'
+    end
```
djkotowski: Given this is just a mapping of symbols to values, why not make this a hash?

overlord/spec/code_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+  end
+
+  let(:number)  { '1234' }
+  let(:letters) { 'abcd' }
+  let(:code)    {}
+  let(:type)    {}
+
+  describe '#valid' do
+    let(:subject) { described_class.new(code, type) }
```
djkotowski: If you want a subject, use

```ruby
subject(:var_name) { described_class.new(code, type) }
```

overlord/spec/code_spec.rb
djkotowski: Use `expect` syntax.

```ruby
it { is_expected.to be_valid }
```

overlord/spec/code_spec.rb
djkotowski: It's better than writing the class inline, as then a class name change will affect more lines of code.

overlord/views/_tubular.erb
```diff
@@ -0,0 +1,3 @@
+<script src="//ajax.googleapis.com/ajax/libs/jquery/1.8.0/jquery.min.js" type="text/javascript"></script>
```
djkotowski: No need anymore to include `type` attributes for `<script>` tags.

overlord/overlord.rb
```diff
@@ -1,14 +1,86 @@
...
+    end
+  end
 end
```
djkotowski: Consider getting some of the business logic out of the Sinatra application and into its own class (e.g. `Bomb`).  It will clean up your application and make it a bit more readable.  Then, you can have your session serialize your bomb, thus having only one session variable (bomb) instead of multiple.

djkotowski:
This is looking good.  I'd refactor a couple things, though.  Your features are very implementation-focused.  The features should be written in English with the business user as the target audience.  Button-clicking and field-filling are too low-level of a concern.  I'd also look at refactoring your Sinatra application.  There's a lot of business logic in the Sinatra app itself that could be extracted to another class.  Let your Sinatra application concern itself only with retrieving and storing from the session, retrieving user input as params, and rendering the view to send back to the user.

Take a look at the UI/JS comments too, but no need to refactor those.

overlord/lib/code.rb
katieschilling: The case statement provides more flexibility.... which I don't need there.

PR #495, opened 2015-02-09
Zb data

zbright:


data_science/calculator.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def conversion_rates_data
+    Hash[cohorts.map { |cohort| [cohort.name, cohort.conversion_rate] }]
+  end
+
+  def sample_size_data
+    Hash[cohorts.map { |cohort| [cohort.name, cohort.sample_size] }]
+  end
+
```
zsyed1: You can change the method names from `def conversions_data` to `def conversions`, `def sample_sizes_data` to `def sample_sizes`, and `def conversion_rates_data` to `def conversion_rates`. Then you can erase lines 12 - 15!

data_science/calculator.rb
```diff
@@ -0,0 +1,74 @@
...
+  def confident?
+    values = {}
+    values[:agroup] = setup_confidence_hash(cohorts.first)
+    values[:bgroup] = setup_confidence_hash(cohorts.last)
+
```
zsyed1: Avoid using data types in variable and method names unless it is directly applicable. Instead of `setup_confidence_hash` you can simply call it `setup_confidences`. An example of where it is directly applicable is a method like `to_hash` that explicitly converts something to a hash (then too it is usually called `to_h`). In your method, the _hash_ is incidental. Let me know if this is unclear - hit me up on slack.

data_science/calculator.rb
```diff
@@ -0,0 +1,74 @@
+require_relative 'test_data_list'
+require_relative 'cohort'
+class Calculator
+  attr_accessor :test_data_list, :cohorts
+
+  def initialize(filepath)
+    @test_data_list = TestDataList.new(filepath).data
+    setup_cohorts(test_data_list)
```
zsyed1: I would remove the word `List` from the class name and the variable name. Please see comment about putting data types in variable names.

data_science/calculator.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def print_stats(sample_sizes, conversions, conversion_rates)
+    puts ["Cohort A has a sample size of #{sample_sizes[:A]}",
+          "Cohort A has #{conversions[:A]} conversions",
+          "Cohort A has a conversion rate between: #{conversion_rates[:A][0]} and #{conversion_rates[:A][1]}",
+          "",
+          "Cohort B has a sample size of #{sample_sizes[:B]}",
+          "Cohort B has #{conversions[:B]} conversions",
+          "Cohort B has a conversion rate between: #{conversion_rates[:B][0]} and #{conversion_rates[:B][1]}",
```
zsyed1: In the method `print_stats` you are repeating the same 4 lines 55-58 and 59-62. So instead you can make a method `def print_stats_for(cohort)` and then call it twice in `print_stats` passing in `:A` and `:B` respectively.

data_science/rspec/calculator_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      expect(calculator.test_data_list).not_to be_empty
+    end
+
+    it "should calculate sample size" do
+      sample_size = calculator.sample_size_data
+      expect(sample_size).to include(:A, :B)
+      expect(sample_size).to include(A: 558)
+      expect(sample_size).to include(B: 641)
+      expect(sample_size).not_to include(:C, :D)
+    end
```
zsyed1: Interesting, I hadn't seen this usage `expect(...).to include(...)` for hashes and keys

data_science/rspec/calculator_spec.rb
```diff
@@ -0,0 +1,47 @@
+require_relative '../calculator'
+describe Calculator do
+  context "happy" do
+    let(:filepath) { 'test_data.json' }
+    subject(:calculator) { Calculator.new(filepath) }
+
+    it "should have a test data list" do
+      expect(calculator.test_data_list).not_to be_empty
```
zsyed1: Just FYI there's a debate about using the word _should_ in tests. One party prefers `it 'has a test data list' do`.

data_science/calculator.rb
```diff
@@ -0,0 +1,74 @@
...
+    raise 'Only 2 Cohorts allowed!' if cohort_names.count != 2
+    self.cohorts = cohort_names.map { |name| Cohort.new(name, test_data_list) }
+  end
+
+  def setup_confidence_hash(cohort)
+    {
+      visitors: cohort.sample_size,
+      converted: cohort.conversion_count,
+    }
+  end
```
zsyed1: This method doesn't really "setup" anything. It creates the confidence-hash. So a better name might be `confidence_hash`. But please see note about putting data types in the names of variables :-)

data_science/calculator.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    ABAnalyzer::ABTest.new(values).different?
+  end
+
+  private
+
+  def setup_cohorts(test_data_list)
+    cohort_names = test_data_list.map(&:cohort).uniq
+    raise 'Only 2 Cohorts allowed!' if cohort_names.count != 2
+    self.cohorts = cohort_names.map { |name| Cohort.new(name, test_data_list) }
```
zsyed1: Use `unless cohort_names.count == 2` instead of the `if !=`

data_science/rspec/calculator_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    it "should calculate sample size" do
+      sample_size = calculator.sample_size_data
+      expect(sample_size).to include(:A, :B)
```
zsyed1: Line 12 should be a `let` statement above `let(:sample_size) { calculator.sample_size }`

data_science/rspec/calculator_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      expect(sample_size).to include(A: 558)
+      expect(sample_size).to include(B: 641)
+      expect(sample_size).not_to include(:C, :D)
+    end
+
+    it "should calculate the number of conversions" do
+      conversions = calculator.conversions_data
+      expect(conversions).to include(:A, :B)
```
zsyed1: Use `let` statement to define `conversions` (see above note on `sample_size`)

data_science/rspec/calculator_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      expect(conversions).to include(A: 22)
+      expect(conversions).to include(B: 41)
+      expect(conversions).not_to include(:C, :D)
+    end
+
+    it "should calculate the conversion rates" do
+      conversion_rates = calculator.conversion_rates_data
+      expect(conversion_rates).to include(:A, :B)
```
zsyed1: Use `let` statement

data_science/rspec/test_data_list_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    let(:filepath) { 'test_data.json' }
+    subject(:test_data_list) { TestDataList.new(filepath).data }
+
+    it "should parse a json file of test result data" do
+      expect(test_data_list).not_to be_empty
+    end
+
+    it "should not have any results without a cohort" do
+      expect(test_data_list.select { |result| result.cohort.nil? }).to be_empty
+    end
```
zsyed1: Perhaps you could add a method to `TestDataList` called `def valid?` that checks that no cohorts are `nil`. Then these three tests would shrink to one (lines 12 -22). This is the fine the way it is, I was just suggesting an alternative that came to mind.

data_science/calculator.rb
zsyed1: You don't use `test_data_list` after the initilizer. Is there a reason why you are storing it? Also making it accessible (`attr_accessor`)?

data_science/test_data_list.rb
```diff
@@ -0,0 +1,15 @@
...
+  TestResult = Struct.new(:cohort, :date, :result)
+  attr_accessor :data
+
+  def initialize(filepath)
+    @data = []
+    raise "File Not Found!" unless File.exist?(filepath)
+
+    json_data = JSON.parse(File.read(filepath))
+    json_data.each do |result|
+      data.push(TestResult.new(result['cohort'], result['date'], result['result']))
```
zsyed1: Nice work overall!

data_science/calculator.rb
djkotowski: This doesn't even involve confidence.  This is just putting visits & conversions in a format recognized by ABAnalyzer.

data_science/calculator.rb
djkotowski: I'd also rather see a heredoc than a joined array of string literals.  I could see joining on output from the two cohorts.

data_science/rspec/calculator_spec.rb
```diff
@@ -0,0 +1,47 @@
+require_relative '../calculator'
+describe Calculator do
+  context "happy" do
```
djkotowski: I generally like to avoid using contexts to separate "happy" and "sad" paths.  When the test results are displayed, the tests will appear as "Calculator happy should have a test data list" and so on.

data_science/rspec/test_data_list.rb
```diff
@@ -0,0 +1,32 @@
...
+describe TestData do
+  context "happy" do
+    let(:filepath) { 'test_data.json' }
+    subject(:test_data) { TestData.new(filepath).data }
+
+    it "should parse a json file of test result data" do
+      expect(test_data).not_to be_empty
+    end
+
+    it "should not have any results without a cohort" do
```
djkotowski: This and the "happy" tests below it are actually "sad" path tests.

data_science/rspec/test_data_list.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    it "should not have any results without a result" do
+      expect(test_data.select { |result| result.result.nil? }).to be_empty
+    end
+  end
+
+  context "sad" do
+    let(:filepath) { 'fake.file' }
+
+    it "should raise an error is the file does not exist" do
```
djkotowski: If you're raising an error, it is a "bad" path.

data_science/rspec/cohort_spec.rb
```diff
@@ -0,0 +1,9 @@
+require_relative '../cohort'
+describe Cohort do
+  context 'sad' do
```
djkotowski: Actually bad path.

data_science/rspec/cohort_spec.rb
```diff
@@ -0,0 +1,9 @@
...
+    it 'should raise an error if initialized with nil values' do
+      expect {  Cohort.new(nil, []) }.to raise_error
+      expect {  Cohort.new("cohort_name", nil) }.to raise_error
+    end
+  end
```
djkotowski: You should test with good data too.  This spec file passes even if `Cohort.new` always raises an error.

djkotowski:
This looks good.  Take a look at the comments, especially about the differences in happy, sad, and bad paths.  I don't think this needs a resubmission.

PR #497, opened 2015-02-09
Dinodex

gguntupalli:
Please review my dinodex exercise.

Thanks
Gopi.

### DIFF
--- a/dino_catalog/dino_catalog.rb
+++ b/dino_catalog/dino_catalog.rb
@@ -1,0 +1,38 @@
+require 'csv'
+require_relative 'dinosaur'
+
+class DinoCatalog
+  attr_reader :dinosaurs
+
+  def initialize
+    CSV::HeaderConverters[:header_converter] = method(:header_converter)
+    dinos_hash = DinoCatalog.parse_dinodex("dinodex.csv")
+                 .concat(DinoCatalog.parse_african_dinosaurs)
+    @dinosaurs = dinos_hash.inject([]) do |dinos, dino|
+      dinos << Dinosaur.new(dino)
+    end
+  end
+
+  def self.parse_dinodex(filename)
+    csv = CSV.read(filename, headers: true,
+          header_converters: [:downcase, :header_converter])
+    csv.map(&:to_hash)
+  end
+
+  def self.parse_african_dinosaurs
+    african_dinos_hash = parse_dinodex("african_dinosaur_export.csv")
+    african_dinos_hash.each do |row|
+      row["continent"] = "Africa"
+      row["diet"] = row["diet"] == "Yes" ? "Carnivore" : "Herbivore"
+    end
+  end
+
+  def header_converter(header)
+    case header
+      when "genus" then "name"
+      when "weight_in_lbs" then "weight"
+      when "carnivore" then "diet"
+      else header
+    end
+  end
+end
--- a/dino_catalog/dinodex.rb
+++ b/dino_catalog/dinodex.rb
@@ -1,0 +1,48 @@
+require_relative 'dino_catalog'
+
+class Dinodex
+  attr_reader :dinos
+
+  def initialize
+    dino_catalog = DinoCatalog.new
+    @dinos = dino_catalog.dinosaurs
+  end
+
+  def parse_input
+    #Initializing the filtered dinos with all dinos initially
+    filtered_dinos = @dinos
+
+    puts "Dinosaurs with #{ARGV}" unless ARGV.empty?
+    ARGV.each do|a|
+      criteria = a.split(':')
+      filtered_dinos = search_by_criteria(criteria[0], criteria[1], filtered_dinos)
+    end
+    print_dinos(filtered_dinos)
+  end
+
+  def search_by_criteria(header, value, filtered_dinos)
+    #Assumes that weight is only called with big or small parameters
+    if header == 'weight'
+      value.downcase == 'big' ? filtered_dinos.select(&:big?)
+                              : filtered_dinos.select(&:small?)
+    else
+      filtered_dinos.select do |dino|
+        #Used include instead of == to not differentiate between early and late Cretaceous
+        dino.send(header).downcase.include? (value.downcase)
+      end
+    end
+  end
+
+  def print_dinos(results)
+    results.each do |dino|
+      puts dino.to_s
+    end
+  end
+end
+
+def main
+  dinodex = Dinodex.new
+  dinodex.parse_input
+end
+
+main
--- a/dino_catalog/dinosaur.rb
+++ b/dino_catalog/dinosaur.rb
@@ -1,0 +1,31 @@
+class Dinosaur
+  attr_accessor :name, :period, :continent, :diet,
+    :weight, :walking, :description
+
+  def initialize(attributes)
+    @name = attributes["name"]
+    @period = attributes["period"]
+    @continent = attributes["continent"]
+    @diet = attributes["diet"]
+    @weight = attributes["weight"].to_i
+    @walking = attributes["walking"]
+    @description = attributes["description"]
+  end
+
+  def big?
+    @weight > 2000
+  end
+
+  def small?
+    @weight <= 2000
+  end
+
+  def to_s
+    s = "#{@name} is a #{@walking} from #{@period} period and #{@continent} "
+    s << "continent with #{@diet} diet"
+    s << " and #{@weight}lbs weight" unless @weight == 0
+    s << ". "
+    s << @description unless @description.nil?
+    s
+  end
+end
--- a/dino_catalog/how_to_run.txt
+++ b/dino_catalog/how_to_run.txt
@@ -1,0 +1,20 @@
+# to print all dinos without any filter
+ruby dinodex.rb
+
+# to print bipeds
+ruby dinodex.rb walking:biped
+
+# to print carnivores
+ruby dinodex.rb diet:carnivore
+
+# to print dinosaurs of specific period
+ruby dinodex.rb period:Cretaceous
+
+# to print big dinosaurs
+ruby dinodex.rb weight:big
+
+# to print big carnivores
+ruby dinodex.rb	weight:big diet:carnivore
+
+# to print all facts for given dino
+ruby dinodex.rb name:Giganotosaurus
--- a/dino_catalog/how_to_run.txt
+++ b/dino_catalog/how_to_run.txt
@@ -1,20 +1,89 @@
-# to print all dinos without any filter
-ruby dinodex.rb
+# to print all facts of a given dino
+ruby dinodex.rb name:Giganotosaurus
+Dinosaurs with ["name:Giganotosaurus"]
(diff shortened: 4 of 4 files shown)


dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
+require_relative 'dino_catalog'
+
+class Dinodex
+  attr_reader :dinos
+
+  def initialize
+    dino_catalog = DinoCatalog.new
```
sshao: creating a DinoCatalog object just to access its `dinosaurs` attribute and then discarding it entirely is weird. more importantly, it suggests that it's not *really* a "dinosaur catalogue" -- especially given that the Dinodex object clearly fulfills that role already.

your DinoCatalog object is actually more of a formatter or parser object (as suggested, too, by its main methods -- `parse_dinodex` and `parse_african_dinosaurs`). you should rename and restructure it accordingly.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,38 @@
+require 'csv'
+require_relative 'dinosaur'
+
+class DinoCatalog
+  attr_reader :dinosaurs
+
+  def initialize
+    CSV::HeaderConverters[:header_converter] = method(:header_converter)
+    dinos_hash = DinoCatalog.parse_dinodex("dinodex.csv")
```
sshao: hardcoding the filenames dinodex.csv and african_dinosaur_export.csv is a bad idea, as it ties your code down to expect those two filenames exactly. what if the filenames change, or what if the user wants to feed more files and more data into the dinodex? better to change this object and its methods so that it accepts an arbitrary filename as an argument.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+  end
+
+  def self.parse_dinodex(filename)
+    csv = CSV.read(filename, headers: true,
+          header_converters: [:downcase, :header_converter])
+    csv.map(&:to_hash)
+  end
+
+  def self.parse_african_dinosaurs
```
sshao: you have two separate parse methods for 'dinodex' and for 'african'; so essentially you have a DinodexParser object and a separate AfricanParser object. instead of hiding them in methods, you should define them explicitly as their own classes

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,38 @@
+require 'csv'
+require_relative 'dinosaur'
+
+class DinoCatalog
+  attr_reader :dinosaurs
+
+  def initialize
+    CSV::HeaderConverters[:header_converter] = method(:header_converter)
+    dinos_hash = DinoCatalog.parse_dinodex("dinodex.csv")
+                 .concat(DinoCatalog.parse_african_dinosaurs)
```
sshao: when you refactor, consider where this `concat` should belong to -- in your formatter/parser, or outside of it? is the `concat` the responsibility of a formatter/parser class?

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,38 @@
...
+    @dinosaurs = dinos_hash.inject([]) do |dinos, dino|
```
sshao: `@dinosaurs = dinos_hash.map { |dino| Dinosaur.new(dino) }`

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,38 @@
...
+  def self.parse_dinodex(filename)
+    csv = CSV.read(filename, headers: true,
+          header_converters: [:downcase, :header_converter])
+    csv.map(&:to_hash)
+  end
+
+  def self.parse_african_dinosaurs
+    african_dinos_hash = parse_dinodex("african_dinosaur_export.csv")
+    african_dinos_hash.each do |row|
+      row["continent"] = "Africa"
```
sshao: this can be taken out and made into a converter method that you can then pass to `CSV.read(converters: <converter_method>)`

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,38 @@
...
+      row["diet"] = row["diet"] == "Yes" ? "Carnivore" : "Herbivore"
```
sshao: this can be taken out and made into a converter method that you can then pass to `CSV.read(converters: <converter_method>)`

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
...
+    @dinos = dino_catalog.dinosaurs
+  end
+
+  def parse_input
+    #Initializing the filtered dinos with all dinos initially
```
sshao: if your method names/variable names/code is clear enough, you shouldn't need to add comments like this

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
...
+    filtered_dinos = @dinos
+
+    puts "Dinosaurs with #{ARGV}" unless ARGV.empty?
```
sshao: `parse_input` is doing too much; the name suggests that all it should be doing is taking some user input and reading it, but you have it taking user input, reading it, then manipulating it to find dinos and output them back to the user. in general you want your methods to be small and to perform a single specific action

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
...
+    ARGV.each do|a|
+      criteria = a.split(':')
+      filtered_dinos = search_by_criteria(criteria[0], criteria[1], filtered_dinos)
```
sshao: what is `criteria[0]` and `criteria[1]`? what actual parts of a dinosaur, or a search query, do they refer to? if you come back to this line in another couple weeks, will you know what `criteria[0]` and `criteria[1]` mean?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+    print_dinos(filtered_dinos)
+  end
+
+  def search_by_criteria(header, value, filtered_dinos)
+    #Assumes that weight is only called with big or small parameters
+    if header == 'weight'
```
sshao: i'd suggest using the `:symbol` converter in `CSV.read` so that all your headers become symbols. symbols make more sense as -- at this point in execution -- headers should be fixed and immutable

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
...
+      value.downcase == 'big' ? filtered_dinos.select(&:big?)
```
sshao: yep, listen to hound on this one

sshao: also, since you've got this whole separate if-case for finding dinosaurs by weight, maybe take out lines 26-27 into their own `search_by_weight(weight)` method

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
...
+                              : filtered_dinos.select(&:small?)
+    else
+      filtered_dinos.select do |dino|
+        #Used include instead of == to not differentiate between early and late Cretaceous
+        dino.send(header).downcase.include? (value.downcase)
```
sshao: does this account for the requirement (in the README) that your dinodex can "Grab all the dinosaurs that were carnivores (fish and insects count)" ?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+    end
+  end
+
+  def print_dinos(results)
```
sshao: instead of print_dinos, this could be a `Dinodex#to_s` method. instead of returning an array of dinosaurs, your find method could return a new Dinodex instead, and then with a to_s method implemented you could simply call `puts dinodex` on it. 

additionally, if your find method returns a new Dinodex, you could then chain searches without having to explicitly pass `filtered_dinos` as an argument: `dinodex.search_by_criteria(...).search_by_criteria(...).search...`

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
...
+    results.each do |dino|
+      puts dino.to_s
+    end
+  end
+end
+
+def main
+  dinodex = Dinodex.new
+  dinodex.parse_input
```
sshao: as mentioned, parse_input is doing too much. i'd actually move a lot of parse_input straight into the `main` method, and then i might even move the main method into its own script/class

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
...
+end
+
+main
```
sshao: if this is the way you invoke your command line, i'd *definitely* abstract out the `main` method and much of the `parse_input` functionality into its own class/script

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,31 @@
...
+    :weight, :walking, :description
+
+  def initialize(attributes)
+    @name = attributes["name"]
+    @period = attributes["period"]
+    @continent = attributes["continent"]
+    @diet = attributes["diet"]
+    @weight = attributes["weight"].to_i
+    @walking = attributes["walking"]
+    @description = attributes["description"]
```
sshao: an alternative to doing this (since you have your attributes defined as attr_accessors):

`attributes.each { |attr, value| send("#{attr}=", value) }` 

on the other hand, what you have already is probably more readable

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def big?
+    @weight > 2000
```
sshao: `2000` is a magic number; define it as a constant instead, then use that constant on this line

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def small?
+    @weight <= 2000
```
sshao: `!big?`

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,31 @@
+class Dinosaur
+  attr_accessor :name, :period, :continent, :diet,
+    :weight, :walking, :description
+
+  def initialize(attributes)
+    @name = attributes["name"]
+    @period = attributes["period"]
+    @continent = attributes["continent"]
+    @diet = attributes["diet"]
+    @weight = attributes["weight"].to_i
```
sshao: some dinos have no weight attribute defined in the CSV. what does this do with empty weight values? 

if an empty weight comes in as `nil`, then `nil.to_i = 0`, so it seems like the results for `search_by_criteria("weight", "small")` would also include dinosaurs that have no weight. that would be incorrect :)

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,31 @@
...
+    @weight <= 2000
+  end
+
+  def to_s
+    s = "#{@name} is a #{@walking} from #{@period} period and #{@continent} "
+    s << "continent with #{@diet} diet"
+    s << " and #{@weight}lbs weight" unless @weight == 0
+    s << ". "
+    s << @description unless @description.nil?
+    s
```
sshao: `String#<<` returns the string with the argument appended to it, so you can remove line 29 and this method will still work as expected

sshao:
review my comments, take a shot at refactoring this, then resubmit for another review :+1:

PR #499, opened 2015-02-10
Dino catalog

shishirtewari:
Looking for initial feedback.

### DIFF
--- a/dino_catalog/dino.rb
+++ b/dino_catalog/dino.rb
@@ -127,6 +127,7 @@ if ARGV.length == 0
   puts ''
   puts '#multiple conditions AND search'
   puts '  walking.biped and diet.Carnivore'
+  puts "  diet.carnivore and walking.biped and period.Cretaceous and continent.'north America'"
   puts ''
   puts '#Dinosaurs facts search'
   puts '  facts.Albertosaurus'
--- a/dino_catalog/dino.rb
+++ b/dino_catalog/dino.rb
@@ -1,0 +1,148 @@
+require 'csv'
+require 'json'
+
+
+class Dinosaurs
+
+  attr_accessor :csv_1, :csv_2,  :final_csv
+
+  def initialize(csv1_path, csv2_path)
+
+    @csv_1 = csv_reader(csv1_path)
+    @csv_2 = csv_reader(csv2_path)
+
+    # before starting merge modify csv2 for diet values
+    @csv_2.each do |row|
+      row['diet'] = row['diet'] == 'Yes' ? 'Carnivore' : nil
+    end
+
+    # merge csv into 1
+    # pass array and return hash for each row.
+    @final_csv =[@csv_1, @csv_2].flat_map(&method(:csv_to_hash_array))
+  end
+
+  def csv_reader(csv_path)
+    CSV.read(csv_path, headers: true, :header_converters => lambda { |h| header_changer(h) })
+  end
+
+  def header_changer(header)
+    case header
+      when 'Genus'
+        header.sub('Genus', 'name')
+      when 'Carnivore'
+        header.sub('Carnivore', 'diet')
+      when 'WEIGHT_IN_LBS'
+        header.sub('WEIGHT_IN_LBS', 'weight')
+      else header.downcase
+    end
+  end
+
+  def csv_to_hash_array(csv)
+    csv.to_a[1..-1].map do |row|
+      Hash[csv.headers.zip(row)]
+    end
+  end
+
+  def search
+    strHash = {}
+    arrARGV = ARGV.map(&:downcase)
+    if arrARGV.length ==1
+      k,v = arrARGV[0].split('.')
+      case
+        when k == 'facts'
+          search_facts(v)
+        when k == 'collection'
+          csv_print(v)
+        when k == 'json'
+          puts final_csv.map { |o| Hash[o.each_pair.to_a] }.to_json
+        when k == 'hashsearch'
+          multiple_search(eval(v.gsub(':','=>')))
+        else
+          strHash[k] = v
+          multiple_search(strHash)
+      end
+    else
+      arrARGV.each.reject{ |s| s=='and' }.each { |str| k, v = str.split('.')
+                                                strHash[k] = v
+                                               }
+      multiple_search(strHash)
+    end
+  end
+
+
+  def multiple_search(search_hash)
+    matches = final_csv.select do |row|
+      match = true
+      search_hash.keys.each do |key|
+        #puts is_row_match(row, search_hash, key)
+        match = match && is_row_match(row, search_hash, key)
+      end
+      match
+    end
+    puts matches.map{|h| h['name']}
+  end
+
+  def is_row_match(row, search_hash, key)
+    if key == 'size' && search_hash[key] == 'big'
+      row['weight'].to_i > 2000
+    elsif key == 'size' && search_hash[key] == 'small'
+      row['weight'].to_i <= 2000
+    else
+      #we have a requirement to count Insectivore and Piscivore in Carnivore so add in
+      if row[key] == 'Insectivore' || row[key] == 'Piscivore'
+        row[key] = 'Carnivore, '+ row[key]
+      end
+      (row[key].to_s.downcase || '').include?(search_hash[key])
+    end
+  end
+
+  def search_facts(name)
+    matches2 = final_csv.select{|a| a['name'].to_s.downcase == name}
+    facts_hash = Hash[*matches2.collect{|h| h.to_a}.flatten]
+    facts_hash.each { |key, value| unless value.nil?
+                                     puts "#{key}: #{value}"
+                                     end }
+  end
+
+  def csv_print(filename)
+    case filename
+      when 'african_dinosaur_export'
+        puts csv_2['name']
+      when 'dinodex'
+        puts csv_1['name']
+    end
+  end
+end
+
+if ARGV.length == 0
+
+  puts 'You can search on following conditions'
+  puts ''
+  puts '#column criteria search'
+  puts '  walking.biped'
+  puts '  diet.Carnivore'
+  puts '  period.Cretaceous'
+  puts '  size.small'
+  puts '  size big'
+  puts ''
+  puts '#multiple conditions AND search'
+  puts '  walking.biped and diet.Carnivore'
+  puts ''
+  puts '#Dinosaurs facts search'
+  puts '  facts.Albertosaurus'
+  puts ''
+  puts '#collection print'
+  puts '  collection.dinodex'
+  puts '  collection.african_dinosaur_export'
+  puts ''
+  puts '#hashsearch'
+  puts " ruby dino.rb hashsearch.\"{'walking': 'Biped', 'continent': 'Europe'}\""
+  puts ''
+  puts '#json print'
+  puts '  json.export'
+  puts ''
+  exit 1
+end
+
+dino = Dinosaurs.new('dinodex.csv', 'african_dinosaur_export.csv')
+dino.search


shishirtewari:
Removed non-dino_catalog exercises from the branch.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
+require 'csv'
+require 'json'
+
+
+class Dinosaurs
```
sshao: your `Dinosaurs` class does *everything* :P in general you want to aim for more modular classes/objects -- each class/object should have one specific responsibility.

right now, `Dinosaurs` is responsible not only for reading/formatting the data, but searching/manipulating it as well. so for a start, you could split out those two responsibilities into their own classes. `Formatter` and `DinoCatalog`, maybe. you should also split out lines 117-149 into their own functionality, too -- even if it's just `run_dinocatalog.rb`. 

doing so will makes more code more maintainable and readable. right now when i see `Dinosaurs`, i expect the class to represent a collection of Dinosaurs that *maybe* includes a search API -- seeing that it parses CSVs and user input is quite surprising to me.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+
+  attr_accessor :csv_1, :csv_2,  :final_csv
```
sshao: these aren't great variable names -- and also, why bother saving csv_1/csv_2 in the first place? after the initial parsing, will they ever be used again?

sshao: as for `final_csv`, that should probably be renamed. think about what that object actually represents. look at line 21 -- it's not even a "csv" after that point.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+
+  def initialize(csv1_path, csv2_path)
```
sshao: this method isn't extensible. it's very specific to the scenario that there are 2 CSV files that you want to parse into the dinocatalog. if you only want to parse one, or three, or more, you'll have to rewrite this entire method -- or copy-and-paste a bunch of lines.

consider instead creating a method/class that handles reading a single CSV file, and then using that to read multiple CSV files. for example, something like:

```ruby
class Formatter
  def self.format(filepath)
    # returns the results of parsing the contents of filepath
  end
end

class DinoCatalog
  def initialize(*filepaths)
    filepaths.each do |fp|
      Formatter.format(fp)
      # do something with the above
    end
  # do more stuff with the above
  end
end
```

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+
+    @csv_1 = csv_reader(csv1_path)
+    @csv_2 = csv_reader(csv2_path)
+
+    # before starting merge modify csv2 for diet values
+    @csv_2.each do |row|
+      row['diet'] = row['diet'] == 'Yes' ? 'Carnivore' : nil
+    end
```
sshao: this could be its own method

```ruby
def convert_diet(csv)
  csv.map do |row|
    row['diet'] = row['diet'] == 'Yes' ? 'Carnivore' : nil
  end
end
```

also, you can go ahead and assume that if a dinosaur isn't a carnivore, it's a herbivore :P

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+
+    # merge csv into 1
+    # pass array and return hash for each row.
+    @final_csv =[@csv_1, @csv_2].flat_map(&method(:csv_to_hash_array))
```
sshao: this works (i assume) but seems overly complicated. `flat_map` isn't a method i see very often. furthermore, i doubt that `csv_to_hash_array` is needed as a method: [CSV::Row#to_hash](http://ruby-doc.org//stdlib-2.2.0.preview1/libdoc/csv/rdoc/CSV/Row.html#method-i-to_hash) exists.

`csv1.map(&:to_hash) + csv2.map(&:to_hash)` should give you the exact same thing, and imo is clearer in purpose

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+
+  def header_changer(header)
+    case header
+      when 'Genus'
+        header.sub('Genus', 'name')
+      when 'Carnivore'
+        header.sub('Carnivore', 'diet')
+      when 'WEIGHT_IN_LBS'
+        header.sub('WEIGHT_IN_LBS', 'weight')
+      else header.downcase
```
sshao: this is ok, but again `header_changer` seems to be doing a lot ... it's explicitly transforming some headers (genus -> name) and also downcasing headers?

you should be able to pass an array of header converter methods to `CSV.read`, and i believe `:downcase` can be one of them.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+
+  def csv_to_hash_array(csv)
+    csv.to_a[1..-1].map do |row|
+      Hash[csv.headers.zip(row)]
+    end
+  end
+
+  def search
+    strHash = {}
+    arrARGV = ARGV.map(&:downcase)
```
sshao: you want better variable names here again. what sort of arguments should be coming in through ARGV? `query`/`queries`, maybe?

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+    if arrARGV.length ==1
+      k,v = arrARGV[0].split('.')
```
sshao: eh... requiring cmd line inputs to come in this "k.v" form is really weird, i don't think i've ever run into a command-line program asking for input in that form. it's not really a big deal, but you might want to consider something more intuitive.

sshao: also, you might want to consider writing a method that takes in the ARGV inputs and transforms them to a more standard form (like a hash, for example)

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+      case
+        when k == 'facts'
+          search_facts(v)
+        when k == 'collection'
+          csv_print(v)
+        when k == 'json'
+          puts final_csv.map { |o| Hash[o.each_pair.to_a] }.to_json
```
sshao: `final_csv.map(&:to_json)`?

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+        when k == 'hashsearch'
+          multiple_search(eval(v.gsub(':','=>')))
```
sshao: oh no, listen to hound on this one! there is absolutely no reason for you to need `eval` here. it's not appropriate, and in general `eval` is [dangerous](http://stackoverflow.com/questions/637421/is-eval-supposed-to-be-nasty).

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+        else
+          strHash[k] = v
+          multiple_search(strHash)
+      end
```
sshao: the logic of this if-else chain is very confusing. the root cause is probably that the categorizations of user inputs is strange:

* why is searching for facts separate from searching by any other dinosaur attribute? 'facts' is just another attribute of dinosaur, after all.
* the 'collection' and 'json' types both relate to *printing* out dinosaurs. the main thing the user wants out of both those functions is that they *print* something -- and the user *may* define a specific format for the output. try writing a method like `print_dinos(format)` and using that instead, where `format` can either be 'default' or 'json'.
* likewise, 'hashsearch' and your regular search are also both related to *searching* dinosaurs, with the difference being what sort of format the user makes their query in. therefore, it should also be handled by the same functionality that handles the other searches. you should take the user input, figure out if it's a search -- of any format (key.value or {'key': 'value'}) -- transform the query into some standard form, and then perform the actual search.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+    else
+      arrARGV.each.reject{ |s| s=='and' }.each { |str| k, v = str.split('.')
```
sshao: agree with hound on this one. you want the `do/end` format instead.

sshao: also, figure out some way of getting rid of this nested case statement/outer 'else' -- this control flow is really hard to parse. 

some things to consider that might help:

* ideally, a `multiple_search` method should also be able to handle a *single* search... they shouldn't need to be separate cases. 
* once more emphasizing that the only 2 real options available to the user are "print some dinos" and "search for some dinos". try taking the initial user input and identifying which of those two options the user is trying to execute, and then structure your control flow from there, ie:

```ruby
raw_input = ARGV
if # raw_input is a search query
  # possibly do something with raw_input
  search(...)
else
  # possibly do something with raw_input
  print(...)
end
```

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+                                                strHash[k] = v
+                                               }
```
sshao: also, too much going on in this one line of code, and too much in this method. 

```
def query_to_hash(query)
  Hash[query.reject { |s| s == 'and' }.map { |s| s.split(".") }]
end

File: dino_catalog/dino.rb [hunk 14]
```diff
@@ -0,0 +1,149 @@
...
+  def multiple_search(search_hash)
+    matches = final_csv.select do |row|
+      match = true
+      search_hash.keys.each do |key|
+        #puts is_row_match(row, search_hash, key)
+        match = match && is_row_match(row, search_hash, key)
+      end
+      match
+    end
+    puts matches.map{|h| h['name']}
```
sshao: a search method shouldn't also be responsible for printing out the search -- you've got a printer method for that. `multiple_search` should just return `matches`.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+
+  def multiple_search(search_hash)
+    matches = final_csv.select do |row|
+      match = true
+      search_hash.keys.each do |key|
+        #puts is_row_match(row, search_hash, key)
+        match = match && is_row_match(row, search_hash, key)
+      end
+      match
+    end
```
sshao: same stuff as before. take out parts of the above code, make them into their own methods. lines 76-79 are prime candidates for this.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+    puts matches.map{|h| h['name']}
+  end
+
+  def is_row_match(row, search_hash, key)
```
sshao: ^ related to what hound's saying: ruby's got a nice coding convention in that methods that return boolean values can/should end in `?`: see [here](http://stackoverflow.com/questions/1345843/ruby-question-mark-usage).

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+    if key == 'size' && search_hash[key] == 'big'
+      row['weight'].to_i > 2000
```
sshao: 2000 is a magic number -- where did it come from? what does it mean? why 2000? assign it to a constant instead, with a descriptive constant variable name.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+    elsif key == 'size' && search_hash[key] == 'small'
+      row['weight'].to_i <= 2000
+    else
+      #we have a requirement to count Insectivore and Piscivore in Carnivore so add in
+      if row[key] == 'Insectivore' || row[key] == 'Piscivore'
+        row[key] = 'Carnivore, '+ row[key]
```
sshao: again, this stuff (and the weight-matching stuff) above could all be their own methods ... `matches_weight?`, `matches_diet?`

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+      (row[key].to_s.downcase || '').include?(search_hash[key])
+    end
+  end
+
+  def search_facts(name)
+    matches2 = final_csv.select{|a| a['name'].to_s.downcase == name}
+    facts_hash = Hash[*matches2.collect{|h| h.to_a}.flatten]
+    facts_hash.each { |key, value| unless value.nil?
+                                     puts "#{key}: #{value}"
+                                     end }
```
sshao: still not convinced that `search_facts` needs to be a separate method from `multiple_search`.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,149 @@
...
+  end
+
+  def csv_print(filename)
+    case filename
+      when 'african_dinosaur_export'
+        puts csv_2['name']
+      when 'dinodex'
+        puts csv_1['name']
```
sshao: this is hard-coded logic, again very dependent upon one specific scenario and one specific usage. note that 'dinodex' and 'african_dinosaur_export' are attributes of *something* -- maybe an attribute of the dinosaur data itself, similar to name/diet/continent. you could add a "source" attribute to your dino hashes, and then this method would become something like:

```ruby
def csv_print(source)
  puts final_csv.select { |dino| dino[:source] == source }
end
```

sshao:
review my comments, take a shot at refactoring your code, and then resubmit for another round of review :+1:

PR #501, opened 2015-02-11
Dino catalogue

theundeadmonk:


dino_catalog/DInodex.rb
```diff
@@ -0,0 +1,132 @@
...
+
+class Dinosaur
+
+    attr_reader :name, :period, :period_prefix, :diet, :weight, :movement, :continent, :description
+    def extract_period(period)
+        if /(.+)\s(.+)/.match(period)
+            @period_prefix = $1
+            @period = $2
+        else
+            @period_prefix = nil
```
zsyed1: Please fix hound errors. If you need help running hound, hit me up on slack.

dino_catalog/DInodex.rb
```diff
@@ -0,0 +1,132 @@
...
+    def initialize(dino)
+        @name = dino[0]
+        extract_period(dino[1])
+        @diet = dino[3]
+        @weight = dino[4]
+        @movement = dino[5]
+        @continent = dino[6]
+        @description = dino[7]
+    end 
+end
```
zsyed1: Normally the constructor (`initialize`) is the first instance method in the file.

dino_catalog/DInodex.rb
```diff
@@ -0,0 +1,132 @@
...
+
+class African_Dinosaur < Dinosaur
+    
+    def is_carnivore?(diet)
+        if diet == 'Yes'
+            "Carnivore"
+        else
+            "Herbivore"
+        end
```
zsyed1: Use the shortened form:
```
return 'Carnivore' if diet == 'Yes'
'Herbivore'
```

zsyed1: This method returns the diet of a dinosaur. So a better name might be `convert_diet` or something. The name `is_carnivore?` implies a boolean is being returns.

dino_catalog/DInodex.rb
```diff
@@ -0,0 +1,132 @@
...
+        @dino=[]
+        @dino[0] = dino['Genus']
+        @dino[1] = dino['Period']
+        @dino[2] = nil
+        @dino[3] = is_carnivore?(dino['Carnivore'])
+        @dino[4] = dino['Weight']
+        @dino[5] = dino['Walking']
+        @dino[6] = "Africa"
+        @dino[7] = nil
+
```
zsyed1: You probably do not want to index the attributes of a dinosaur by integers. For example, what does index 2 store? Kind of hard to read. It would be better to store the attributes in a hash keyed by symbol:

```
@dino = {}
@dino[:genus] = dino['Genus']
@dine[:period} = dino['Period']
...
```

Also, if later on someone tries to access a key that does not exist in the hash, it will return `nil`. So lines like 56 and 61 are not needed.

dino_catalog/DInodex.rb
```diff
@@ -0,0 +1,132 @@
...
+    search_terms = eval(params.gsub(':','=>'))
+    $Dinodex.each do|key, dinosaur|
+        @flag = false
+        search_terms.each do|search_key, search_value|
+            if search_value == dinosaur.send(search_key.to_s)
+                @flag = true
+            else
+                @flag = false
+            end
+        end
```
zsyed1: I think your logic is off here. What you want to do is set `@flag = true` in line 114. If the condition in line 116 is ever false then you want to set `@flag = false`. The way you have it now, it will print the dinosaur only based on the last criterion.

About the coding:
The variable `@flag` does not need to be a member variable so no need for the `@` sign.
Lines like 116 - 120 can be replaces with a single line: `@flag = (search_value == dinosaur.send(search_key.to_s))`
Lines 114 -121 can be replaced with:
```
flag = search_terms.all? { |key, value| value == dinosaur.send(key.to_s) }
```
And finally, in line 122 you can write `if @flag`. There is no need for `== true` for booleans. Likewise to check for false you do `if !@flag`.

djkotowski:
New PR opened (see #503)

PR #503, opened 2015-02-12
Implements Dinodex

theundeadmonk:
I deleted my old repo. Problems with git.

### DIFF
--- a/dino_catalog/Dino_Methods.rb
+++ b/dino_catalog/Dino_Methods.rb
@@ -1,126 +1,139 @@
 module Dino_Methods
 
+$OPTION = [1,2,3,4]
+
 require 'json'    
 
-    def print_menu(option)
-        printer = Hash.new
-        printer[1] = "1) All dinosaurs that were bipeds"
-        printer[2] = "2) All dinosaurs that were carnivores"
-        printer[3] = "3) Grab dinosaurs by time period"
-        printer[4] = "4) Grab dinosaurs by size" 
-        puts("What would you like to search by?")
-        for i in option
-            puts(printer.fetch(i))
-        end
+  def print_menu(option)
+    printer = Hash.new
+    printer[1] = "1) All dinosaurs that were bipeds"
+    printer[2] = "2) All dinosaurs that were carnivores"
+    printer[3] = "3) Grab dinosaurs by time period"
+    printer[4] = "4) Grab dinosaurs by size" 
+    puts("What would you like to search by?")
+    for i in option
+      puts(printer.fetch(i))
     end
-        
-    def menu(dinos)
-        option = [1,2,3,4]
-        user_option = []
-        @size = nil
-        @period = nil
-        loop do      
-            option_temp = option
-            option_temp = option_temp - user_option
-            print_menu(option_temp)
-            temp = gets.chomp.to_i
-            if temp < 1 ||temp > 4
-                puts("Please pick a valid option")
-                next
-            else
-                user_option << temp
-                case temp
-                when 3
-                    puts("Please enter the period")
-                    @period = gets.chomp
-                when 4
-                    loop do
-                        puts("Big or small? B/S")
-                        @size = gets.chomp
-                        if @size == 'B' || @size == 'b'
-                            @size = 'Big'
-                            break
-                        elsif @size == 'S' || @size == 's'
-                            @size = 'Small'
-                            break
-                        else
-                            puts("Please enter a valid size B/S")
-                        end
-                    end
-                end
-                if user_option.length < 4
-                    puts("Would you like to chain another option Y/N")
-                    user_selection = gets.chomp
-                end
-                if user_selection == 'Y'
-                    next
-                else
-                    search(dinos, user_option, @period, @size)
-                    user_option = []
-                end
-            end
-        end
+  end
+  
+  def get_period
+    puts("Please enter the period")
+    period = gets.chomp
+  end
+  
+  def get_size
+    puts("Big or Small? B/S")
+    loop do 
+      size = gets.chomp
+      if size == 'B' || size == 'b'
+        size = 'Big'
+        break
+      elsif size == 'S' || size == 's'
+        size = 'Small'
+        break
+      else
+        puts("Valid sizes are B/S Please re-enter an appropriate size")  
+      end
     end
-    
-    def print_dino(dinosaur)
-        puts("Name: "+dinosaur.name)
-        if dinosaur.period_prefix != nil
-            puts("Period: "+dinosaur.period_prefix+" "+dinosaur.period)
-        else
-            puts("Period: "+dinosaur.period)
-        end
-        puts("Diet: "+dinosaur.diet) if dinosaur.diet!=nil
-        puts("Weight: "+dinosaur.weight.to_s)if dinosaur.weight!=nil
-        puts("Movement: "+dinosaur.movement) if dinosaur.movement != nil
-        puts("Continent: "+dinosaur.continent)
-        puts("Description: "+dinosaur.description) if dinosaur.description != nil
-        puts
-    end        
+    size
+  end
 
-    def search(dino, option, period, size)
-        @json = nil
-        dino.each do |key, dinosaur|
-            @flag = false
-            option.each do|indx|
-                case indx
-                when 1
-                    if  dinosaur.movement == 'Biped'
-                        @flag = true
-                    else
-                        @flag = false
-                    end
-                when 2
-                    if dinosaur.diet == 'Carnivore'||dinosaur.diet == 'Insectivore'||dinosaur.diet == 'Piscivore'
-                        @flag = true
-                    else
-                        @flag = false
-                    end                        
-                when 3 
-                    if dinosaur.period == period
-                        @flag = true
-                    else
-                        @flag = false
-                    end
-                when 4
-                    if size == 'Small' && dinosaur.weight.to_i <= 2000
-                        @flag = true
-                    elsif size == 'Big' && dinosaur.weight.to_i > 2000
-                        @flag = true
-                    else
-                        @flag = false
-                    end
-                end
-            end
-            if @flag == true
-                print_dino(dino.fetch(key))
-                @json = @json .to_s + dino.fetch(key).to_json
-            end
-        end
-        File.open('output.json', 'w') do |file|
-            file.puts @json
-            puts "JSON output file output.json has been created"
-        end
+  def get_user_selection
+    puts("Would you like to chain another option Y/N")
+    user_selection = gets.chomp
+    if user_selection == 'Y'|| user_selection == 'y'
     end
+  end
+    
+      
+  def menu(dinos)
+    user_option = []
+      @size = nil
+      @period = nil
+      loop do      
+      option_temp = option_temp - user_option
+      print_menu(option_temp)
+      temp = gets.chomp.to_i
+      if temp < 1 || temp > 4
+        puts("Please pick a valid option")
+        next
+      else
+        user_option << temp
+        case temp
+          when 3
+            @period = get_period
+          when 4
+            @size = get size
+        end
+      end
+        if user_option.length < 4
+          if get_user_selection
+            next
+          else
+            search(dinos, user_option, @period, @size)
+            user_option = []
+          end
+        end
+      end
+  end
+    
+  def print_dino(dinosaur)
+    puts("Name: "+dinosaur.name)
+    puts("Diet: "+dinosaur.diet) if dinosaur.diet!=nil
+    puts("Weight: "+dinosaur.weight.to_s)if dinosaur.weight!=nil
+    puts("Movement: "+dinosaur.movement) if dinosaur.movement != nil
+    puts("Continent: "+dinosaur.continent)
+    puts("Description: "+dinosaur.description) if dinosaur.description != nil
+    puts
+  end        
+
+  def match_period(dino, period)
+    dino.period.each do |index|
+      if dino.period[index].include? period
+        @flag = true
+      end
+    end
+    @flag ||= false
+  end
+
+  def match_size(dinosaur, size)
+    @flag = (size == 'Small' && dinosaur.weight.to_i <= 2000 ) || (size == 'Big' && dinosaur.weight.to_i > 2000)
+  end
+
+  def search(dino, option, period, size)
+    @json = nil
+    dino.each do |key, dinosaur|
+      @flag = false
+      option.each do|indx|
+        case indx
+        when 1
+          @flag = (dinosaur.movement == 'Biped')
+        when 2
+          @flag = (dinosaur.diet == 'Carnivore'||dinosaur.diet == 'Insectivore'||dinosaur.diet == 'Piscivore')
+        when 3 
+          @flag = match_period(dinosaur, period)
+        when 4
+          @flag = match_size(dinosaur, size)
+        end
+      end
+      if @flag 
+        print_dino(dino.fetch(key))
+        @json = @json .to_s + dino.fetch(key).to_json
+      end
+    end
+    File.open('output.json', 'w') do |file|
+      file.puts @json
+      puts "JSON output file output.json has been created"
+    end
+  end
+  def split_period period
+    unless /(.+)\sor\s(.+)/.match(period)
+      @period << dino['PERIOD']
+    else
+      @period << Regexp.last_match(1)
+      split_period Regexp.last_match(2)
+    end
+  end 
 end
                 
                         
--- a/dino_catalog/Dino_Methods.rb
+++ b/dino_catalog/Dino_Methods.rb
@@ -1,0 +1,130 @@
+module Dino_Methods
+
+require 'json'    
+
+    def print_menu(option)
+        printer = Hash.new
+        printer[1] = "1) All dinosaurs that were bipeds"
+        printer[2] = "2) All dinosaurs that were carnivores"
+        printer[3] = "3) Grab dinosaurs by time period"
+        printer[4] = "4) Grab dinosaurs by size" 
+        puts("What would you like to search by?")
+        for i in option
+            puts(printer.fetch(i))
+        end
+    end
+        
+    def menu(dinos)
+        option = [1,2,3,4]
+        user_option = []
+        @size = nil
+        @period = nil
+        loop do      
+            option_temp = option
+            option_temp = option_temp - user_option
+            print_menu(option_temp)
+            temp = gets.chomp.to_i
+            if temp < 1 ||temp > 4
+                puts("Please pick a valid option")
+                next
+            else
+                user_option << temp
+                case temp
+                when 3
+                    puts("Please enter the period")
+                    @period = gets.chomp
+                when 4
+                    loop do
+                        puts("Big or small? B/S")
+                        @size = gets.chomp
+                        if @size == 'B' || @size == 'b'
+                            @size = 'Big'
+                            break
+                        elsif @size == 'S' || @size == 's'
+                            @size = 'Small'
+                            break
+                        else
+                            puts("Please enter a valid size B/S")
+                        end
+                    end
+                end
+                if user_option.length < 4
+                    puts("Would you like to chain another option Y/N")
+                    user_selection = gets.chomp
+                end
+                if user_selection == 'Y'
+                    next
+                else
+                    search(dinos, user_option, @period, @size)
+                    user_option = []
+                end
+            end
+        end
+    end
+    
+    def print_dino(dinosaur)
+        puts("Name: "+dinosaur.name)
+        if dinosaur.period_prefix != nil
+            puts("Period: "+dinosaur.period_prefix+" "+dinosaur.period)
+        else
+            puts("Period: "+dinosaur.period)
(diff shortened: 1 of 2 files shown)


dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
+module Dino_Methods
```
jmmastey: Yep, hound is correct. Also, the filename for `DinoMethods` should `dino_methods.rb`.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+  def print_menu(option)
+    printer = Hash.new
+    printer[1] = "1) All dinosaurs that were bipeds"
+    printer[2] = "2) All dinosaurs that were carnivores"
+    printer[3] = "3) Grab dinosaurs by time period"
+    printer[4] = "4) Grab dinosaurs by size" 
+    puts("What would you like to search by?")
+    for i in option
+      puts(printer.fetch(i))
+    end
```
jmmastey: For this sort of thing, it's usually better just to puts all the statements:

```
puts "What do you want?"
puts "1) ..."
puts "2) ..."
```

I get that it's about option deletion, but we can probably handle that other ways.

Also, `for i in hash` is not idiomatic ruby. Typically, we'd use `.each`:

```
messages.each { |msg| puts msg unless option_taken?(msg) }
```

That last bit is my lame attempt at the same output filtering.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
+module Dino_Methods
+
+$OPTION = [1,2,3,4]
+
+require 'json'    
+
+  def print_menu(option)
+    printer = Hash.new
```
jmmastey: You *can* do a hash with integer keys, but it's really non-idiomatic. That's really honestly an array. It looks like this was so that you can have non-zero-based indices, but it'd be better to use a semantic string key (e.g. `"bipeds"`) and then kill the options that way.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+    printer[4] = "4) Grab dinosaurs by size" 
+    puts("What would you like to search by?")
+    for i in option
+      puts(printer.fetch(i))
+    end
+  end
+  
+  def get_period
+    puts("Please enter the period")
+    period = gets.chomp
```
jmmastey: This is true. The last statement on a line is automatically returned.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+  end
```
jmmastey: In general, in ruby we try not to name methods `get_` and `set_`. Getters are "queries" in the Command / Query Separation view of the world, and they should be given a noun name that reflects their contents. In this case `user_supplied_period` would be a decent choice of a name.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+        break
+      elsif size == 'S' || size == 's'
+        size = 'Small'
+        break
+      else
+        puts("Valid sizes are B/S Please re-enter an appropriate size")  
+      end
+    end
+    size
+  end
```
jmmastey: loop/do is somewhat unidiomatic. take a look at using something like `retry` to accomplish this. Example:

```
def size_filter
  puts "Big or Small?"
  case gets.chomp
    when /b/i then "Big"
    when /s/i then "Small"
    else raise InvalidArgumentError
  end
rescue InvalidArgumentError
  puts "Valid sizes are B/S Please re-enter an appropriate size"
  retry
end
```

Notice that most of the cruft (like `break` statements) just go away. Also, don't use parens for puts, or mostly any method with zero or one arg.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  def get_user_selection
+    puts("Would you like to chain another option Y/N")
+    user_selection = gets.chomp
+    if user_selection == 'Y'|| user_selection == 'y'
+    end
+  end
```
jmmastey: Does this actually work? I think this is broken.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+        if user_option.length < 4
+          if get_user_selection
+            next
+          else
+            search(dinos, user_option, @period, @size)
+            user_option = []
+          end
+        end
+      end
+  end
```
jmmastey: So this method has some structural issues. Mainly, it's way too long and has too much nested logic. This becomes obvious when you see and realize that the indentation is muxed up in multiple places but it's hard to see. Split out this method into multiple smaller methods that each handle one concern. For instance, repeatedly trying for a user option until they give a real one is one logical method.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+      if temp < 1 || temp > 4
+        puts("Please pick a valid option")
+        next
+      else
+        user_option << temp
+        case temp
+          when 3
+            @period = get_period
+          when 4
+            @size = get size
```
jmmastey: This is definitely broken.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+      @size = nil
+      @period = nil
+      loop do      
+      option_temp = option_temp - user_option
+      print_menu(option_temp)
+      temp = gets.chomp.to_i
+      if temp < 1 || temp > 4
+        puts("Please pick a valid option")
+        next
+      else
```
jmmastey: In the arrowhead exercise, you'll see a fix for this pattern. The `if if if else else else` pattern is really tough to read.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+    
+  def print_dino(dinosaur)
+    puts("Name: "+dinosaur.name)
+    puts("Diet: "+dinosaur.diet) if dinosaur.diet!=nil
+    puts("Weight: "+dinosaur.weight.to_s)if dinosaur.weight!=nil
+    puts("Movement: "+dinosaur.movement) if dinosaur.movement != nil
+    puts("Continent: "+dinosaur.continent)
+    puts("Description: "+dinosaur.description) if dinosaur.description != nil
+    puts
+  end
```
jmmastey: Define a `to_s` on `Dinosaur` and then just do `puts dinosaur`. In this case, the DinoMethods module knows too much about the details of manipulating dinosaurs.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+        end
+      end
+  end
+    
+  def print_dino(dinosaur)
+    puts("Name: "+dinosaur.name)
+    puts("Diet: "+dinosaur.diet) if dinosaur.diet!=nil
+    puts("Weight: "+dinosaur.weight.to_s)if dinosaur.weight!=nil
+    puts("Movement: "+dinosaur.movement) if dinosaur.movement != nil
+    puts("Continent: "+dinosaur.continent)
```
jmmastey: Why do you assume continent will never be nil if everything else might be?

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+  end        
+
+  def match_period(dino, period)
+    dino.period.each do |index|
+      if dino.period[index].include? period
+        @flag = true
+      end
+    end
+    @flag ||= false
+  end
```
jmmastey: A few possible suggestions. The direct replacement for what you're doing:

```
def matches_period?(dino, period)
  dino.period.any? { |chunk| chunk.include? period }
end
```

Or if you're comfortable with regex (and switch `dino.period` to be a string, which is probably reasonable):

```
def matches_period?(dino, period)
  dino.period =~ period
end
```

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+    puts("Name: "+dinosaur.name)
+    puts("Diet: "+dinosaur.diet) if dinosaur.diet!=nil
+    puts("Weight: "+dinosaur.weight.to_s)if dinosaur.weight!=nil
+    puts("Movement: "+dinosaur.movement) if dinosaur.movement != nil
+    puts("Continent: "+dinosaur.continent)
+    puts("Description: "+dinosaur.description) if dinosaur.description != nil
+    puts
+  end        
+
+  def match_period(dino, period)
```
jmmastey: If a method returns a boolean, we call it a Predicate Method. In ruby, we name predicate methods with a question-mark suffix. That lets other folks know what sort of value to expect back, and once you get into testing it gives you some nice extra abilities with those methods.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+    dino.period.each do |index|
+      if dino.period[index].include? period
+        @flag = true
+      end
+    end
+    @flag ||= false
+  end
+
+  def match_size(dinosaur, size)
+    @flag = (size == 'Small' && dinosaur.weight.to_i <= 2000 ) || (size == 'Big' && dinosaur.weight.to_i > 2000)
```
jmmastey: This `@flag` stuff is confusing, as it gets set in a ton of places and then used in one place (with that loop/do construct). Instead, in your loop, use methods like `.select` and then just return a value from these predicate methods.

If you can keep your methods from mutating state (like setting instance vars), they're 100x easier to test. In this case, this class would become very unwieldy to test as it has tons of side effects that are not obvious or intuitive until you read the whole class.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  def get_user_selection
+    puts("Would you like to chain another option Y/N")
+    user_selection = gets.chomp
+    if user_selection == 'Y'|| user_selection == 'y'
+    end
+  end
+    
+      
+  def menu(dinos)
```
jmmastey: The name `menu` is really deceptive here, as the only thing this method _doesn't_ do is display a menu.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+      end
+    end
+    @flag ||= false
+  end
+
+  def match_size(dinosaur, size)
+    @flag = (size == 'Small' && dinosaur.weight.to_i <= 2000 ) || (size == 'Big' && dinosaur.weight.to_i > 2000)
+  end
+
+  def search(dino, option, period, size)
```
jmmastey: This method duplicates a ton of functionality from the `menu` method. Instead, try to use only this method for searching, and remove all the duplicated search functionality from  `menu`.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+    @json = nil
```
jmmastey: Rather than aggregating here, the method should be arranged for clarity as follows:

```
def search
  // get all dinos that match the criteria
  // iterate over them using .map to retrieve the attributes we want
  // call .to_json on the whole thing
end
```

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+      end
+      if @flag 
+        print_dino(dino.fetch(key))
+        @json = @json .to_s + dino.fetch(key).to_json
+      end
+    end
+    File.open('output.json', 'w') do |file|
+      file.puts @json
+      puts "JSON output file output.json has been created"
+    end
```
jmmastey: This should be in a separate method.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+  end
+  def split_period period
```
jmmastey: Add blank lines between method definitions.

dino_catalog/Dino_Methods.rb
```diff
@@ -0,0 +1,143 @@
...
+    unless /(.+)\sor\s(.+)/.match(period)
+      @period << dino['PERIOD']
+    else
+      @period << Regexp.last_match(1)
+      split_period Regexp.last_match(2)
+    end
+  end
```
jmmastey: This method appears to be unused in this file. Why does it exist here?

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
+require 'csv'
+require 'json'
+require_relative 'Dino_Methods'
```
jmmastey: `Dinosaur` should not depend on `DinoMethods`, since the latter appears to be just a runner for the whole program.

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+
+$Dinodex = Hash.new
```
jmmastey: Nope. No `$Globals` anywhere, please. They make debugging truly impossible. Also, use `{}` for a hash rather than `Hash.new`

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+
+class Dinosaur
+
+ attr_reader :name, :period, :diet, :weight, :movement, :continent, :description
+  def extract_period(period)
```
jmmastey: Fix indents, also add blank line between attr declarations and methods.

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+    period.each do |row, column|
+      if /(.+)\s(.+)/.match(period)
+        @period << [Regexp.last_match(1), Regexp.last_match(2)]
+      else
+        @period << ["",period]
+      end
+    end
+  end
```
jmmastey: As I mentioned above, make this method return a value rather than setting an instance variable. That would make your usage of it mimic the other setters in the initializer. Honestly, though, there's no reason why you'd need to split these values.

jmmastey: Also, I think this might actually be broken, as `@period` never appears to be set to be an empty array.

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+  def get_period period
+    period.each do |index|
+      @output = @output + "" + index
+    end
+      @output
+  end
+    
+  def to_json
+    {'name' => @name,
+     'period' => (get_period period).to_s,
```
jmmastey: No need to add `to_s` to a method that already returns a string.

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+     'diet' => @diet,
+     'weight' => @weight,
+     'movement' => @movement,
+     'continent' => @continetn,
+     'description' => @description}.to_json
```
jmmastey: Define a `to_h` (which you've done implicitly here) and then this method is just `to_h.to_json`

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  def initialize(dino)
+    @name = dino[0]
+    extract_period(dino[1])
+    @diet = dino[2]
+    @weight = dino[3]
+    @movement = dino[4]
+    @continent = dino[5]
+    @description = dino[6]
+  end
```
jmmastey: Index based initializers are problematic. If someone adds a single extra column to the CSV the whole class breaks. Alternatively, if you get input from another source (like a JSON file), the entire class breaks.

So, instead, you can (and should) isolate CSV parsing and standardization inside of one bit of parsing code, which `Dinosaur` should be entirely ignorant of.

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
+require 'csv'
+require 'json'
+require_relative 'Dino_Methods'
+
+$Dinodex = Hash.new
+
+class Dinosaur
```
jmmastey: The class `Dinosaur` belongs in `dinosaur.rb`.

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+    extract_period(dino[1])
+    @diet = dino[2]
+    @weight = dino[3]
+    @movement = dino[4]
+    @continent = dino[5]
+    @description = dino[6]
+  end 
+end
+
+class African_Dinosaur < Dinosaur
```
jmmastey: If I had a dinosaur from Africa in the `dinodex.csv` file, it would just be a `Dinosaur`, not an `AfricanDinosaur`. That doesn't make a lot of sense. See above about regularizing data elsewhere.

If this were to stay, it would need to change to `AfricanDinosaur` (no underscore) and also belong in `african_dinosaur.rb`.

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+    @dino=[]
+    @dino[0] = dino['Genus']
+    @dino[1] = split_period dino['PERIOD']
+    @dino[2] = is_carnivore?(dino['Carnivore'])
+    @dino[3] = dino['Weight']
+    @dino[4] = dino['Walking']
+    @dino[5] = "Africa"
+    @dino[6] = nil
+    super(@dino)
+  end
```
jmmastey: This doubles the fragility of the above initializer, as a change to an unrelated CSV format would break this class as well.

jmmastey: The nice thing is that clearly you have a hash with string keys, which means you could amend your dinosaur to use those.

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+end
+
+class Normal_Dinosaur < Dinosaur
+     def split_period period
+     unless /(.+)\sor\s(.+)/.match(period)
+       @period << dino['PERIOD']
+     else
+       @period << Regexp.last_match(1)
+       split_period Regexp.last_match(2)
+     end
```
jmmastey: Fix indents. Also, I don't understand why this logic is duplicated multiple times elsewhere?

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+  initialize(dino)
```
jmmastey: This is almost certainly broken. Initialize needs a `def`. Also, add a newline above it.

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+      @dino[6] = dino['DESCRIPTION']
+      super(@dino)
+  end
+end        
+
+#CSV.foreach('african_dinosaur_export.csv', converters: :numeric, headers:true) do |row|
+ #    $Dinodex[row['Genus']] = African_Dinosaur.new(row)
+#end
+
+CSV.foreach('dinodex.csv', converters: :numeric, headers:true) do |row|
```
jmmastey: When you intermingle class definitions and driver code, it becomes impossible to use the library anywhere but within the context of this script. If I wanted to use this data for a Rails website, for instance, I wouldn't be able to, as this code only runs from command line. Instead, move driver code into its own file and leave class definitions in their own (well named) files.

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+      @dino[4] = dino['WALKING']
+      @dino[5] = dino['CONTINENT']
+      @dino[6] = dino['DESCRIPTION']
+      super(@dino)
+  end
+end        
+
+#CSV.foreach('african_dinosaur_export.csv', converters: :numeric, headers:true) do |row|
+ #    $Dinodex[row['Genus']] = African_Dinosaur.new(row)
+#end
```
jmmastey: Why commented?

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+end 
+
+def match_period(dino, period)
+  dino.period.each do |index|
+    if dino.period[index].include? period
+      @flag = true
+    end
+  end
+  @flag ||= false
+end
```
jmmastey: Third duplication of period logic?

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+
+def match_attribute(dino, attribute, key)
+  attribute == dino.send(key.to_s)
+end
+  
+
+def find(params)
+  search_terms = eval(params.gsub(':','=>'))
+  $Dinodex.each do|key, dinosaur|
```
jmmastey: Your driver script should keep track of an array of dinosaurs as a local var, not a global. There's no advantage to the global here, but lots of disadvantages.

dino_catalog/Dinodex.rb
```diff
@@ -0,0 +1,135 @@
...
+    @flag = false
```
jmmastey: This flag thing showed up again. Don't do that. Also, the name is confusing.

jmmastey:
Hey there. A few comments added, centering around organization, idiomatic usage, and duplication. Let me know if any of it is unclear and we can walk through it together.

Give it a refactor and submit again, please. :)

PR #505, opened 2015-02-12
Refactored code for readability

theundeadmonk:


robot_name/robot_name.rb
```diff
@@ -3,22 +3,33 @@ class NameCollisionError < RuntimeError; end
...
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def is_error?
+    if !(@name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
+      raise NameCollisionError, 'There was a problem generating the robot name!'
     end
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
+  def initialize(args = {})
```
ashtonthomas: I'm used to initialize being the first thing in the class. Whatcha think?

robot_name/robot_name.rb
```diff
@@ -3,22 +3,33 @@ class NameCollisionError < RuntimeError; end
...
+    @@registry ||= []
+    @name_generator = args[:name_generator]
```
ashtonthomas: what do we gain from having a generic hash as a param? Can we increase the readability by just directly having a name_generator parameter?

robot_name/robot_name.rb
```diff
@@ -3,22 +3,33 @@ class NameCollisionError < RuntimeError; end
...
+    @name = name_generator
```
ashtonthomas: I'm not sure I understand what you are trying to do here. We have a parameter named the same as a method that you call and that parameter is stored as an instance method.

If you were just to read:
```
@name = name_generator 
# and eventually 
@name = @name_generator.call
```
Would you have any concerns?

robot_name/robot_name.rb
```diff
@@ -3,22 +3,33 @@ class NameCollisionError < RuntimeError; end
...
+      raise NameCollisionError, 'There was a problem generating the robot name!'
     end
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
+  def initialize(args = {})
+    @@registry ||= []
+    @name_generator = args[:name_generator]
+    @name = name_generator 
+    is_error?
```
ashtonthomas: Do you think `is_error?` belongs in the initialize (what kind or error? related to name generation? why don't put it in the method that deals with name generation?)

`is_error?` suggest that it returns a boolean. However it does something. It raises an exception. 

We need to fix this. They way it is being used would suggest a name like: `validate_name`

robot_name/robot_name.rb
```diff
@@ -3,22 +3,33 @@ class NameCollisionError < RuntimeError; end
...
     @@registry << @name
```
ashtonthomas: this should probably go into the name generator method as well

ashtonthomas:
@theundeadmonk, I provided some comments. Take a look (and the other PR's as well) and take another stab. I'm sure you will have some ideas after the other reviews.

Keep it up!!

- A

ashtonthomas:
Also, if you had a separate class for name generation, you could move all the registry, duplicate, pattern stuff into it and leave the robot class clean and focusing on something more important than names ;-)

PR #507, opened 2015-02-12
Robot Name Changes

shishirtewari:
Please review changes for Robot Name.

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -1,25 +1,51 @@
 class NameCollisionError < RuntimeError; end
+class InvalidFormatError < RuntimeError; end
 
+# Generates robot's name
 class Robot
   attr_accessor :name
-
-  @@registry
 
   def initialize(args = {})
     @@registry ||= []
     @name_generator = args[:name_generator]
+  end
 
+  def name
     if @name_generator
       @name = @name_generator.call
+      fail InvalidFormatError, 'Not supported format' unless pattern_check?
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+      generate_name
     end
+    repeat_check
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
+  def random_num
+    rand(10)
+  end
+
+  def random_char
+    ('A'..'Z').to_a.sample
+  end
+
+  def generate_name
+    str_arr = []
+    (1..5).each do |i|
+      str_arr << (i == 1 || i == 2 ? random_char : random_num.to_s)
+    end
+    @name = str_arr.join
+  end
+
+  def pattern_check?
+    @name =~ /[[:alpha:]]{2}[[:digit:]]{3}/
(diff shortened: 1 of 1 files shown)


robot_name/robot_name.rb
```diff
@@ -1,25 +1,51 @@
 class NameCollisionError < RuntimeError; end
+class InvalidFormatError < RuntimeError; end
 
+# Generates robot's name
 class Robot
   attr_accessor :name
 
-  @@registry
-
   def initialize(args = {})
```
ashtonthomas: There is a case for accepting a hash as an argument and that is for flexibility and being able to pass args in varying order. However, I think it improved readability if you just explicitly declar the parameter instead of wrapping in a hash.

robot_name/robot_name.rb
```diff
@@ -1,25 +1,51 @@
...
     @@registry ||= []
     @name_generator = args[:name_generator]
```
ashtonthomas: What if you don't pass a generator?

robot_name/robot_name.rb
```diff
@@ -27,6 +53,6 @@ def initialize(args = {})
 puts "My pet robot's name is #{robot.name}, but we usually call him sparky."
 
 # Errors!
-# generator = -> { 'AA111' }
-# Robot.new(name_generator: generator)
-# Robot.new(name_generator: generator)
+generator = -> { 'AB111' }
```
ashtonthomas: Don't want to have more than one robot?

robot_name/robot_name.rb
```diff
@@ -1,25 +1,51 @@
...
+  def random_num
+    rand(10)
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def random_char
+    ('A'..'Z').to_a.sample
+  end
+
+  def generate_name
```
ashtonthomas: oh just kidding you do have an internal name generator. Maybe there is a better way of structuring this.

robot_name/robot_name.rb
ashtonthomas: I see now that you can bypass the name_generator. Maybe it will help if you do an `initialize(name_generator = NameGenerator.new)` (?)

robot_name/robot_name.rb
```diff
@@ -1,25 +1,51 @@
...
     end
+    @name = str_arr.join
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
+  def pattern_check?
+    @name =~ /[[:alpha:]]{2}[[:digit:]]{3}/
+  end
+
+  def repeat_check
```
ashtonthomas: This method name makes me think that we are doing the action of calling a check again.

Or perhaps this was supposed to return a boolean and be repeat_name?

or probably: validate_name (?)

Either way, the flow of #name and this method seem a bit odd

robot_name/robot_name.rb
```diff
@@ -1,25 +1,51 @@
...
+class InvalidFormatError < RuntimeError; end
 
+# Generates robot's name
 class Robot
   attr_accessor :name
 
-  @@registry
-
   def initialize(args = {})
     @@registry ||= []
```
ashtonthomas: I kinda wish that there was a class for generating names and the registry was property of that class. thoughts?

ashtonthomas:
@shishirtewari, I hope my comments help!

PR #508, opened 2015-02-13
Zb overlord

zbright:


overlord/features/boot_bomb.feature
```diff
@@ -0,0 +1,43 @@
...
+  Scenario Outline: Invalid activation code
+    When I enter <activation> for the activation code
+    And I enter <deactivation> for the deactivation code
+    And I boot the bomb
+    Then the bomb is not booted
+
+    Examples:
+      |  activation  |  deactivation  |
+      |  "9999"      |  "9876"        |
+      |  "9999"      |  "9876"        |
```
djkotowski: Running the same test twice?

overlord/features/boot_bomb.feature
```diff
@@ -0,0 +1,43 @@
...
+  Scenario Outline: Invalid deactivation code
+    When I enter <activation> for the activation code
+    And I enter <deactivation> for the deactivation code
+    And I boot the bomb
+    Then the bomb is not booted
+
+    Examples:
+      |  activation  |  deactivation  |
+      |  "2222"      |  "9999"         |
+      |  "2222"      |  "9999"        |
```
djkotowski: Ditto my comment on line 32.

overlord/features/activate_bomb.feature
```diff
@@ -0,0 +1,28 @@
...
+    Given the bomb is booted with "4444"
+    When I enter "4444" for the activation code
+    And I press the activate button
+    Then the bomb is active
+
+  Scenario: Bomb was activated with custom codes, but wrong code is entered
+    Given the bomb is booted with "4444"
+    When I enter "9999" for the activation code
+    And I press the activate button
+    Then to bomb warns me that I was wrong
```
djkotowski: Typo

overlord/bomb.rb
```diff
@@ -0,0 +1,81 @@
...
+    if activated? || invalid?
+      300 - (Time.now - time_activated)
+    else
+      0
+    end
+  end
+
+  private
+
+  def valid?(code)
```
djkotowski: I'd probably name this more along the lines of `valid_code?`.  If it's just `valid?`, the name implies the method checks if the _bomb_ is valid, not the code.

overlord/features/boot_bomb.feature
zbright: copy paste errors. One should only have 3 numbers in each example

overlord/overlord.rb
```diff
@@ -1,14 +1,61 @@
-# run `ruby overlord.rb` to run a webserver for this app
-
+require_relative 'bomb'
 require 'sinatra'
+# require 'pry'
 
 enable :sessions
+set :haml, format: :html5
```
ashtonthomas: I didn't even know you could run a sinatra app like this outside of a class that extends something within sinatra? are you able to have private methods? do you need private methods? I dunno

overlord/overlord.rb
```diff
@@ -1,14 +1,61 @@
...
+  haml :explode
+end
+
+get '/reset' do
+  session.clear
+  develop_bomb
+  haml :index, locals: { bomb: bomb }
+end
+
+def develop_bomb
```
ashtonthomas: `develop_bomb` is an odd name

overlord/overlord.rb
```diff
@@ -1,14 +1,61 @@
...
+  session[:bomb] = Bomb.new
```
ashtonthomas: what if we already have a bomb in session? is this going to erase/reset it?

overlord/overlord.rb
```diff
@@ -1,14 +1,61 @@
...
+end
+
+def bomb
+  session[:bomb]
```
ashtonthomas: we could probably manipulate this bomb by just navigating to a url...? like reset? or the home path?

ashtonthomas:
you have enough config files at your root mixed with app files that you could benefit from having an app/ or lib/ directory (maybe?)

overlord/bomb.rb
```diff
@@ -0,0 +1,81 @@
...
+      transition [:activated, :invalid] => :exploded
+    end
+  end
+
+  def initialize
+    @attempts_remaining = MAX_ATTEMPTS
+    @activation_code = DEFAULT_ACTIVATION_CODE
+    @deactivation_code = DEFAULT_DEACTIVATION_CODE
+    super() # Needed for the state machine to initialize
+    self.build!
```
ashtonthomas: I've never used state_machine so do you need the self's here and other places?

overlord/features/boot_bomb.feature
```diff
@@ -0,0 +1,43 @@
...
+  Scenario: Visit home page
+    Then I see fields to enter activation/deactivation codes
+    And I see a switch to boot the bomb
+
+  Scenario: Enter default activation/deactivation codes
+    When I boot the bomb
+    Then the bomb is booted, but deactivated
+
+  Scenario: Enter custom activation/deactivation codes
+    When I enter "1234" for the activation code
```
ashtonthomas: It is probably best to rephrase this to "..enter a valid/invalid code for activation/deactivation"

Cause I can't really tell what the intent is. Do i want valid? do I want invalid? What if I change the requirements of valid and invalid? Cukes are supposed to leave out as much implementation as possible. Your comment will probably be that you are just reusing the same step that is used in a different feature where you pass in a table of examples. But even then I don't think you are really gaining much. Maybe just call your template step from the step def of "I enter a valid code..." or better yet, have a step that enters both act and deact codes plus boots the bomb all in one step - use a factory so you know what the codes will be.

ashtonthomas: deleting  a comment I made earlier

ashtonthomas:
@zbright I think it is all very reasonable. See the comments, and I don't think there is enough to do it again unless you want me to review the UI :-)

overlord/bomb.rb
jmmastey: I don't think so, fwiw. As long as it's within the class, there's no way to mistake it for a local so the self should be unnecessary.

If not, I'd be confused.

PR #509, opened 2015-02-16
Bte03 overlord

magicalbanana:
@jmmastey please review.


overlord/mediators/interface.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  def deploy
+    # When arming a bomb, event: arm
+    # A disarming code must be set, if not it will not arm
+    # the bomb.
+    @payload.arm if validate!
+  end
+
+  def disable(code)
+    # when disabling a bomb, event :disarm
```
jmmastey: These various commends aren't helpful. Use self-documenting code for this sort of thing.

overlord/mediators/interface.rb
```diff
@@ -0,0 +1,125 @@
+require 'virtus'
+require 'active_model'
+require './models/bombs.rb'
+
+ValidationError = Class.new RuntimeError
+
+class Interface
```
jmmastey: This is not an awesome name for a class, fwiw. :)

overlord/mediators/interface.rb
```diff
@@ -0,0 +1,125 @@
...
+  end
+
+  private
+
+  def add_attempts
+    @payload.attempts += 1
+    @payload.save
+  end
+
+  def disarming_code=(disarming_code)
```
jmmastey: The implication here is correct, the arg needs to be used.

overlord/mediators/interface.rb
```diff
@@ -0,0 +1,125 @@
...
+  def get_payload
+    # This gets a current bomb. There can only be one bomb in the record at all
+    # times. If no bomb exists, it will return a new instance of bomb that will be
+    # saveable.
+    bomb = Bombs.all - Bombs.all(:state => :deactivated)
+    if bomb.count >= 1
+      bomb.first
+    else
+      Bombs.new
+    end
```
jmmastey: If you're using active record anyway, this should be something like `first_or_create`. This is too much work.

overlord/mediators/interface.rb
```diff
@@ -0,0 +1,125 @@
...
+  def valid_disarming_code?(code)
+    code == @payload.disarming_code
+  end
+
+  def validate!
+    if valid?
+      true
+    else
+      raise ValidationError, errors
+    end
```
jmmastey: `valid? || raise ValidationError, errors`

overlord/overlord.rb
```diff
@@ -1,14 +1,134 @@
...
+      erb :boot_device
+    end
+  end
+
+  get '/create_bomb' do
+    unless valid_event?('activated')
+      redirect to payload_state_redirect
+    end
+    erb :create_bomb
+  end
```
jmmastey: `get create_bomb` is an odd route. it's not the purpose of this exercise, but make sure that when you do rails you look into restful / resourceful routes.

overlord/overlord.rb
```diff
@@ -1,14 +1,134 @@
...
+    if session_interface.disable(disarming_code.to_i)
+      redirect to payload_state_redirect
+    end
+
+    session_interface.count_disarming_attempts
+
+    if session_interface.max_attempts_reached?
+      session_interface.payload.detonate
+      redirect to payload_state_redirect
+    end
```
jmmastey: Given you already have models and mediators, one of those layers should probably handle the max-attempts logic.

overlord/overlord.rb
```diff
@@ -1,14 +1,134 @@
...
+
+  post '/detonate_bomb' do
+    # set to detonate then delete
+    # redirect to show page where bomb detonated
+    if session_interface.trigger_explosion
+      status 200
+    else
+      status 500
+    end
+  end
```
jmmastey: I cannot imagine why I would ever call this route :)

overlord/public/js/bootstrap.js
```diff
@@ -0,0 +1,2306 @@
...
+    offset: 0,
+    target: window
+  }
+
+  Affix.prototype.getState = function (scrollHeight, height, offsetTop, offsetBottom) {
+    var scrollTop    = this.$target.scrollTop()
+    var position     = this.$element.offset()
+    var targetHeight = this.$target.height()
+
+    if (offsetTop != null && this.affixed == 'top') retur
```
jmmastey: Toooooooo much code. Get the .min of this stuff. I can't tell if you modified any of this from the stock libraries.

overlord/public/js/overlord.js
```diff
@@ -0,0 +1,42 @@
+// When clicking the mustached overlord image
+$(function() {
+  $("#overlord").on('click', function() {
+    window.location = '/boot_device';
+  });
```
jmmastey: Typically, this sort of thing should just be a link. Directly setting `window.location` is somewhat surprising.

overlord/mediators/interface.rb
magicalbanana: I know, sorry about that. Forgot to change that to something else.

overlord/mediators/interface.rb
magicalbanana: Yeah, I understand.

overlord/mediators/interface.rb
magicalbanana: Yep,  one liner. Should have done that.

overlord/overlord.rb
magicalbanana: Yeah, in rails there are built in helpers for create and what not and here I thought what approach I can do to differentiate it.

But yeah, clearly it would've been best if I just left it as /bomb

overlord/overlord.rb
magicalbanana: Yeah, I could make another class for handling that. I was thinking I would do that but since it's only used on this part of the app I opted not to.

overlord/overlord.rb
magicalbanana: Hehe. I know it's bad route naming. Shame on me.

overlord/public/js/bootstrap.js
magicalbanana: Ah no, I did not but yeah should've gotten the .min

overlord/public/js/overlord.js
magicalbanana: Yeah, but the image on the html side has no set link that's why I did it here instead. Do links normally stay in the html? Or is it acceptable in JavaScripts?

overlord/spec/features/bomb_activation/bomb_activation.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Bomb Activation
+
+  Scenario: Activating a bomb with a valid code
+    Given I am on the home page
+    Then I should see the image "overlord" within "container"
```
jmmastey: This is a bit of an odd one. I get that it's a CSS selector, but `I see overlord within container` is a nonsensical statement.

overlord/spec/features/bomb_activation/bomb_activation.feature
```diff
@@ -0,0 +1,32 @@
...
+      And I should see "Valentine's Bomb" within "intro-text"
+      And I should see "Desperate - Single - Ready to Mingle" within "intro-text"
+    When hovering on the Overlord image
```
jmmastey: Dan and I talked about this one a bit in person. 90% of the time, a single set of Given -> When -> Then is best. There's one workflow here (`activate a bomb with a valid code`), which can make sense for a combo workflow (Given -> When -> Then -> When -> Then etc), but in this case the former steps (seeing Overlord, for instance) are not related to activation.

So, for this test, I'd recommend having separate setup tests that check for intro-text etc, and keeping the activation test here.

overlord/spec/features/bomb_activation/bomb_activation.feature
```diff
@@ -0,0 +1,32 @@
...
+      And clicking the image
+    Then I am on bomb activation page
+      And should see the field "boot_code" within "row"
+      And should see the button "submit" within "row"
+    When I will enter my boot code:
+      | code | 6969 |
```
jmmastey: This is a strange format. Just do `When I enter the boot code "6969"`

overlord/spec/features/bomb_activation/bomb_activation.feature
```diff
@@ -0,0 +1,32 @@
...
+      And I click the submit button
+    Then I should be on bomb creation page
+
+  Scenario: Activating a bomb with an invalid code
+    Given I am on the home page
+    Then I should see the image "overlord" within "container"
+      And I should see "Valentine's Bomb" within "intro-text"
+      And I should see "Desperate - Single - Ready to Mingle" within "intro-text"
```
jmmastey: Duplicate assertions to the other test. This is part of why I think this bit is actually a different test.

overlord/spec/features/bomb_creation/bomb_creation.feature
```diff
@@ -0,0 +1,22 @@
...
+  Scenario: Configuring the bomb settings with all required fields
+    Given I am on the bomb creation page
+    Then I should see the field "disarming_code" within "row"
+      And I should see the field "timer" within "row"
+      And I should see the field "max_attempts" within "row"
+    When I fill in the following:
+      | disarming_code | 1234 |
+      | timer          | 30 |
+      | max_attempts   | 5 |
+      And I click the submit button
```
jmmastey: This is a fine use of tables. Right on.

overlord/spec/features/bomb_creation/bomb_creation.feature
```diff
@@ -0,0 +1,22 @@
...
+    Then I should be on bomb disarming page
+
+  Scenario: Configuring the bomb settings with missing required fields
+    Given I am on the bomb creation page
+    Then I should see the field "disarming_code" within "row"
+      And I should see the field "timer" within "row"
+      And I should see the field "max_attempts" within "row"
```
jmmastey: Something missing here? I don't see any configuration happening. (e.g. there's no `When` whatsoever)

overlord/spec/features/bomb_disarmament/bomb_disarmament.feature
```diff
@@ -0,0 +1,43 @@
+Feature: Bomb Disarmament
+
+  Background:
+    Given A bomb has already been deployed
```
jmmastey: Lowercase the `A` here.

overlord/spec/features/bomb_disarmament/bomb_disarmament.feature
```diff
@@ -0,0 +1,43 @@
...
+      And I click the submit button
+    Then I should be on bomb disarmed page
+
+  Scenario: When I enter the incorrect code for disarmament
+    Given I am on the bomb disarming page
+    Then I should see the field "disarming_code" within "row"
+    When I fill in the following:
+      | disarming_code | 3333 |
+      And I click the submit button
+    Then I should not be on bomb disarmed page
```
jmmastey: Is there a way to assert `And the bomb is still active`?

overlord/spec/features/bomb_disarmament/bomb_disarmament.feature
```diff
@@ -0,0 +1,43 @@
...
+  
+  Scenario: When I run use up all attempts to disarm it
+    Given I am on the bomb disarming page
+    Then I should see the field "disarming_code" within "row"
+    # Just did it 3 times instead of looping it, eh.
+    # The default max attempts is 3 anyway
```
jmmastey: Ehhhhhhhh, this should just be one step. `When I fill in an incorrect code three times`

overlord/spec/features/bomb_state_routing/bomb_state_routing.feature
```diff
@@ -0,0 +1,21 @@
+Feature: Bomb State Routing
+
+  Scenario: Accessing the bomb activation page
+    Given A bomb has already been activated
+    When I am on the bomb activation page
```
jmmastey: So, `When` steps should be actions. `When I visit the bomb activation page`

overlord/spec/features/step_definitions/background_steps.rb
```diff
@@ -0,0 +1,30 @@
+Given /^(?:|A) bomb has already been activated$/ do
+  bomb = Bombs.new
+  bomb.activate
+end
+
+Given /^(?:|A) bomb has already been armed$/ do
+  bomb = Bombs.new(:disarming_code => 1234, :max_attempts => 3)
```
jmmastey: The convention with `data_mapper` (and definitely with `active_record`) is to use the singular for the class name. `s/Bombs/Bomb`

overlord/spec/features/step_definitions/bomb_activation_steps.rb
```diff
@@ -0,0 +1,13 @@
+When(/^hovering on the Overlord image$/) do
+  page.find('#overlord').trigger(:mouseover)
+end
+
+When(/^clicking the image$/) do
+  page.find('#overlord').trigger(:click)
+end
+
+When(/^I will enter my boot code:$/) do |table|
+  # table is a Cucumber::Ast::Table
```
jmmastey: Pointless comment

overlord/spec/features/step_definitions/bomb_disarmament_steps.rb
```diff
@@ -0,0 +1,5 @@
+When /^The timer runs out$/ do
+  bomb = Bombs.first
+  timer = bomb.timer
+  sleep(timer + 3)
+end
```
jmmastey: Be really wary of this sort of test. It makes your test suite REALLY slow as you add these sorts of tests.

overlord/spec/mediators/interface_spec.rb
```diff
@@ -0,0 +1,391 @@
...
+  describe "#configure_settings" do
+    context "when an attribute is set" do
+      { :disarming_code => rand(1000..6000),
+        :timer => rand(10..29), :max_attempts => rand(1..2) }.each do |attribute, value|
+        it "changes the value of #{attribute} to #{value}" do
+          interface.configure_settings(attribute => value)
+          expect(interface.payload.send(attribute)).to eq(value)
+        end
+      end
+    end
```
jmmastey: A test with this much abstraction becomes really hard to use as documentation, even though it serves as very good verification. Consider for most simple attributes moving to a simpler hard-coded declaration that serves as good documentation.

overlord/spec/mediators/interface_spec.rb
```diff
@@ -0,0 +1,391 @@
...
+          interface.configure_settings(nil)
+        end
+        it "returns false" do
+          interface.configure_settings(nil)
+          expect(interface.turn_on(code)).to be_falsey
+        end
+        it "does not change the payload state" do
+          state = interface.payload.state
+          interface.turn_on(code)
+          expect(interface.payload.state).to eq(state)
```
jmmastey: Are you a one-assertion sort of person? I always found this to be much more difficult to work with overall than having a few assertions. Purely a matter of taste, nothing to change here.

overlord/spec/mediators/interface_spec.rb
```diff
@@ -0,0 +1,391 @@
...
+        expect(interface.destroy_payload).to be_falsey
+      end
+    end
+  end
+
+  # Private Methods
+  #
+  # Ideally you do not test Private methods with integration testing
+  # you could do this in a unit test capacity but then again, Svajone likes testing.
+  # But really, private method testing in Rails convention is not a good thing to do.
```
jmmastey: Keep in mind that even though you like testing, this makes your code more brittle for the next person who has to maintain it. Maybe write them and then delete them before pushing :)

overlord/spec/model/bombs_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  describe "#state_machine" do
+    subject(:bomb) { Bombs.new }
+
+    describe "#states" do
+      [:deactivated, :activated, :armed, :disarmed, :detonated].each do |state|
+        it { expect(Bombs).to have_state(state) }
+      end
+    end
```
jmmastey: This sort of test typically strikes me as low value. The odds of catching anything with this test are not very high, and the functionality is tested elsewhere using transitions and such.

That said, it's clearly a very cheap test to write, so it might be worth it on that count. Not sure.

overlord/spec/overlord_spec.rb
```diff
@@ -0,0 +1,309 @@
...
+  describe "#get '/create_bomb'" do
+    subject { get '/create_bomb' }
+
+    context "when the paylod.state = active" do
+      before :each do
+        b = Bombs.new
+        b.activate
+      end
+      it "renders the create_bomb.erb template" do
+        expect(subject.body).to include(IO.binread("./views/create_bomb.erb"))
```
jmmastey: These are a very strange convention. Rails does have controller tests to verify that views are rendered, but they also tend to be low-value tests (and are duplicated in integration tests).

Overall, what you want to assert _isn't_ that `create_bomb.erb` was rendered. Nobody (least of all you) cares which file was rendered. What you care about is the content of the page (mostly what's in your integration tests). This is a source of fragility: I cannot refactor these files without breaking all the tests.

overlord/spec/overlord_spec.rb
```diff
@@ -0,0 +1,309 @@
...
+      end
+      it "returns a status of 200" do
+        expect(subject.status).to eq(200)
+      end
```
jmmastey: Unless there's something unexpected about the return status, also low value.

overlord/spec/overlord_spec.rb
```diff
@@ -0,0 +1,309 @@
...
+
+  describe "#get '/disarmed_bomb'" do
+    subject { get '/disarmed_bomb' }
+
+    context "when the validate_event? == true" do
+      before :each do
+        b = Bombs.new
+        b.activate
+        b.arm
+        b.disarm
```
jmmastey: I've seen this code a ton of times by now. In rails, use a factory. In sinatra, just abstract it into a few methods. Or a much-much higher level `let`

overlord/views/boot_device.erb
```diff
@@ -0,0 +1,30 @@
+<section id="boot_device">
+    <div class="container">
```
jmmastey: Fix your tabs. indent of 2 :)

jmmastey:
Code is well separated and well tested, nicely done. No need for resubmission. Two overall concerns:

1. Your love for testing is creating a lot of overlap coverage, which is hard to read and maintain. Work on providing unique value which each test, and emphasize the business requirements that the tests satisfy, rather than the mechanical aspects of the code.
2. This was overkill, but you knew that :) For reference, this branch is 5x the size of all of LevelUpRails.

Looking forward to seeing Rails!

PR #510, opened 2015-02-16
Arrowhead

theundeadmonk:
Refactored code as per review comments

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -1,36 +1,43 @@
 class Arrowhead
-  # This seriously belongs in a database.
+# This seriously belongs in a database.
   CLASSIFICATIONS = {
     far_west: {
       notched: "Archaic Side Notch",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Agate Basin",
-      bifurcated: "Cody",
-    },
+               stemmed: "Archaic Stemmed",
+               lanceolate: "Agate Basin",
+               bifurcated: "Cody",
+              },
     northern_plains: {
       notched: "Besant",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Humboldt Constricted Base",
-      bifurcated: "Oxbow",
-    },
-  }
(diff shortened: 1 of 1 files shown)


arrowhead/arrowhead.rb
```diff
@@ -1,34 +1,47 @@
 class Arrowhead
-  # This seriously belongs in a database.
+
+  UNKNOWN_METHOD = "Unknown region, please provide a valid region."
+  UNKNOWN_SHAPE = %"Unknown shape value. Are you sure you know what you're
+    talking about?"
+# This seriously belongs in a database.
```
ashtonthomas: Hey, @theundeadmonk what does the `%` do before the string?

arrowhead/arrowhead.rb
```diff
@@ -1,34 +1,47 @@
...
-      shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        puts "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
+    arrowhead = %"You have a(n) '#{CLASSIFICATIONS[region][shape]}'
```
ashtonthomas: Do you think it is better to have the query after the shape and region checks?

ashtonthomas:
@theundeadmonk, It would be cool if all of this conformed to hound/rubocop, but I don't think we need to do another review. I made a couple of comments/questions. Address as you see fit.

LGTM

arrowhead/arrowhead.rb
theundeadmonk: So the % lets me span the string across multiple lines. That way, line length won't exceed 80 characters.

arrowhead/arrowhead.rb
theundeadmonk: Yeah, I've made the changes.

PR #513, opened 2015-02-16
Bte09 shopping cart

magicalbanana:
@jmmastey please review.

### DIFF
--- a/shopping_cart/features/authentication/registration.feature
+++ b/shopping_cart/features/authentication/registration.feature
@@ -1,0 +1,45 @@
+Feature: Registration
+
+  Scenario: Registration is successful
+    Given I follow the registration link
+    Then I should see the registration form
+      And I should see a field for my username
+      And I should see a field for my password
+      And I should see a field for my password confirmation
+      And I should see a field for my email
+    When I fill in the username with "svajone"
+      And I fill in the password with "svajone69"
+      And I fill in the password confirmation with "svajone69"
+      And I fill in the email with "svajone@svajone.lt"
+      And click on the submit button
+    Then I should be on the dashboard page
+      And I should see a message thanking me for registering
+
+  Scenario: Registration has missing require fields
+    Given I follow the registration link
+    Then I should see the registration form
+      And I should see a field for my username
+      And I should see a field for my password
+      And I should see a field for my password confirmation
+      And I should see a field for my email
+    When I fill in the username with "svajone"
+      And I fill in the password with "svajone69"
+      And I fill in the password confirmation with "svajone69"
+      And click on the submit button
+    Then I should still be on the registration page
+      And I should see a message telling me to fill in the email
+
+  Scenario: Registration has bad data
+    Given I follow the registration link
+    Then I should see the registration form
+      And I should see a field for my username
+      And I should see a field for my password
+      And I should see a field for my password confirmation
+      And I should see a field for my email
+    When I fill in the username with "svajone"
+      And I fill in the password with "svajone69"
+      And I fill in the password confirmation with "svajone69"
+      And I fill in the email with "svajone@@@svajone.lt"
+      And click on the submit button
+    Then I should still be on the registration page
+      And I should see a message telling me that the email I entered is not valid
--- a/shopping_cart/features/shopping/checkout.feature
+++ b/shopping_cart/features/shopping/checkout.feature
@@ -1,0 +1,33 @@
+Feature: Checkout
+
+  Scenario: Checkout is successful
+    Given I am signed in
+      And The page is SSL Secured
+      And I have items in my cart
+      And I am on the shopping cart page
+    Then I should on the shopping cart page
+      And I should see the items I added
+    When I click on checkout
+    Then I should be on the checkout page
+      And I should see payment options field
+    When I click on credit card
+    Then I should see a form to entere my credit card information with valid info
+    When I complete my credit card information
+      And I click submit
+    Then I should be directed to the confirmation page
+
+  Scenario: Checkout was not successful
+    Given I am signed in
+      And The page is SSL Secured
+      And I have items in my cart
+      And I am on the shopping cart page
+    Then I should on the shopping cart page
+      And I should see the items I added
+    When I click on checkout
+    Then I should be on the checkout page
+      And I should see payment options field
+    When I click on credit card
+    Then I should see a form to entere my credit card information
+    When I complete my credit card information with invalid info
+      And I click submit
+    Then I should see a message stating my credit card is not valid
--- a/shopping_cart/features/shopping/coupon.feature
+++ b/shopping_cart/features/shopping/coupon.feature
@@ -1,0 +1,40 @@
+Feature: Coupon
+
+  Scenario: Coupon is valid
+    Given I am signed in
+      And I have items in my cart
+      And I am on the checkout page
+    Then I should be on the checkout page
+      And I should see the items I added
+      And I should see the total cost of all the items
+      And I should see a field for coupon
+    When I fiill in the coupon with "svajonelove"
+      And click on submit
+    Then I should see a message telling me my coupon was accepted
+      And I should see my price change
+
+  Scenario: Coupon is expired
+    Given I am signed in
+      And I have items in my cart
+      And I am on the checkout page
+    Then I should be on the checkout page
+      And I should see the items I added
+      And I should see the total cost of all the items
+      And I should see a field for coupon
+    When I fiill in the coupon with "svajonelovemetoo"
+      And click on submit
+    Then I should see a message telling me my coupon is expired
+      And I should see my price unchanged
+
+  Scenario: Coupon is not valid
+    Given I am signed in
+      And I have items in my cart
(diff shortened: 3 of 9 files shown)


shopping_cart/features/authentication/registration.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Registration
+
+  Scenario: Registration is successful
+    Given I follow the registration link
+    Then I should see the registration form
+      And I should see a field for my username
+      And I should see a field for my password
+      And I should see a field for my password confirmation
+      And I should see a field for my email
```
djkotowski: Don't add `Then` tests before `When` actions.

These are also unnecessary, as lines 10-14 will fail if a field is missing.

shopping_cart/features/authentication/registration.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Registration
+
+  Scenario: Registration is successful
```
djkotowski: This is the expected end result.  The scenario should describe what is happening to reach the expected result.

Example:  "User registers with valid information"

shopping_cart/features/authentication/registration.feature
```diff
@@ -0,0 +1,45 @@
...
+    Then I should see the registration form
+      And I should see a field for my username
+      And I should see a field for my password
+      And I should see a field for my password confirmation
+      And I should see a field for my email
+    When I fill in the username with "svajone"
+      And I fill in the password with "svajone69"
+      And I fill in the password confirmation with "svajone69"
+      And I fill in the email with "svajone@svajone.lt"
+      And click on the submit button
```
djkotowski: Lines 10-14 have a lot of knowledge about the implementation.  Business users (the intended audience) don't care exactly what field is filled with which value.  You can summarize this process in a single step.

Example:
```cucumber
When I submit valid registration information
```

shopping_cart/features/shopping/checkout.feature
```diff
@@ -0,0 +1,33 @@
+Feature: Checkout
+
+  Scenario: Checkout is successful
+    Given I am signed in
+      And The page is SSL Secured
```
djkotowski: What action occurs here?

shopping_cart/features/shopping/coupon.feature
```diff
@@ -0,0 +1,40 @@
...
+
+  Scenario: Coupon is valid
+    Given I am signed in
+      And I have items in my cart
+      And I am on the checkout page
+    Then I should be on the checkout page
+      And I should see the items I added
+      And I should see the total cost of all the items
+      And I should see a field for coupon
+    When I fiill in the coupon with "svajonelove"
```
djkotowski: Typo

shopping_cart/features/shopping/item_details.feature
```diff
@@ -0,0 +1 @@
+Feature: Item Details
```
djkotowski: Empty feature?

shopping_cart/features/shopping/search.feature
```diff
@@ -0,0 +1,28 @@
+Feature: Search Item
+
+  Scenario: My search yields matches equal to the keyword entered
+    Given I am on the shopping dashboard page
+    Then I should see a field for item search
+    When I fill the item search with "dresser"
+      And click on the submit button
+    Then I should be the search result page
+      And should be presented with items equal and equivalent to my search
```
djkotowski: In 6, you manually specify the search terms.  This step implies knowledge about the details specified in step 6.  No steps should have knowledge about other steps.

shopping_cart/features/shopping/session_cart.feature
```diff
@@ -0,0 +1,29 @@
...
+    Then I should see a flash message asking me to sign in
+      And I should see a link for signing in
+    Then I should on the sign in page
+      And I should see a field for my username
+      And I should see a field for my password
+    When I fill in the username with "svajone"
+      And I fill in the password with "svajone69"
+      And click on the submit button
+    Then I should on the shopping cart page
+      And I should see the items I added
```
djkotowski: This is an extremely complex test and should be split into multiple other tests.

As a rule, you should follow "Given, When, Then" order.

shopping_cart/features/authentication/registration.feature
```diff
@@ -0,0 +1,45 @@
...
+      And I should see a field for my email
+    When I fill in the username with "svajone"
+      And I fill in the password with "svajone69"
+      And I fill in the password confirmation with "svajone69"
+      And click on the submit button
+    Then I should still be on the registration page
+      And I should see a message telling me to fill in the email
+
+  Scenario: Registration has bad data
+    Given I follow the registration link
```
djkotowski: When you have the same `Given` step occurring with every scenario in a feature, add it as a background to the top of the feature.

```cucumber
Background:
  Given I am on the registration page
```

shopping_cart/features/authentication/registration.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Registration
+
+  Scenario: Registration is successful
+    Given I follow the registration link
```
djkotowski: This implies you are already on a page with a visible registration link.  Why not just

```cucumber
Given I am on the registration page
```

shopping_cart/features/shopping/session_cart.feature
djkotowski: This step in particular also implies knowledge about previous steps and will not work as a standalone step.

djkotowski: What about cart merging scenarios?
- User account has x of item in cart
- Current user (not logged in) has y of item in cart
- Current user logs in to account
- Expect cart to have x + y of item

djkotowski:
Good job on covering a lot of scenarios here.  Your features in general have too much knowledge about your view implementation.  Try to keep the features and the implementations separate.  The step definitions are where you link the two.

Some of your steps also have too much knowledge about other steps.  By keeping your steps rather generic (e.g. `When I register with missing fields`), then this becomes a non-issue.

Your scenarios are also too complex.  By following a simple `Given`, `When`, `Then` order, you ensure they don't get too convoluted.  When you have a complex structure in your scenario, it's usually a sign that the scenario has unnecessary tests *or* could be split into multiple steps.

Take a look at the comments, refactor, and resubmit.  Feel free to ask me any questions you might have.

shopping_cart/features/shopping/checkout.feature
magicalbanana: Did not do this, but it is standard to check when dealing with payments.

shopping_cart/features/shopping/item_details.feature
magicalbanana: Yeah, forgot about this one. Realized too much testing already.

shopping_cart/features/shopping/checkout.feature
djkotowski: In that case, it would be a separate scenario.  All tests should be in `Then` statements.  `Given` and `When` statements are reserved for actions.

shopping_cart/features/shopping/session_cart.feature
magicalbanana: Misundertstoosd the requirements, but yeah I could add that easily as when I approached this test that was what I had in mind, similar to Amazon's cart feature.

And yeah, at first glance it looks complex but in the real world this would be moved to a test harness, I wanted to write the harness but decided to lay it out here instead.

And the Given, When, Then is not really a standard format as in reality it does not matter what I put infront of my test. When I write my feature test, I follow the purpose of Given (Which is the state which the state should be in), When (the key event) and Then (the observable).

shopping_cart/features/authentication/registration.feature
magicalbanana: True, I am familiar with backgorund/background outline. But since I always do red-green-refactor-green, I did not set it up as such. But ideally, like I said before, I would actually move all the fillins or whatnot in a harness and include them in the test itself.

shopping_cart/features/authentication/registration.feature
magicalbanana: Because of how I named my feature. I also want to test the functionality of my links, just because the test is for registration I want to step through the zero steps till completion.

shopping_cart/features/authentication/registration.feature
magicalbanana: Please refer to my comment regarding Given, When, Then.

But my concept here is I want to make sure the fields are existent, ideally I would move this all to a harness to limit the number of lines I have written. 

There's a logic as to why I need to look at the fields prior because in the event that the page loads slow or the driver is not fast enough, you would have issues with the fill in methods.

shopping_cart/features/shopping/coupon.feature
magicalbanana: True. :dancer:

shopping_cart/features/authentication/registration.feature
djkotowski: In that case, why not test the links separately?  `Given` statements tend to be worded to reflect the current state rather than actions.

PR #514, opened 2015-02-16
Showmetheweather

arrubin:
I am looking for some initial feedback before I refactor further.


event_system/showmetheweather/app/assets/javascripts/application.js
```diff
@@ -0,0 +1,16 @@
+// This is a manifest file that'll be compiled into application.js, which will include all the files
+// listed below.
+//
+// Any JavaScript/Coffee file within this directory, lib/assets/javascripts, vendor/assets/javascripts,
+// or any plugin's vendor/assets/javascripts directory can be referenced here using a relative path.
+//
+// It's not advisable to add code directly here, but if you do, it'll appear at the bottom of the
+// compiled file.
+//
+// Read Sprockets README (https://github.com/sstephenson/sprockets#sprockets-directives) for details
```
jmmastey: These are a pain sometimes, but once the files are generated and checked into your codebase, they're yours to fix. You may want to truncate or remove the lines entirely as appropriate.

event_system/showmetheweather/app/assets/javascripts/forecasts.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
```
jmmastey: Delete unused files.

event_system/showmetheweather/app/assets/stylesheets/application.scss
```diff
@@ -0,0 +1,16 @@
+@import "bootstrap-sprockets";
+@import "bootstrap";
+
+#search_form {
+  text-align: center;
+}
+
+#forecast {
```
jmmastey: This is a new one. Huh. I'm not really buying it, let me see how to turn off this particular warning.

event_system/showmetheweather/app/controllers/forecasts_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+      redirect_to action: "show", zip_code: current_user.zip_code
+    end
+  end
+
+  def search
+    unless params[:forecast][:zip_code] =~ /^\d{5}$/
+      flash[:danger] = "Invalid ZIP code"
+      redirect_to action: "index"
+      return
+    end
```
jmmastey: This is possibly better off as a `before_filter` or at least abstracted into another method. Not a big deal either way, though.

event_system/showmetheweather/app/controllers/forecasts_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+    redirect_to action: "show", zip_code: params[:forecast][:zip_code]
+  end
+
+  def show
+    @forecast = Forecast.recent(params[:zip_code]).last ||
+      Forecast.retrieve(params[:zip_code])
```
jmmastey: I'd prefer to see this select-or-fetch behavior wrapped within the Forecast object. There's no reason why the controller should understand those cache semantics, as opposed to some other class (like the model)

event_system/showmetheweather/app/controllers/forecasts_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+    respond_to do |format|
+      format.html
+      format.json { render json: @forecast.to_json }
+    end
```
jmmastey: You may be able to use `respond_with` here, which would remove the whole respond block.

event_system/showmetheweather/app/models/forecast.rb
```diff
@@ -0,0 +1,59 @@
+require "nws/rest_service"
+
+class Forecast < ActiveRecord::Base
+  has_many :periods, -> { order "start ASC" }, dependent: :destroy
+  scope :recent, (lambda do |zip_code|
+    includes(periods: :predictions)
+      .where(zip_code: zip_code)
+      .where(["created_at > ?", 12.hours.ago])
```
jmmastey: I don't think these array brackets are necessary, are they?

event_system/showmetheweather/app/models/forecast.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def prediction_for_period(prediction_label, period_name)
+    period = periods.find { |p| p.name == period_name }
+    return "--" unless period
+    prediction = period.predictions.find do |p|
+      p.label == prediction_label
+    end
+    return "--" if prediction.nil? || prediction.value.empty?
+    prediction.value
+  end
```
jmmastey: Overall, this method is a bit too complex. Can you split it into a few smaller private methods?

event_system/showmetheweather/app/models/forecast.rb
```diff
@@ -0,0 +1,59 @@
...
+                 include: { predictions: { only: [:label, :value] } } },
+    }
+    super(options)
+  end
+
+  private
+
+  def self.retrieve_forecast(zip_code)
+    NWS::RestService.forecast_by_zip_code(zip_code, days: 7)
+  end
```
jmmastey: Nice. I dig this. Fwiw, it's not _actually_ private if it's a class method.

event_system/showmetheweather/app/models/forecast.rb
```diff
@@ -0,0 +1,59 @@
...
+  def self.store_forecast(forecast)
+    new_forecast = nil
+    ActiveRecord::Base.transaction do
+      new_forecast = Forecast.create!(zip_code: forecast.zip_code)
+      forecast.periods.each_pair do |name, attributes|
+        store_period(new_forecast, name, attributes)
+      end
+    end
+    new_forecast
+  end
```
jmmastey: So the return value of a transaction is similar to a method (it returns the last value from the transaction), so would this be a possible refactor?

```
def self.store_forecast(forecast_data)
  ActiveRecord::Base.transaction do
    Forecast.create!(zip_code: forecast_data.zip_code).tap do |forecast|
      forecast_data.periods.each_pair do |name, attributes|
         store_period(forecast, name, attributes)
      end
    end
end
```

Even still the nesting is a little deep, but it seems more readable to me.

event_system/showmetheweather/app/views/devise/mailer/confirmation_instructions.html.erb
```diff
@@ -0,0 +1,5 @@
+<p>Welcome <%= @email %>!</p>
+
+<p>You can confirm your account email through the link below:</p>
+
+<p><%= link_to 'Confirm my account', confirmation_url(@resource, confirmation_token: @token) %></p>
```
jmmastey: I generally avoid generating the devise views, because now you have to manage them and upgrade them when devise changes. If you don't need to modify this email, just remove it from source control.

event_system/showmetheweather/app/views/devise/passwords/edit.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    <%= f.label :password_confirmation, "Confirm new password" %><br />
+    <%= f.password_field :password_confirmation, autocomplete: "off" %>
+  </div>
+
+  <div class="actions">
+    <%= f.submit "Change my password" %>
+  </div>
+<% end %>
+
+<%= render "devise/shared/links" %>
```
jmmastey: Same here as with other devise files. If not changed, don't keep them.

As a side note, a huge chunk of your app appears to currently comprise of default devise views :D

event_system/showmetheweather/app/views/forecasts/show.html.erb
```diff
@@ -0,0 +1,81 @@
+<div id="forecast">
+  <div class="row">
+    <div class="col-xs-12">
+      <h1>Forecast for <%= @forecast.zip_code %></h1>
+    </div>
+  </div>
+  <div class="row">
+    <div class="col-xs-2">&nbsp;</div>
```
jmmastey: `&nbsp;` is a frontend anti-pattern. try to remove it, let me know if you need suggestions.

event_system/showmetheweather/app/views/forecasts/show.html.erb
```diff
@@ -0,0 +1,81 @@
...
+    </div>
+    <div class="col-xs-2" id="<%= tonight_name.parameterize + "-low" %>">
+      <h2><%= @forecast.prediction_for_period("Low Temperature", tonight_name) %></h2> low
+    </div>
+    <div class="col-xs-2" id="<%= tonight_name.parameterize + "-precip" %>">
+      <h2><%= @forecast.prediction_for_period("Chance of Precipitation", tonight_name) %>%</h2> chance of precipitation
+    </div>
+    <div class="col-xs-2">&nbsp;</div>
+  </div>
+  <br />
```
jmmastey: `<br />` is also an antipattern, same as `&nbsp;`

event_system/showmetheweather/app/views/forecasts/show.html.erb
```diff
@@ -0,0 +1,81 @@
...
+  <table class="table table-striped" id="week-forecast">
```
jmmastey: This view has gotten pretty large at by the time I get to this code. Can you split this stuff into partials?

event_system/showmetheweather/config/initializers/inflections.rb
```diff
@@ -0,0 +1,16 @@
...
+#   inflect.plural /^(ox)$/i, '\1en'
+#   inflect.singular /^(ox)en/i, '\1'
+#   inflect.irregular 'person', 'people'
+#   inflect.uncountable %w( fish sheep )
+# end
+
+# These inflection rules are supported but not enabled by default:
+# ActiveSupport::Inflector.inflections(:en) do |inflect|
+#   inflect.acronym 'RESTful'
+# end
```
jmmastey: Delete unused files.

event_system/showmetheweather/config/routes.rb
```diff
@@ -0,0 +1,7 @@
+Rails.application.routes.draw do
+  devise_for :users
+  root 'forecasts#index'
+  get 'forecasts/index'
+  post 'forecasts/search'
+  get 'forecasts/show/:zip_code', to: 'forecasts#show', as: 'forecasts_show'
+end
```
jmmastey: Overall, I tend to prefer a `resource` style declaration, but this is probably honestly a better fit for this particular usage. Still, you could move towards `get forecasts/:zip_code` as a resourceful route, for instance, and improve clarity.

event_system/showmetheweather/db/migrate/20150202211602_create_periods.rb
```diff
@@ -0,0 +1,13 @@
+class CreatePeriods < ActiveRecord::Migration
+  def change
+    create_table :periods do |t|
+      t.string :name
+      t.datetime :start
+      t.datetime :end
```
jmmastey: Are these fields nullable? It's probably worth declaring `null: false` to enforce consistency.

event_system/showmetheweather/db/migrate/20150211234148_add_zip_code_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddZipCodeToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :zip_code, :string
```
jmmastey: Smart keeping it as a string.

event_system/showmetheweather/features/email_me_the_weather.feature
```diff
@@ -0,0 +1,20 @@
+Feature: email me the weather
+
+  As a user
+  I would like to have the weather emailed to me
+  So that I do not have to visit the site each day
+
+  Scenario: sign up
+    Given I am not signed in 
+    When I sign up 
+    Then I see that I am signed in
```
jmmastey: This scenario is probably not salient to the "email me the weather" feature. The very last test is, though. Generally speaking, this test and the second one are more properly features of devise, which makes them not-terribly-useful tests.

event_system/showmetheweather/features/show_me_the_weather.feature
```diff
@@ -0,0 +1,17 @@
+Feature: show me the weather
+
+  As a human
+  I would like to see a weather forecast
+  So that I can protect my fragile meatbag with the appropriate clothing
```
jmmastey: Smart move.

event_system/showmetheweather/features/show_me_the_weather.feature
```diff
@@ -0,0 +1,17 @@
...
+
+  Scenario: weather forecast is shown for zip code
```
jmmastey: Capitalize the first word of a scenario.

event_system/showmetheweather/features/show_me_the_weather.feature
```diff
@@ -0,0 +1,17 @@
...
+    Given I have navigated to the site
+    When I enter my zip code
+    Then I see today's weather forecast
+    And I see tonight's weather forecast
+    And I see the week's weather forecast
```
jmmastey: These are good test specifications, btw.

event_system/showmetheweather/features/show_me_the_weather.feature
```diff
@@ -0,0 +1,17 @@
...
+
+  Scenario: error is shown for invalid zip code
+    Given I have navigated to the site
+    When I enter an invalid zip code
+    Then I see an error
```
jmmastey: This one might be a little more specific? What kind of error, maybe? What page I'm on afterward?

event_system/showmetheweather/features/step_definitions/email_steps.rb
```diff
@@ -0,0 +1,22 @@
+Given(/^I have a user account$/) do
+  @user = FactoryGirl.create(:user)
+end
+
+When(/^the daily email forecast is sent to me$/) do
+  forecast = Forecast.find_by(zip_code: @user.zip_code)
+  ForecastMailer.daily_email(forecast, @user).deliver_now
```
jmmastey: This may verge on a tautological test. The most valuable part of the test is that you get the right zip code worth of weather, but you've manually specified the "correct" zip here. The code that sends daily emails should probably be in a class, and that class should test the wiring of the user to the zip.

event_system/showmetheweather/features/step_definitions/email_steps.rb
```diff
@@ -0,0 +1,22 @@
...
+  expect(body).to have_css("#forecast")
+  expect(body.find("#forecast")).to have_content("Today")
+  expect(body.find("#forecast")).to have_content("Tonight")
+  expect(body.find("#forecast")).to have_content("100")
+  expect(body.find("#forecast")).to have_content("75")
+  expect(body.find("#forecast")).to have_content("50%")
+  expect(body.find("#forecast")).to have_content("25%")
+  expect(body.find("#forecast")).to have_content("Mostly Sunny")
+  expect(body.find("#forecast")).to have_content("Mostly Cloudy")
+end
```
jmmastey: I believe strongly that there's a method call that lets you scope into `#forecast`. Looking for it. Try...

```
within("#forecast") do
  expect(page).to have_content("Today")
end
```

Also, the very first assertion is redundant since the other ones will only succeed if `#forecast` exists.

event_system/showmetheweather/features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,75 @@
...
+When(/^I enter my zip code$/) do
+  fill_in "forecast[zip_code]", :with => "60606"
+  click_button "Show Me The Weather!"
+end
+
+Then(/^I see today's weather forecast$/) do
+  expect(find("#forecast")).to have_content("Today")
+  expect(find("#today-high")).to have_content("100")
+  expect(find("#today-precip")).to have_content("50%")
+  expect(find("#today-conditions")).to have_content("Mostly Sunny")
```
jmmastey: Fwiw, these are probably more appropriate as classes than IDs. Very few elements on a page actually deserve an ID, as they jack up the CSS specificity rules really badly.

event_system/showmetheweather/features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,75 @@
...
+  visit edit_user_registration_path
+  fill_in "ZIP code", with: "90210"
+  fill_in "Current password", with: "abcd1234"
+  click_button "Update"
+end
+
+Then(/^my profile is updated$/) do
+  visit edit_user_registration_path
+  expect(find("#user_zip_code").value).to eq("90210")
+end
```
jmmastey: Typically, try to organize your step definition files to put all `Given` up top, then `When`, and `Then`.

event_system/showmetheweather/features/support/hooks.rb
```diff
@@ -0,0 +1,4 @@
+Before do
+  FactoryGirl.create(:forecast)
+  FactoryGirl.create(:another_forecast)
+end
```
jmmastey: Why? This probably belongs in a `Given` step, which can be invoked as a `Background: Given there are forecasts`

event_system/showmetheweather/lib/nws/dwml_parser.rb
```diff
@@ -0,0 +1,87 @@
+require "nokogiri"
+
+module NWS
+  class DWMLParser
+    def initialize(xml)
+      @forecast = Nokogiri::XML(xml)
```
jmmastey: Whenever I see XML in code review I take a deep breath, thanks for coping with that shit :D

event_system/showmetheweather/lib/nws/dwml_parser.rb
```diff
@@ -0,0 +1,87 @@
...
+    def time_layouts
+      return @time_layouts if @time_layouts
+
+      time_layouts = {}
+      @forecast.xpath("//time-layout").each do |time_layout_element|
+        key = time_layout_element.at_xpath("layout-key").text
+        start_times = time_layout_element.xpath("start-valid-time")
+        end_times = time_layout_element.xpath("end-valid-time")
+        periods = []
+        start_times.each_with_index do |start_time, i|
```
jmmastey: This probably deserves an `each_with_object`, and possibly its own method. Since we're dealing with XML there's always a bit of a mess, but I'd say this method is slightly too large.

event_system/showmetheweather/lib/nws/dwml_parser.rb
```diff
@@ -0,0 +1,87 @@
...
+      parse_values("//temperature[@type='maximum']", "High Temperature")
+    end
+
+    def parse_low_temps
+      parse_values("//temperature[@type='minimum']", "Low Temperature")
+    end
+
+    def parse_precipitation
+      parse_values("//probability-of-precipitation", "Chance of Precipitation") 
+    end
```
jmmastey: These are queries, not commands. Probably deserve a noun name.

event_system/showmetheweather/lib/nws/rest_service.rb
```diff
@@ -0,0 +1,44 @@
+require "nws/forecast"
+require "nws/dwml_parser"
+require "uri"
+require "open-uri"
+
+module NWS
+  class RestService
+    SERVICE_HOST = "graphical.weather.gov"
+    SERVICE_PATH = "/xml/sample_products/browser_interface/ndfdBrowserClientByDay.php"
```
jmmastey: This stuff should probably move into app configuration somewhere.

event_system/showmetheweather/lib/nws/rest_service.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    private
+
+    def self.retrieve_forecast(zip_code, days)
+      uri = build_uri({
+        "zipCodeList" => zip_code,
+        "format" => "12 hourly",
+        "startDate" => Time.new.strftime("%Y-%m-%d"),
+        "numDays" => days,
+        "Unit" => "e"
```
jmmastey: Good old `"Unit" => "e"`. ;) Not your fault, not yours to change, but man fuck APIs...

event_system/showmetheweather/spec/controllers/forecasts_controller_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+RSpec.describe ForecastsController, :type => :controller do
+
+  describe "GET index" do
+    it "returns http success" do
+      get :index
+      expect(response).to have_http_status(:success)
+    end
+  end
+
+end
```
jmmastey: Is this the default generated test? It doesn't seem like a very useful test and probably retreads your integration tests. Delete or expand it.

event_system/showmetheweather/spec/factories/forecast.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  factory :another_forecast, class: "Forecast" do
+    zip_code "90210"
+    after(:create) do |forecast|
+      create(:period_today, forecast: forecast)
+      create(:period_tonight, forecast: forecast)
+    end
+  end
+end
```
jmmastey: Usually you should leave your factories general. For this case, for example, when you currently invoke `FactoryGirl.create(:another_forecast)`, you could just do `FactoryGirl.create(:forecast, zip_code: "90210")`, which prevents the factories from needing to understand the exigencies of the test suite.

event_system/showmetheweather/spec/factories/period.rb
```diff
@@ -0,0 +1,23 @@
...
+    name "Tonight"
+    start "2015-02-02T18:00:00-06:00"
+    self.end "2015-02-03T06:00:00-06:00"
+    after(:create) do |period|
+      create(:prediction_low_temp, period: period)
+      create(:prediction_precipitation_25, period: period)
+      create(:prediction_conditions_cloudy, period: period)
+    end
+  end
+end
```
jmmastey: As an alternative strategy, try:

```
factory :period do
  name "etc"
  ...

  trait :for_today do
    name "Today"
    start "whatever"
  end
end
```

Then your invocation is `create(:period, :for_today)`, which is more predictable for other developers. To see no factory named `:period` is somewhat surprising if you're factory-ing `Period` objects.

event_system/showmetheweather/spec/helpers/forecasts_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+# describe ForecastsHelper do
+#   describe "string concat" do
+#     it "concats two strings with spaces" do
+#       expect(helper.concat_strings("this","that")).to eq("this that")
+#     end
+#   end
+# end
+RSpec.describe ForecastsHelper, :type => :helper do
+  skip
+end
```
jmmastey: Delete unused files.

event_system/showmetheweather/spec/nws/forecast_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+      end
+      
+      context "when the fields of two forecast objects are not the same" do
+        it "returns false" do
+          expect(Forecast.new("90210", {})).not_to eq(Forecast.new("60606", {}))
+        end
+      end
+    end
+  end
+end
```
jmmastey: This never tests the period matching. It's probably not as important, but it _is_ more likely to break (with recursive `==` matching and such, especially since hashes are ordered and different orders probably don't match correctly).

jmmastey:
Hey, added some notes, but this is really solid. There's nothing significantly missing or misplaced. Take a look at the suggestions, let me know if any are wrong or unclear, and move on. Also come talk to me about joining a team :)

PR #515, opened 2015-02-16
Cyoa (need discussion)

jfinzel:
Working so far: the user registration/login, and the weather API is fully functioning and updating when needed.  Joe/Dan - please see my email upcoming before assigning.


cyoa/app/assets/javascripts/home.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
```
jmmastey: Delete unused files.

cyoa/app/assets/javascripts/users.js.coffee
```diff
@@ -0,0 +1,14 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
+jQuery ->
+  $('form .remove_fields').click (event) ->
+    $(this).prev('input[type=hidden]').val('1')
```
jmmastey: Not sure what the intent is here? Using a clearly named method may help (same thing we do for ruby)

cyoa/app/controllers/home_controller.rb
```diff
@@ -0,0 +1,10 @@
+require './app/models/forecast'
+require './app/models/api_data_transfer/weather_refresher'
+
+class HomeController < ApplicationController
+  def index
+    WeatherRefresher.refresh_if_needed
+    @forecast_daily = Forecast.daily_for_next_week
+    @forecast_three_hour = Forecast.three_hour_for_next_week
+  end
```
jmmastey: I dig this in general. I might move the `refresh_if_needed` stuff to a before action, since it's not _really_ related to the displaying of the data.

cyoa/app/controllers/users_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+  end
+
+  def user_params
+    params.require(:user)
+          .permit(:phone,
+                  user_notifications_attributes: [:id,
+                                                  :notification_time,
+                                                  :active])
+  end
```
jmmastey: Dem indents. Thanks rubocop...

This controller is clean, nicely done.

cyoa/app/helpers/application_helper.rb
```diff
@@ -0,0 +1,10 @@
+module ApplicationHelper
+  def link_to_add_fields(name, f, association)
+    new_object = f.object.send(association).klass.new
```
jmmastey: This line is a bit monstrous. If I wanted to remove an association from one of my classes, for instance, I might grep around the codebase to find any uses of that association. This makes it pretty tough. Also I think the `klass` is redundant, technically.

cyoa/app/helpers/application_helper.rb
```diff
@@ -0,0 +1,10 @@
...
+    id = new_object.object_id
+    fields = f.fields_for(association, new_object, child_index: id) do |builder|
+      render(association.to_s.singularize + "_fields", f: builder)
+    end
+    link_to(name, '#', class: "add_fields", data: {id: id, fields: fields.gsub("\n", "")})
```
jmmastey: The rest of the method is pretty rough, too. It may almost be worth a comment. If there are only a few (maybe <= 4) types of associations, though, I'd probably recommend just leaving duplication and having explicit methods for different types instead.

cyoa/app/helpers/home_helper.rb
```diff
@@ -0,0 +1,2 @@
+module HomeHelper
+end
```
jmmastey: Delete all unused classes.

cyoa/app/helpers/private_attr_accessor.rb
```diff
@@ -0,0 +1,6 @@
+module PrivateAttrAccessor
+  def private_attr_accessor(*names)
+    private
+    attr_accessor *names
+  end
+end
```
jmmastey: These always make me squicky. Waiting to see usage :)

cyoa/app/mailers/user_notification_mailer.rb
```diff
@@ -0,0 +1,12 @@
+class UserNotificationMailer < ApplicationMailer
+  def user_notification_email(user_notification)
+    @user = User.find(user_notification.user_id)
+    @weather_url = home_index_path
+    st = user_notification.notification_time_today
+    et = user_notification.notification_time_today_end
+    @forecast_daily = Forecast.daily_for_range(st, et)
+    @forecast_three_hour = Forecast.three_hour_for_range(st, et)
+    binding.pry
```
jmmastey: Remove debugging statements.

cyoa/app/models/api/lat_lon_zipcode_client.rb
```diff
@@ -0,0 +1,33 @@
+require 'nori'
+require 'savon'
+require './app/models/api/nws_client.rb'
+require './app/models/api/lat_lon_zipcode_client_response'
+
+class LatLonZipcodeClient
```
jmmastey: This classname is a bit odd. Probably because it reveals too much of its implementation via the classname. Maybe that's fine? Anyway, feels awkward.

cyoa/app/models/api/lat_lon_zipcode_client.rb
```diff
@@ -0,0 +1,33 @@
...
+  extend Savon::Model
+
+  client wsdl: NWSClient::WSDL, convert_request_keys_to: NWSClient::REQUEST_KEYS
+  operations :lat_lon_list_zip_code
+
+  def request(zip_code_list)
+    LatLonZipcodeClientResponse.new(lat_lon_list_zip_code(zip_code_list), zip_code_list).zip_code_lat_lon
+  end
```
jmmastey: In ruby, we usually use `.call` as the name for this sort of generic "execute" method.

Count of words in line #13: zip(code): 5, list: 3, lat: 3, lon: 3. Work on naming.

cyoa/app/models/api/lat_lon_zipcode_client_response.rb
```diff
@@ -0,0 +1,57 @@
+require 'nori'
+
+class LatLonZipcodeClientResponse
```
jmmastey: This is a very java-esque name. What's the reasoning for this class to exist?

cyoa/app/models/api/lat_lon_zipcode_client_response.rb
```diff
@@ -0,0 +1,57 @@
...
+  def response_parameters
+    body_response_hash.fetch(:list_lat_lon_out).fetch(PARAMETER_CONTAINER)
+  end
+
+  def lat_lon_list
+    response_parameters.fetch(:lat_lon_list)
+  end
+
+  def zip_code_lat_lon
+    merge_zip_code_with_lat_lon.each_with_object({}) do |(key, value), all|
```
jmmastey: this line is pretty complicated. consider name changes or structuring this code differently to reduce complexity.

cyoa/app/models/api/parameters/cloud_cover_parameter.rb
```diff
@@ -0,0 +1,2 @@
+class CloudCoverParameter < TemperatureTypeParameter
+end
```
jmmastey: ??

Inheritance makes for a lot of complexity. If you can avoid it reasonably, do so.

cyoa/app/models/api/parameters/conditions_icon_parameter.rb
```diff
@@ -0,0 +1,10 @@
+class ConditionsIconParameter
+  attr_reader :name, :icon_link, :type, :time_layout
+
+  def initialize(conditions_icon)
+    @name = conditions_icon[:name]
+    @icon_link = conditions_icon[:icon_link]
+    @type = conditions_icon[:@type]
+    @time_layout = conditions_icon[:@time_layout]
+  end
+end
```
jmmastey: This and the next few classes are currently just `Struct`s, basically. Consider moving to those unless you need to add behavior to them.

cyoa/app/models/api/parameters/temperature_parameter.rb
```diff
@@ -0,0 +1,32 @@
+require './app/helpers/private_attr_accessor'
+require './app/models/api/parameters/temperature_type_parameter'
+
+class TemperatureParameter
+  extend PrivateAttrAccessor
+  private_attr_accessor :temperature
```
jmmastey: So, given the relative simplicity of the class, maybe it's worth it just to use `@temperature` in this class. Certainly class variables are acceptable within the class. I don't think `private_attr_accessor` made anything better in this case.

cyoa/app/models/api/parameters/weather_conditions_parameter.rb
```diff
@@ -0,0 +1,26 @@
...
+    @coverages = get_value(weather_condition, :@coverage)
+    @intensities = get_value(weather_condition, :@intensity)
+    @weather_types = get_value(weather_condition, :@weather_type)
+    @qualifiers = get_value(weather_condition, :@qualifier)
+    @additives = get_value(weather_condition, :@additive)
+  end
+
+  private
+
+  def get_value(weather_condition, key)
```
jmmastey: `get_value` is not a good name for a method. also see if you can collapse the cases in the if/else here.

cyoa/app/models/api/weather_client.rb
```diff
@@ -0,0 +1,101 @@
+require 'time'
```
jmmastey: Overall, this belongs in `/lib`, not `/models`. Think of it as being a separated gem.

cyoa/app/models/api/weather_client.rb
```diff
@@ -0,0 +1,101 @@
...
+require 'nori'
+require 'savon'
+require './app/models/api/nws_client'
+require './app/models/api/weather_client_response'
```
jmmastey: All of these things should be available automatically by virtue of using bundler.

cyoa/app/models/api_data_transfer/forecast_refresher.rb
```diff
@@ -0,0 +1,140 @@
...
+    forecast.start_time = load_params[:time_key][:start_time]
+    forecast.end_time = load_params[:time_key][:end_time]
+    forecast.maxt = load_params[:maxt]
+    forecast.mint = load_params[:mint]
+    forecast.cloud_cover = load_params[:cloud_cover]
+    forecast.icon_link = load_params[:icon_link]
+    forecast.save!
+    load_params[:forecast] = forecast
+    add_forecast_weather!(response, load_params)
+  end
```
jmmastey: Use `Forecast.new.tap` here.

jmmastey: Or `Forecast.create(foo: '', bar: '')`

cyoa/app/models/api_data_transfer/forecast_refresher.rb
```diff
@@ -0,0 +1,140 @@
...
+                              forecast_type_id: load_params[:time_key][:forecast_type_id],
+                              start_time: load_params[:time_key][:start_time],
+                              end_time: load_params[:time_key][:end_time])
+    if forecast.empty?
+      insert_forecast(response, load_params)
+    else
+      load_params[:forecast] = forecast.first
+      update_forecast(response, load_params)
+    end
+  end
```
jmmastey: Use `Forecast.first_or_create`, probably.

cyoa/app/models/api_data_transfer/forecast_refresher.rb
```diff
@@ -0,0 +1,140 @@
...
+
+  def self.add_forecast_weather!(response, load_params)
+    indexes = param_indexes(response, load_params, response.weather_data.weather[load_params[:point_index]].time_layout)
+    raise WeatherLoaderError, "Unhandled multiple weather times" if indexes.count > 1
+    return false if indexes.count == 0
+    ForecastWeatherType.destroy_all(forecast_id: load_params[:forecast][:id])
+    load_params.merge!(indexes: indexes)
+    insert_weather!(response, load_params)
+  end
```
jmmastey: I feel like we talked over the methods in this class in person. Let me know if you need a refresher on it.

cyoa/app/models/api_data_transfer/forecast_type_refresher.rb
```diff
@@ -0,0 +1,37 @@
+require './app/models/forecast_type'
```
jmmastey: Rails really does do good autoloading, I promise :)

cyoa/app/models/api_data_transfer/forecast_type_refresher.rb
```diff
@@ -0,0 +1,37 @@
...
+  def self.times_by_forecast_type(times)
+    times.each_with_object([]) do |(key, value), final|
+      final << key
+    end.group_by { |row| row[:forecast_type_id] }
+  end
+
+  def self.update_refresh_start_times(forecast_times)
+    forecast_times.map do |key, value|
+      value.min_by { |h| h[:start_time] }
+    end.each do |forecast_refresh|
```
jmmastey: This construction is really awkward (chaining off an `end`). Just use a local for it, or give the start_time part a method of its own.

```
min_times = forecast_times.map { |key, value| value.min_by { |h| h[:start_time] } }
min_times.each do
 ...
end
```

cyoa/app/models/api_data_transfer/point_loader.rb
```diff
@@ -0,0 +1,41 @@
...
+    end.join(" ")
+  end
+
+  def self.call_api(zip_code_list)
+    requester.new.request(zip_code_list)
+  end
+
+  def self.requester
+    LatLonZipcodeClient
+  end
```
jmmastey: I'm not positive that this abstraction is really helpful. If you really did want to inject the dependency, I might use it like this:

```
def self.call_api(zipcodes, client = LatLongZipcodeClient.new)
  client.call(zipcodes)
end

File: cyoa/app/models/api_data_transfer/time_key_builder.rb [hunk 26]
```diff
@@ -0,0 +1,95 @@
...
+        if tk.count > 1
+          tk.each do |k|
+            final = add_value_pair(k, pair, final)
+          end
+        else
+          final = add_value_pair(tk[0], pair, final)
+        end
+      end
+    end
+  end
```
jmmastey: Very overly complicated method. What are you trying to do here?

cyoa/app/models/api_data_transfer/time_key_builder.rb
```diff
@@ -0,0 +1,95 @@
...
+    else
+      st_1 = start_time_for_24_hour_start(st)
+      et_1 = end_time_for_24_hour_start(st)
+      st_2 = start_time_for_24_hour_end(et)
+      et_2 = end_time_for_24_hour_end(et)
+      return [{ forecast_type_id: id, start_time: st_1, end_time: et_1 }] unless st_1 != st_2
+      [{ forecast_type_id: id, start_time: st_1, end_time: et_1 },
+       { forecast_type_id: id, start_time: st_2, end_time: et_2 }]
+    end
+  end
```
jmmastey: Good lord, this method is a mess. Also refactor. If nothing else, the two-char variable names make it really hard to understand what's happening.

cyoa/app/models/api_data_transfer/weather_loader.rb
```diff
@@ -0,0 +1,46 @@
+require 'active_support/all'
+require './app/models/point'
+require './app/models/forecast'
+require './app/models/weather_type'
+require './app/models/forecast_weather_type'
+require './app/models/api_data_transfer/time_key_builder'
+require './app/models/api_data_transfer/forecast_refresher'
+require './app/models/api_data_transfer/forecast_type_refresher'
```
jmmastey: Stahp.

cyoa/app/models/api_data_transfer/weather_loader.rb
```diff
@@ -0,0 +1,46 @@
...
+        point = Point.where("round(lat, 2) = :lat AND round(lon, 2) = :lon",
+                            { lat: location.latitude,
+                              lon: location.longitude }).first
+        load_point_times(response, { point: point, point_index: index })
+      end
+    end
+  end
+
+  def self.location_by_key(response, location_key)
+    response.weather_data.locations.select { |location| location.location_key == location_key }.first
```
jmmastey: `.find` or `.detect` == `.select.first`

cyoa/app/models/forecast.rb
```diff
@@ -0,0 +1,122 @@
...
+    .where("end_time <= :end", { end: et })
+    .group("forecasts.id",
+           "forecasts.start_time",
+           "forecasts.end_time",
+           "forecasts.maxt",
+           "forecasts.mint",
+           "forecasts.cloud_cover",
+           "forecasts.icon_link")
+    .order(:start_time)
+  end
```
jmmastey: I'd probably move something like this out to another class. Or, alternatively, I'd build a view and then just select from that view instead. This is just a ton of code for a select (or for a model)

cyoa/app/models/forecast.rb
```diff
@@ -0,0 +1,122 @@
...
+        c = "#{c} #{self.coverages[index]}"
+        final << "#{c} #{wt}"
+      else
+        wt = weather_type
+        c = self.coverages[index]
+        final << "#{assign_additive(index)}#{c} #{wt}"
+      end
+    end
+    final.join(", ")
+  end
```
jmmastey: Display logic belongs in Helpers or Presenters

cyoa/app/models/forecast_weather_type.rb
```diff
@@ -0,0 +1,8 @@
+class ForecastWeatherType < ActiveRecord::Base
+  belongs_to :forecast
+  belongs_to :weather_type
+
+  def formatted_weather
+    
+  end
```
jmmastey: ?

cyoa/app/views/partials/_three_hour_weather.html.erb
```diff
@@ -0,0 +1,20 @@
+<div>Detail Forecast Every 3 Hours</div>
+<ul id="three_hour_forecasts">
+<% @forecast_three_hour.group_by { |row| row.start_date_in_app_time_zone }.each do |date, forecasts| %>
```
jmmastey: This sort of logic is too much to put into a view. Consider moving into a helper.

cyoa/config/locales/devise.en.yml
```diff
@@ -0,0 +1,60 @@
...
+  errors:
+    messages:
+      already_confirmed: "was already confirmed, please try signing in"
+      confirmation_period_expired: "needs to be confirmed within %{period}, please request a new one"
+      expired: "has expired, please request a new one"
+      not_found: "not found"
+      not_locked: "was not locked"
+      not_saved:
+        one: "1 error prohibited this %{resource} from being saved:"
+        other: "%{count} errors prohibited this %{resource} from being saved:"
```
jmmastey: If you aren't modifying the default devise partials, just delete them. Otherwise you have to maintain them, which is a pain.

cyoa/config/routes.rb
```diff
@@ -0,0 +1,64 @@
+Rails.application.routes.draw do
+  get 'home/index'
+
+  devise_for :users
+  resources :users, only: [:edit]
+  get 'users/:id', :to => 'users#show', :as => :user
+  patch 'users/:id', :to => 'users#update'
```
jmmastey: These three (maybe all four?) lines all belong in one block, grouped together.

cyoa/config/routes.rb
```diff
@@ -0,0 +1,64 @@
...
+  #   end
+  #   resources :posts, concerns: :toggleable
+  #   resources :photos, concerns: :toggleable
+
+  # Example resource route within a namespace:
+  #   namespace :admin do
+  #     # Directs /admin/products/* to Admin::ProductsController
+  #     # (app/controllers/admin/products_controller.rb)
+  #     resources :products
+  #   end
```
jmmastey: Delete all this example code.

cyoa/db/migrate/20150206173451_add_unique_constraint_to_forecast_type.rb
```diff
@@ -0,0 +1,16 @@
+class AddUniqueConstraintToForecastType < ActiveRecord::Migration
+  def up
+    update <<-SQL
+      ALTER TABLE forecast_types
+      ADD CONSTRAINT unique_forecast_type 
+      UNIQUE (forecast_type);
+    SQL
+  end
```
jmmastey: Would an index not be helpful here? It might be worth just creating a unique index instead.

cyoa/db/migrate/20150213150736_modify_forecast_weather_type_constraint.rb
```diff
@@ -0,0 +1,11 @@
+class ModifyForecastWeatherTypeConstraint < ActiveRecord::Migration
+  def change
+    # This index was unique, but the data is not necessarily unique.
+    # There can be multiple of the same weather type for one forecast,
+    # having a different coverage, but we are not confident enough to add
+    # another unique constraint based on this which would cause the API refresh
+    # to fail if violated.
```
jmmastey: Interesting. I like this comment, actually :)

cyoa/db/seeds.rb
```diff
@@ -0,0 +1,7 @@
+require './app/models/api_data_transfer/point_loader'
+
+ForecastType.create(forecast_type: "3-hour")
+ForecastType.create(forecast_type: "24-hour")
+# PointLoader.load("60606 60532")
+Point.create(lat: 41.8370, lon: -87.6850, zip: "60606")
+Point.create(lat: 41.7918, lon: -88.0878, zip: "60532")
```
jmmastey: This is probably not seed data so much as test data, in which case it belongs in tests. Seed data would include, for instance, a list of US states (if you needed those for your app). It's stuff that needs to exist in prod for the site to operate.

cyoa/features/landing.feature
```diff
@@ -0,0 +1,19 @@
...
+    And I see a register now option
+    And I see a login option
+
+  Scenario: Logged in landing
+    Given some registered users
+    And I am logged in
+    When I navigate to the main URL
+    Then I see weather information
+    And I see a logout option
+    And I see a settings option
```
jmmastey: Your feature descriptions are really solid. Nice

cyoa/features/step_definitions/login_steps.rb
```diff
@@ -0,0 +1,36 @@
...
+
+Then(/^I receive a login error message$/) do
+  expect(page).to have_content("Invalid email or password")
+end
+
+When(/^I submit an invalid login$/) do
+  fill_in "Email", with: "joetestzzer"
+  fill_in "Password", with: "va"
+  click_button "Log in"
+end
```
jmmastey: Try to organize your step definitions to do _all_ Givens, then _all_ Whens, then _all_ Thens.

cyoa/features/step_definitions/login_steps.rb
```diff
@@ -0,0 +1,36 @@
...
+  fill_in "Password", with: @user.password
+  click_button "Log in"
+end
+
+Given(/^I am on the login page$/) do
+  visit new_user_session_path
+end
+
+When(/^I submit a valid login$/) do
+  step "I am logged in"
```
jmmastey: Calling steps from each other is actually a bit of an anti-pattern. Better is to define a method that both steps call.

cyoa/features/step_definitions/registration_steps.rb
```diff
@@ -0,0 +1,53 @@
+Given(/^I am not a registered user$/) do
+
```
jmmastey: Add `# NOOP` comment like the other ones. I like these steps.

cyoa/features/step_definitions/weather_steps.rb
```diff
@@ -0,0 +1,11 @@
+Given(/^I am on the weather page$/) do
+  visit home_index_path
+end
+
+Then(/^I can see the weather for the next (\d+) hours$/) do |number_of_hours|
+  pending # express the regexp above with the code you wish you had
```
jmmastey: ?

cyoa/spec/controllers/home_controller_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+    it "returns http success" do
+      get :index
+      expect(response).to have_http_status(:success)
+    end
+  end
+
+  it "returns the default lat lon in the right format" do
+    expect(HomeController.default_lat_lon).to match(/[\-]?\d{1,2}\.\d+,[\-]?\d{1,3}\.\d+( [\-]?\d{1,2}\.\d+,[\-]?\d{1,3}\.\d+)*/)
+  end
+end
```
jmmastey: Is this a generated test file, or are you using it? If the former, delete. If the latter... possibly still delete, unless that format test is really really useful. In which case, probably move it to your integration test, and still still delete.

cyoa/spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe UsersController, :type => :controller do
+
+end
```
jmmastey: This one definitely delete.

cyoa/spec/factories.rb
```diff
@@ -0,0 +1,30 @@
...
+    phone    '630-849-2299'
+    password 'secret123'
+  end
+
+  factory :user_notification do
+    user
+    notification_time '06:00:00'
+    active            true
+  end
+end
```
jmmastey: Better to have separate files for your factories: `factories/users.rb`, `factories/points.rb`. Be careful with picking attribute values as well: at the moment it appears that you cannot generate two users within one test.

Finally, do those forecast/point/etc factories work? They have no attributes.

cyoa/spec/mailers/user_notification_mailer_spec.rb
```diff
@@ -0,0 +1,5 @@
+require "rails_helper"
+
+RSpec.describe UserNotificationMailer, :type => :mailer do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
jmmastey: Delete all unused files.

cyoa/spec/models/api/lat_lon_zipcode_client_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  describe "for multiple zip codes", vcr: { cassette_name: "multiple zip codes" } do
+    let(:client) { LatLonZipcodeClient.new }
+    let(:response) { client.request("60606 60532") }
+    it "has the lat_lon_list parameter in the response" do
+      expect(response.response_parameters).to have_key(:lat_lon_list)
+    end
+    it "returns the expected latitude and longetude format" do
+      expect(response.lat_lon_list).to match(/[\-]?\d{1,2}\.\d+,[\-]?\d{1,3}\.\d+( [\-]?\d{1,2}\.\d+,[\-]?\d{1,3}\.\d+)*/)
```
jmmastey: Eventually, I'd probably write a new rspec matcher called `be_a_lat_long_pair` or something. This regex is pretty complicated.

cyoa/spec/models/api/weather_client_spec.rb
```diff
@@ -0,0 +1,217 @@
+require 'savon'
+require './app/models/api/weather_client'
+
+def keys(hash)
+  hash.each_with_object([]) do |(key, value), keys|
+    keys << key
+  end
+end
```
jmmastey: ???

I think maybe you're just looking for `hash.keys`?

cyoa/spec/models/api/weather_client_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+                    :weather,
+                    :cloud_amount,
+                    :conditions_icon,
+                    :@applicable_location]
+      parameters.each do |parameter|
+        expect(weather_data.data[:parameters]).to have_key(parameter)
+      end
+    end
+
+    it "returns expected attributes as arrays of 1 object" do
```
jmmastey: This is a somewhat odd design to me. Why do attributes return as arrays of one?

cyoa/spec/models/api/weather_client_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+        expect(weather_data.data).to have_key(parameter)
+      end
+    end
+
+    it "has the expected parameters nested under :parameters" do
+      parameters = [:temperature,
+                    :weather,
+                    :cloud_amount,
+                    :conditions_icon,
+                    :@applicable_location]
```
jmmastey: Why the single instance-var-style key?

cyoa/spec/models/api/weather_client_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+    it "returns time_layouts as an array of at least 1 object" do
+      expect(weather_data.time_layouts).to be_an(Array)
+      expect(weather_data.time_layouts.count).to be > 0
+    end
+
+    it "returns time_layouts with layout_key, start_valid_time, end_valid_time" do
+      expect(weather_data.time_layouts[0]).to respond_to(:layout_key,
+                                                         :start_valid_time,
+                                                         :end_valid_time)
+    end
```
jmmastey: Possibly low value test. Respond-to is the weakest kind of assertion. Is there anything better we can assert?

cyoa/spec/models/api/weather_client_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+    it "returns maxt with temperature type parameters" do
+      expect(weather_data.temperatures[0].maxt).to respond_to(:name, :value, :type, :units, :time_layout)
+    end
+
+    it "returns mint with temperature type parameters" do
+      expect(weather_data.temperatures[0].mint).to respond_to(:name, :value, :type, :units, :time_layout)
+    end
+
+    it "returns cloud cover with correct parameters" do
+      expect(weather_data.cloud_covers[0]).to respond_to(:name, :value, :type, :units, :time_layout)
```
jmmastey: `weather.cloud_covers` is already a weird construction. adding `cloud_covers[0]` is even weirder. I think your tests are telling you that the design is not quite right.

cyoa/spec/models/api/weather_client_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+      expect(weather_data.conditions_icons[0]).to respond_to(:name, :icon_link, :type, :time_layout)
+    end
+
+    it "returns applicable locations that are one of the location_keys" do
+      location_keys = weather_data.locations.each_with_object([]) do |location, keys| 
+                        keys << location.location_key
+                      end
+      difference = weather_data.applicable_locations - location_keys
+      expect(difference).to be_empty
+    end
```
jmmastey: Since arrays are ordered, it shouldn't be necessary to do the difference thing. Something like this should suffice:

```
expect(weather.applicable_locations).to eq(weather.locations.map(&:location_key))

File: cyoa/spec/models/api/weather_client_spec.rb [hunk 56]
```diff
@@ -0,0 +1,217 @@
...
+
+    it "returns locations with location_key, latitude, and longitude" do
+      expect(weather_data.locations[0]).to respond_to(:location_key, :latitude, :longitude)
+      expect(weather_data.locations[1]).to respond_to(:location_key, :latitude, :longitude)
+    end
+
+    it "returns time_layouts as an array of at least 1 object" do
+      expect(weather_data.time_layouts).to be_an(Array)
+      expect(weather_data.time_layouts.count).to be > 0
```
jmmastey: `expect(weather.time_layouts).not_to be_empty`

cyoa/spec/models/api/weather_client_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+    end
+
+    it "returns cloud cover with correct parameters" do
+      expect(weather_data.cloud_covers[0]).to respond_to(:name, :value, :type, :units, :time_layout)
+      expect(weather_data.cloud_covers[1]).to respond_to(:name, :value, :type, :units, :time_layout)
+    end
+
+    it "returns conditions icons with correct parameters" do
+      expect(weather_data.conditions_icons[0]).to respond_to(:name, :icon_link, :type, :time_layout)
+      expect(weather_data.conditions_icons[1]).to respond_to(:name, :icon_link, :type, :time_layout)
```
jmmastey: So the problem I'm having is that the number of mechanical tests of this sort are obscuring the actual logic of the library. Since the objects are in fact classes anyway, I'm not sure these tests are contributing much value relative to the amount of noise they create.

jmmastey:
Overall really well done. Two things:

1. A few organizational suggestions (`/lib`, stuff in helpers).
2. The code in lib seems overly complicated for the usage. You don't need to prove to me that you can refactor, but if this codebase were long lived, I'd spend a bunch of time paring down and reorganizing that code.

No need to resubmit unless any of the changes are interesting. :)

cyoa/app/models/api/lat_lon_zipcode_client_response.rb
jfinzel: The XML I am parsing is really complicated.  I tried to simplify this, but I'm not sure how I'd do it.  I have to do a few of these initial things just to get something reasonable to work with....

cyoa/app/models/api/lat_lon_zipcode_client_response.rb
jfinzel: It is for the response of the API call to the Lat Lon Zipcode service from NWS.  It's complicated enough to be in its own class rather than the client itself, I think.  This is something Dan and I discussed quite a bit - separating out the response from the client.

cyoa/app/models/api/parameters/cloud_cover_parameter.rb
jfinzel: This parameter follows the identical format of the one referenced.  Essentially there are two parameters in the weather service response that are this way, so I was trying to eliminate the duplication.  What would you suggest?  It's hard to know how the superclass should be named, but they have the same parameters with the same return types.

cyoa/app/models/api/parameters/conditions_icon_parameter.rb
jfinzel: That's helpful to know.  I recognize a lot of my classes are just structural, and I had to do this to not go insane and to have an organized way to eventually take this data in a meaningful state and put it into the database.  If I put this in a Struct, could I still reference them easily when I need to, i.e. conditions_icon.icon_link, etc. ?

cyoa/app/models/api/parameters/weather_conditions_parameter.rb
jfinzel: What would be better for what I'm doing?  Also, it's hard to collapse the logic here, because weather_condition could either be an array of hashes (in which case I can use each) or simply one hash.  Or do you mean moving to other methods?

cyoa/app/models/api/weather_client.rb
jfinzel: do you mean the gems, or my own files?

cyoa/app/models/api_data_transfer/point_loader.rb
jfinzel: Yeah, I missed this one.  I had intended to remove it.  Not sure I really need to inject the dependency, as this is all so specific to this API.

cyoa/app/models/api_data_transfer/time_key_builder.rb
jfinzel: If you look at the XML you may know why... this was one of the most challenging parts of it.  The layout of the times that go with each piece of weather information is exceedingly complex, and this is all my attempt to reorganize it into something meaningful.  It may help for me to go over this one in person.  I'd be interested in improvements.

cyoa/app/models/api_data_transfer/time_key_builder.rb
jfinzel: I worked on refactoring this a lot but again, really tricky formatting.  I opted to change to st and et because I think that is clear enough, and makes it a little easier to read than something like start_time and end_time everywhere.  I will take a look if I can do anything else here.  I think I could probably move the else part to a different method.

cyoa/app/models/api_data_transfer/weather_loader.rb
jfinzel: thank you!

cyoa/app/models/forecast.rb
jfinzel: It actually can't be a view, because the start and end time are parameters.  I could filter after the view but it would not less optimal, selecting all times for the table first. It could be a table-function but for this kind of a thing, it's overkill and has other challenges.  But is there some kind of convention for doing a query that is related to a model in a separate class?  How would I write that?

I really like it being a class method because it makes a lot of sense to just do forecast.query.

cyoa/app/models/forecast.rb
jfinzel: I don't know what you mean.  Do you mean create a separate class for this?

cyoa/app/models/forecast_weather_type.rb
jfinzel: whoops.

cyoa/app/views/partials/_three_hour_weather.html.erb
jfinzel: Something like @forecast_three_hour.by_start_date.each  or something?

cyoa/config/locales/devise.en.yml
jfinzel: Hmm... I just don't know which files I really need.  I would have thought these files were needed.  How do I know which files I can delete?  Is everything just in the gem?

cyoa/config/routes.rb
jfinzel: I don't think I can group the devise for users with the resource for users.  I'm not sure how that works.  In one block, is it something like this?
```
resources :users do
only: [:edit]
get :id, to: :show, as: :user
patch :id, to: update
end
```

Or am I way off?

cyoa/db/migrate/20150206173451_add_unique_constraint_to_forecast_type.rb
jfinzel: I don't remember why I wanted to only do a unique constraint and not an index honestly.  But since this table only has two rows, I don't think it's a good candidate for an index.

cyoa/db/seeds.rb
jfinzel: Yeah, both of them are seed data, not test data.  The 3-hour and 24-hour values have meaning in my ForecastType model.  Actually, I don't really need 60532, but the other point is (currently) the only one I am planning to use in production.

cyoa/features/step_definitions/login_steps.rb
jfinzel: Good to know.

cyoa/features/step_definitions/registration_steps.rb
jfinzel: Okay, that's really helpful to know.  I thought there had to be some practice around a step where you don't need to do anything.

cyoa/features/step_definitions/weather_steps.rb
jfinzel: Yeah... I haven't implemented this.  Not sure if it's a sensible test but I want to test that the weather is showing the right date/time range.

cyoa/spec/controllers/home_controller_spec.rb
jfinzel: It was generated, then I added this test to it, then I actually moved the method to a different class!  So my bad - I will remove it.

cyoa/spec/factories.rb
jfinzel: No, I need to work on this.  I got on the track of figuring out the API but never returned to these.  Also, yeah, I suppose you are right about not generating two users.  I need to figure out how to do that.

cyoa/spec/models/api/weather_client_spec.rb
jfinzel: I'm not sure if I understand your question, but it returns an array because if you make a request with more than one lat lon point, it returns an array of multiple objects.  So for one lat lon point, this is just the expected format of the XML for one lat lon point.  If it were different, that would not be what I am expecting to parse.  Dan was telling me I should make assertions on the format of the expected XML return.

cyoa/spec/models/api/weather_client_spec.rb
jfinzel: I don't understand the question...

cyoa/spec/models/api/weather_client_spec.rb
jfinzel: It's tough for layout_key, because I don't know the entire range of time formats that they give.  I could do a check that the times are actually time formattable?

cyoa/spec/models/api/weather_client_spec.rb
jfinzel: I don't know what you mean, honestly.  "Cloud cover" is part of the weather information, and covers is just a pluralization (admittedly, an awkward 1).  But then cloud_covers[0] makes sense because cloud_covers is an array of cloud_cover values.

cloud_coverage and cloud_coverages may sound better.  Does that make sense or am I missing your point?

cyoa/spec/models/api/weather_client_spec.rb
jfinzel: Nice.

PR #516, opened 2015-02-16
Dinodex

gguntupalli:
Dinodex for second review.

Thanks
Gopi.


dino_catalog/run_dinodex.rb
```diff
@@ -0,0 +1,33 @@
...
+      @dinodex = @dinodex.find_by(header.to_sym, value.downcase)
+    end
+  end
+end
+
+def main
+  run_dinodex = RunDinodex.new
+  run_dinodex.find
+  puts run_dinodex.dinodex
+end
```
zsyed1: Instead of making these three calls in succession on `run_dinodex` (lines 28, 29 and 30), it would be better to only have to call one line like: `RunDinodex.new(ARGV).run`. After all, your class is supposed to do one thing - and that is _run_ Dinodex over the command-line arguments. Please let me know if this is unclear - hit me up on Slack.

dino_catalog/run_dinodex.rb
```diff
@@ -0,0 +1,33 @@
...
+class RunDinodex
+  attr_accessor :dinodex
+  def initialize
+    @dinodex = Dinodex.new(parse_dinos)
+  end
+
+  def parse_dinos
+    parser = Parser.new
+    dinos_hash = parser.parse_dinodex("dinodex.csv")
+         .concat parser.parse_african_dinosaurs("african_dinosaur_export.csv")
```
zsyed1: Do not include variable types in variable names. Instead of `dinos_hash` you can simply call this `dinos`. That way in the future if you choose to change the underlying data type (e.g. to array), you don't have to change the variable name everywhere.

dino_catalog/run_dinodex.rb
```diff
@@ -0,0 +1,33 @@
...
+require_relative 'parser'
+
+class RunDinodex
+  attr_accessor :dinodex
+  def initialize
+    @dinodex = Dinodex.new(parse_dinos)
+  end
+
+  def parse_dinos
+    parser = Parser.new
```
zsyed1: Since you never call `parse_dinos` except within the class `RunDinodex`, the method should be `private`.

dino_catalog/parser.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def parse_african_dinosaurs(filename)
+    african_dinos_hash = parse_dinodex(filename)
+    african_dinos_hash.each do |row|
+      row[:continent] = "Africa"
+      row[:diet] = row[:diet] == "Yes" ? "Carnivore" : "Herbivore"
+    end
+  end
+
```
zsyed1: You can compress lines 15 and 16: `parse_dinodex(filename).each do |row|`. It's short enough that you don't really need the intermediate variable `african_dinos_hash`.

dino_catalog/parser.rb
```diff
@@ -0,0 +1,30 @@
...
+class Parser
+  def initialize
+    CSV::HeaderConverters[:header_converter] = method(:header_converter)
+  end
+
+  def parse_dinodex(filename)
+    csv = CSV.read(filename, headers: true,
+          header_converters: [:downcase, :header_converter, :symbol])
+    csv.map(&:to_hash)
+  end
```
zsyed1: Nice way to leverage the CSV gem!

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,37 @@
...
+    @continent = attributes[:continent]
+    @diet = attributes[:diet]
+    @weight = attributes[:weight]
+    @walking = attributes[:walking]
+    @description = attributes[:description]
+  end
+
+  def big?
+    !@weight.nil? && @weight.to_i > WEIGHT_DIFFERENTIATOR
+  end
```
zsyed1: A shorter way to say `!@weight.nil` is simply `@weight`. So line 18 would look like: `@weight && @weight.to_i > WEIGHT_DIFFERENTIATOR`. Same with line 22.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def to_s
+    s = "#{@name} is a #{@walking} from #{@period} period and #{@continent} "
+    s << "continent with #{@diet} diet"
+    s << " and #{@weight}lbs weight" unless @weight.nil?
+    s << ". "
+    s << @description unless @description.nil?
+    s
+  end
```
zsyed1: You can avoid repeating the `s` at the beginning of every line by writing this as:
```
["#{@name} is a #{@walking}  from ...",
 "continent with #{@diet} diet?",
 ...
 ". "].join(' ')
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
+class Dinodex
+  attr_accessor :dinos
+
+  def initialize(dinos)
+    @dinos = dinos
+  end
+
+  def find_by(header, value)
+    dinodex = Dinodex.new(@dinos)
+    case header
```
zsyed1: This is an interesting design decision. You send back a _child_ dinodex that contains only those dinos that pass the search criteria.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
...
+  def search_by_criteria(header, value, dinodex)
+    dinodex.dinos.select do |dino|
+      dino.send(header).downcase.include? value
+    end
+  end
+
+  def to_s
+    @dinos.each do |dino|
+      puts dino
+    end
```
zsyed1: The method name `to_s` implies the method will return a string. It shouldn't print anything. Instead you can write this method like:
```
def to_s
  @dinos.map(&:to_s).join('\n')
end
```

zsyed1: If you really want to print something you could simply change the name of the method to `display`.

dino_catalog/run_dinodex.rb
zsyed1: Line 30 is invoking the `to_s` method on `Dinodex`. Please see comment above about the `to_s` method on `Dinodex`.

zsyed1: Nice work overall! I think this is closed @djkotowski

dino_catalog/dinodex.rb
djkotowski: And I recommend against printing to stdout in Dinodex.  Save output for the runner.

dino_catalog/dinosaur.rb
djkotowski: Or just use the `<<` operator after each line.

dino_catalog/run_dinodex.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+end
+
+def main
+  run_dinodex = RunDinodex.new
+  run_dinodex.find
+  puts run_dinodex.dinodex
+end
+
+main
```
djkotowski: You can extract the contents of `main` and put them inline here.  Having a `main` method doesn't provide you any benefit here.

djkotowski:
This is looking good.  Please review the comments, but no need for resubmission.

gguntupalli:
Thanks Zamir and Dan.

PR #517, opened 2015-02-17
ruby style guide exercise: triangle_facts

gguntupalli:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,59 @@
...
-		return [angleA, angleB, angleC]
-	end
-
-	def radians_to_degrees(rads)
-		return (rads * 180 / Math::PI).round
-	end
+  attr_accessor :side1, :side2, :side3
+
+  def initialize(side1, side2, side3)
+    @side1 = side1
```
incazteca: This could have also been fixed by doing `@side1, @side2, @side3 = side1, side2, side3`

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,59 @@
...
+  def isosceles?
+    [side1, side2, side3].uniq.length == 2
+  end
+
+  def scalene?
+    side1 != side2 && side2 != side3
+  end
+
+  def recite_facts
+    if equilateral?
```
incazteca: This conditional can be shorted to 3 single line conditionals.

incazteca:
@gguntupalli This can be closed. A resubmission is not necessary.

triangle_facts/triangle_facts.rb
ashtonthomas: @incazteca, I was thinking we were following the style guide that recommends against multi assignment per line. no?

triangle_facts/triangle_facts.rb
ashtonthomas: I've heard arguments that execution may be faster with this style of if/else, but I agree with @incazteca about preferring the 3 single lines with ending conditional (we're hip ruby devs here ;-)

PR #520, opened 2015-02-17
Robot name

shishirtewari:
Ashton, I have incorporated your review comments. Can you please have a quick look?

### DIFF
--- a/robot_name/namegenerator.rb
+++ b/robot_name/namegenerator.rb
@@ -1,0 +1,46 @@
+class NameCollisionError < RuntimeError; end
+class InvalidFormatError < RuntimeError; end
+
+# Generates name for Robot
+class NameGenerator
+  attr_accessor :name
+  @@registry = []
+
+  def self.name(args)
+    @name = args[:k]
+    if @name
+      fail InvalidFormatError, 'Not supported format' unless pattern?
+    else
+      generate_name
+    end
+    repeated?
+  end
+
+  def self.random_num
+    rand(10)
+  end
+
+  def self.random_char
+    ('A'..'Z').to_a.sample
+  end
+
+  def self.generate_name
+    str_arr = []
+    (1..5).each do |i|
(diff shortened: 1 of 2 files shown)


ashtonthomas:
Just an FYI, you ca use a self block instead of appending self. to all the class methods. I actually prefer appending self to all the class methods, So I like what you have here. Just and FYI FWIW FTW

robot_name/namegenerator.rb
```diff
@@ -0,0 +1,46 @@
+class NameCollisionError < RuntimeError; end
+class InvalidFormatError < RuntimeError; end
+
+# Generates name for Robot
+class NameGenerator
+  attr_accessor :name
+  @@registry = []
+
+  def self.name(args)
```
ashtonthomas: We already talked in person about why or why not to have a hash a parameter. All up to you at some point, so this is fine

robot_name/robot_name.rb
```diff
@@ -1,32 +1,25 @@
-class NameCollisionError < RuntimeError; end
+load 'namegenerator.rb'
 
+# Prints robot's name
 class Robot
-  attr_accessor :name
+  attr_accessor :robotname
```
ashtonthomas: Super minor, I would probably just left this as name so your code would be: `robot.name` instead of `robot.robotname` (in which case robot_name would be better than robotname

robot_name/namegenerator.rb
```diff
@@ -0,0 +1,46 @@
+class NameCollisionError < RuntimeError; end
+class InvalidFormatError < RuntimeError; end
+
+# Generates name for Robot
+class NameGenerator
+  attr_accessor :name
+  @@registry = []
```
ashtonthomas: I like how all of this is in a NameGenerator class and robot just cares about things directly related to a robot

ashtonthomas:
LGTM

PR #521, opened 2015-02-17
Refactor using Active Support

arrubin:


supportive/supportive.rb
```diff
@@ -9,26 +9,19 @@ class BlagPost
...
-      @categories = []
+    @categories = args.fetch(:categories, []).reject do |category|
+      category.in? DISALLOWED_CATEGORIES
     end
 
     @comments = args[:comments] || []
-    @body = args[:body].gsub(/\s{2,}|\n/, ' ').gsub(/^\s+/, '')
-    @publish_date = (args[:publish_date] && Date.parse(args[:publish_date])) || Date.today
+    @body = args[:body].squish
+    @publish_date = args[:publish_date].to_date || Date.today
```
incazteca: If there is no key publish_date this will fail since `nil` does not have a `to_date` method. Have it use try so that it doesn't fail.

supportive/supportive.rb
```diff
@@ -38,63 +31,34 @@ def to_s
...
+    publish_date + 3.years > Date.today
   end
 
   def abstract
-    if body.length < 200
-      body
-    else
-      body[0..200] + "..."
-    end
+    body.truncate(204)
```
incazteca: Is there a reason you are using 204? Just curious.

supportive/supportive.rb
```diff
@@ -38,63 +31,34 @@ def to_s
...
-    if categories.length > 1
-      last_category = categories.pop
-      suffix = " and #{as_title(last_category)}"
-    else
-      suffix = ""
-    end
-
-    label + ": " + categories.map { |cat| as_title(cat) }.join(", ") + suffix
+    "Category".pluralize(categories.length) +
+      ": " + categories.map { |cat| as_title(cat) }.to_sentence
```
incazteca: This can be done without mapping. Just have the sentence created and then pass that to `as_title`

supportive/supportive.rb
arrubin: I misread the documentation. to_date returns nil for blank receivers, but not nil.

supportive/supportive.rb
arrubin: The original code actually returns 204 characters (201 characters from the body and the ellipsis). Changing this to 200 makes sense, but would not do exactly what the original code did.

incazteca:
@arrubin This can be closed and no resubmission is necessary.

PR #523, opened 2015-02-17
Dino catalog

shishirtewari:
Sophia, I have incorporated your review comments.
I have divided existing dino.rb into three files:
1- dino_catalog/dinodex.rb
2- dino_catalog/formatter.rb
3- dino_catalog/run_dinocatalog.rb 
Please review.

### DIFF
--- a/dino_catalog/dino.rb
+++ b/dino_catalog/dino.rb
@@ -127,6 +127,7 @@ if ARGV.length == 0
   puts ''
   puts '#multiple conditions AND search'
   puts '  walking.biped and diet.Carnivore'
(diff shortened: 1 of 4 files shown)


incazteca:
@shishirtewari If you are not using dino.rb please remove it. I assumed it was a model for a single dinosaur. It seems to be the older version of your dino catalog.

incazteca:
@shishirtewari I would like to see a model created for dinosaurs. A lot of the code that is used for searching will be greatly simplified just by having a dinosaur object. I would like to see a resubmission of this exercises with a dinosaur class created. Also would like to schedule a time for us to pair on this.

PR #525, opened 2015-02-20
robot_name exercise for review.

gguntupalli:
Thanks
Gopi.

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -1,32 +1,52 @@
+class InvalidNameError < RuntimeError; end
 class NameCollisionError < RuntimeError; end
 
 class Robot
   attr_accessor :name
-
-  @@registry
+  INVALID_NAME_ERROR_MSG = 'Invalid name for the robot!'
+  COLLISION_ERROR_MSG = 'There was a collision generating the robot name!'
 
   def initialize(args = {})
     @@registry ||= []
-    @name_generator = args[:name_generator]
-
-    if @name_generator
-      @name = @name_generator.call
+    if args[:name_generator]
+      @name = args[:name_generator].call
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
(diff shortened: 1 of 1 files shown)


robot_name/robot_name.rb
```diff
@@ -1,32 +1,52 @@
+class InvalidNameError < RuntimeError; end
 class NameCollisionError < RuntimeError; end
 
 class Robot
   attr_accessor :name
-
-  @@registry
+  INVALID_NAME_ERROR_MSG = 'Invalid name for the robot!'
+  COLLISION_ERROR_MSG = 'There was a collision generating the robot name!'
```
ashtonthomas: Such a nitpick and probably personal preference, but I'd figure I'd mention: I would tab out the bottom constant so the `=`'s line up. I think it makes it just a lil prettier but I may be wrong for doing this... so FWIW lolz

robot_name/robot_name.rb
```diff
@@ -1,32 +1,52 @@
...
 
   def initialize(args = {})
```
ashtonthomas: So the starting code had this `args = {}` in it, but I would at least bring up the fact that we only have one parameter right now. So we could improve readability by just having the one parameter be the `name_generator`. However, having this `args` param provides flexibility and is a topic that we can expand on when we can safely use the new hotness of keyword arguments to their fullest.

I think the code you have for this entire exercise is really great and satisfies ALL requirements for this exercise. I think going any further may be over engineering. However, if we were to go further I would suggest looking to extract name generation into it's own class. This would allow you to remove the responsibility of `@@registry` and stuff to a new context leaving `robot` to only focus on robot things (which would be nothing really in this case). Then you could have your argument be an optional name_generator with a default value of an instance of the new class.

That is totally FWIW. Your coded is solid.

ashtonthomas:
Totally satisfies all requirements. :+1: 

Some comments to think about but no need to do a resubmission.

- A

gguntupalli:
Thanks Ashton.

PR #526, opened 2015-02-20
dinodex changes

shishirtewari:
Franco, I have incorporated your review comments. Please review.

### DIFF
--- a/dino_catalog/dinodex.rb
+++ b/dino_catalog/dinodex.rb
@@ -1,0 +1,64 @@
+require 'json'
+require 'table_print'
+autoload? 'formatter.rb'
+
+# Class Dinodex will have dino collection and search api
+class Dinodex
+  attr_accessor :dinos
+
+  def initialize(*filepaths)
+    @dinos = []
+    filepaths.each do |fp|
+      @dinos += Formatter.format(fp)
+    end
+  end
+
+  def search_by_walking(walking)
+    @dinos = @dinos.select { |dino| dino.walking == walking }
+    self
+  end
+
+  def search_by_name(name)
+    @dinos = @dinos.select { |dino| dino.name == name }
+    self
+  end
+
+  def search_by_source(source)
+    @dinos = @dinos.select { |dino| dino.source == source }
+    self
+  end
+
+  def search_by_diet(diet)
+    @dinos = @dinos.select { |dino| dino.diet == diet }
+    self
+  end
+
+  def search_carnivores
+    @dinos = @dinos.select(&:carnivore?)
+    self
+  end
+
+  def search_by_period(period)
+    @dinos = @dinos.select { |dino| dino.period_within?(period) }
+    self
+  end
+
+  def search_big
+    @dinos = @dinos.select(&:big?)
+    self
+  end
+
+  def search_small
+    @dinos = @dinos.select(&:small?)
+    self
+  end
+
+  def print
+    tp dinos, :name, :period, :diet, :weight,
+       :walking, :continent, :description, :source
+  end
+
+  def print_json
+    puts @dinos.to_json
+  end
+end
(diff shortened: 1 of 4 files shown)


ashtonthomas:
@shishirtewari I'm gonna steal this review from Franco, but I have reviewed his comments of your previous submission

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,64 @@
+require 'json'
+require 'table_print'
+autoload? 'formatter.rb'
+
+# Class Dinodex will have dino collection and search api
```
ashtonthomas: comment not very helpful. I normally try to talk myself out of comments at all cost. there may be some times when you need them, but I don't think this is one of them

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,64 @@
...
+class Dinodex
+  attr_accessor :dinos
+
+  def initialize(*filepaths)
+    @dinos = []
+    filepaths.each do |fp|
+      @dinos += Formatter.format(fp)
```
ashtonthomas: `Formatter.format(fp)` returns a list of dinos.. It's not totally obvious from this line so maybe we could rename something to make it more intention revealing TM..  

Perhaps: `DinoFormatter.parse(fp)` or `Formatter.parse_dinos(fp)` I dunno really know...

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,47 @@
+require 'json'
+
+# dinasour object
+class Dinosaur
+  SMALL = 2000
+  attr_accessor :name, :period, :diet, :weight,
+                :walking, :description, :continent, :source
+
+  def initialize(row)
```
ashtonthomas: I don't really care that this arg is a `row` of something. I just care that is is a hash of attributes. I am just confident (re: Avdi's Confident Ruby) that this is a hash with the appropriate attributes by the time this initializer gets called (that the job of something other than the Dino i.e. Formatter).

I would rename this to something like `attrs` or `attributes`. This is such a small point but it does remove any outside knowledge/suggestions that we shouldn't care about once inside of Dinosaur

dino_catalog/formatter.rb
```diff
@@ -0,0 +1,53 @@
...
+    convert_data(rows, csvfile)
+    rows.map { |row| Dinosaur.new(row) }
+  end
+
+  def self.convert_data(rows, csvfile)
+    convert_rows(rows, csvfile)
+    convert_weight(rows)
+  end
+
+  def self.convert_rows(rows, csvfile)
```
ashtonthomas: Did we take into account not differentiating between early and late Cretaceous? This is just another cruel trick by @jmmastey ;-)

dino_catalog/formatter.rb
```diff
@@ -0,0 +1,53 @@
+require 'csv'
+load 'dinosaur.rb'
+
+# Class Formatter will format input csv files
+class Formatter
```
ashtonthomas: So Formatter has some really ugly edge cases. But that is totally cool because we have abstracted it out from our core app code. So our important files can just be confident.

dino_catalog/formatter.rb
ashtonthomas: NVM! I see where you did that `dinosaur.period_within?` I guess that's cool cause you don't strip information. I will buy it!

dino_catalog/formatter.rb
```diff
@@ -0,0 +1,53 @@
+require 'csv'
+load 'dinosaur.rb'
+
+# Class Formatter will format input csv files
```
ashtonthomas: comment not adding anything

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,47 @@
+require 'json'
+
+# dinasour object
```
ashtonthomas: Come one, man. Really? Lolz comment :-)

ashtonthomas:
@shishirtewari great job :+1: 

No need to resubmit. Please see my comments and ask me if you have any questions.

- A

PR #527, opened 2015-02-20
Dinodex

theundeadmonk:
Updated and refactored dinodex


theundeadmonk:
I'm not sure why the arrowhead exercise got submitted. I had commits on my dinodex files.

ashtonthomas:
It's okay about the git issue. This happens. Check your master branch to make sure you have no changes in it. That way you can always cut new feature branches off of master and not have other commits. I would suspect that you added something to master at some point or you cut this branch while on arrowhead. Let me know if you need help cleaning it up (if master has commits on it)

- A

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,32 @@
+require 'json'
+
+class Dinosaur
+
+  attr_reader :name, :period, :diet, :weight, :movement, :continent,
+    :description
+
+  def initialize(dino)
+    @name = dino.fetch(:name)
```
ashtonthomas: I assume you are using `fetch` so it throws an exception if name is not there? I would take that to mean that all the other parameters are optional? We do a lot of processing outside of this class, and it would be nice to be able to assume the parameters are correct by the time it gets here. otherwise, we may want to make our validations more explicit and consistent among the other params.

I'm not going to push the issue too much on this exercise, but FWIW. Maybe we can just have confident code and rely on our dino parser (named load_dino.rb or DinoCollection?) to provide the correct attributes

ashtonthomas:
What is`.dino_methods.rb.swp`?

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,32 @@
...
+      period: period_to_string.to_s,
+      diet: @diet,
+      weight: @weight,
+      movement: @movement,
+      continent: @continent,
+      description: @description }.to_json
+  end
+
+  def period_to_string
+    @period.join(" or ")
```
ashtonthomas: So, it's not really clear what is going on here and why.. I'm sure I could follow the logic in the parser and how you use it. I feel like this is too coupled with other parts of the program maybe?

dino_catalog/load_dino.rb
```diff
@@ -0,0 +1,60 @@
+require 'csv'
+require_relative 'Dinosaur'
+
+class DinoCollection
```
ashtonthomas: the name of this file and the name of the class should be consistent

dino_catalog/load_dino.rb
```diff
@@ -0,0 +1,60 @@
...
+  
+  attr_reader :dino_collection
+
+  def initialize
+  @dino_collection ||= {}
```
ashtonthomas: indention?

ashtonthomas:
@theundeadmonk, I've added some comments, but I'm gonna stop going through it right now because I see two files which should probably be three:

dinodict.rb and load_dino.rb seem to be doing the work of: 
1) Setting up a DinoDex, 
2) Parsing Data (and handling exceptions, edge cases, etc) 
3) Running the CLI
* maybe a little coupling with the dino file here and there too

These should all be separate pieces that work together, but should also be reusable in other contexts.

- Can we make the dinosaur class be usable in other contexts with more generic properties (I'm actually not sure if things will change too much from what you currently have - but maybe)
- The main dinodex/dinodict class should off load all the parsing and formatting to a special file and then never worry about the exceptions and edge cases again
- Then you should have a separate class that acts as the CLI. This file should only access the API of the main dinodex/dinodict class. You should be able to submit your code without this CLI class and allow any other user/program to use the other code. The CLI class should just be a convenient way for us to run the program from the command line.

Try to break these things apart, and I think you will see a lot of other things fall into place

- A

If any part of this doesn't make sense or you have trouble implementing parts of this, please let me know.

PR #534, opened 2015-03-02
Yadda 2 and 3

arrubin:


yadda/data_seeder.rb
```diff
@@ -0,0 +1,154 @@
...
+    print "Adding #{@total_users} users"
+    @total_users.times do |n|
+      add_user
+      print "." if ten_percent?(@total_users, n + 1)
+    end
+    puts
+  end
+
+  def add_user
+    first_name = DataGenerator.first_name
```
incazteca: Just use the Faker Gem. No more having to create giant lists of names and other stuff. https://github.com/stympy/faker

yadda/data_seeder.rb
```diff
@@ -0,0 +1,154 @@
...
+    last_name = DataGenerator.last_name
+    created_on = DateTime.now
+    @user_ids << conn[:users].insert(
+      created_on: created_on,
+      updated_on: created_on,
+      name: "#{first_name} #{last_name}",
```
incazteca: Use parameters for your insertion queries.

yadda/data_seeder.rb
```diff
@@ -0,0 +1,154 @@
...
+    print "Adding #{@total_beers} beers"
+    @total_beers.times do |n|
+      add_beer
+      print "." if ten_percent?(@total_beers, n + 1)
+    end
+    puts
+  end
+
+  def add_beer
+    brewery = @breweries.sample
```
incazteca: Kind of seems unnecessary to create an array of breweries. Could just have the brewery_id set to a random integer between 1 and the count of breweries.

yadda/data_seeder.rb
```diff
@@ -0,0 +1,154 @@
...
+    style = @beer_styles.sample
+    name = "#{brewery[:name]} #{style[:name]}"
```
incazteca: Use parameters

yadda/data_seeder.rb
```diff
@@ -0,0 +1,154 @@
...
+class DataSeeder
+  def initialize(database, host = "localhost",
+    total_users = 100, total_ratings = 1000,
+    total_breweries = nil, total_beers = nil)
+    @connection_string = "postgres://#{host}/#{database}"
+    @total_users = total_users.to_i
+    @total_ratings = total_ratings.to_i
+    @total_beers = (total_beers || @total_ratings / 10).to_i
+    @total_breweries = (total_breweries || @total_beers / 10).to_i
+    @user_ids = []
```
incazteca: These arrays don't really seem necessary. Since we're just randomly assigning things we could just directly set the id to a random integer that is between 1 and the count of that table.

arrubin: That is usually true if starting from a newly created database, which I suppose will typically be the case with this script.

yadda/data_seeder.rb
arrubin: I use the brewery name in the beer name for added authenticity. I suppose that this is not completely necessary.

yadda/data_seeder.rb
```diff
@@ -0,0 +1,154 @@
+require "sequel"
+require_relative "data_generator"
+
+class DataSeeder
+  def initialize(database, host = "localhost",
```
incazteca: For simplicity just set the total_users and other totals to values that you are using for testing.

yadda/data_generator.rb
```diff
@@ -0,0 +1,29 @@
+class DataGenerator
```
incazteca: Using Faker this is mostly unnecessary.

yadda/views.sql
```diff
@@ -0,0 +1,67 @@
+DROP VIEW IF EXISTS beer_scores CASCADE;
+
+CREATE VIEW beer_scores AS
+  SELECT
+    beer_id,
+    created_on,
+    ((appearance + aroma + taste + texture + overall) * 2) AS score
```
incazteca: What's the reasoning behind doubling the score?

yadda/data_seeder.rb
```diff
@@ -0,0 +1,154 @@
...
+    puts
+  end
+
+  def add_beer
+    brewery = @breweries.sample
+    style = @beer_styles.sample
+    name = "#{brewery[:name]} #{style[:name]}"
+    created_on = DateTime.now
+    created_by = @user_ids.sample
+    @beer_ids << conn[:beers].insert(
```
incazteca: Populate the year brewed and throw some random text in for a description.

yadda/views.sql
arrubin: Doubling them makes the total score a 100 point scale. I considered using some kind of weighted score, but did not figure that the score itself was terribly important in this exercise.

yadda/data_seeder.rb
arrubin: Making this a one-off script? Am I understanding you correctly?

yadda/views.sql
```diff
@@ -0,0 +1,67 @@
...
+    ROUND(AVG(bs.score)) AS score
+  FROM
+    beers AS b,
+    beer_scores AS bs
+  WHERE
+    b.id = bs.beer_id
+  GROUP BY
+    b.id;
+
+DROP VIEW IF EXISTS top_beers CASCADE;
```
incazteca: Have your view also show the value for `beer_style_id` and `updated_by`. A view should be something that a user uses so that they could easily look up information.

incazteca:
I would like to see fixes for the seeding script. Part 3 is mostly fine in identifying where indexes need to go and for providing explanations.

yadda/views.sql
arrubin: b.* includes all columns from beers. Are you suggesting that I limit which columns are included?

yadda/data_seeder.rb
incazteca: This is really mostly a nitpick.

PR #536, opened 2015-03-02
Datascience

shishirtewari:
Please review data science changes.


data_science/spec/fileparser_spec.rb
```diff
@@ -0,0 +1,17 @@
+load 'fileparser.rb'
+require 'spec_helper'
+require 'rspec'
+
+describe FileParser do
+  context 'raise an error when' do
```
zsyed1: Add a _happy_ path test: `expect { FileParser.parse_data('valid.json') }.not_to raise_error(RuntimeError)`

data_science/datascience.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def conversion_miss_count(name)
+    @cohorts.count { |cohort| cohort.name == name && cohort.result == 0 }
+  end
+
+  def chisqaure_p
+    @analyzer_data = {}
+    @cohorts.uniq(&:name).each do |row|
+      @analyzer_data[row.name] ||=
```
zsyed1: Since you are calling `uniq` you shouldn't need the `||=` correct?

data_science/datascience.rb
```diff
@@ -0,0 +1,65 @@
+require 'ABAnalyzer'
+load 'fileparser.rb'
+
+class DataScience
+  attr_accessor :cohorts, :ab_analyzer, :analyzer_data
+
```
zsyed1: In line 5, `ab_analyzer` can be removed. I don't think it's used in the code.

zsyed1: In line 5, `analyzer_data` is not being used anywhere so you can remove it. Also see note below about the same variable name.

data_science/datascience.rb
```diff
@@ -0,0 +1,65 @@
...
+    @cohorts.count { |cohort| cohort.name == name && cohort.result == 1 }
+  end
+
+  def conversion_miss_count(name)
+    @cohorts.count { |cohort| cohort.name == name && cohort.result == 0 }
+  end
+
+  def chisqaure_p
+    @analyzer_data = {}
+    @cohorts.uniq(&:name).each do |row|
```
zsyed1: The variable `analyzer_data` is local to this method `chisquare_p`. So you don't need the `@` symbol in its name.

data_science/datascience.rb
```diff
@@ -0,0 +1,65 @@
...
+  def conversion_count(name)
+    @cohorts.count { |cohort| cohort.name == name && cohort.result == 1 }
+  end
+
+  def conversion_miss_count(name)
+    @cohorts.count { |cohort| cohort.name == name && cohort.result == 0 }
+  end
+
+  def chisqaure_p
+    @analyzer_data = {}
```
zsyed1: Spelling: `chisquare_p`

data_science/datascience.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+    leader.max_by { |_k, v| v }[:name]
+  end
+
+  def distribution_limit(name)
+    r = conversion_rate(name)
+    eb = error_bar(name)
+    [(r - eb).round(4), (r + eb).round(4)]
+  end
+
```
zsyed1: Nice `distribution_limit(name)`

data_science/datascience.rb
```diff
@@ -0,0 +1,65 @@
...
+    @cohorts = FileParser.parse_data(filename)
+  end
+
+  def cohort_count(name)
+    @cohorts.count { |cohort| cohort.name == name }
+  end
+
+  def conversion_count(name)
+    @cohorts.count { |cohort| cohort.name == name && cohort.result == 1 }
+  end
```
zsyed1: I link the way you've implemented this (line 16 and others like it). Just wanted to point out that another way to do this is to move the logic `cohort.result == 1` into the `Cohort` class and call it a method:

```
def conversion?
  @result == 1
end
```
Then here line 16 could change to: `@cohorts.count { |cohort| cohort.name == name && cohort.conversion?}`. The same goes with `conversion_miss_count` line 20.

data_science/cohort.rb
```diff
@@ -0,0 +1,9 @@
+class Cohort
+  attr_accessor :name, :result, :date
+
+  def initialize(row)
+    @date = row['date']
+    @name = row['cohort']
+    @result = row['result']
+  end
+end
```
zsyed1: Just FYI, your `Cohort` class really represents a `Visit` more than a `Cohort`. Here there are two cohorts: A and B. Each of them visits the website many times, once for each row.

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,24 @@
+load 'cohort.rb'
+require 'spec_helper'
+require 'rspec'
+
+describe Cohort do
+  cohort = {}
+  cohort['date'] = '2014-03-20'
+  cohort['cohort'] = 'B'
+  cohort['result'] = 0
+
```
zsyed1: Probably want to put lines 6 through 9 into the `let` on line 11. I would also change the name from `cohortobj` to simply `cohort`.

data_science/spec/datascience_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    it 'Should calculate distribution range for each cohort' do
+      expect(ds.distribution_limit('A')).to match_array([0.005, 0.2056])
+      expect(ds.distribution_limit('B')).to match_array([0.1896, 0.3022])
+    end
+    it 'Should calculate chisqaure in sample size' do
+      expect(ds.chisqaure_p).to eq(0.9058)
+    end
+    it 'Should check who is leader' do
+      expect(ds.leader).to eq('B')
+    end
```
zsyed1: Nice test. Btw there is much debate about whether or not to use the word `Should` in tests.

data_science/datascience.rb
zsyed1: In line 5, `cohorts` is never used outside this class. So it doesn't have to be in `attr_accessor`.

data_science/spec/spec_helper.rb
```diff
@@ -0,0 +1,3 @@
+require_relative '../fileparser.rb'
+require_relative '../cohort.rb'
+require_relative '../datascience.rb'
```
zsyed1: It was a pleasure reading your code. Nice work.

data_science/spec/datascience_spec.rb
djkotowski: I prefer not having `should` in descriptions.

djkotowski: Also, given these lines describe methods, why not `describe` the methods?

e.g. `describe "#conversion_rate"`

djkotowski:
This is looking good. Take a look at the notes and reactor. You should only need one resubmission on that.

PR #543, opened 2015-03-04
message for joe

katieschilling:


joe.txt
```diff
@@ -0,0 +1 @@
+WE MISS YOU, JOE! <3 Katie & Zach
```
ashtonthomas: You're missing a few signatures here. Plz fix

djkotowski:
Insufficient signatures.

PR #545, opened 2015-03-06
finished triangle facts

wyaeiga:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,63 @@
...
-		return [angleA, angleB, angleC]
-	end
-
-	def radians_to_degrees(rads)
-		return (rads * 180 / Math::PI).round
-	end
+  attr_accessor :side1, :side2, :side3
+
+  def initialize(side1, side2, side3)
+    @side1, @side2, @side3 = side1, side2, side3
```
ashtonthomas: multi line assignment is generally frowned upon (says the style guide)

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,63 @@
...
+  def equilateral
+    side1 == side2 && side2 == side3
+  end
+
+  def isosceles
+    [side1, side2, side3].uniq.length == 2
+  end
+
+  def scalene
+    if equilateral || isosceles
```
ashtonthomas: I think you can do this without the if/else and all on one short line ;-)

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,63 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def recite_facts
+    case
```
ashtonthomas: Hrm case statements stink..

ashtonthomas:
@wyaeiga, this is pretty close. Take a look at my comments and let me know if  you have any questions. I don't think there is enough left to warrant a new PR, but go ahead and make some changes on your end.

PR #547, opened 2015-03-06
Robot name

geekaholic:
Any feedback?

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -1,25 +1,40 @@
-class NameCollisionError < RuntimeError; end
+require 'pry'
+class NameCollisionError < RuntimeError
+  def to_s
+    "There was a problem generating the robot name!"
+  end
+end
 
 class Robot
-  attr_accessor :name
+  attr_reader :name
 
   @@registry
 
   def initialize(args = {})
     @@registry ||= []
-    @name_generator = args[:name_generator]
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
-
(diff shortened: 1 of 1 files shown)


robot_name/robot_name.rb
```diff
@@ -1,25 +1,39 @@
...
+  end
+end
 
 class Robot
-  attr_accessor :name
+  attr_reader :name
 
   @@registry
 
   def initialize(args = {})
```
ashtonthomas: We really only need one parameter. The hash param may be overkill when all we need is a `name_generator`

robot_name/robot_name.rb
```diff
@@ -1,25 +1,39 @@
...
     @@registry ||= []
-    @name_generator = args[:name_generator]
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+    name_generator = args[:name_generator]
```
ashtonthomas: There is a lot of logic in this initialize method that may feel better in a a separate class. It's almost like you have another _thing_ that could be its own object. 

Perhaps your initialize could look like: `def initialize(name_generator = RobotNameGenerator.new)`

This new class could then encapsulate `@@registry` and the collision logic, etc. This might provide a good way to reduce duplicate while promoting extensibility

ashtonthomas:
:+1: 

But I did leave one comment about possibly reducing duplication while promoting extensibility

geekaholic:
Thanks for feedback. I did some refactoring but thought of not sending
another PR.

https://github.com/geekaholic/level_up_exercises/commit/1131befc742c08f84af97719c6f59a7d98a20235

Bud

On Mon, Mar 9, 2015 at 10:04 AM, Ashton Thomas <notifications@github.com>
wrote:

> [image: :+1:]
>
> But I did leave one comment about possibly reducing duplication while
> promoting extensibility
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jmmastey/level_up_exercises/pull/547#issuecomment-77870277>
> .
>



-- 
Bud

Geek with an Attitude
Blog: http://geekaholic.org
twitter: @geekaholic

ashtonthomas:
Cool, just made one comment on the new PR. I hope that _feels_ a lot better :-)

PR #548, opened 2015-03-06
fixed arrowhead programming

wyaeiga:


arrowhead/arrowhead.rb
```diff
@@ -15,19 +15,18 @@ class Arrowhead
...
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
+    unless CLASSIFICATIONS.include? region
       raise "Unknown region, please provide a valid region."
     end
+
+    shapes = CLASSIFICATIONS[region]
+    unless shapes.include? shape
+      raise "Unknown shape value. Are you sure you know what you're talking about?"
```
ashtonthomas: If you moved that error message into a constant you could do something like:
`raise SOMEERRORR unless CLASSIFICATIONS.include? region`

ashtonthomas:
LGTM

:+1: 

@wyaeiga, think about improving the readability (or the prettiness of the code - if that's even a thing) by using constant and the ending conditional

PR #549, opened 2015-03-06
finished robot name

wyaeiga:


robot_name/robot_name.rb
```diff
@@ -9,17 +9,32 @@ def initialize(args = {})
     @@registry ||= []
     @name_generator = args[:name_generator]
```
ashtonthomas: I know the starting example had `args` as a hash, but we really only have one parameter, so we could totally just have one named parameter (a hash as a param is great when you have multiple params and you don't want order to matter and you need to make some optional, etc)

robot_name/robot_name.rb
```diff
@@ -9,17 +9,32 @@ def initialize(args = {})
...
 
+    generate_name
```
ashtonthomas: What happens if you remove generate_name or if you remove the lines below that raise an error? It seems like there is too much coupling between the order and presence of these lines..

When this happens, it makes it harder to easily digest the code at a glance. Our goal in these training exercises will be to promote easy to comprehend and, therefore, easy to _change_ code.

Now, to take this one step further, wouldn't it be nice if Robot only cared about things that a single Robot should care about? Maybe we can create a new Class that handles the name generation and worries about collisions as well as the `@@registry`. Your robot init could then look like: `def initialize(name_generator = RobotNameGenerator.new)` 

Think about how this approach will clean up that conditional in `generate_name`. And then think about how this will affect the reusabilityTM, readabilityTM, and drinkabilityTM (yea, robots be partiers)

ashtonthomas:
@wyaeiga think through these comments and see if you can improve your design. Please reach out to me if you have any questions. 

A good goal for the new hires may be to track down your code reviewers and meet them ;-)

You might even meet others in the process haha

PR #552, opened 2015-03-09
Dinodex

wyaeiga:


dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,99 @@
...
+        walking: dino[:walking],
+        description: dino[:description]
+      }
+
+      add Dino.new(restructured_dino)
+    end
+  end
+
+  def self.print(dino_catalog)
+    dino_catalog.each { |dino| dino.print }
```
ashtonthomas: clever hound is right here

ashtonthomas:
@wyaeiga, how do you feel about this code? I think it is a great improvement. I would imagine it feels pretty good. 

I think the next step would be to look at your naming (naming is an art - and takes time to master)

Can you describe each class in one sentence:
- DinodexHandler
- Dinodex
- DinoCatalog

I'm thinking that maybe we can rename these classes to make them more intention revealing. 

- Dinodex -> DinodexRunner
- DinoCatalog -> (well, we still have parsing/conversion along with normal dinodex/catalog behavior...)
- DinodexHandler -> not really sure how much value we are getting here

Also, DinodexHandler has nothing but class methods but utilizes instance variables? Why not just make all the methods instance methods?

I'm going to make some more comments inline and then maybe we can go over it this week

dino_catalog/dino.rb
```diff
@@ -0,0 +1,24 @@
...
+    @period = args[:period]
+    @continent = args[:continent]
+    @diet = args[:diet]
+    @weight = args[:weight]
+    @walking = args[:walking]
+    @description = args[:description]
+  end
+
+  def print
+    puts "Name: #{@name}" unless name.nil?
```
ashtonthomas: This is very unconfident code (see Avdi's Confident Ruby). You can always validate that no attributes are null and then be confident (unless, of course, you want to support optional values)

ashtonthomas: But if we are using a 'Parser' we might be able to just make some assumptions that this part of the code will do its job (?)

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,100 @@
...
+  include Enumerable
+  extend Forwardable
+
+  def_delegators :@dino_catalog, :each, :<<
+  attr_reader :dino_catalog
+
+  def initialize(catalog = [])
+    @dino_catalog = catalog
+
+    if @dino_catalog.empty?
```
ashtonthomas: I think it would be cleaner and more open to extension if we kept a list of file_names and then had a method to return a clean list of dinos all at once. This does, however, satisfy our requirement, so there is no need to make a change if you feel your solution is satisfactory.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,100 @@
...
+require './dino'
+
+class DinoCatalog
+  include Enumerable
+  extend Forwardable
+
+  def_delegators :@dino_catalog, :each, :<<
+  attr_reader :dino_catalog
+
+  def initialize(catalog = [])
```
ashtonthomas: So you have a 'catalog' class that accepts a catalog as a parameter? If this were true why would we need to wrap it? Perhaps it is better to instead accept a list of filenames (variable length). the default names could be the two files we already know about. If you are concerned with what will happen if we try to pass some other file other than what we know we can handle: good point. Perhaps we can handle it when it happens or maybe this solution is fine. Up to you

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  def &(rvalue)
+    DinoCatalog.new(@dino_catalog & rvalue.dino_catalog)
+  end
+
+  def empty?
+    @dino_catalog.empty?
+  end
+
+  def get_bipeds
```
ashtonthomas: no need to prepend `get_` us ruby hipsters prefer to write methods in noun form is possible so just have `def bipeds` eventually your code will read like normal er'day speak

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  def get_small
+    @dino_catalog.select { |dino| dino.weight.to_i < 4001 }
+  end
+
+  def get_big
+    @dino_catalog.select { |dino| dino.weight.to_i > 4000 }
+  end
+
+  def get_by_name(name)
```
ashtonthomas: `dino_by_name` perhaps

ashtonthomas:
Take a look at all these comments, lets talk, and we can decide where to go from here

:+1:

PR #557, opened 2015-03-12
Initial Dino Catalog implementation

jmtai:


dino_catalog/dino_test.rb
```diff
@@ -0,0 +1,62 @@
...
+  puts "   Print dino data for the results of a find operation"
+  puts ""
+  puts "=> export"
+  puts "   Display dino data in JSON format"
+  puts ""
+  puts "=> exit"
+  puts "   Exit the program"
+  puts ""
+end
+
```
zsyed1: Do this like:
```
['Usage:',
'=> find category ....',
...
].join("\n")
```

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,138 @@
...
+  def match_one?(key, val, dino)
+    case key
+      when "DIET"
+        match_diet?(val, dino)
+      when "WEIGHT"
+        match_weight?(val, dino)
+      else
+        dino[key] ? dino[key].downcase.index(val) : false
+    end
+  end
```
zsyed1:
```
return match_diet?(val, dino) if key == "DIET"
return match_weight?(val, dino) if key == "WEIGHT"
return dino[key].downcase.index(val) if dino[key]
false
```

dino_catalog/dino_csv_parser.rb
```diff
@@ -0,0 +1,51 @@
...
+    @dinos = []
+    @categories = nil
+
+    File.open(filename) do |file|
+      while (line = file.gets)
+        parse_line(line)
+      end
+    end
+  end
+
```
zsyed1: Other than `parse_file`, the methods below are not used except within this class - so the should probably be `private`.

dino_catalog/dino_csv_parser.rb
```diff
@@ -0,0 +1,51 @@
...
+    @categories = line.strip.upcase.split(',')
+    normalize_categories!
+  end
+
+  def normalize_categories!
+    @categories[0] = "NAME" if @categories[0] == "GENUS"
+
+    idx = @categories.index("WEIGHT_IN_LBS")
+    @categories[idx] = "WEIGHT" if idx
+  end
```
zsyed1: You can make lines 31 and 32 more ruby-esque using a single line `map`: `@categories.map { |c| c == 'WEIGHT_IN_LBS' ? 'WEIGHT' : c }`

dino_catalog/dino_csv_parser.rb
```diff
@@ -0,0 +1,51 @@
+class DinoCsvParser
+  attr_reader :dinos
+
+  def parse_file(filename)
+    @dinos = []
+    @categories = nil
```
zsyed1: You can just have `parse_file` return an array of dinosaurs. No need to store it as an instance member. So you can remove line 2 and then make the variable @dinos a local variable. Also, you should probably initialize `@categories` to `[]` instead of `nil`.

dino_catalog/dino_csv_parser.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+  end
+
+  def parse_line(line)
+    if @categories.nil?
+      parse_categories(line)
+    else
+      @dinos << parse_data(line)
+    end
+  end
```
zsyed1: I suppose you could do
```
return parse_categories(line) unless @categories
@dinos << parse_data(line)
```
The return value is not used.

dino_catalog/dino_csv_parser.rb
```diff
@@ -0,0 +1,51 @@
...
+    Hash[paired_params]
+  end
+
+  def category_data_pair(param, category)
+    if category == "CARNIVORE"
+      param == "Yes" ? %w(DIET Carnivore) : %w(DIET Non-Carnivore)
+    else
+      [category, param]
+    end
+  end
```
zsyed1: You can reduce lines 45-49 to two lines.
```
return [category, param] unless category == "CARNIVORE"
param == "YES" ...
```

dino_catalog/dino_test.rb
```diff
@@ -0,0 +1,62 @@
...
+      dc.print_results(params)
+    when "find"
+      dc.find(params)
+    when "export"
+      dc.to_json
+    when "exit"
+      exit
+    else
+      show_help
+  end
```
zsyed1: You can replace this `case` statement with a map from `command` to `method` - something like
```
actions = {}
actions["print"] = method(:print_results)
...
```
And then dispatch them like `actions[command].invoke(params)`. I may have some of the syntax/keywords wrong. Hit me up on Slack if this is unclear.

dino_catalog/dino_printer.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  def print_entry(category, dino, verbose)
+    col_width = @col_widths[category]
+
+    if verbose
+      entry = dino[category]
+    else
+      entry = truncate_entry(dino[category], col_width - 2)
+    end
+
```
zsyed1:
```
entry = dino[category]
entry = truncate_entry(entry, col_width - 2) if verbose
```

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,138 @@
...
+      return true if diet == dino_val
+    end
+
+    false
+  end
+
+  def carnivore?(val)
+    if val == "carnivore" || val == "insectivore" || val == "piscivore"
+      return true
+    end
```
zsyed1: `return true if val == "carnivore" || ...` squashes lines 115 to 117 down to one line,

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,138 @@
...
+    dino_val = dino["DIET"].downcase
+
+    if diet == "carnivore"
+      return true if carnivore?(dino_val)
+    else
+      return true if diet == dino_val
+    end
+
+    false
+  end
```
zsyed1: Lines 105 - 111 can be reduce to:
```
return (carnivore?(dino_val) || diet == dino_val) if diet == 'carnivore'
```

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,138 @@
...
+    matches = @results.map { |dino| dino["NAME"] }
+    puts matches.join(", ")
+  end
+
+  def match?(params, dino)
+    params.each do |param|
+      key = param[0].upcase
+      val = param[1].downcase
+      return false unless match_one?(key, val, dino)
+    end
```
zsyed1: You can send `(param, dino)` to `match_one?` and do the conversion (lines 83 - 84) there.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,138 @@
...
+      puts "Invalid parameters"
+      []
+    end
+  end
+
+  def map_to_key_val_pairs(params)
+    params.map do |param|
+      args = param.split("=")
+
+      raise ArgumentError if args[0].nil? || args[1].nil?
```
zsyed1: No need for these blank lines. Sometimes there good, here not so necessary (line 68 and 70). This is highly subjective though....

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,138 @@
...
+    print_find_result
+  end
+
+  def print_results(params)
+    if params.downcase == "all"
+      @printer.display(@dinos)
+    else
+      return printf "\nNo results to print\n" if @results.empty?
+      @printer.display(@results)
+    end
```
zsyed1: You can squash lines 30 - 32 into `return @printer.display(@dinos) if params.downcase == "all"`

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,138 @@
...
+  def find(criteria)
+    @results = []
+    params = parse_search_params(criteria)
+
+    return if params.length == 0
+
+    @dinos.each do |dino|
+      @results << dino if match?(params, dino)
+    end
+
```
zsyed1: Lines 22-24 should be done with `select`: `@results = @dinos.select { |dino| match?(params, dino) }`

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,138 @@
...
+    @printer = DinoPrinter.new
+
+    load_all_csv_files
+  end
+
+  def find(criteria)
+    @results = []
+    params = parse_search_params(criteria)
+
+    return if params.length == 0
```
zsyed1: Remove blank line 19.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,138 @@
...
+  def match_weight?(weight, dino)
+    dino_weight = dino["WEIGHT"].to_i
+
+    return false if dino_weight == 0
+
+    return true if big_or_small?(weight, dino_weight)
+    return true if weight.to_i == dino_weight
+
+    false
+  end
```
zsyed1: Lines 127 - 130 can be replaced with: `return big_or_small?(...) || weight_to_i == dino_weight`

dino_catalog/dino_printer.rb
```diff
@@ -0,0 +1,77 @@
...
+  def get_categories(list)
+    categories = []
+
+    list.each do |dino|
+      categories << dino.keys
+    end
+
+    categories.flatten.uniq
+  end
+
```
zsyed1: Lines 11 - 17 can be reduces to one line: `list.map { |dino| dino.keys }.flatten.uniq`

dino_catalog/dino_test.rb
zsyed1: Nice work implementing the catalog features. This will probably need another PR.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,138 @@
...
+    @results = []
+    params = parse_search_params(criteria)
+
+    return if params.length == 0
+
+    @dinos.each do |dino|
+      @results << dino if match?(params, dino)
+    end
+
+    print_find_result
```
djkotowski: Should your catalog really be responsible for console output?

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,138 @@
...
+      @printer.display(@dinos)
+    else
+      return printf "\nNo results to print\n" if @results.empty?
+      @printer.display(@results)
+    end
+  end
+
+  def to_json
+    json = JSON.pretty_generate(@dinos)
+    puts json
```
djkotowski: A method called `#to_json` sounds like it should return a JSON string, not return `nil` and output JSON to the console.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,138 @@
...
+  end
+
+  def match_one?(key, val, dino)
+    case key
+      when "DIET"
+        match_diet?(val, dino)
+      when "WEIGHT"
+        match_weight?(val, dino)
+      else
+        dino[key] ? dino[key].downcase.index(val) : false
```
djkotowski: A case statement _and_ a ternary operator makes this pretty confusing at first glance.

dino_catalog/dino_catalog.rb
djkotowski: Not quite.  This would reduce to

```ruby
return carnivore?(dino_val) if diet == "carnivore"
diet == dino_val
```

dino_catalog/dino_catalog.rb
djkotowski: Avoid returning `true` or `false`.  Just return the result of the condition.

djkotowski:
Much of the matching logic in your `DinoCatalog` class really should be handled by the individual dinosaurs.  Why have

```ruby
catalog.match_weight(weight, dino)
```

when you could have

```ruby
dinosaur.match_weight(weight)
# OR
dinosaur.weight == weight
```

dino_catalog/dino_printer.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+  end
+
+  def print_table_header(list)
+    print_categories(list)
+    print_separators(list)
+  end
+
+  def print_categories(list)
+    printf "\n"
```
djkotowski: Why not just use `puts`?

dino_catalog/dino_test.rb
```diff
@@ -0,0 +1,62 @@
...
+  puts ""
+  puts "=> print"
+  puts "   Print dino data for the results of a find operation"
+  puts ""
+  puts "=> export"
+  puts "   Display dino data in JSON format"
+  puts ""
+  puts "=> exit"
+  puts "   Exit the program"
+  puts ""
```
djkotowski: This really belongs in a [Heredoc](http://ruby-doc.org/core-2.2.0/doc/syntax/literals_rdoc.html#label-Here+Documents) or a text file that you can just read from and output.

dino_catalog/dino_test.rb
djkotowski: I recommend using Heredocs instead.  Why make an array just to make a string out of a series of literals?

dino_catalog/dino_test.rb
djkotowski: Regarding string-to-method mappings, you could have a constant actions hash mapping input to symbols, then use `#send` to call those symbols.

```ruby
ACTIONS = {
  print: :print_results,
  find: :find,
  # etc, etc
}

command = (args[0] || "").to_sym
params = args[1] || ""

action = ACTIONS.fetch(command, :show_help) # The fetch method accepts a default value if the key does not exist
dc.send(action, params)
```

This would need to be tuned up a bit, but that's the gist of it.

djkotowski:
This is a good start.  Take a look at the comments and work on refactoring it a bit.  Make sure your classes do not have too much responsibility.  Does the catalog really need to handle loading files *and* console output *and* filtering?

Feel free to talk to Zamir or myself with any questions.

PR #558, opened 2015-03-12
Triangle facts exercise

jmtai:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,85 @@
...
-	end
-
-	def radians_to_degrees(rads)
-		return (rads * 180 / Math::PI).round
-	end
-end
+  attr_accessor :side1, :side2, :side3
+
+  def initialize(side1, side2, side3)
+    @side1, @side2, @side3 = side1, side2, side3
```
ashtonthomas: multi line assignment is generally discouraged (with our referenced style guides)

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,85 @@
...
+  end
+
+  def equilateral
```
ashtonthomas: A good ruby idiom is to prepend `?` to methods that return a boolean so: `def equlateral?`

Same for all the other methods below - because ruby is cool and we are too! :D

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,85 @@
...
+
+    puts "This triangle is also a right triangle!" if angles.include? 90
+    puts ""
+  end
+
+  def recite_equilateral?
+    puts "This triangle is equilateral!" if equilateral
+  end
+
+  def recite_isosceles?
```
ashtonthomas: Well, I'm not actually sure why we have the `?` in these cases that print (?)

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,85 @@
...
+  def equilateral
+    side1 == side2 && side2 == side3
+  end
+
+  def isosceles
+    [side1, side2, side3].uniq.length == 2
+  end
+
+  def scalene
+    if equilateral || isosceles
```
ashtonthomas: you really don't need an if/else here. ruby will just evaluate and return `equilateral || isoseles`

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,85 @@
...
+    angle_c = radians_to_degrees(calculate_angle_c(a, b, c))
+
+    [angle_a, angle_b, angle_c]
+  end
+
+  def radians_to_degrees(rads)
+    (rads * 180 / Math::PI).round
+  end
+
+  def calculate_angle_a(a, b, c)
```
ashtonthomas: Pulling these math equations into helper methods is cool, but this may be a case where you could argue that you don't gain anything by adding a helper (the original may, in fact, be easier to maintain. Honestly, I think this argument is VERY week, so either way is puurrrfectly fine :cat:

ashtonthomas:
I made some comments, but I don't think they warrant another submission. Please see the comments and let me know if you have any questions about them.

- A

triangle_facts/triangle_facts.rb
jmtai: Yeah, I wasn't sure about that one.  The functions probably should have been renamed, since I was just trying to capture the idea that it will only print something if the condition is true.

triangle_facts/triangle_facts.rb
jmtai: Agreed, that was just a cosmetic thing to make the lines shorter and more readable.

triangle_facts/triangle_facts.rb
ashtonthomas: Perhaps you don't need a method for this. You could just call in `recite_facts`:

`puts "foo asd asdf asdf asdf " if equilateral?`
`puts "foo asd asdf asdf asdf " if scalene?`
`puts "foo asd asdf asdf asdf " if isosceles?`

triangle_facts/triangle_facts.rb
jmtai: Question - is there a style preference something like this?

!(equilateral || isosceles)
  vs
!equilateral && !isosceles

ashtonthomas: I'm not really sure on this. I think I would prefer the former, but I don't really have any reason for that :-/

PR #560, opened 2015-03-12
DinoDex Project

ckhadilkar:


dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+      puts "WALK: #{@walk}\n"
+    end
+    if !@descrp.nil?
+      puts "DESCRIPTION: #{@descrp}\n"
+    end
+  end
+end
+
+
+class DinosaurCollection
```
sshao: split  classes into separate files, ie `dinosaur.rb` and `dinosaur_collection.rb`

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
+require 'csv'
+
+class Dinosaur
+  attr_reader :name
+  attr_reader :period
+  attr_reader :continent
+  attr_reader :diet
+  attr_reader :weight
+  attr_reader :walk
+  attr_reader :descrp
```
sshao: compact lines 4-10: `attr_reader :name, :period, :continent, ... :descrp`. also, might as well use `description` over `descrp`, it's not that much longer, especially considering that you're not shortening `continent`.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+    end
+    if !@weight.nil?
+      puts "WEIGHT: #{@weight} lbs\n"
+    end
+    if !@walk.nil?
+      puts "WALK: #{@walk}\n"
+    end
+    if !@descrp.nil?
+      puts "DESCRIPTION: #{@descrp}\n"
+    end
```
sshao: these can all be shortened to `puts "NAME: #{@name}" if @name`; `if @name` will evaluate to false if `@name` is nil.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+    @continent = continent
+    @diet = diet
+    @weight = weight
+    @walk = walk
+    @descrp = descrp
+  end
+
+  def display
+    if !@name.nil?
+      puts "NAME: #{@name}\n"
```
sshao: why the extra newline? `puts` adds a newline by default.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+    @name = name
+    @period = period
+    @continent = continent
+    @diet = diet
+    @weight = weight
+    @walk = walk
+    @descrp = descrp
+  end
+
+  def display
```
sshao: `display` is the same thing as a `to_s` method -- it's the string representation of a Dinosaur object. might as well rename this `to_s` and refactor it to return a string instead of `puts`ing. 

then to print a dinosaur `dino` you can just call `puts dino`, and the `to_s` will kick in.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+  end
+end
+
+
+class DinosaurCollection
+  attr_accessor :dinosaurs
+  def initialize 
+    @dinosaurs = {}
+
+    dino = CSV.new(File.read("dinodex.csv"), headers: true, :header_converters => :symbol, :converters => :all)
```
sshao: you're tying your implementation down to the hardcoded file paths "dinodex.csv" and "african_dinosaur_export.csv", which means that if those files change, so must your code. refactor this method so that it can take an arbitrary file name (or arbitrary file names) instead

sshao: also, mixing ruby 1.8.7's hash syntax with ruby 1.9 syntax here (`key => value` vs `key: value`). pick one and stick with it :)

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+    dino_hash = dino.to_a.map {|row| row.to_hash}
```
sshao: `dino.to_a.map(&:to_hash)` should work here

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+    dino_hash.each do |dino|
```
sshao: yup, shadowing outer local variables is weird and tricky, and i wouldn't do it. rename the block argument to something else.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+      @dinosaurs[dino[:name]] = Dinosaur.new(dino[:name], dino[:period], 
+                                dino[:continent], dino[:diet], 
+                                dino[:weight_in_lbs], dino[:walking], dino[:description])
```
sshao: instead of this, maybe pass the entire `dino` hash to `Dinosaur.new` and let the `Dinosaur` class figure out how to extract attributes from it? seems cleaner that way.

also, then you could get clever and muck around with the keys of the hash and [Object#instance_variable_set](http://ruby-doc.org/core-2.2.1/Object.html#method-i-instance_variable_set) -- note that the hash keys and the names of your Dinosaur attributes are almost identical.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+      puts "DESCRIPTION: #{@descrp}\n"
+    end
+  end
+end
+
+
+class DinosaurCollection
+  attr_accessor :dinosaurs
+  def initialize 
+    @dinosaurs = {}
```
sshao: why a hash of dinosaurs, and not an array?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+
+    dino = CSV.new(File.read("dinodex.csv"), headers: true, :header_converters => :symbol, :converters => :all)
+    dino_hash = dino.to_a.map {|row| row.to_hash}
+    dino_hash.each do |dino|
+      @dinosaurs[dino[:name]] = Dinosaur.new(dino[:name], dino[:period], 
+                                dino[:continent], dino[:diet], 
+                                dino[:weight_in_lbs], dino[:walking], dino[:description])
+    end
+    
+    african_dino = CSV.new(File.read("african_dinosaur_export.csv"), headers: true, :header_converters => :symbol, :converters => :all)
```
sshao: this is a lot of work for one `initialize` method, especially considering that it's doing the same thing twice -- just for two different cases (dinodex vs african_dinodex).

you can extract out the majority of this method into its own class, that will be solely responsible for reading/parsing files. you can then use that class to figure out how to deal with the differences between dinodex.csv and african_dinosaur_export.csv in a cleaner way -- subclassing, maybe?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+    african_dino_hash = african_dino.to_a.map {|row| row.to_hash}
+
+    african_dino_hash.each do |h|
+      h[:continent] = nil
+      h[:description] = nil
+    end
```
sshao: look a little deeper into the CSV class -- it provides converters for both headers and bodies. so this block (and the block on lines 69-78) can be converted into procs and passed into `CSV.new` as converters.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+        dino[:carnivore] = "Carnivore"
+      else
+        dino[:carnivore] = nil  
+      end
+        @dinosaurs[dino[:genus]] = Dinosaur.new(dino[:genus], dino[:period], 
+                                  dino[:continent], dino[:carnivore], 
+                                  dino[:weight], dino[:walking],dino[:description])
+    end
+  end
+end
```
sshao: not sure if you need a `DinosaurCollection` class -- the concept of a "dinosaur object" seems like it could be folded in as an hash/array attribute of a Finder/Dinodex class

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+
+class Finder
+  def initialize
+    @dinoCol = DinosaurCollection.new
+    search
```
sshao: i'm not sure about calling the search method directly from initialize, personally more fond of an explicit `finder = Finder.new; finder.search`

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+  end
+
+  def getBipeds
```
sshao: what if a user wants to get Quadrupeds? might as well write a method for that too -- or a more general search method that takes in an attribute ("walking") and a criteria ("biped" or "quadruped" or ...)

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+    @dinoCol.dinosaurs = @dinoCol.dinosaurs.select { |k, v| v.walk == "Biped" }
```
sshao: hm... you're overwriting `@dinoCol.dinosaurs`? that's surprising -- no one expects a search to be destructive D:

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+
+  def getSpecificDinosaur(param)
+    @dinoCol.dinosaurs = @dinoCol.dinosaurs.select { |k, v| k == param }
+    if @dinoCol.dinosaurs.empty?
+      puts "No records found matching your criteria for #{param}"
+    end 
+  end
+ 
+  def getBigDinosaurs(param)
+    @dinoCol.dinosaurs = @dinoCol.dinosaurs.select { |k, v| v.weight.to_i > param.to_i * 2000 }
```
sshao: assign `2000` to a constant and use that constant instead. the constant should be named such that someone skimming through your code knows what it means -- right now, i see "2000" and i have no idea why that number was picked, or what it really stands for

sshao: also, instead of having to convert `weight` to integer here, just convert it to an integer when it's assigned to the `Dinosaur` attribute. then you can drop the `to_i` here.

dino_catalog/dinodex.rb
sshao: in fact, consider generalizing all your searches -- `getWalking` instead of `getBipeds`, `getDiet` instead of `getCarnivores`, etc.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+  end
+
+  def getSpecificDinosaur(param)
+    @dinoCol.dinosaurs = @dinoCol.dinosaurs.select { |k, v| k == param }
+    if @dinoCol.dinosaurs.empty?
+      puts "No records found matching your criteria for #{param}"
+    end 
+  end
+ 
+  def getBigDinosaurs(param)
```
sshao: style note: camelcasing in ruby is usually reserved for class/module names -- use snake case for methods and variables (`get_big_dinosaurs` instead of `getBigDinosaurs`)

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+  end
+  
+  def getPeriodSpecificDinosaurs(param)
+    @dinoCol.dinosaurs = @dinoCol.dinosaurs.select { |k, v| v.period =~ /#{param}/ }
+    if @dinoCol.dinosaurs.empty?
+      puts "No records found matching your criteria for #{param}"
+    end
+  end
+  
+  def getDinosaurCollection
```
sshao: is this doing anything? if not, get rid of dead code :bomb:

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+  end
+  
+  def displayUsage
+    puts "Usage:\nruby dinodex.rb <search_options>\n-b\t\t\t\treturn all biped dinosaurs\n-c\t\t\t\treturn all carnivore dinosaurs"
+    puts "-a\t\t\t\treturn dinosaur collection\n-n  <NAME>\t\t\treturn facts for this dinosaur\n"
+    puts "-wg <WEIGHT IN TONS>\t\treturn dinosaurs weighing more than <WEIGHT IN TONS>\n-wl <WEIGHT IN TONS>\t\treturn dinosaurs weighing less or equal to <WEIGHT IN TONS                >\n-p  <PERIOD>\t\t\treturn dinosaurs in this <PERIOD>"
+    puts "You can also combine the criteria\nExample 1: ruby dinodex.rb -bcp Late\twill return all biped dinosaurs that are carnivores in 'Late' period."
+    puts "Example 2: ruby dinodex.rb -bwg 2\twill return all biped dinosaurs who weigh more than 2 tons."
```
sshao: this is ok, but since you already have a README with the exact same content... `puts File.read("DINODEX_README")`?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+
+  end
+ 
+  def displayDinosaurData
+    @dinoCol.dinosaurs.each do |k, v|
+      v.display
+      puts "\n"
+    end
+  end
```
sshao: as mentioend with your Dinosaur class, this is essentially a `to_s` method on your Finder, and should be converted to that.

also, once you implement `Dinosaur#to_s`, you could probably do something as simple as `puts @dinoCol.dinosaurs` to implement `Finder#to_s`.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+      v.display
+      puts "\n"
+    end
+  end  
+  
+  def search
+    option = ARGV[0]
+    param = ARGV[1]
+    
+    if !option.nil?
```
sshao: you might find (OptionParser)[http://ruby-doc.org/stdlib-2.2.0/libdoc/optparse/rdoc/OptionParser.html] useful here.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,162 @@
...
+      getPeriodSpecificDinosaurs(param) if option =~ /p/
+      getDinosaurCollection if option =~ /a/
+      puts "\n"
+      displayDinosaurData
+    else
+      displayUsage
+    end   
+   end
+end
+obj = Finder.new
```
sshao: might be overkill for this exercise, but usually try to keep object class implementations separate from the actual program execution (so i'd move this last line into its own file)

dino_catalog/dinodex.rb
sshao: related to the above -- if you nix your DinosaurCollection class and just fold it into Finder as a hash/array attribute, then you don't have to do all the `@dinoCol.dinosaurs` stuff -- you can just use `@dinoCol` directly instead

sshao:
take a look at my comments & refactor, then resubmit for another review :+1:

dino_catalog/dinodex.rb
ckhadilkar: fixed it

dino_catalog/dinodex.rb
ckhadilkar: removed it

dino_catalog/dinodex.rb
ckhadilkar: done.

dino_catalog/dinodex.rb
ckhadilkar: done. looks much cleaner :)

dino_catalog/dinodex.rb
ckhadilkar: done.

PR #561, opened 2015-03-12
Robot name exercise

jmtai:


robot_name/robot_name.rb
```diff
@@ -1,25 +1,62 @@
...
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+    @name = generate_unique_name(generator, MAX_RETRIES)
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+    @@registry << @name
+  end
+
+  def generate_unique_name(generator, retries)
+    (1..retries).each do
```
sshao: `retries.times do`

robot_name/robot_name.rb
```diff
@@ -1,25 +1,62 @@
...
+      new_name = generator ? generator.call : generate_name
```
sshao: `generate_name` doesn't make it super explicit that it's the default, fall-through name generator -- maybe change this method name to indicate that's the case.

optionally, refactor `generate_name` into a proc and assign it to some variable `generator` (possibly an instance variable) so that you're always calling a proc to generate a name. then instead of doing this ternary here, this line could just become `new_name = generator.call`

robot_name/robot_name.rb
```diff
@@ -1,25 +1,62 @@
...
-    @@registry << @name
+    raise NameCreateError, ERR_CREATE
+  end
+
+  def generate_name
+    "#{generate_char_str(2)}#{generate_num_str(3)}"
+  end
+
+  def unique?(name)
+    @@registry.include?(name) ? false : true
```
sshao: `!@@registry.include?(name)` ?

robot_name/robot_name.rb
```diff
@@ -1,25 +1,62 @@
...
+  end
+
+  def valid?(name)
+    valid = name.match(/[[:alpha:]]{2}[[:digit:]]{3}/)
+    raise NameInvalidError, ERR_INVALID unless valid
```
sshao: don't raise here, just return false if the name is invalid. `valid?` is a boolean method, as indicated by the ending `?` in its name -- so its only responsibility should be to return true or false, not catch erroneous input and then raise errors. the way it's written right now, there's no scenario in which `valid` would actually return false. kinda weird for a boolean method.

if you must raise, then do something like `raise NameInvalidError, ERR_INVALID unless valid?(name)` further above -- maybe above line 25.

robot_name/robot_name.rb
```diff
@@ -1,25 +1,62 @@
...
+    valid
+  end
+
+  def generate_char_str(length)
+    str = ""
+
+    length.times do
+      str << ('A'..'Z').to_a.sample
+    end
```
sshao: for a line this short you could compact it down to one line. `length.times { str << ('A'..'Z').to_a.sample }`

also, maybe `str = ('A'..'Z').to_a.sample(length).join`?

robot_name/robot_name.rb
```diff
@@ -1,25 +1,62 @@
...
+
+    str
+  end
+
+  def generate_num_str(length)
+    str = ""
+
+    length.times do
+      str << rand(10).to_s
+    end
```
sshao: likewise, `length.times { str << rand(10).to_s }`. normally the `do/end` block is used when the block body is more complex, or spans longer/multiple lines.

robot_name/robot_name.rb
jmtai: Yeah, I thought I read in one of the style guides that the use of ! was not preferred but I got it confused with !!

robot_name/robot_name.rb
```diff
@@ -1,25 +1,62 @@
...
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+    @name = generate_unique_name(generator, MAX_RETRIES)
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+    @@registry << @name
+  end
+
```
sshao: seems like every method (other than `initialize`) should be a `private` method :)

sshao:
looks good! review the comments i've made, but otherwise you're all set -- no need to resubmit :+1:

robot_name/robot_name.rb
jmtai: Yeah, I originally wrote it in a functional style since I'd normally write like that in Scala, but I wasn't sure if that was the preferred style here.

Is the rule of thumb something like if you can express it in less than 80 characters, go ahead and implement it with chained methods?

robot_name/robot_name.rb
jmtai: Whoops!  Good catch.  Are exceptions used exclusively here or are return values also used for error checking?

robot_name/robot_name.rb
sshao: The rule of thumb is a little more vague than that -- I'd say it depends more on how many methods you chain, and the main thing you're looking for is readability :) For example, something like `str.to_a.sample(length).join.split(/regex/).gsub(stuff, more_stuff).join()...` is less than 80 characters, but I'd still split it up!

In fact, I'd be a little leery of even the `('A'..'Z').to_a.sample(length).join` that I suggested, but I threw it out there since it seems a little more readable than what you have -- as I'd argue that `sample(length)` makes it clearer that you're looking to grab `length` number of random characters, as opposed to `length.times { str << ... }`.

robot_name/robot_name.rb
sshao: Not certain that I understand what you're asking here, could you clarify? 

In the case I suggested above, you actually would be using both -- you'd be using the return value of `valid?` *to* determine whether or not to `raise NameInvalidError` :)

jmtai: Sorry, the question was unrelated to this specific code.

Just asking in general if the code written here typically uses only exceptions for signaling errors or if return values from methods (as you'd typically see in C/C++) are also used.

sshao: oh, i see. yeah, i think we only use exceptions to signal errors (fatal errors, anyway)

PR #563, opened 2015-03-12
Data science

wyaeiga:


ashtonthomas:
Updating the PR title to "Data science" FYI

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,17 @@
...
+    tester = ABAnalyzer::ABTest.new values
+    confidence_level = 1 - tester.chisquare_p
+    confidence_level.round(2)
+  end
+
+  def self.conversion_rate(successes, trials, confidence_level = 0.95)
+    confidence_interval = ABAnalyzer.confidence_interval(successes, trials, confidence_level)
+    conversion_rate = (confidence_interval[0] + confidence_interval[1]) / 2
+    error = (confidence_interval[1] - confidence_interval[0]) / 2
+
```
zsyed1: Nice usage of `ABAnalyzer`

data_science/data_import.rb
```diff
@@ -0,0 +1,62 @@
+require 'json'
+require './data_analyzer'
+
+class DataImport
+  attr_reader :data_hash
+
+  def initialize(file_name = 'data_export_2014_06_20_15_59_02.json')
+    data_file = File.read(file_name)
+    @data_hash = JSON.parse(data_file)
+  end
```
zsyed1: Avoid putting variable types in variable names. The variable `@data_hash` is better named by `@data`. That way you don't lock down the _type_ of the variable.

data_science/data_import.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def confidence
+    DataAnalyzer.confidence(format)
+  end
+
+  def conversion_rates
+    formatted_data = format
+    conversion_rates = {}
```
zsyed1: You can remove the caching of `format` with `formatted_data`. Just refer to it as `format` in line 19 and remove line 17.

data_science/data_import.rb
```diff
@@ -0,0 +1,62 @@
...
+      total = successes + fails
+      conversion_rates[cohort.to_sym] = DataAnalyzer.conversion_rate(successes, total)
+    end
+
+    conversion_rates
+  end
+
+  def format
+    formatted_data = {}
+    formatted_data[:cohort_a] = { conversions: 0, fails: 0 }
```
zsyed1: I would remove the prefix `formatted_` and just call it `data`.

data_science/data_import.rb
```diff
@@ -0,0 +1,62 @@
...
+    formatted_data[:cohort_b] = { conversions: 0, fails: 0 }
+    @data_hash.each do |entry|
+      cohort = "cohort_#{entry['cohort'].downcase}".to_sym
+      formatted_data[cohort][:conversions] += 1 if entry['result'] == 1
```
zsyed1: You might want to move the right side of line 34 into a separate method. Same with line 44 below. The name of the method will help the reader understand what the purpose of that formatting is. For example: `cohort = normalized_cohort_name(entry)` or something like that.

data_science/data_import.rb
```diff
@@ -0,0 +1,62 @@
...
+      formatted_cohort = cohort.to_s.split('_').map(&:capitalize).join(' ')
+      cohort_conversion = rates[:conversion_rate]
+      cohort_error = rates[:error]
+      puts "The conversion rate for #{formatted_cohort} "\
+        "is #{cohort_conversion} +/- #{cohort_error}"
+    end
+
+    puts "The confidence level is #{confidence}."
+  end
+end
```
zsyed1: The method `print_statistics` looks like it's doing two things:

Display Conversions
Display Rates

Since there is no sharing of information from the first to the second, it's better to put each of these into separate methods so that this method becomes 3 lines:

```
def print_statistics
  display_conversions
  display_rates
  puts "The confidence level is #{confidence}."
end
```

data_science/data_import.rb
```diff
@@ -0,0 +1,62 @@
...
+      cohort_conversion = stats[:conversions]
+      cohort_fails = stats[:fails]
+      cohort_total = cohort_conversion + cohort_fails
+      puts "The conversions for #{formatted_cohort} "\
+        "is #{cohort_conversion} out of #{cohort_total}"
+    end
+
+    conversion_rates.each do |cohort, rates|
+      formatted_cohort = cohort.to_s.split('_').map(&:capitalize).join(' ')
+      cohort_conversion = rates[:conversion_rate]
```
zsyed1: Lines 44 and 53 are doing the exact same thing - better to move this into a method with a descriptive name say: `def pretty_name(cohort)`. You can then remove line 44 and inline `pretty_name(cohort)` in line 48.

data_science/rspec/data_analyzer.rb
```diff
@@ -0,0 +1,23 @@
+require_relative '../data_analyzer'
+
+describe 'data analyzer' do
+  context 'when conducting pre-calculated chi-squared test' do
+    it 'should return 0.96' do
+      test_values = {}
+      test_values[:cohort_a] = { conversions: 41, fails: 259 }
+      test_values[:cohort_b] = { conversions: 55, fails: 545 }
+      response = DataAnalyzer.confidence(test_values)
```
zsyed1: Lines 6 - 9 should live inside a `let` block. That's where you typically want to initialize/setup objects that you want to test.

data_science/rspec/data_analyzer.rb
```diff
@@ -0,0 +1,23 @@
...
+      expect(response).to eql(0.96)
+    end
+  end
+
+  context 'when calculating pre-calculated conversion rate' do
+    it 'should return .18 +/- 0.01' do
+      successes = 711
+      trials = 4000
+      response = DataAnalyzer.conversion_rate(successes, trials)
+      expect(response[:conversion_rate]).to eql(0.18)
```
zsyed1: See comment above about `let` blocks. If this is unclear, please feel free to hit me up on Slack.

data_science/rspec/data_import.rb
```diff
@@ -0,0 +1,35 @@
...
+    it 'should return .18 +/- .01 for both groups' do
+      data_set = DataImport.new('test_data_2.json')
+      response = data_set.conversion_rates
+      expect(response[:cohort_a][:conversion_rate]).to eql(0.18)
+      expect(response[:cohort_a][:error]).to eql(0.01)
+      expect(response[:cohort_b][:conversion_rate]).to eql(0.18)
+      expect(response[:cohort_b][:error]).to eql(0.01)
+    end
+  end
+end
```
zsyed1: Your code is nicely written, easy to read. I don't know if you need another pull-request, this is up to your manager @djkotowski

data_science/data_import.rb
djkotowski: I don't think this clarification is bad.

data_science/data_import.rb
djkotowski: It wouldn't even be horrible to make L60 into its own one-line method.

```ruby
def print_statistics
  print_conversions
  print_rates
  print_confidence_level
end
```

data_science/rspec/data_analyzer.rb
```diff
@@ -0,0 +1,23 @@
+require_relative '../data_analyzer'
```
djkotowski: RSpec files typically live in the `spec/` directory and filenames end with the `_spec` suffix.  Example:

```
data_science/spec/data_analyzer_spec.rb
```

data_science/rspec/data_analyzer.rb
```diff
@@ -0,0 +1,23 @@
...
+
+describe 'data analyzer' do
```
djkotowski: Just reference the class here.

```ruby
describe DataAnalyzer do
  # ...
end
```

data_science/rspec/data_analyzer.rb
```diff
@@ -0,0 +1,23 @@
...
+  context 'when conducting pre-calculated chi-squared test' do
```
djkotowski: Contexts generally describe preconditions.  Examples could include "when there are no views" or "when valid data is provided".

In this case, create another `describe` statement to indicate you're testing the `#confidence` method.

```ruby
describe "#confidence" do
  # ...
end
```

data_science/rspec/data_analyzer.rb
```diff
@@ -0,0 +1,23 @@
...
+    it 'should return 0.96' do
```
djkotowski: `it` statements should act as documentation for your class.  As it stands now, this test would read
```
Data Analyzer when conducting pre-calculated chi-squared test should return 0.96
```

The description should be independent of your test data.  Also, avoid the use of "should" in your descriptions.  The word "should" does not imply confidence in your code.

data_science/rspec/data_import.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  context 'when calculating conversion rate' do
+    it 'should return .18 +/- .01 for both groups' do
+      data_set = DataImport.new('test_data_2.json')
+      response = data_set.conversion_rates
+      expect(response[:cohort_a][:conversion_rate]).to eql(0.18)
+      expect(response[:cohort_a][:error]).to eql(0.01)
+      expect(response[:cohort_b][:conversion_rate]).to eql(0.18)
+      expect(response[:cohort_b][:error]).to eql(0.01)
```
djkotowski: If `:conversion_rate` and `:error` are the only results in each cohort, you could easily do

```ruby
expect(response[:cohort_a]).to eq({ conversion_rate: 0.18, error: 0.01 })
```

Furthermore, if the response hash only contains `:cohort_a` and `:cohort_b`, you could just make one hash to compare to.

djkotowski:
This is a good start.  You'll definitely want to cover more edge cases.  What happens when there aren't any views?  What happens if the test data is different?  Also look at the use of `let` and `subject` statements; they'll help clean up your code.

Please take a look at comments, refactor, and resubmit.  Let me know if you have any questions.

PR #564, opened 2015-03-13
Shopping cart features

theundeadmonk:


shopping_cart/features/cart_manipulation.feature
```diff
@@ -0,0 +1,30 @@
+Feature: Add and remove items from cart
+  Once I am logged in
+  I want to be able to view and add items to my cart
```
djkotowski: I recommend sticking with the `As a ______, I want to _______, in order to _______.` syntax.

Also, the feature name does not match the description.  The name implies adding and removing, while the description implies adding and viewing.  These seem like separate features anyway.

shopping_cart/features/cart_manipulation.feature
```diff
@@ -0,0 +1,30 @@
...
+    Then I should see the list of items
+
+  Scenario: View item
+    Given I am on the "items" page
+    When I click on an item
+    Then I should be taken to the page for that item
+
+  Scenario: Add item to cart
+    Given I am on a page for a particular item
+    When I click "+"
```
djkotowski: I recommend avoiding using the web steps such as `When I click "button_content"` as it couples the feature with the view implementation.  Let your step definitions handle that binding and leave your features generic.

shopping_cart/features/checkout.feature
```diff
@@ -0,0 +1,51 @@
...
+    And the page should display "Nothing to checkout"
+
+  Scenario: Checkout with items in cart
+    When I click the checkout button
+    And there are items in my cart
+    Then I should be taken to the checkout page
+
+  Scenario: Enter coupon
+    Given I am on the checkout page
+    Then I should be given the option to enter a coupon
```
djkotowski: This scenario isn't necessary.  If there's no coupon option, all your succeeding steps will fail.  What value does this step really provide?

shopping_cart/features/shopping_cart.feature
```diff
@@ -0,0 +1,35 @@
...
+    When I click "-"
+    Then that item should vanish from my cart
+
+  Scenario: Increase quantity of item in cart
+    Given I am on the shopping cart page
+    And there is an item on that page
+    When I click "+" next to that item
+    Then a single unit of that item should be added to my cart
+
+
```
djkotowski: Nitpick: Extra newlines

shopping_cart/features/user_logs_in.feature
```diff
@@ -0,0 +1,25 @@
...
+  Scenario: Sign in with an invalid username and password
+    When I enter username "Invalid"
+    And I enter password "Invalid"
+    And I click "Sign In"
+    Then I should be taken to the home page again
+    And the page should display "Invalid Credentials"
+
+  Scenario: Continue as guest
+    When I click "Guest"
+    Then I should be able to make a purchase
```
djkotowski: What happens when you log in with an anonymous cart?  How do you handle cart merging?

shopping_cart/features/shopping_cart.feature
djkotowski: This entire project is about a shopping cart.  Why is this one `shopping_cart.feature` but the others aren't?  This feature seems a bit unfocused.

shopping_cart/features/checkout.feature
```diff
@@ -0,0 +1,51 @@
...
+    When I enter an invalid address
+    Then I should be taken to the checkout page
+    And the page should say "Invalid address"
+
+  Scenario: Enter valid address
+    Given I am on the checkout page
+    When I enter a valid address
+    Then I should get the estimated shipping time
+
+
```
djkotowski: Coupons and shipping estimates seem like they should be in separate files.

shopping_cart/features/cart_manipulation.feature
```diff
@@ -0,0 +1,30 @@
...
+  Scenario: Remove item from cart
+    Given I am on a page for a particular item
+    And that item exists in my cart
+    When I click "-"
+    Then a single unit of that item should be removed from my cart
+
+  Scenario: Attempt to remove invalid item from my cart
+    Given I am on a page for a particular item
+    And that item does not exist in my cart
+    Then I should not see "-"
```
djkotowski: What about adding an item to your cart that already exists there?

djkotowski:
This is a good start.  I'd look at handling cart merges (anonymous user w/ cart --> log in) and also organizing your tests a bit more.  Some of your features seem unfocused.  Let the feature name and description guide what scenarios you put in it.  If a scenario doesn't seem to fit, extract it to another feature.  Please add a few more tests, work on the organization, and resubmit.

PR #565, opened 2015-03-13
Dinodex

wyaeiga:


dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,56 @@
+require './dino'
+
+class DinoCatalog
+  attr_reader :dino_catalog
```
ashtonthomas: does DinoCatalog have a dino_catalog? Perhaps we should rename `:dino_catalog` to just `:dinos`?

dino_catalog/dinodex_filter.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def update_filtered_catalog
+    @filtered_catalog = @last_result unless @last_result.empty?
+    puts 'chain updated'
+  end
+
+  def reset_filtered_catalog
+    @filtered_catalog = @dino_catalog
+    puts 'chain reset'
```
ashtonthomas: I don't know i if I like your puts here in dinodex_filter. If this message is important to the program, you should return the message but not print it out.

Either return some value and handle the printing in the interface class or just assume it executed correctly and output in the interface class without returning any value.

ashtonthomas:
@wyaeiga, this is a really great improvement. I hope you enjoyed this back and forth as it has produced code that is easy to read and understand and, therefore, easy to change. It brought up a lot of really great topics.

Take down the link to this PR, and come back to it when you complete training. It will be interesting to see if you make any further changes.

I only had two short comments but no other PR is required for this exercise

:+1:

PR #566, opened 2015-03-16
Dino Catalog refactor

jmtai:
What changed:

- Created a "Dino" class and moved matching logic into it
- DinoCatalog class now only handles storing/searching the dino database
- Moved all printing logic in DinoPrinter
- Moved all application level logic into "DinoApp" class
- Other cleanup based on previous review

### DIFF
--- a/dino_catalog/dino.rb
+++ b/dino_catalog/dino.rb
@@ -46,7 +46,6 @@ class Dino
   def match_other?(param)
     key = param[0]
     val = param[1]
-
     dino_val = send(key)
     dino_val.downcase.index(val)
   end
@@ -61,5 +60,15 @@ class Dino
 
   def small?
     weight.to_i.between?(1, SMALL_WEIGHT)
+  end
+
+  def to_hash
+    hash = {}
+    self.instance_variables.each do |var|
+      category = var.to_s.delete("@").upcase
+      val = self.instance_variable_get(var)
+      hash[category] = val if val.to_s.length > 0
+    end
+    hash
   end
 end
--- a/dino_catalog/dino.rb
+++ b/dino_catalog/dino.rb
@@ -1,0 +1,65 @@
+class Dino
+  attr_accessor :name, :period, :diet, :weight, :walking, :continent,
+                :description
+
+  SMALL_WEIGHT = 4000
+  CARNIVORE_DIETS = %w(carnivore insectivore piscivore)
+
+  def initialize(values = {})
+    @name = values[:name]
+    @period = values[:period]
+    @diet = values[:diet]
+    @weight = values[:weight]
+    @walking = values[:walking]
+    @continent = values[:continent]
+    @description = values[:description]
+  end
+
+  def diet=(val)
+    @diet = val
+    @diet = "Carnivore" if val == "Yes"
+    @diet = "Herbivore" if val == "No"
+  end
+
+  def match?(params)
+    params.each do |param|
+      return false unless match_one?(param)
+    end
+  end
+
+  def match_one?(param)
+    key = param[0]
+    matcher = (key == "weight" || key == "diet") ? key : "other"
+    send("match_#{matcher}?", param)
+  end
+
+  def match_diet?(param)
+    val = param[1]
+    ((val == "carnivore") && carnivore?) || (val == diet.downcase)
+  end
+
+  def match_weight?(param)
+    val = param[1]
+    (val == "big" && big?) || (val == "small" && small?) || (weight == val)
+  end
+
+  def match_other?(param)
+    key = param[0]
+    val = param[1]
+
+    dino_val = send(key)
+    dino_val.downcase.index(val)
+  end
+
+  def carnivore?
+    CARNIVORE_DIETS.include?(diet.downcase)
+  end
+
+  def big?
+    weight.to_i > SMALL_WEIGHT
+  end
+
+  def small?
+    weight.to_i.between?(1, SMALL_WEIGHT)
+  end
+end
--- a/dino_catalog/dino_catalog.rb
+++ b/dino_catalog/dino_catalog.rb
@@ -1,0 +1,126 @@
+require_relative "dino_csv_parser"
+require_relative "dino_printer"
+
+class DinoCatalog
+  def initialize
+    @dinos = []
+    @categories = nil
+    @results = []
+
+    @csv_parser = DinoCsvParser.new
+    @printer = DinoPrinter.new
+
+    load_all_csv_files
+  end
+
+  def load_all_csv_files
+    filenames = Dir["./*.csv"]
+
+    filenames.each do |filename|
+      @csv_parser.parse_file(filename)
+      @dinos << @csv_parser.dinos
+    end
+
+    @dinos.flatten!
+  end
+
+  def parse_search_params(criteria)
+    begin
+      params = criteria.split(",")
+      parsed_params = params.map do |param|
+        args = param.split(/[=<>]/)
+        [args[0], args[1]]
+      end
+    rescue
+      puts "Invalid parameters"
+      parsed_params = []
+    end
+
+    parsed_params
+  end
+
+  def find(criteria)
+    if criteria.length == 0
+      puts "You must enter search criteria for the find command"
+      return
+    end
+
+    @results = []
+    params = parse_search_params(criteria)
+
+    @dinos.each do |dino|
+      @results << dino if match?(params, dino)
+    end
+
+    print_find_result
+  end
+
+  def print_find_result
+    puts ""
+    print "Found #{@results.length} match(es): "
+    matches = @results.map { |dino| dino["NAME"] }
+    puts matches.join(", ")
+  end
+
+  def match?(params, dino)
+    params.each do |param|
+      key = param[0].upcase
+      val = param[1].downcase
+      return false unless match_one?(key, val, dino)
+    end
+
+    true
+  end
+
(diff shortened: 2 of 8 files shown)


dino_catalog/dino.rb
```diff
@@ -0,0 +1,74 @@
...
+    @diet = values[:diet]
+    @weight = values[:weight]
+    @walking = values[:walking]
+    @continent = values[:continent]
+    @description = values[:description]
+  end
+
+  def diet=(val)
+    @diet = val
+    @diet = "Carnivore" if val == "Yes"
```
incazteca: This logic of turning a yes or a no into carnivore or herbivore can be done in the CSV Parser. This way if additional inputs are added with different diet headings and values it can have the values normalized at that step.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,74 @@
...
+    @diet = "Herbivore" if val == "No"
+  end
+
+  def match?(params)
+    params.each do |param|
+      return false unless match_one?(param)
+    end
+  end
+
+  def match_one?(param)
```
incazteca: `match_one?` what? If this is matching a dino name the method `match_one_dino?` for better readability.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def diet=(val)
+    @diet = val
+    @diet = "Carnivore" if val == "Yes"
+    @diet = "Herbivore" if val == "No"
+  end
+
+  def match?(params)
+    params.each do |param|
```
incazteca: Could make the block split param into key and value variables so that you don't have to assign them in the other methods. You could then just have `params.each do |k, v|`

dino_catalog/dino.rb
```diff
@@ -0,0 +1,74 @@
...
+    @description = values[:description]
+  end
+
+  def diet=(val)
+    @diet = val
+    @diet = "Carnivore" if val == "Yes"
+    @diet = "Herbivore" if val == "No"
+  end
+
+  def match?(params)
```
incazteca: All these match methods should go into the DinoCatalog since it should take the responsibility of finding and matching Dino objects.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def parse_search_params(criteria)
+    params = criteria.split(",")
+    map_to_key_val_pairs(params)
+  rescue
+    puts "Invalid parameters"
+    []
+  end
+
+  def map_to_key_val_pairs(params)
```
incazteca: Try to fix this method so that spaces in between the `=` is ignored. That way the command `find diet = carnivore` will work.

dino_catalog/dino_printer.rb
```diff
@@ -0,0 +1,84 @@
...
+    matches = results.map { |dino| dino.name }
+    puts matches.join(", ")
+  end
+
+  private
+
+  def display_table(list)
+    categories = get_categories(list)
+    print_table_header(categories)
+    verbose = list.length == 1 ? true : false
```
incazteca: what is verbose being used for?

incazteca:
@jmtai: Please close this PR and resubmit with comments from Hound and myself addressed. Also fix the printing of the dinosaur so that it meets the requirements that is listed in the README. Currently when dinosaurs are printed out only the names are printed and not all it's facts.

dino_catalog/dino.rb
jmtai: Oops, I meant to move that out but forgot.  There's actually an unused "DIET_REMAP" hash in the CSV parser.

dino_catalog/dino.rb
jmtai: I actually thought this was pretty understandable from the context.

match?(params) --> does a dino match all the [category, value] entries in the params list
match_one?(param --> does a dino match one [category, value] parameter entry

If this is failing the 5 second rule, perhaps "match?" could be renamed to "match_all?"

dino_catalog/dino.rb
jmtai: I tried it that way initially but didn't like the resulting code.  Since I'm dispatching to a specific matcher method in match_one? it required creating methods with unused parameters.  Personally, I found that more objectionable than having to extract the key and/or value in each of the matcher methods.

I'm ok with either implementation - I just found this one more aesthetically pleasing.

dino_catalog/dino.rb
jmtai: Ha!  I originally implemented it that way and it was suggested in the first code review that the match methods should be put into the Dino class.  After implementing it both ways, I think it produces cleaner code this way and still works logically.

dino_catalog/dino_catalog.rb
jmtai: It's designed to ignore extraneous whitespace.  Further down you can see "strip" is called for args[0].  args[1] also used to have strip applied but for some reason it's not there any more.  I think I must have forgotten to save the change at some point.

jmtai:
@incazteca: The printing works as described in the README.  The "print" command will print out all the dino facts after a find command is processed.  The wording in the description was slightly ambiguous so I chose to print a compact result list for the find command, but allow the user to retrieve the full printout manually if desired.

dino_catalog/dino.rb
djkotowski: I think this really belongs in the Dino class.  The DinoCatalog should have the responsibility of filtering, but its selection logic could just use Dino's `#match?` method.

dino_catalog/dino.rb
djkotowski: `#match_one?` is a bit hard to understand, because `one` could mean anything.  Also, it's a bit confusing to accept a two-element array.  I'd prefer

```ruby
def match_param?(key, value)
  # your code here
end

def match?(params)
  params.all? { |key, value| match_param?(key, value) }
end
```

This new block also takes advantage of the queries available from the `Enumerable` module.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def to_hash
+    hash = {}
+    self.instance_variables.each do |var|
+      category = var.to_s.delete("@").upcase
+      val = self.instance_variable_get(var)
+      hash[category] = val if val.to_s.length > 0
+    end
+    hash
+  end
```
djkotowski: In Ruby, I've found `#to_hash` often refers to a hash value for equality testing, whereas `#to_h` returns a `Hash`.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,45 @@
...
+    JSON.pretty_generate(hash_array)
+  end
+
+  private
+
+  def parse_search_params(criteria)
+    params = criteria.split(",")
+    map_to_key_val_pairs(params)
+  rescue
+    puts "Invalid parameters"
```
djkotowski: Avoid using `puts` in collection classes.  Save the output for your runner script or a class that deals solely with output formatting.  Either return `[]`, or raise an `ArgumentError` with the message "Invalid parameters".

dino_catalog/dino_test.rb
```diff
@@ -0,0 +1,119 @@
...
+=> quit
+   Quit the program
+    HEREDOC
+
+    puts help
+  end
+end
+
+app = DinoApp.new
+app.run
```
djkotowski: Creating a runner class then running `#run` on that class is considered "un-rubylike".

dino_catalog/dino_test.rb
```diff
@@ -0,0 +1,119 @@
...
+
+=> print
+   Print dino data for the results of a find operation
+
+=> export
+   Display dino data in JSON format
+
+=> quit
+   Quit the program
+    HEREDOC
```
djkotowski: Nice to see the Heredoc.

dino_catalog/dino_test.rb
```diff
@@ -0,0 +1,119 @@
...
+    send(action, params)
+  end
+
+  def print_results(params)
+    if params.downcase == "all"
+      @printer.display_dinos(params, @catalog.dinos)
+    else
+      @printer.display_dinos(params, @catalog.results)
+    end
+  end
```
djkotowski: Why not `#show_results` to stay consistent with the rest of your methods?

dino_catalog/dino.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def match?(params)
+    params.each do |param|
+      return false unless match_one?(param)
+    end
+  end
+
+  def match_one?(param)
+    key = param[0]
+    matcher = (key == "weight" || key == "diet") ? key : "other"
```
djkotowski: Avoid ternary operators when possible.  They can get long and unruly fast.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,45 @@
...
+  def map_to_key_val_pairs(params)
+    params.map do |param|
+      args = param.split("=")
+      raise ArgumentError if args[0].nil? || args[1].nil?
+      [args[0].strip.downcase, args[1].downcase]
+    end
+  end
+
+  def convert_dinos_to_hashes
+    @dinos.map { |dino| dino.to_hash }
```
djkotowski: This capability makes the code even easier to read.
```ruby
@dinos.map(&:to_hash)
```

djkotowski:
This is looking good.  Take a look at the suggestions and refactor a bit.  I don't think a resubmission is necessary.  Good work!

PR #567, opened 2015-03-16
Data science exercise

jmtai:


data_science/file_parser.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def load(filename)
+    data = File.read(filename)
+    import_data(data)
+  end
+
+  private
+
+  def import_data(data)
+    hash_array = JSON.parse(data)
```
zbright: In general it is not a good idea to name variables based on their data structure. In this case something like `parsed_data` would be preferable

data_science/file_parser.rb
```diff
@@ -0,0 +1,44 @@
...
+    split_cohorts(hash_array)
+    validate_cohorts
+  end
+
+  def split_cohorts(hash_array)
+    @cohort_a = hash_array.select { |hash| hash["cohort"].downcase == "a" }
```
zbright: See above comment about variable names

data_science/file_parser.rb
```diff
@@ -0,0 +1,44 @@
...
+    @cohort_b = hash_array.select { |hash| hash["cohort"].downcase == "b" }
+  end
+
+  def valid_cohort_id?(cohort_id)
+    cohort_id == 'a' || cohort_id == 'b'
```
zbright: Instead of doing an or here, you could do something like `['a','b'].include?(cohort_id)`

data_science/analyzer.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def confidence_level
+    ((1 - @tester.chisquare_p) * 100).round(DECIMAL_PLACES)
+  end
+
+  private
+
+  def run
+    create_cohorts
+    ab_data = ab_analyzer_data
```
zbright: No need to create a variable for this, just pass it in directly

data_science/cohort.rb
```diff
@@ -0,0 +1,54 @@
...
+  private
+
+  def standard_error
+    p = conversion_rate
+    Math.sqrt(p * (1 - p) / sample_size)
+  end
+
+  def parse_data(data)
+    data.each do |hash|
+      result = hash["result"]
```
zbright: variables that you only use once dont really serve a purpose, In general, just avoid it. See variable naming comment as well for `hash`

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,42 @@
+require_relative "spec_helper"
+
+describe Cohort do
+  before :each do
```
zbright: You should use lets here instead of doing a before block. You wont be making changes to the cohorts so there is really no need to re-calculate each time you hit a new it block

data_science/analyzer.rb
```diff
@@ -0,0 +1,57 @@
...
+    @tester = ABAnalyzer::ABTest.new(ab_data)
+  end
+
+  def create_cohorts
+    @cohort_a = Cohort.new(@parser.cohort_a)
+    @cohort_b = Cohort.new(@parser.cohort_b)
+  end
+
+  def ab_analyzer_data
+    values = {}
```
zbright:
```
values = {
               agroup: @cohort_a.to_ab_hash,
               bgroup: @cohortb.to_ab_hash
              }
```

data_science/analyzer.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def print_cohorts
+    puts "Cohort A:"
+    puts @cohort_a
+    puts "Cohort B:"
+    puts @cohort_b
+  end
+
+  def print_confidence_level
+    puts "Confidence level: #{confidence_level.round(2)}%"
```
zbright: use your global for `DECIMAL_PLACES` here

data_science/analyzer.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def print_stats
+    print_cohorts
+    print_confidence_level
+  end
+
+  def confidence_level
+    ((1 - @tester.chisquare_p) * 100).round(DECIMAL_PLACES)
+  end
```
djkotowski: `Analyzer` has a little too much responsibility, handling both confidence level calculations _and_ console output.

data_science/cohort.rb
```diff
@@ -0,0 +1,54 @@
...
+    { :pos => conversions, :neg => non_conversions }
+  end
+
+  def to_s
+    "  sample_size     = #{sample_size}\n" \
+    "  conversions     = #{conversions}\n" \
+    "  non-conversions = #{non_conversions}\n" \
+    "  conversion_rate = #{conversion_rate.round(DECIMAL_PLACES)}\n" \
+    "  error_bars      = #{error_bars}\n" \
+    "\n"
```
djkotowski: You can use a Heredoc here.  Heredocs support interpolation.

data_science/file_parser.rb
djkotowski: It doesn't matter too much for 2 cases, but definitely keep this in mind for larger sets.

data_science/spec/analzyer_spec.rb
```diff
@@ -0,0 +1,13 @@
+require_relative "spec_helper"
+
+describe Analyzer do
+  TEST_FILE = "data_simple_test.json"
+
+  it "should initialize successfully" do
+    expect { Analyzer.new(TEST_FILE) }.not_to raise_error
```
djkotowski: Since you're reusing it, extract
```ruby
Analyzer.new(TEST_FILE)
```
into a `let`.

data_science/spec/spec_helper.rb
```diff
@@ -0,0 +1,95 @@
+require_relative "../file_parser"
+require_relative "../cohort"
+require_relative "../analyzer"
+
+# This file was generated by the `rspec --init` command. Conventionally, all
```
djkotowski: Since this file won't ever be _re_-generated, please delete unused comments.

djkotowski:
I'd cover a few more edge cases in your tests.  What happens if you get bad data?

djkotowski:
This is looking very good, especially in your implementation.  It's very clean.  The tests could use a bit of work, though.  Please take a look at the comments and refactor.  I don't think this really needs another submission, though.

PR #569, opened 2015-03-16
shoppingcart changes

shishirtewari:
Please review shopping cart changes.

### DIFF
--- a/shopping_cart/features/add_product.feature
+++ b/shopping_cart/features/add_product.feature
@@ -1,0 +1,59 @@
+Feature: Adding product to shopping cart
+
+  Scenario Outline: I add a product in shopping cart
+    Given I have empty shopping cart
+    When I add <product> in shopping cart
+    Then I see <quantity> and <price> and <total>
+
+  @happy
+  Examples:
+    | product      |  quantity   |  price    |  total  |
+    | ipad         |  2          |  400      |  800    |
+    | iphone       |  1          |  300      |  300    |
+
+  @bad
+  Examples:
+    | product        | quantity   | price    | total |
+    | watch          | 0          | 400      | 0     |
+
+  @bad
+  Scenario: Adding product with no availability
+    Given I have empty shopping cart
+    When I add "watch" to shopping cart
+    Then I see 0 "watch" in shopping cart
+
+  @happy
+  Scenario: Adding same product twice in cart
+    Given I have empty shopping cart
+    When I add "ipad" to shopping cart twice
+    Then I see 1 "ipad" but quantity is 2
+
+  @happy
+  Scenario: Adding two different products to shopping cart
+    When I add 1 "ipad" to shopping cart
+    And 1 "iphone" to shopping cart
+    Then I see total 2 products in shopping cart
+    And I see quantity price and total
+
+  @happy
+  Scenario: Adding products as anonymous user and log in to account
+    Given I am not login in to my account
+    When  I add "ipad" to shopping cart
+    And after adding product I login
+    Then I see the "ipad" in my shopping cart
+
+  @happy
+  Scenario: Adding products as an existing user and log out
+    Given I login as existing user
+    When I add 1 "ipad" to shopping cart
+    And I log out
+    Then I see empty shopping cart
+
+  @happy
+  Scenario: Adding products as an existing user and log out and login
+    Given I login as existing user
+    When I add 1 "iphone" to shopping cart
+    And I log out
+    And I log in
+    Then I see 1 total products in the shopping cart
+    And I see "iphone" with total and quantity in shopping cart
--- a/shopping_cart/features/coupon.feature
+++ b/shopping_cart/features/coupon.feature
@@ -1,0 +1,39 @@
+Feature: apply to coupon to shopping cart
+
+  Background:
+    Given I have 2 "ipad" in shopping cart and they cost $800
+    And I have  "iphone" in shopping cart and it cost $300
+    And the total for these items are $1100
+
+  Scenario Outline:
+    Given I have <quantity> <product> to shopping cart
+    When I applied <coupon_code>
+    And I press "Apply coupon"
+    Then I see a discount amount $<discount>
+    And see a total of $<total>
+
+  @happy
+  Examples:
+    | product      | quantity | coupon_code   | discount  | total   |
+    | ipad         | 2        | 50OFF         | 100.00    | 700.00  |
+    | iphone       | 1        | 10OFF         | 10.00     | 290.00  |
+
+  @bad
+  Examples:
+    | product      | quantity | coupon_code  | discount | total     |
+    | ipad         | 2        | test         | 0.00     | 800.00    |
+    | iphone       | 1        | xvz          | 0.00     | 300.00    |
+
+  @happy
+    Scenario: Active coupon
+    When I applied coupon code "50OFF"
+    And press "Apply coupon"
+    Then I see discount "100"
+    And total "700"
+
+  @bad
+  Scenario: Expired coupon
+    When I add 1 of "iphone" to shopping cart
+    And I fill in "coupon" with "xvz"
+    And I press "Apply Coupon"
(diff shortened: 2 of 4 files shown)


shopping_cart/features/add_product.feature
```diff
@@ -0,0 +1,59 @@
...
+  @happy
+  Examples:
+    | product      |  quantity   |  price    |  total  |
+    | ipad         |  2          |  400      |  800    |
+    | iphone       |  1          |  300      |  300    |
+
+  @bad
+  Examples:
+    | product        | quantity   | price    | total |
+    | watch          | 0          | 400      | 0     |
```
djkotowski: Until I see the scenario below detailing that this has no availability, I have no idea why this is "bad" path.  Also, this seems more like "sad" path.

Sad = result where nothing happens (e.g. product w/ no availability, login w/o password, etc.)
Bad = result where error occurs (e.g. bad login attempt, non-existent product code, etc.)

shopping_cart/features/add_product.feature
```diff
@@ -0,0 +1,59 @@
...
+
+  @bad
+  Scenario: Adding product with no availability
+    Given I have empty shopping cart
+    When I add "watch" to shopping cart
+    Then I see 0 "watch" in shopping cart
```
djkotowski: This seems identical to the scenario outline above.

shopping_cart/features/add_product.feature
```diff
@@ -0,0 +1,59 @@
...
+
+  @happy
+  Scenario: Adding same product twice in cart
+    Given I have empty shopping cart
+    When I add "ipad" to shopping cart twice
```
djkotowski: Suggestion:  Replace "twice" with "2 times".  It makes the step easier to reuse.

Better suggestion:  Have the first item added to the cart in the `Given` statement.  It is really more of a scenario setup step than an action step.  Then, you can get rid of `Given I have an empty shopping cart`, as that isn't really relevant to this test.

shopping_cart/features/add_product.feature
```diff
@@ -0,0 +1,59 @@
...
+    Then I see 1 "ipad" but quantity is 2
+
+  @happy
+  Scenario: Adding two different products to shopping cart
+    When I add 1 "ipad" to shopping cart
```
djkotowski: Why no `Given I have an empty shopping cart` here, but you use it elsewhere?

shopping_cart/features/add_product.feature
```diff
@@ -0,0 +1,59 @@
...
+    And 1 "iphone" to shopping cart
+    Then I see total 2 products in shopping cart
+    And I see quantity price and total
+
+  @happy
+  Scenario: Adding products as anonymous user and log in to account
+    Given I am not login in to my account
```
djkotowski: Typo:  `logged in`

shopping_cart/features/add_product.feature
```diff
@@ -0,0 +1,59 @@
...
+    When  I add "ipad" to shopping cart
+    And after adding product I login
```
djkotowski: What relevance does "after adding product" have in this step?

shopping_cart/features/add_product.feature
djkotowski: Also, how is this different from the scenarios above?  What step is being executed here?

shopping_cart/features/coupon.feature
```diff
@@ -0,0 +1,39 @@
...
+    | product      | quantity | coupon_code  | discount | total     |
+    | ipad         | 2        | test         | 0.00     | 800.00    |
+    | iphone       | 1        | xvz          | 0.00     | 300.00    |
+
+  @happy
+    Scenario: Active coupon
+    When I applied coupon code "50OFF"
+    And press "Apply coupon"
+    Then I see discount "100"
+    And total "700"
```
djkotowski: `And I see total "700"`

Each step needs to be grammatically independent of each other.  You shouldn't have any steps that require another step to be grammatically correct.

djkotowski: Also, this is different from L13 above.  Stay consistent.

shopping_cart/features/shipping.feature
```diff
@@ -0,0 +1,34 @@
+Feature: Shipping
+
+Scenario Outline: Shipping estimate
+Given I fill out the shipping address and ready to checkout
```
djkotowski: Isn't this step happening below?  Why are you filling out the address in a `Given` and a `When` statement?

shopping_cart/features/shipping.feature
```diff
@@ -0,0 +1,34 @@
...
+@sad
+Examples:
+| address               | city        | state | zip     | shipping_cost | total       | message |
+| 14612 NE 36 ST        | Bellevue    | WA    | 98007   | 0.00          |   800.00    | Sorry we dont ship this item currently in WA |
+| 559 Kha 42 Srinagar   | Lucknow     | UP    | 382007  | 0.00          |   300.00    | Sorry we dont ship this to outside of USA    |
+
+@bad
+Examples:
+| address               | city        | state | zip     | shipping_cost| total        | message |
+| 200 W jackson Blvd    | Chicago     | IL    |  588888 | 0.00         |   800.00     | invalid zip code  |
```
djkotowski: Great happy/sad/bad organization here.

shopping_cart/features/update_remove_product.feature
```diff
@@ -0,0 +1,37 @@
...
+    | product      |  quantity   |  price    |  total  |
+    | ipad         |  2          |  400      |  800    |
+  @sad
+  Examples:
+    | product      |  quantity   |  price    |  total  |
+    | surface      |  1          |  200      |  200    |
+  @bad
+  Examples:
+    | product      |  quantity   |  price    |  total  |
+    | iphone       |  -1         |  300      |  0    |
```
djkotowski: Based on your scenario outline above, `price` is irrelevant.

shopping_cart/features/update_remove_product.feature
```diff
@@ -0,0 +1,37 @@
...
+  Scenario: Remove product
+    When I remove "ipad"
+    Then I see only "iphone" in the cart
+    And quantity and total gets changed
+
+  @happy
+  Scenario: Remove all products from shopping cart
+    When I empty shopping cart
+    Then I see 0 products in shopping cart
+    And I total as 0
```
djkotowski: Grammatically incorrect.

shopping_cart/features/update_remove_product.feature
```diff
@@ -0,0 +1,37 @@
...
+  @happy
+  Scenario:
+    Given I add "ipad" to cart
+    And I add "iphone" to cart
+
+  @happy
+  Scenario: Remove product
+    When I remove "ipad"
+    Then I see only "iphone" in the cart
+    And quantity and total gets changed
```
djkotowski: You express detail in other steps when describing the displayed quantities and totals, yet here you don't.  Be consistent.

shopping_cart/features/update_remove_product.feature
djkotowski: Given this is a bad step, should the product even be in the cart with a -1 qty?

shopping_cart/features/add_product.feature
djkotowski: If you have 0 qty, shouldn't the item be absent from the cart?

djkotowski:
This is a good first submission.  A few notes to keep in mind...
- Avoid including too many details about the implementation in the feature.
- Be consistent with your steps.  You'll want to reuse the definitions in different combinations, so don't name the same step differently without reason to.
- Make sure each of your steps can stand on their own.  A step that says
```cucumber
Given I have product X in my cart
And product Y
```
doesn't stand on its own when you have

```cucumber
Given product Y
```
- Avoid unnecessary steps that are simply describing the status quo.  Steps should alter the state of your application somehow.

Please take a look at the notes throughout, refactor, and resubmit.

PR #570, opened 2015-03-18
DinoDex Project Refactor

ckhadilkar:
I wanted dinodex.rb to be a like a script file that runs the program. So, did not make it into a class.

### DIFF
--- a/dino_catalog/DINODEX_README
+++ b/dino_catalog/DINODEX_README
@@ -1,0 +1,14 @@
+How to run dinodex
+
+Usage:
+ruby dinodex.rb <search_options>
+-b        return all biped dinosaurs
+-c        return all carnivore dinosaurs
+-a        return dinosaur collection
+-n  <NAME>      return facts for this dinosaur
+-wg <WEIGHT IN TONS>    return dinosaurs weighing more than <WEIGHT IN TONS>
+-wl <WEIGHT IN TONS>    return dinosaurs weighing less or equal to <WEIGHT IN TONS                >
+-p  <PERIOD>      return dinosaurs in this <PERIOD>
+You can also combine the criteria
+Example 1: ruby dinodex.rb -bcp Late  will return all biped dinosaurs that are carnivores in 'Late' period.
+Example 2: ruby dinodex.rb -bwg 2 will return all biped dinosaurs who weigh more than 2 tons.
--- a/dino_catalog/dino_catalog.rb
+++ b/dino_catalog/dino_catalog.rb
@@ -1,0 +1,49 @@
+class DinoCatalog
+  WEIGHT_IN_TONS = 4000
+  def initialize(dino_data)
+    @dino_catalog = dino_data
+  end
+
+  def get_walking(type)
+   @dino_catalog.select{ |k, v| type.include?(v.walk) if v.walk }
+  end
+
+  def get_diet(type)
+    @dino_catalog.select { |k, v| type.include?(v.diet) if v.diet }
+  end
+
+  def get_specific_dinosaur(param)
+    @dino_catalog.select { |k, v| param.include?(v.name) if v.name }
+  end
+ 
+  def get_big_dinosaurs
+    @dino_catalog.select { |k, v| v.weight > WEIGHT_IN_TONS}
+  end
+
+  def get_small_dinosaurs
+    @dino_catalog.select { |k, v| v.weight <= WEIGHT_IN_TONS unless v.weight.zero?}
+  end
+  
+  def get_period_specific_dinosaurs(param)
+    @dino_catalog.select { |k, v| v.period.split(" ").any?{|item| param.include? item} }
+  end
+  
+  def get_dinosaur_collection
+     @dino_catalog 
+  end
+
+  def print_data 
+    @dino_catalog.each do |k, v|
+      v.display
+      puts "\n"
+    end
+  end
+  
+  def print_filtered_catalog
+    if @dino_catalog.empty?
+      puts "No records found matching your criteria"
+    else
+      print_data
+    end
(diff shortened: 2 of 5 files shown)


ckhadilkar:
I wanted dinodex.rb to be a like a script file that runs the program. So, did not make it into a class.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
...
+load 'dinosaur_data_parser.rb'
+load 'dino_catalog.rb'
+
+files_to_read = ["dinodex.csv", "african_dinosaur_export.csv"]
+dino_catalog = DinoCatalog.new(DinosaurDataReader.new(files_to_read).dinosaurs)
+option = ARGV[0]
+criteria = ARGV.to_s
+
+def displayUsage
+  puts "Usage:\nruby dinodex.rb <search_options>\n-w  <WALK_TYPE>\t\t\treturn all <WALK_TYPE> dinosaurs\n-d  <DIET_TYPE>\t\t\treturn all <DIET_TYPE> dinosaurs"
```
incazteca: You can use a HEREDOC for this.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
...
+  puts "-a\t\t\t\treturn dinosaur collection\n-n  <NAME>\t\t\treturn facts for this dinosaur\n"
+  puts "-b\t\t\t\treturn dinosaurs weighing more than 2 tons\n-s\t\t\t\treturn dinosaurs weighing less or equal to 2 tons\n-p  <PERIOD>\t\t\treturn dinosaurs in this <PERIOD>"
+  puts "You can also combine the criteria\nExample 1: ruby dinodex.rb -w Biped -d Carnivore -p Late\twill return all biped dinosaurs that are carnivores in 'Late' period."
+  puts "Example 2: ruby dinodex.rb -w Biped -b\twill return all biped dinosaurs who weigh more than 2 tons."
+end
+
+def is_valid_option(criteria)
+  if criteria.include?("-w") ||
```
incazteca: You could use the OptionParser module to handle arguments instead of doing this yourself. http://ruby-doc.org/stdlib-2.2.0/libdoc/optparse/rdoc/OptionParser.html Documentation for Ruby 2.2.0

dino_catalog/DINODEX_README
```diff
@@ -0,0 +1,14 @@
+How to run dinodex
```
incazteca: Take Sophia's advice and have this printed out for the usage. It seems that this is pretty much the same text in there.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,28 @@
...
+    @period = dino[:period]
+    @continent = dino[:continent] || 'African'
+    @diet = dino[:diet] || convert_to_carnivore(dino[:carnivore])
+    @weight = dino[:weight_in_lbs] || dino[:weight]
+    @weight = @weight.to_i
+    @walk = dino[:walking]
+    @description = dino[:description]
+  end
+
+  def convert_to_carnivore(arg)
```
incazteca: This could be made into a private method.

incazteca:
@ckhadilkar Looks mostly good. Take into consideration the notes made by Hound as well. You have some indentation that's off it seems. Also know that the Optionparser exists and is there so you don't have to reimplement CLI argument parsing. A resubmission is not necessary.

dino_catalog/dinosaur_data_parser.rb
```diff
@@ -0,0 +1,23 @@
...
+    files_to_read.each do |file|
+      dino = CSV.new(File.read(file), headers: true, header_converters: :symbol, converters: :all)
+      dino_hash = dino.to_a.map(&:to_hash)
+      build_dinosaurs(dino_hash)
+    end
+  end
+  
+  def build_dinosaurs(dino_hash)
+    dino_hash.each do |dinosaur|
+      @dinosaurs[dinosaur[dinosaur.keys[0]]] = Dinosaur.new(dinosaur)
```
djkotowski: Interesting to make this a hash instead of an array.  Is it really needed as a hash?

The code to assign to the new hash is a bit confusing.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,49 @@
...
+      puts "\n"
+    end
+  end
+  
+  def print_filtered_catalog
+    if @dino_catalog.empty?
+      puts "No records found matching your criteria"
+    else
+      print_data
+    end
```
djkotowski: Keep in mind that guard clauses can shrink this down from 5 lines to 2.

```ruby
return print_data unless @dino_catalog.empty?
puts "No records found matching your criteria"
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,48 @@
...
+  puts "\n"
+
+  if is_valid_option(criteria)
+    dino_catalog.print_filtered_catalog
+  else
+    displayUsage
+  end
+else
+  displayUsage
+end
```
djkotowski: There's a bit of arrowhead code here.

djkotowski:
Looking good.  Take a look at the comments and consider how you would refactor.  No need for a resubmission, though.

ckhadilkar:
Thanks Franco and Dan for the comments. I'll definitely look into
OptionParser. Also, I might have missed out on refactoring some of the
arraowhead code :)

About the hash to store dino objects, When i started working on the
solution, i had this idea that hash of dino objects with 'name' as the key
will be efficient to fetch a dino object. But, yes the code there is a bit
tricky. I can see how array will make it simple to read.

PR #571, opened 2015-03-18
Overlord

cnorrisjr:


overlord/overlord.rb
```diff
@@ -1,14 +1,71 @@
 # run `ruby overlord.rb` to run a webserver for this app
+require "sinatra"
+require_relative "models/Bomb"
```
incazteca: should be requiring `models/bomb`. Server not starting due to it unable to find bomb model.

overlord/features/BadAct.feature
```diff
@@ -0,0 +1,27 @@
...
+  When I use the default codes
+  Then I land on the Activate page
+  When I enter '1235' to activate the bomb
+  Then the bomb status is 'inactive'
+  When I enter '1234' to activate the bomb
+  Then the bomb status is 'active'
+
+Scenario: Mastermind Lands on the main page and enters values for the codes and puts the wrong activation code in
+  Given I am on the home page
+  When I enter '6789' and '9876' into the bomb
```
incazteca: Be a little more descriptive with your steps. This can be `I set '6789' as the activation code` and `I set '9876' as the deactivation code`.

overlord/features/BadAct.feature
```diff
@@ -0,0 +1,27 @@
...
+  Then I land on the Activate page
+  When I enter '1235' to activate the bomb
+  Then the bomb status is 'inactive'
+  When I enter '9876' to activate the bomb
```
incazteca: If testing for a wrong activation code there is no need to input the correct code.

overlord/features/BadAct.feature
incazteca: actually not sure which code is which.... so yeah be more descriptive.

overlord/features/BadAct.feature
```diff
@@ -0,0 +1,27 @@
+Feature: Evil mastermind configs a bomb, but enters bad data in the activation field and bomb is not activated.
```
incazteca: Seems that it would make more sense to have an activation feature that would test a good and bad activation code for default and non-default activation codes. Also change the file names to snake case to follow conventions.

overlord/features/ConfigActDeactBomb.feature
```diff
@@ -0,0 +1,69 @@
+Feature: Evil mastermind wants to config a bomb,
+activate the bomb, and deactivate the bomb
+
+Scenario: Mastermind Lands on the main page and uses the default codes deactivates 1st time
+  Given I am on the home page
```
incazteca: The first three steps can be placed in a background block.

overlord/features/ConfigActDeactBomb.feature
```diff
@@ -0,0 +1,69 @@
...
+  Then My bomb is activiated and I have '3' attempts to disarm
+  When I enter '0001' as my deactivation code
+  Then My bomb is activiated and I have '2' attempts to disarm
+  When I enter '0001' as my deactivation code
+  Then My bomb is activiated and I have '1' attempts to disarm
+  When I enter '0000' as my deactivation code
+  Then I the bomb is deactivated and I can config a new one
+
+Scenario: Mastermind Lands on the main page and uses the custom codes deactivates 1st time
+  Given I am on the home page
```
incazteca: Could probably just separate this out into another feature file where the tests are done with non-default codes and have the first three steps here in a background block. Or you could just make a step such as `Given default activation and deactivation codes` and it will basically do set you up and leave you on the activate page. The same could also then be done for the non-default codes.

overlord/features/ConfigActExplodes.feature
```diff
@@ -0,0 +1,28 @@
+Feature: Evil mastermind wants to config a bomb,
+activate the bomb, and sets the bomb off
+
+Scenario: Mastermind Lands on the main page and uses the default codes
+  Given I am on the home page
+  When I use the default codes
+  Then I land on the Activate page
+  When I enter '1234' to activate the bomb
+  Then My bomb is activiated and I have '3' attempts to disarm
+  When I enter '0001' as my deactivation code
```
incazteca: You could just have the step modified to accept the number of times an input is entered so that you don't have to keep writing the same steps. Example: `I enter '0001' as my deactivation code 1 time`

overlord/features/ConfigActExplodes.feature
```diff
@@ -0,0 +1,28 @@
...
+  Then My bomb is activiated and I have '1' attempts to disarm
+  When I enter '0001' as my deactivation code
+  Then My bomb goes off and everyone dies
+
+Scenario: Mastermind Lands on the main page and uses the custom codes
+  Given I am on the home page
+  When I enter '6789' and '9876' into the bomb
+  Then I land on the Activate page
+  When I enter '6789' to activate the bomb
+  Then My bomb is activiated and I have '3' attempts to disarm
```
incazteca: Also please ensure that `activiated` is changed to `activated`

overlord/features/SadCases.feature
```diff
@@ -0,0 +1,34 @@
...
+  Then I see the alert box and I accept the alert
+  When I enter '1236' in the 'activation' box
+  Then I have '1236' in the 'Activation' box
+
+Scenario: Mastermind Lands on the main page and enters nothing in the Deactivation code
+  Given I am on the home page
+  When I enter no code in the 'deactivation' box
+  Then I see the alert box and I accept the alert
+  When I enter '1236' in the 'deactivation' box
+  Then I have '1236' in the 'Deactivation' box
```
incazteca: Is this step basically just making sure that entering in a text field works? Doesn't really seem to be a necessary test.

overlord/models/bomb.rb
```diff
@@ -0,0 +1,43 @@
...
+  attr_reader :activate_code, :deactivate_code, :attempts
+
+  def initialize(activate_code: '1234', deactivate_code: '0000')
+    @activate_code = check_code?(activate_code)
+    @deactivate_code = check_code?(deactivate_code)
+    @status = 'inactive'
+    @attempts = 0
+  end
+
+  def activate(act_user)
```
incazteca: what is `act_user`? Is this the activation code? If so just name the variable `activation_code`

overlord/models/bomb.rb
```diff
@@ -0,0 +1,43 @@
...
+    if @activate_code.eql?(act_user) && @status.eql?('inactive')
+      @status = 'active'
+    else
+      @status
+    end
+  end
+
+  def deactivate(dact_user)
```
incazteca: Same comment regarding `act_user`

overlord/models/bomb.rb
```diff
@@ -0,0 +1,43 @@
...
+    @attempts += 1
+    explode if @attempts >=  3
+    @status = 'inactive' if @deactivate_code.eql?(dact_user)
+  end
+
+  def explode
+    return @status unless @status.eql?('active')
+    @status, @activate_code, @deactivate_code = 'explode', 'explode', 'explode'
```
incazteca: Is it necessary to set the codes to explode?

overlord/overlord.rb
```diff
@@ -1,14 +1,71 @@
...
+end
+
+get '/config' do
+  erb :config
+end
+
+post '/config' do
+  @act_code = params[:act]
+  @dact_code = params[:deact]
+  bomb = Bomb.new(activate_code: @act_code, deactivate_code: @dact_code)
```
incazteca: Could just use params[:act] and params[:deact] here instead of having to put them into other variables.

overlord/overlord.rb
```diff
@@ -1,14 +1,71 @@
...
+  session[:activate_code] = bomb.activate_code
```
incazteca: You could just save the bomb object into the session. Then you wouldn't have to save each attribute into the session hash.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+    end
+
+    it 'Create a bomb with a custom Activation Code and default deactivate code' do
+      my_bomb = Bomb.new(activate_code: '9999')
+      checkbomb(%w(inactive 9999 0000), my_bomb)
+    end
+
+    it 'Create a bomb with default activation code and custom deactivate code' do
+      my_bomb = Bomb.new(deactivate_code: '1234')
+      checkbomb(%w(inactive 1234 1234), my_bomb)
```
incazteca: It would be easier to have this first argument a hash so that we would know what the values are supposed to mean.

overlord/image.txt
```diff
@@ -0,0 +1,26 @@
+
```
incazteca: If this file is not being used it should be removed.

incazteca:
@cnorrisjr Please resubmit with changes addressing the comments from Hound and myself. Also please have the Gemfile up to date so that it's easy to start running the tests.

overlord/features/ConfigActDeactBomb.feature
```diff
@@ -0,0 +1,69 @@
+Feature: Evil mastermind wants to config a bomb,
+activate the bomb, and deactivate the bomb
+
+Scenario: Mastermind Lands on the main page and uses the default codes deactivates 1st time
+  Given I am on the home page
+  When I use the default codes
+  Then I land on the Activate page
```
djkotowski: This test looks like it should be on its own.  No need for a repeated Given When Then for every test.  Only test what the scenario says you're testing.

overlord/features/ConfigActDeactBomb.feature
```diff
@@ -0,0 +1,69 @@
...
+  Then My bomb is activiated and I have '3' attempts to disarm
+  When I enter '0000' as my deactivation code
+  Then I the bomb is deactivated and I can config a new one
+
+Scenario: Mastermind Lands on the main page and uses the default codes deactivates 2nd time
+  Given I am on the home page
+  When I use the default codes
+  Then I land on the Activate page
+  When I enter '1234' to activate the bomb
+  Then My bomb is activiated and I have '3' attempts to disarm
```
djkotowski: This is tested above.  No need to retest here.

overlord/features/ConfigActDeactBomb.feature
```diff
@@ -0,0 +1,69 @@
...
+  When I enter '0000' as my deactivation code
+  Then I the bomb is deactivated and I can config a new one
+
+Scenario: Mastermind Lands on the main page and uses the default codes deactivates 3rd time
+  Given I am on the home page
+  When I use the default codes
+  Then I land on the Activate page
+  When I enter '1234' to activate the bomb
+  Then My bomb is activiated and I have '3' attempts to disarm
+  When I enter '0001' as my deactivation code
```
djkotowski: Ditto on re-testing.

overlord/features/SadCases.feature
djkotowski: Definitely unnecessary to test the engine.

overlord/features/step_definitions/general.rb
```diff
@@ -0,0 +1,61 @@
...
+  find(:xpath, "//input[@value='Submit']").click
+end
+
+Then(/^My bomb is activiated and I have '(\d+)' attempts to disarm$/) do |count|
+  expect(page).to have_content("You have #{count} attempts")
+end
+
+When(/^I enter '(\d+)' as my deactivation code$/) do |de_act|
+  fill_in 'de_code_in', with: de_act
+  find(:xpath, "//input[@value='Submit']").click
```
djkotowski: - Favor CSS over XPath when possible.
- Since this exact step reappears elsewhere, extract it to a method.

```ruby
def submit_form
  find("input[value='Submit']").click
end
```

overlord/models/bomb.rb
djkotowski: If you want all the codes to appear as "explode", that's a view concern, not a `Bomb` concern.

overlord/models/bomb.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def valid_code?(pin)
+    /^\d{4}$/ === pin
+  end
+
+  def check_code?(pin)
+    return pin if valid_code?(pin) == true
+    return 'invalid' if valid_code?(pin) == false
```
djkotowski: Don't compare to `true` and `false`.  Just use `if` or `unless`.

Also, you're running the same comparison twice, resulting in L37 being a tautology.  If execution reaches L37, then the code is definitely invalid.

overlord/overlord.rb
```diff
@@ -1,14 +1,71 @@
...
+  erb :activation
+end
+
+post '/activate' do
+  @user_submit = params[:user_submit]
+  bomb_status = bomb.activate(@user_submit)
+  session[:status] = bomb_status
+  puts session[:status]
+  redirect "/activate" if bomb_status.eql?('inactive')
+  redirect "/deactivate" if bomb_status.eql?('active')
```
djkotowski: This is really confusing.  You're redirecting to GET "/activate" from POST "/activate".

overlord/overlord.rb
```diff
@@ -1,14 +1,71 @@
...
+get '/activate' do
+  @status = session[:status]
+  erb :activation
+end
+
+post '/activate' do
+  @user_submit = params[:user_submit]
+  bomb_status = bomb.activate(@user_submit)
+  session[:status] = bomb_status
+  puts session[:status]
```
djkotowski: Get rid of debugging statements.

overlord/overlord.rb
```diff
@@ -1,14 +1,71 @@
...
+end
+
+get '/deactivate' do
+  bomb.status = session[:status]
+  erb :deactivation
+end
+
+post '/deactivate' do
+  bomb.status = session[:status]
+  @user_submit_de = params[:user_submit_de]
```
djkotowski: What is `@user_submit_de` and why is it an instance variable?

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,104 @@
+require_relative '../models/bomb'
+
+describe Bomb do
+
+  context 'Create a bomb' do
+    it 'Create a bomb with the default values' do
```
djkotowski: Your tests should read like documentation.  This will read like
```
Bomb Create a bomb Create a bomb with the default values
```

Consider how the tests will read out when you're writing them.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+      my_bomb = Bomb.new
+      checkbomb(%w(inactive 1234 0000), my_bomb)
```
djkotowski: What does this array mean?

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+      my_bomb.deactivate('1')
+      expect(my_bomb.status).to eql('active')
+      my_bomb.deactivate('2')
+      expect(my_bomb.status).to eql('active')
+      my_bomb.deactivate('3')
+      expect(my_bomb.status).to eql('explode')
+    end
+  end
+
+  def checkbomb(status, my_bomb)
```
djkotowski: Move methods to the start of the spec.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+    expect(my_bomb.status).to eql(status[0])
+    expect(my_bomb.activate_code).to eql(status[1])
+    expect(my_bomb.deactivate_code).to eql(status[2])
```
djkotowski: As mentioned earlier, the status array is confusing.  A hash would work better.

overlord/views/activation.erb
```diff
@@ -0,0 +1,30 @@
...
+// Form validation code will come here.
+function validate()
+{
+  if (/^\d{4}$/.test(document.activate.user_submit.value) == false)
+  {
+    alert( "Please provide an Activation Code that is four numberic characters long" );
+    document.activate.user_submit.focus() ;
+    return false;
+  }
+  return( true );
```
djkotowski: Why parens?

overlord/views/activation.erb
```diff
@@ -0,0 +1,30 @@
+<script type="text/javascript">
+<!--
+// Form validation code will come here.
+function validate()
+{
+  if (/^\d{4}$/.test(document.activate.user_submit.value) == false)
```
djkotowski: While it is a short regex, I suggest extracting it to its own variable.  Also, don't compare expressions to `true`/`false`.

overlord/views/activation.erb
```diff
@@ -0,0 +1,30 @@
...
+  <table style="width:25%">
+
+  <tr>
+   <td> Activation Code  </td>
+   <td><input type="text" id='activation' name="user_submit" value="Enter a number"></td>
+  </tr>
+  <tr>
+  <td><input type="submit" value='Submit'></td>
+  </tr>
+</table>
```
djkotowski: Avoid table layouts.  This isn't critical for Overlord, but you'll get many comments in Pmail if you do this.

overlord/views/config.erb
```diff
@@ -0,0 +1,52 @@
...
+</script>
+<title> Configure a bomb</title>
+<form action="/config" method="post" name='setup' onsubmit="return(validate());">
+<br> Configure a bomb
+<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;,--.!,
+<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;__/&nbsp;&nbsp;&nbsp;-*-
+<br>&nbsp;&nbsp;,d08b.&nbsp;&nbsp;&nbsp;'|`
+<br>0088MM
+<br>`9MMP'
+
```
djkotowski: Maybe extract this out to a partial view.  It's confusing that this is ASCII art just looking at the code.

djkotowski:
This is a good start.  Take a look at the comments, refactor, and please resubmit.

PR #572, opened 2015-03-19
Overlord

wyaeiga:


overlord/bomb.rb
```diff
@@ -0,0 +1,55 @@
+class Bomb
+  attr_accessor :active, :detonated,
+   :deactivation_attempts, :activation_code, :deactivation_code
+
+  def initialize(activation_code = '1234', deactivation_code = '0000')
+    @active = false
```
incazteca: You can make a `status` attribute and then create methods such as `active?` and `detonated?`

overlord/bomb.rb
```diff
@@ -0,0 +1,55 @@
+class Bomb
+  attr_accessor :active, :detonated,
```
incazteca: If you're going to be tracking the status of the bomb it'll make better sense that they have a `:status` accessor

overlord/bomb.rb
```diff
@@ -0,0 +1,55 @@
...
+   :deactivation_attempts, :activation_code, :deactivation_code
+
+  def initialize(activation_code = '1234', deactivation_code = '0000')
+    @active = false
+    @detonated = false
+    @deactivation_attempts = 0
+
+    @activation_code = correct_format?(activation_code) ? activation_code : '1234'
```
incazteca: Don't use magic numbers. Use a constant to refer to the activation code and deactivation code.

overlord/bomb.rb
```diff
@@ -0,0 +1,55 @@
...
+    @detonated
+  end
+
+  def activate(code)
+    return false if active?
+
+    if code == @activation_code
+      @active = true
+      @deactivation_attempts = 0
+      return true
```
incazteca: The `return` is not necessarily needed. You could just have this line as `true` since the method would return the last line.

overlord/bomb.rb
```diff
@@ -0,0 +1,55 @@
...
+    else
+      return false
+    end
+  end
+
+  def deactivate(code)
+    return false if !(active?)
```
incazteca: There is no need for paranthesis. This line can be written as `return false if !active?` or `return false unless active?`

overlord/bomb.rb
```diff
@@ -0,0 +1,55 @@
...
+        @active = false
+        return false
+      else
+        return false
+      end
+    end
+  end
+
+  def correct_format?(code)
+    (code !~ /\D/) && (code.length > 3)
```
incazteca: You can use a single regex statement to check if the code is all digits and at at least greater than 3 digits. Here's a page regarding Regex in Ruby. http://ruby-doc.org/core-2.2.0/Regexp.html

overlord/overlord.rb
```diff
@@ -1,14 +1,40 @@
-# run `ruby overlord.rb` to run a webserver for this app
-
+require 'json'
 require 'sinatra'
+require 'haml'
+require './bomb'
 
 enable :sessions
+$bomb ||= nil
```
incazteca: The use of sigils is not necessary. The bomb can be saved into the session.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+  let (:correct_deactivation_code) { '0000' }
+  let (:incorrect_activation_code) { '1235' }
+  let (:incorrect_deactivation_code) { '0001' }
+  let (:custom_activation_code) { '4321' }
+  let (:custom_deactivation_code) { '1111' }
+  let (:invalid_code) { 'asdf' }
+
+  context 'when initializing' do
+    it 'is not active' do
+      expect(Bomb.new.active?).to eql(false)
```
incazteca: You could use the `bomb` you've already initialized in line 4. Also don't check that it's equal to false. Use the rspec matcher `to_be`. This line can be `expect(bomb.active?).to_be falsy`. Falsy will check if the result is either false or nil.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+    end
+
+    it 'is not detonated' do
+      expect(Bomb.new.detonated?).to eql(false)
+    end
+
+    it 'has no activation attempts' do
+      expect(Bomb.new.deactivation_attempts).to eql(0)
```
incazteca: remember to use `bomb` that's already been setup.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+      expect(deactivation).to eql(false)
+
+      expect(bomb.deactivation_attempts).to eql(0)
+    end
+
+    it 'explodes when deactivation code is entered incorrectly 3 times' do
+      bomb.activate(correct_activation_code)
+      expect(bomb.active?).to eql(true)
+
+      deactivation = bomb.deactivate(incorrect_deactivation_code)
```
incazteca: You can have this looped instead of having to write similar statements multiple times.

incazteca:
@wyaeiga This is a pretty good first submission however you are lacking cucumber tests. Please add them in. Also here's a link regarding rspec's matcher https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers. Also please provide fixes based on the comments.

overlord/bomb.rb
djkotowski: You could also combine the condition with L23.

```ruby
return false if active? || code != @activation_code

@active = true
@deactivation_attempts = 0
true
```

overlord/bomb.rb
djkotowski: Prefer `unless condition` over `if !condition`

overlord/bomb.rb
```diff
@@ -0,0 +1,55 @@
...
+      @active = false
+      return true
+    else
+      @deactivation_attempts += 1
+      if @deactivation_attempts == 3
+        @detonated = true
+        @active = false
+        return false
+      else
+        return false
```
djkotowski: Again, unnecessary returns here.  Also, you return false in both branches, whereas you could just put that part outside the `if` statement.

```ruby
if @deactivation_attempts == 3
  @detonated = true
  @active = false
end

return false
```

Better yet, extract the detonation procedure into its own method.

overlord/spec/bomb_spec.rb
djkotowski: Falsy is good to check for false or nil, but if you want to check with false, use predicate matchers.

```ruby
expect(bomb).not_to be_active
```

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+  let (:custom_deactivation_code) { '1111' }
+  let (:invalid_code) { 'asdf' }
+
+  context 'when initializing' do
+    it 'is not active' do
+      expect(Bomb.new.active?).to eql(false)
+    end
+
+    it 'is not detonated' do
+      expect(Bomb.new.detonated?).to eql(false)
```
djkotowski: Again, predicate matcher here.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+
+    it 'defaults to 1234/0000 as codes when provided invalid codes' do
+      bomb = Bomb.new(invalid_code, invalid_code)
+      expect(bomb.activation_code).to eql(correct_activation_code)
+      expect(bomb.deactivation_code).to eql(correct_deactivation_code)
+    end
+
+    it 'allows custom activation and deactivation codes' do
+      bomb = Bomb.new(custom_activation_code, custom_deactivation_code)
+      expect(bomb.activate(custom_activation_code)).to eql(true)
```
djkotowski:
```ruby
expect(method_call).to be true
```

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+
+  context 'when rearming' do
+    it 'resets the deactivation_attempt counter' do
+      bomb.activate(correct_activation_code)
+      bomb.deactivate(incorrect_deactivation_code)
+      expect(bomb.deactivation_attempts).to eql(1)
+
+      bomb.deactivate(correct_deactivation_code)
+      bomb.activate(correct_activation_code)
+      expect(bomb.deactivation_attempts).to eql(0)
```
djkotowski: These are two tests.  The first checks if the deactivation attempt counter increments.  The second checks for reset.  The first doesn't belong in this step.

djkotowski:
This is a good first submission, though it wasn't developed using BDD.  Please add the cucumber tests, review the comments, update and resubmit.

PR #573, opened 2015-03-19
Overlord exercise

jmtai:


overlord/bomb.rb
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  def initialize(arm_code = ARM_CODE_DFLT, disarm_code = DISARM_CODE_DFLT)
+    unless Bomb.code_valid?(arm_code) && Bomb.code_valid?(disarm_code)
+      raise InvalidCodeError
+    end
+
+    @arm_code = arm_code
+    @disarm_code = disarm_code
+    @state = STATES[:inactive]
```
djkotowski: Why not just use the symbol itself?

```ruby
@state = :inactive
```

overlord/bomb.rb
```diff
@@ -0,0 +1,91 @@
...
+    @disarm_retries = DISARM_RETRIES_MAX
+  end
+
+  def self.code_valid?(code)
+    code.match(/[^0-9]/) ? false : true
```
djkotowski: This could be represented as
```ruby
!code.match(/[^0-9]/)
```

or better yet

```ruby
code.match(/^[0-9]+$/)
```

Use the regular expressions to describe the code, not what isn't a code.

overlord/bomb.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    disarm_code_fail
+    false
+  end
+
+  def arm(code)
+    return false unless inactive?
+    @state = STATES[:armed] if arm_code_match?(code)
+    true
+  end
```
djkotowski: If the arming code is not a match, this method returns `true` still.

overlord/bomb.rb
jmtai: This is me trying to create a pseudo-enum, even though Ruby doesn't support enums.  I suppose this is probably un-Rubylike code.

overlord/features/configure.feature
```diff
@@ -0,0 +1,37 @@
...
+
+  Scenario: Submit invalid activation AND deactivation codes
+    Given I am on the configure page
+    When I submit an invalid activation and deactivation code
+    Then I should see "You entered an invalid code"
+
+  Scenario: Boot bomb with default codes
+    Given I am on the configure page
+    When I do not enter any codes
+    And I click the Save Codes button
```
djkotowski: This is right on the edge of being too implementation-based.  I'm fine with this here, but avoid referring too much to the implementation details.

Example:   `And I click the Save Codes button` vs. `And I save the codes`

overlord/features/configure.feature
```diff
@@ -0,0 +1,37 @@
...
+    Then I should be on the home page
+
+  Scenario: Boot bomb with custom codes
+    Given I am on the configure page
+    When I fill in arm_code with "1111"
+    And I fill in disarm_code with "2222"
```
djkotowski: This _is_ too implementation based.  Let your step definitions identify the names `arm_code` and `disarm_code`.  Your feature shouldn't know that.

overlord/features/configure.feature
```diff
@@ -0,0 +1,37 @@
+Feature: Booting the bomb
+  As a super villain
+  I want to configure the bomb
+  So I can blow stuff up
```
djkotowski: Good format here.

overlord/bomb.rb
jmtai: Yeah, this is questionable.  Since the problem description states that entering the correct activation code after the bomb has been activated should do nothing, I'm returning true here to indicate that the bomb is still armed.  I should probably handle that special case outside of this method so the logic stays consistent with the disarm() method.

overlord/bomb.rb
djkotowski: It is un-Rubylike.  Also, using the `[]` method means an invalid symbol will set `@state = nil`.  `Hash#fetch` will raise an error instead.

overlord/features/home_page.feature
```diff
@@ -0,0 +1,63 @@
...
+  As a super villain
+  I want to see the home page of the bomb app
+  So I can operate the bomb
+
+  Scenario: View bomb status
+    Given The bomb has been booted
+    When I visit the home page 
+    Then I should see "Bomb is currently: INACTIVE"
+
+  Scenario: View bomb control form
```
djkotowski: No need for a view test.

overlord/features/home_page.feature
```diff
@@ -0,0 +1,63 @@
+Feature: Using the bomb
```
djkotowski: This feature is a combination of multiple features (activating, deactivating, exploding, etc).  By splitting up the feature into multiple features, you'll be able to navigate the scenarios more easily.  You'll also be able to use `Background` to define `Given` steps used throughout all scenarios.

overlord/features/step_definitions/home_page.rb
```diff
@@ -0,0 +1,60 @@
...
+  submit_incorrect_deactivation_code
+end
+
+When(/^I submit an incorrect deactivation code two times$/) do
+  2.times { submit_incorrect_deactivation_code }
+end
+
+When(/^I submit an incorrect deactivation code three times$/) do
+  3.times { submit_incorrect_deactivation_code }
+end
```
djkotowski: You can provide an argument to combine the 3 steps above.
Example:
```ruby
When(/^I submit an incorrect deactivation code ([0-9]+) times?$/) do |count|
# Matches "1 time", "2 times", etc.
```

overlord/features/step_definitions/home_page.rb
```diff
@@ -0,0 +1,60 @@
...
+
+def submit_correct_activation_code
+  fill_in('code', :with => '1234')
+  click_button('submit')
+end
+
+def submit_incorrect_deactivation_code
+  fill_in('code', :with => '9999')
+  click_button('submit')
+end
```
djkotowski: Move your methods to the top of the definitions file.

overlord/features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,219 @@
+# Taken from the cucumber-rails project.
```
djkotowski: *Don't* use web steps.  They encourage [poor scenario-writing](http://aslakhellesoy.com/post/11055981222/the-training-wheels-came-off).

overlord/overlord.rb
```diff
@@ -1,14 +1,111 @@
...
+    @disarm_code_default = Bomb.disarm_code_default
+  end
+
+  def codes_valid?(arm_code, disarm_code)
+    Bomb.code_valid?(arm_code) && Bomb.code_valid?(disarm_code)
+  end
+
+  def set_home_page_vars
+    bomb = session[:bomb]
+    @bomb_state = bomb.state
```
djkotowski: Just pass the bomb to the view.

overlord/overlord.rb
```diff
@@ -1,14 +1,111 @@
...
+    @bomb_state_class = @bomb_state.downcase
```
djkotowski: The view or view helper should handle the CSS class, not the "controller".

overlord/features/home_page.feature
```diff
@@ -0,0 +1,63 @@
...
+  As a super villain
+  I want to see the home page of the bomb app
+  So I can operate the bomb
+
+  Scenario: View bomb status
+    Given The bomb has been booted
```
djkotowski: As a nitpicky nitpick, don't capitalize "The" here.  Step descriptions should read as a full sentence starting with `Given`, `When`, `Then` or `And`.

overlord/features/step_definitions/configure.rb
```diff
@@ -0,0 +1,47 @@
+Given(/^I have not booted the bomb$/) do
+  # not sure if this is kosher but I'm not sure how to test this condition
+  true
+end
```
djkotowski: Don't test Givens.  If you're not doing anything in the step, delete the step.

overlord/features/step_definitions/home_page.rb
```diff
@@ -0,0 +1,60 @@
...
+  expect(page).to have_selector('input')
+end
+
+Then(/^I should see a submit button$/) do
+  expect(page).to have_selector('input[type=submit]')
+end
+
+Given(/^The bomb has been configured with default codes$/) do
+  true
+end
```
djkotowski: If nothing happens, get rid of the step.

overlord/overlord.rb
```diff
@@ -1,14 +1,111 @@
...
+
+  def process_code(code)
+    bomb = session[:bomb]
+    # ignore empty codes
+    # ignore the correct activation code after the bomb is armed
+    return true if code.strip == "" || (bomb.armed? && code == bomb.arm_code)
+    return bomb.arm(code) if bomb.inactive?
+    return bomb.disarm(code) if bomb.armed?
+    false
+  end
```
djkotowski: `process_code` really belongs in the bomb

djkotowski:
This is looking good.  Just a few issues to be resolved.  Take a look at the comments, refactor, and please resubmit.  Let me know if you have any questions.

PR #575, opened 2015-03-20
Supportive

jfinzel:


supportive/supportive.rb
```diff
@@ -38,68 +34,39 @@ def to_s
...
-      else "th"
-    end
-    "You will be the #{comments.length}#{ordinal} commenter"
+    return '' unless comments_allowed? && comments.length > 0
+    "You will be the #{comments.length.ordinalize} commenter"
   end
 
   def comments_allowed?
-    publish_date + (365 * 3) > Date.today
+    publish_date + 3.years > Date.today
```
incazteca: You can use the `ago` method on years to have the following statement: `publish_date > 3.years.ago`

supportive/supportive.rb
```diff
@@ -38,68 +34,39 @@ def to_s
...
-    string = String(string)
-    words = string.gsub('_', ' ').split(' ')
+  def categories_label
+    "Category".pluralize(categories.length)
+  end
 
-    words.map!(&:capitalize)
-    words.join(' ')
+  def categories_title
+    categories.map { |cat| String(cat).titleize }.to_sentence
```
incazteca: There's no need to use `map` here. You can have categories use the `to_sentence` method and then chain `titleize` onto it. `categories.to_sentence.titleize` will give the same output you have here.

incazteca:
Looks good. Just a few minor things but no resubmission is needed.

PR #576, opened 2015-03-23
Overlord

cnorrisjr:
Revised based on Franco and Dan's CR.


overlord/features/step_definitions/general.rb
```diff
@@ -0,0 +1,72 @@
+
+def submit
+  find("input[value='Submit']").click
+end
+
+def content_check(text)
+  expect(page).to have_content(text)
+end
+
+Given(/^I am on the home page$/) do
```
incazteca: Should probably have the step name changed to `I am on the config page`

overlord/features/step_definitions/general.rb
```diff
@@ -0,0 +1,72 @@
...
+  fill_in 'deact', with: deact
+  submit
+end
+
+When(/^I use the default codes$/) do
+  submit
+end
+
+Then 'I land on the Activate page' do
+  find(:xpath, "//td[contains(text(),'Activation Code')]")
```
incazteca: Is there a reason you're using xpath here? Also are you just checking if the activation code is in the text box? There isn't a reason to test if a text box can hold text.

overlord/features/step_definitions/general.rb
```diff
@@ -0,0 +1,72 @@
...
+  fill_in 'activation', with: act_code
+  submit
+end
+
+Then(/^My bomb is activated and I have '(\d+)' attempts to disarm$/) do |count|
+  expect(page).to have_content("You have #{count} attempts")
+end
+
+When(/^I enter '(\d+)' as my deact code '(\d+)' times$/) do |de_act, times|
+  1.upto(times.to_i) do
```
incazteca: Rename your variable `times` to something else like attempts. Then you can have the following for setting the loop: `attempts.times do`

overlord/features/bad_activation_code.feature
```diff
@@ -0,0 +1,17 @@
+Feature: Evil mastermind configs a bomb, but enters bad data in the activation field and bomb is not activated.
+Whomp whomp
+
+Scenario: Mastermind Lands on the main page and enters default values
+and puts the wrong activation code in
+  Given I am on the home page
```
incazteca: These first three steps can be consolidated to a step `Given I am on activate page with default codes`

overlord/general.rb
```diff
@@ -0,0 +1,72 @@
+
```
incazteca: Baleet this.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  def checkbomb(expected, my_bomb)
+    expect(my_bomb.status).to eql(expected[:status])
+    expect(my_bomb.activate_code).to eql(expected[:activate])
+    expect(my_bomb.deactivate_code).to eql(expected[:deactivate])
+  end
+
+  context 'Create a bomb' do
+    it 'With the default values' do
+      my_bomb = Bomb.new
```
incazteca: Set `Bomb` with a let block so that you don't have to instantiate it here.

incazteca:
@cnorrisjr There are a few small fixes to be made but another pull is not needed.

overlord/features/config_act_explodes.feature
```diff
@@ -0,0 +1,20 @@
+Feature: Evil mastermind wants to config a bomb,
+activate the bomb, and sets the bomb off
```
djkotowski: This is a long feature name that looks like three in one.  Simplify the feature to better organize your tests.

overlord/features/deactivate_with_custom.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Evil mastermind wants to config a bomb,
+activate the bomb, and deactivate the bomb
```
djkotowski: This has the same feature name as config_act_explodes.feature

overlord/features/bad_activation_code.feature
```diff
@@ -0,0 +1,17 @@
+Feature: Evil mastermind configs a bomb, but enters bad data in the activation field and bomb is not activated.
+Whomp whomp
```
djkotowski: This feature reads like a scenario.  A feature is a story about what the bomb is expected to do.  This could be "Activate the bomb" or "Prevent unauthorized activations".

Also, please use the format
```cucumber
As a <role>
I want to <action>
In order to <reason>
```
to identify and focus your feature.

overlord/general.rb
djkotowski: Follow step definition convention.

`bomb_steps.rb` maybe?

overlord/models/bomb.rb
```diff
@@ -0,0 +1,42 @@
...
+    @status = 'inactive'
+    @attempts = 0
+  end
+
+  def activate(user_inputted_code)
+    if @activate_code.eql?(user_inputted_code) && @status.eql?('inactive')
+      @status = 'active'
+    else
+      @status
+    end
```
djkotowski: You could get rid of the `else` branch since you're always returning the latest value of `@status`.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+      checkbomb(expected, my_bomb)
+    end
+
+    it 'With default activation code and deactivate code' do
+      my_bomb = Bomb.new(deactivate_code: '1234')
+      expected = { status: 'inactive',
+                   activate: '1234',
+                   deactivate: '1234' }
+      checkbomb(expected, my_bomb)
+    end
```
djkotowski: L20 and L28 are identical (except for the capitalization), yet the expected result is different.  What's going on?

Format your `it` statements as if you are describing whatever is in the `describe` statement.  If it's a class, `it` should be a statement describing the class.  If it is a method, describe the method or method output.  Don't treat `it` as a command, treat it as the beginning of your description.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+
+describe Bomb do
+
+  def checkbomb(expected, my_bomb)
+    expect(my_bomb.status).to eql(expected[:status])
+    expect(my_bomb.activate_code).to eql(expected[:activate])
+    expect(my_bomb.deactivate_code).to eql(expected[:deactivate])
+  end
+
+  context 'Create a bomb' do
```
djkotowski: Treat contexts as a precondition, not a test categorizer.

The statement "when entering valid codes" provides context; "create a bomb" does not.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+                   deactivate: '0000' }
+      checkbomb(expected, my_bomb)
+    end
+  end
+
+  context "Deactivate a bomb" do
+    it 'Using the default activation code' do
+      my_bomb = Bomb.new
+      my_bomb.activate('1234')
+      expect(my_bomb.status).to eql('active')
```
djkotowski: This expectation belongs in a test checking if activation works, not if deactivation works.

djkotowski:
While I agree another pull isn't needed, please take a look at the comments and refactor.  Include user stories in your feature (As a ____, in order to _____, I want _____).  Also work on the text describing your tests.  Think about `it`, `context`, and `describe` as verbs beginning a sentence or sentence fragment, not merely commands.

overlord/features/step_definitions/general.rb
cnorrisjr: The test sets the codes into the bomb and will go to the activate page next.  I am checking for the text 'Activation Code to verify I land on the activate page.  XPath is can check for text directly, whereas CSS selectors are a little messy.

overlord/models/bomb.rb
cnorrisjr: @djkotowski 
I have two versions of this written:

First uses a guard clause:
  def activate(user_inputted_code)
    return @status unless @activate_code.eql?(user_inputted_code) && @status.eql?('inactive')
    return @status = 'active'
  end

Second uses the condition statement:
  def activate(user_inputted_code)
    if @activate_code.eql?(user_inputted_code) && @status.eql?('inactive')
     @status = 'active'
    end
  end

I was leaning toward using the guard clause version based on rubocop/style convention.  

Which would be the correct one to use?

cnorrisjr: @djkotowski I rewrote this using a guard close, looks similar to my explode method.

def activate(user_input)
    return @status unless @activate_code.eql?(user_input)
    @status = 'active'
 end

I also wrote another version: 

def activate(user_input)
    if @activate_code.eql?(user_inputted_code) && @status.eql?('inactive')
      @status = 'active'
    end
 end
Would the the guard close be a better implementation than the condition statement?

djkotowski: I think so, though you're always returning `@status`, so I don't know why you need it.  You could also write predicate methods for your statuses, to simplify `@status.eql('inactive')` to `#inactive?`.  Also, why are you using strings instead of symbols for your statuses.  Each time you write a string `'active'` or `'inactive'`, you're creating a new string.  If you use `:active` and `:inactive`, you reuse the same symbol each time.

overlord/features/step_definitions/general.rb
djkotowski: Maybe the HTML/CSS could be reworked a bit to make it less messy?  Fetching a `td` that contains 'Activation Code' and checking its content further seems a bit too confusing and brittle.

PR #580, opened 2015-03-27
Overlord refactor

jmtai:
Cleaned up the code per review.

What changed:
- Moved process_code method out of the controller and into the Bomb
- Moved some view logic out of the controller
- Split the "Using the bomb" feature into separate Activate/Deactivate/Explode features
- Removed web_helpers.rb and split up the spec definitions appropriately
- Other assorted cleanup per the review comments


overlord/features/step_definitions/common.rb
```diff
@@ -0,0 +1,9 @@
+Then(/^I should be on the (.*?) page$/) do |arg1|
```
incazteca: Have your variable names be something other than `arg1`. `arg1` does not give much of a hint what this variable is.

overlord/features/step_definitions/common.rb
```diff
@@ -0,0 +1,9 @@
...
+  current_path = URI.parse(current_url).path
+  page = (arg1 == "home") ? "/" : "/#{arg1}"
+  expect(current_path).to eql(page)
+end
+
+Then(/^I should see "(.*?)"$/) do |arg1|
```
incazteca: Same comment as above. `arg1` could be `value`.

incazteca:
@jmtai Everything looks pretty good however you're missing tests for your bomb class. Be sure to make an rspec test for your bomb class and resubmit.

overlord/bomb.rb
```diff
@@ -0,0 +1,100 @@
+class InvalidCodeError < ArgumentError; end
+
+class Bomb
+  attr_reader :arm_code, :disarm_retries, :state
```
djkotowski: Why is `arm_code` visible but `disarm_code` is not?

overlord/bomb.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  ARM_CODE_DFLT = "1234"
+  DISARM_CODE_DFLT = "0000"
+  DISARM_RETRIES_MAX = 3
+
+  # Possible Bomb states are
+  #  :inactive
+  #  :armed
+  #  :exploded
```
djkotowski: Something to watch out for:  If you add comments describing restrictions on fields without coding validations, you have untested information regarding your class' functionality.  If you (or someone else) adds or removes a possible state, you have no guarantee that this information will be updated to reflect that.

overlord/bomb.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  def initialize(arm_code = ARM_CODE_DFLT, disarm_code = DISARM_CODE_DFLT)
```
djkotowski: Nice use of default values / constants.

overlord/bomb.rb
```diff
@@ -0,0 +1,100 @@
...
+    @state = :inactive
+    @disarm_retries = DISARM_RETRIES_MAX
+  end
+
+  def self.code_valid?(code)
+    code.match(/^[0-9]+$/)
+  end
+
+  def self.arm_code_default
+    ARM_CODE_DFLT
```
djkotowski: You can also access this without class methods.

```ruby
Bomb::ARM_CODE_DFLT
```

overlord/bomb.rb
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  def self.disarm_code_default
+    DISARM_CODE_DFLT
```
djkotowski: Ditto above.

overlord/bomb.rb
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  def process_code(code)
+    # ignore empty codes
+    # ignore the correct activation code after the bomb is armed
```
djkotowski: Similar to my caveat above.  Comments are lies waiting to happen.

overlord/bomb.rb
```diff
@@ -0,0 +1,100 @@
...
+    return true if code.strip == "" || (armed? && code == arm_code)
+    return arm(code) if inactive?
+    return disarm(code) if armed?
+    false
```
djkotowski: Returning true/false is a bit confusing.

overlord/bomb.rb
```diff
@@ -0,0 +1,100 @@
...
+    state == :inactive
+  end
+
+  def armed?
+    state == :armed
+  end
+
+  def exploded?
+    state == :exploded
+  end
```
djkotowski: Good use of predicate methods.  Also makes testing clean and very readable.

overlord/features/activate.feature
```diff
@@ -0,0 +1,26 @@
...
+    When I submit an incorrect activation code
+    Then I should see "Bomb is currently: INACTIVE"
+
+  Scenario: Enter correct activation code
+    Given the bomb has been booted with default codes
+    When I submit the correct activation code
+    Then I should see "Bomb is currently: ARMED"
+
+  Scenario: Enter correct activation code after bomb is activated
+    Given the bomb has been booted with default codes
```
djkotowski: When you reuse `Given` steps for all your scenarios like this, extract them out to a `Background`.

overlord/features/configure.feature
```diff
@@ -0,0 +1,30 @@
...
+    Then I should see "You entered an invalid code"
+
+  Scenario: Boot bomb with default codes
+    Given I am on the configure page
+    When I do not enter any codes
+    And I submit the codes
+    Then I should be on the home page
+
+  Scenario: Boot bomb with custom codes
+    Given I am on the configure page
```
djkotowski: Ditto above comment about `Background`.

overlord/features/step_definitions/deactivate.rb
```diff
@@ -0,0 +1,13 @@
...
+  fill_in('code', :with => '9999')
+  click_button('submit')
+end
+
+When(/^I submit the correct deactivation code$/) do
+  fill_in('code', :with => '0000')
+  click_button('submit')
+end
+
+When(/^I submit an incorrect deactivation code ([0-9]+) times?$/) do |count|
```
djkotowski: Great use of the regex here.

overlord/features/support/paths.rb
```diff
@@ -0,0 +1,35 @@
+# Taken from the cucumber-rails project.
```
djkotowski: Remove generated comments when there's no danger of them being re-generated.

overlord/bomb.rb
jmtai: That's actually leftover code from the previous version.  The only thing that should actually be visible is `disarm_retries`.

djkotowski:
You need your RSpec tests still, but otherwise it looks very good.  Please take a look at the comments, add some unit tests and resubmit.

overlord/bomb.rb
jmtai: Yeah, I agree. This could probably be reworked into a cleaner form.

overlord/features/activate.feature
jmtai: If the `Given` is extracted, do the scenarios need to be rewritten so they still fit the `Given/When/Then` format?

djkotowski: Yes and no.  In the case of this scenario, the `And` should become `Given`.  Any steps that are `When` and `Then` should remain that way, regardless if the `Given` is extracted.

Example:
```cucumber
Scenario: Enter correct activation code
  Given the bomb has been booted with default codes
  When I submit the correct activation code
  Then I should see "Bomb is currently: ARMED"
```
becomes
```cucumber
Background:
  Given the bomb has been booted with default codes

Scenario: Enter correct activation code
  When I submit the correct activation code
  Then I should see "Bomb is currently: ARMED"
```

PR #581, opened 2015-03-27
Avadino

avavita:
I'm not sure why thewillofone.txt is in this request. What? Whatever.

### DIFF
--- a/dino_catalog/Dinodex_Wizard.rb
+++ b/dino_catalog/Dinodex_Wizard.rb
@@ -53,7 +53,7 @@ def self.prompt_user_for_constraints
 end
 
 def self.choose_function
-  puts "Select function to run:\nQUERY\nBIG\nSMALL\n\n"
+  puts "Select function to run:\nQUERY\nBIG\nSMALL\nCARNIVORES\n"
   input = STDIN.gets.strip.downcase
 end
 
@@ -71,6 +71,10 @@ def self.run(files)
     pp dex.get_big_dinos
   when "small"
     pp dex.get_small_dinos
+  when "carnivores"
+    pp dex.get_carnivores
+  else
+    abort("You're not even reading the instructions are you?")
   end
 end
 
--- a/dino_catalog/Dinodex_Wizard.rb
+++ b/dino_catalog/Dinodex_Wizard.rb
@@ -5,24 +5,21 @@ class Dinodex_Wizard
 require_relative 'dinodex'
 require 'pp'
 
-def check_args(args)
+def self.check_args(args)
   if args == []
     abort("Please include CSV files as arguments. Exiting...")
   end
 end
 
-def prompt_user_for_constraint
+def self.prompt_user_for_constraint
 
-  special_commands = ["SMALL","BIG"]
-  joeFormat = ["NAME","PERIOD","CONTINENT","DIET","WEIGHT_IN_LBS","WALKING","DESCRIPTION"]
+  joe_format = ["NAME","PERIOD","CONTINENT","DIET","WEIGHT_IN_LBS","WALKING","DESCRIPTION"]
 
-  puts "Valid fields listed below:"
-  puts joeFormat
-  puts "Please type a field to search and the value to look for, separated by a =. BIG and SMALL are special cases."
-  puts "{examples: name = velociraptor, BIG, SMALL}"
+  puts "Valid fields listed below:\n#{joe_format}"
+  puts "Please type a field to search and the value to look for, separated by ="
+  puts "{ex: name = velociraptor}"
 
   input = STDIN.gets.split("=", 2)
-  return input if special_commands.include? input[0].strip
 
   if input.count != 2
     abort("You aren't paying attention to the format at all. I bet you don't even like dinosaurs. Get out of here.")
@@ -32,13 +29,13 @@ def prompt_user_for_constraint
     x.strip!
   end
 
-  unless joeFormat.include? input[0]
+  unless joe_format.include? input[0]
     abort("That's not a valid field. I bet you don't even like dinosaurs. Get out of here.")
   end
   return input
 end
 
-def prompt_user_for_constraints
+def self.prompt_user_for_constraints
 
   filter_further = true
   constraints = []
@@ -47,7 +44,7 @@ def prompt_user_for_constraints
     puts "add a constraint? (y/n)"
     input = STDIN.gets.strip.downcase
     if input == "y" 
-      constraints += prompt_user_for_constraint
+      constraints << prompt_user_for_constraint
     else
       filter_further = false
     end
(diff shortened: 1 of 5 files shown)


djkotowski:
`thewillofone.txt` is in because you probably created this branch off of the branch `thewillofone.txt` was committed to.

dino_catalog/Dinodex_Wizard.rb
```diff
@@ -0,0 +1,83 @@
...
+end
+
+def self.run(files)
+  check_args(files)
+  dex = Dinodex.new(files)
+  pp dex.array_of_dinos
+  pp "Here's the entire set."
+  choice = choose_function
+  case choice
+  when "query"
```
avavita: the github ruby style guide disagrees, hound.
https://github.com/styleguide/ruby

Indent when as deep as case.

case
when song.name == "Misty"
  puts "Not again!"
when song.duration > 120
  puts "Too long!"
when Time.now.hour > 21
  puts "It's too late"
else
  song.play
end

djkotowski: `case` and `when` should be at same level.  Contents of `when` should be indented.
```ruby
case
when condition
  do_stuff
when other_condition
  do_stuff
else
  do_stuff
end
```

dino_catalog/Dinodex_Wizard.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  joe_format = ["NAME","PERIOD","CONTINENT","DIET","WEIGHT_IN_LBS","WALKING","DESCRIPTION"]
+
+  puts "Valid fields listed below:\n#{joe_format}"
+  puts "Please type a field to search and the value to look for, separated by ="
+  puts "{ex: name = velociraptor}"
+
+  input = STDIN.gets.split("=", 2)
+
+  if input.count != 2
```
zbright: above, you limit the split to 2. I understand that you need this check to be sure there are no less than two, but it just feels redundant.

Also, in this case using an unless would be more readable imo.

`abort(message) unless input.count == 2`

dino_catalog/Dinodex_Wizard.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  input.each do |x|
+    x.strip!
+  end
+
+  unless joe_format.include? input[0]
+    abort("That's not a valid field. I bet you don't even like dinosaurs. Get out of here.")
+  end
+  return input
```
zbright: I would flip these around
```
return input if joe_format.include? input[0]
about(message)
```

dino_catalog/Dinodex_Wizard.rb
```diff
@@ -0,0 +1,83 @@
...
+def self.prompt_user_for_constraints
+
+  filter_further = true
+  constraints = []
+
+  while filter_further
+    puts "add a constraint? (y/n)"
+    input = STDIN.gets.strip.downcase
+    if input == "y" 
+      constraints << prompt_user_for_constraint
```
zbright: having methods with names as similar as these is rough to read. I spent a minute trying to figure out the recursion before I realized it was calling a different method

dino_catalog/Dinodex_Wizard.rb
```diff
@@ -0,0 +1,83 @@
...
+    else
+      filter_further = false
+    end
+  end
+  return constraints
```
zbright: no need to use return here. Ruby magic allows for just `constraints`

zbright: Also, since this is the only thing happening after you break from the loop, why not just move the return in and lose the `filter_further` var.

dino_catalog/Dinodex_Wizard.rb
```diff
@@ -0,0 +1,83 @@
...
+end
+
+def self.choose_function
+  puts "Select function to run:\nQUERY\nBIG\nSMALL\nCARNIVORES\n"
+  input = STDIN.gets.strip.downcase
```
zbright: Have to agree with hound here, was there some reason you left the variable assignment?

dino_catalog/Dinodex_Wizard.rb
avavita: alright I'm good with that

dino_catalog/Dinodex_Wizard.rb
avavita: left over from a previous refactor, really. Just slipped my mind. I've clean up a lot of my returns in my current iteration.

zbright: Do you want to submit those changes?

dino_catalog/Dinodex_Wizard.rb
```diff
@@ -0,0 +1,83 @@
...
+
+def self.check_args(args)
+  if args == []
+    abort("Please include CSV files as arguments. Exiting...")
+  end
+end
+
+def self.prompt_user_for_constraint
+
+  joe_format = ["NAME","PERIOD","CONTINENT","DIET","WEIGHT_IN_LBS","WALKING","DESCRIPTION"]
```
djkotowski: This should be a constant.

dino_catalog/Dinodex_Wizard.rb
djkotowski: Also note the redundant `return`.

dino_catalog/Dinodex_Wizard.rb
avavita: working on it. Still cleaning a few things. soon.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,46 @@
...
+      return true if @continent.include? value
+    when "DIET"
+      return true if @diet.include? value
+    when "WEIGHT_IN_LBS" # ...I guess if you really want to string match on a number, sure.
+      return true if @weight_in_lbs.include? value
+    when "WALKING"
+      return true if @walking.include? value
+    when "DESCRIPTION"
+      return true if @description.include? value
+    end
```
djkotowski: `field` matches the name of the reader, just with different capitalization.  Why not take advantage of that?  You can verify that `field` is a valid option, downcase `field`, and use `#send` to retrieve the value.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,46 @@
...
+    return false
+  end
+
+  def heavier_than?(pounds)
+    return false if @weight_in_lbs.nil? # in case they send a negative number that would succeed against nil.to_i (zero)
+    return @weight_in_lbs.to_i > pounds.to_i ? true : false
```
djkotowski: This equates to
```ruby
if condition is true
  return true
else # if condition is false
  return false
end
```

dino_catalog/dino.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def heavier_than?(pounds)
+    return false if @weight_in_lbs.nil? # in case they send a negative number that would succeed against nil.to_i (zero)
+    return @weight_in_lbs.to_i > pounds.to_i ? true : false 
+  end
+
+  def lighter_than?(pounds)
+    return false if @weight_in_lbs.nil? # again, because nil.to_i == 0
+    return @weight_in_lbs.to_i < pounds.to_i ? true : false
```
djkotowski: Ditto above.

djkotowski:
Take a look at the comments, finish cleaning up, and resubmit.  @zbright will review after.

dino_catalog/dino.rb
avavita: oh wow. #send is cool.

dino_catalog/Dinodex_Wizard.rb
avavita: well that took longer than expected. new pull req is up.

PR #582, opened 2015-03-30
Overlord

wyaeiga:


incazteca:
@wyaeiga Can you provide an updated Gemfile? I am unable to get the tests to run.

overlord/bomb.rb
```diff
@@ -0,0 +1,76 @@
+class Bomb
+  attr_accessor :status, :deactivation_attempts, :activation_code, :deactivation_code
+
+  ACTIVATION_CODE = '1234'
+  DEACTIVATION_CODE = '0000'
+  INACTIVE = 0
```
incazteca: Statuses are internally used as integers. However I see them used as symbols outside of this class. Pick a type and remain consistent.

incazteca:
@wyaeiga Fix up the issue regarding the status as we earlier discussed. However there is no need for resubmission.

overlord/bomb.rb
djkotowski: I _highly_ recommend using symbols instead of integers.  It's more ruby-like.

overlord/features/initialize_bomb.feature
```diff
@@ -0,0 +1,18 @@
...
+
+  Scenario: initializing bomb
+    Given there is currently no bomb
+    When I enter an activation code in the activation field
+    And I enter a deactivation code in the deactivation field
+    And I click on the initialize button
+    Then I should be redirected to a page with an inactive bomb
+
+  Scenario: initializing bomb with invalid codes
+    Given there is currently no bomb
```
djkotowski: This should be in a `Background`.

overlord/features/step_definitions/deactivate_bomb_steps.rb
```diff
@@ -0,0 +1,18 @@
+And(/^it has been activated$/) do
```
djkotowski: Don't use `And` in step definitions.  Stick with `Given`, `When`, or `Then`.  `And` is used in features only to avoid saying
```cucumber
Given ...
Given ...
When ...
When ...
Then ...
```

overlord/features/step_definitions/deactivate_bomb_steps.rb
```diff
@@ -0,0 +1,18 @@
...
+  step 'I click on the activate button'
+  step 'I should be redirected to a page with an active bomb'
+end
+
+And(/^I click on the deactivate button$/) do
+  click_button('deactivate')
+end
+
+When(/^I deactivate the bomb incorrectly three times$/) do
+  step 'I enter an incorrect deactivation code in the code field'
```
djkotowski: It's bad form to call other steps within a step.  If you're reusing logic, make a method.

overlord/views/_bomb_keypad.haml
```diff
@@ -0,0 +1,8 @@
+.bomb-keypad
+  - 10.times do |button|
+    - button_value = (button+1)%10
+    - button_id = "button_"+button_value.to_s
+    - if keypad_enabled
+      %button.bomb-button{id: button_id} #{button_value}
+    - else
+      %button.bomb-button.disabled{id: button_id} #{button_value}
```
djkotowski: This file is nice and clean.  I like it.

overlord/overlord.rb
```diff
@@ -1,14 +1,39 @@
...
+require 'json'
 require 'sinatra'
+require 'haml'
+require './bomb'
 
 enable :sessions
 
 get '/' do
-  "Time to build an app around here. Start time: " + start_time
+  if session[:bomb]
```
djkotowski: Extract `session[:bomb]` out to a method.  `#bomb` maybe?

overlord/overlord.rb
```diff
@@ -1,14 +1,39 @@
...
+    if session[:bomb].detonated?
+      haml :index, locals: { bomb_view: :_bomb_detonated, default_codes: nil }
```
djkotowski: Why not just pass the bomb to the view, and let the view decide what it does and does not need?

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    end
+
+    it 'defaults to 1234/0000 as codes when provided invalid codes' do
+      bomb = Bomb.new(invalid_code, invalid_code)
+      expect(bomb.activation_code).to eql(correct_activation_code)
+      expect(bomb.deactivation_code).to eql(correct_deactivation_code)
+    end
+
+    it 'sets a flag to let user know if the bomb is using default codes' do
+      expect(bomb.default_codes?).to eql(true)
```
djkotowski: Avoid comparing to `true`.  Use expectations for predicate methods when possible.

For example, if the method were `using_default_codes?`, you could use the syntax
```ruby
expect(bomb).to be_using_default_codes
```

djkotowski: Otherwise,
```ruby
expect(bomb.default_codes?).to be true
```
still works better.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    end
+
+    it 'allows custom activation and deactivation codes' do
+      bomb = Bomb.new(custom_activation_code, custom_deactivation_code)
+      expect(bomb.activate(custom_activation_code)).to be(true)
+      expect(bomb.deactivate(custom_deactivation_code)).to be(true)
+    end
+
+    it 'does not allow activation and deactivation codes shorter than 4 characters' do
+      expect(bomb.activation_code).to eql(correct_activation_code)
```
djkotowski: This test is confusing.  It doesn't look like the test description and the process match.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      expect(bomb.deactivation_code).to eql(correct_deactivation_code)
+    end
+  end
+
+  context 'when inputting codes' do
+    it 'activates when activation code is entered' do
+      activation = bomb.activate(correct_activation_code)
+      expect(activation).to be(true)
```
djkotowski: Is this assertion necessary?  Won't the line below provide the correct assertion for this test?

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    end
+
+    it 'does nothing when deactivation code is entered if bomb not active' do
+      deactivation = bomb.deactivate(correct_deactivation_code)
+      expect(deactivation).to be(false)
+
+      deactivation = bomb.deactivate(incorrect_deactivation_code)
+      expect(deactivation).to be(false)
+
+      expect(bomb.deactivation_attempts).to be(0)
```
djkotowski: Use `eq` for numeric comparisons.  In this case, I might prefer `be_zero`.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+  let (:correct_activation_code) { '1234' }
+  let (:correct_deactivation_code) { '0000' }
+  let (:incorrect_activation_code) { '1235' }
+  let (:incorrect_deactivation_code) { '0001' }
+  let (:custom_activation_code) { '4321' }
+  let (:custom_deactivation_code) { '1111' }
+  let (:invalid_code) { 'asdf' }
+  let (:inactive) { :inactive }
+  let (:active) { :active }
+  let (:detonated) { :detonated }
```
djkotowski: L12 - L14 provide an unnecessary layer of abstraction.  Just use the symbols throughout.  If the symbols change, so should the documentation (specs).

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+
+      deactivation = bomb.deactivate(incorrect_deactivation_code)
+      expect(deactivation).to be(false)
+
+      expect(bomb.deactivation_attempts).to be(0)
+    end
+
+    it 'explodes when deactivation code is entered incorrectly 3 times' do
+      bomb.activate(correct_activation_code)
+      expect(bomb.status).to be(active)
```
djkotowski: You repeat this same expectation frequently.  Have one test to check this.  Otherwise, if this expected functionality ever changes, you have to alter all of your tests.

djkotowski:
Avoid repeating the same assertions in your tests.  What is described in the `it` portion of your test is what you should test.  If an assertion isn't described there, it should be in a different test.

djkotowski:
This is looking good.  I would look at refactoring some of the tests.  Use `describe` when describing methods in RSpec to make the functionality of your classes more clear.

I think only one more resubmission is necessary.

PR #583, opened 2015-03-30
Avadino

avavita:
...fixed a billion things.


dino_catalog/Dinodex_Wizard.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    input = STDIN.gets.split("=", 2)
+
+    abort("You aren't paying attention") unless input.count == 2
+
+    input.each do |x|
+      x.strip!
+    end
+
+  	return input if SUPPORTED_FIELDS.include? input[0]
```
avavita: oh damn it sublime

dino_catalog/Dinodex_Wizard.rb
```diff
@@ -0,0 +1,76 @@
...
+  def self.prompt_user_for_constraint
+    puts "Valid fields listed below:\n#{SUPPORTED_FIELDS}"
+    puts "Please type a field to search and the value to look for, separated by ="
+    puts "{ex: name = velociraptor}"
+
+    input = STDIN.gets.split("=", 2)
+
+    abort("You aren't paying attention") unless input.count == 2
+
+    input.each do |x|
```
zbright: Good idea hound!

dino_catalog/dinoConverter.rb
```diff
@@ -0,0 +1,52 @@
...
+  "Weight" => "WEIGHT_IN_LBS", 
+  "Walking" => "WALKING"
+  }
+
+  def self.csv_file_to_dino_array(file)
+    csv = CSV::parse(File.open(file, 'r') {|f| f.read })
+    fields = csv.shift
+    hash_array = csv.map { |record| Hash[*fields.zip(record).flatten] }
+    fixed_headers_array = hash_array.map { |hash| fix_pirate_hash(hash) }
+    dino_array = []
```
zbright: check out `each_eith_object` for something like this

dino_catalog/dinoConverter.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+    dino_array
+  end
+
+  def self.csv_files_to_dino_array(files)
+    abort("No files given") if files == []
+    arrays = files.map do |file|
+      csv_file_to_dino_array(file)
+    end
+    dino_array = []
```
zbright: `each_with_object`

dino_catalog/dinoConverter.rb
```diff
@@ -0,0 +1,52 @@
...
+    arrays.each do |array|
+      dino_array += array
+    end
+    dino_array
+  end
+
+  def self.fix_pirate_hash(hash)
+    if hash.keys.include? "Genus"
+      fixed_h = Hash[hash.map {|k, v| [MAPPINGS[k], v] }]
+      case fixed_h["DIET"]
```
zbright: this case seems to be over kill. Maybe a ternary instead?

`fiexed_h['DIET'] == 'Yes' ? 'Carnivore' : 'Unknown'`

dino_catalog/dinoConverter.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+    dino_array = []
+    arrays.each do |array|
+      dino_array += array
+    end
+    dino_array
+  end
+
+  def self.fix_pirate_hash(hash)
+    if hash.keys.include? "Genus"
```
zbright: invert this if with a return to simplify. 
`return hash unless hash.keys.include? "Genus"`

dino_catalog/dino.rb
```diff
@@ -0,0 +1,32 @@
...
+    @period = attr_hash["PERIOD"]
+    @continent = attr_hash["CONTINENT"]
+    @diet = attr_hash["DIET"]
+    @weight_in_lbs = attr_hash["WEIGHT_IN_LBS"]
+    @walking = attr_hash["WALKING"]
+    @description = attr_hash["DESCRIPTION"]
+  end
+
+  def like?(field, value)
+    if respond_to? field.downcase
```
zbright: in general, if you find yourself returning from an if statement its best to invert it and have an explicit return

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,49 @@
...
+  require_relative 'dinoConverter'
+  require 'pp'
+
+  BIG = 4000
+
+  def initialize(files)
+    @array_of_dinos = DinoConverter.csv_files_to_dino_array(files)
+  end
+
+  def array_of_dinos
```
zbright: just make this variable readable - let ruby do the magic :)

`attr_reader array_of_dinos`

dino_catalog/dino.rb
```diff
@@ -0,0 +1,32 @@
...
+      (send field.downcase).include? value
```
zbright: Why use include here? I may just be tired, but from what I can tell, this is calling one of the variables defined in the initializer. In which case, there would be only one value returned? 

Also, comparing strings without (at least) downcasing scares me

dino_catalog/dinoConverter.rb
```diff
@@ -0,0 +1,52 @@
...
+      fixed_h = Hash[hash.map {|k, v| [MAPPINGS[k], v] }]
+      case fixed_h["DIET"]
+        when "Yes"
+          fixed_h["DIET"] = "Carnivore"
+        else
+          fixed_h["DIET"] = "Unknown"
```
zbright: Instead of unknown, go with 'Herbivore'. Its just one of of the Joe-isms

dino_catalog/dino.rb
zbright: This will fail if you query on description since african dinos dont have that field

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,49 @@
...
+  def big_dinos
+    @array_of_dinos.select {|dino| dino.heavier_than?(BIG)}
+  end
+
+  def small_dinos
+    @array_of_dinos.select {|dino| dino.lighter_than?(BIG)}
+  end
+
+  def carnivores
+    carnivores = @array_of_dinos.select { |dino|
```
djkotowski: Unnecessary assignment.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def small_dinos
+    @array_of_dinos.select {|dino| dino.lighter_than?(BIG)}
+  end
+
+  def carnivores
+    carnivores = @array_of_dinos.select { |dino| 
+      dino.like?("DIET", "Carnivore") || 
+      dino.like?("DIET", "Piscivore") || 
+      dino.like?("DIET", "Insectivore")
```
djkotowski: Could simplify to
```ruby
["Carnivore", "Piscivore", "Insectivore"].any? { |diet| dino.like?("DIET", diet) }
```

dino_catalog/Dinodex_Wizard.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def self.prompt_user_multiple_constraints(constraint_array = [])
+    constraint_array << prompt_user_for_constraint
+    puts "add a constraint? (y/n)"
+    input = STDIN.gets.strip.downcase
+    if input == "y"
+      prompt_user_multiple_constraints(constraint_array)
+    else
+      constraint_array
+    end
```
djkotowski: Use guard clauses to shrink this.

dino_catalog/dinoConverter.rb
djkotowski: Don't name variables after the data type.  Name them after what they represent.

dino_catalog/dinodex.rb
djkotowski:
```ruby
attr_reader :array_of_dinos
```

djkotowski:
Chaining queries would be difficult here.  Instead of having Dinodex's filters return an array of filtered dinosaurs, why not have it return a new, filtered Dinodex?

djkotowski:
Good job using `Enumerable` methods!  You'll find those will save you a lot of trouble.
Please look at the comments, refactor and resubmit.  Good work!

PR #584, opened 2015-03-30
| is good > is bad

avavita:


arrowhead/arrowhead.rb
```diff
@@ -15,19 +15,19 @@ class Arrowhead
...
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
+    unless CLASSIFICATIONS.include? region
       raise "Unknown region, please provide a valid region."
     end
+
+    shapes = CLASSIFICATIONS[region]
+
```
djkotowski: As `Hash#[]` returns `nil` when the key is missing, you could make the assignment, and then raise an error `unless region`.

djkotowski:
This is good.  No need for resubmission.

PR #587, opened 2015-04-01
Sometimes I feel like the only real programming lesson is componentizing things.

avavita:
Also, componentizing should be a word.


robot_name/name_generator.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def self.generate_char
+    ('A'..'Z').to_a.sample
+  end
+
+  def self.generate_num
+    rand(10)
+  end
+
+  def self.generate_name
```
paulh16: Consider trying to generate the random name with `inject`. Something like this:

```
  def generate_name
    "#{random_letter_sequence(2)}#{random_number_sequence(3)}"
  end

  def random_letter_sequence(number_of_characters)
    number_of_characters.times.inject('') { |string, iteration| string << CHARACTERS.sample }
  end

  def random_number_sequence(number_of_digits)
    number_of_digits.times.inject('') { |string, iteration| string << DIGITS.sample }
  end
```

robot_name/name_generator.rb
```diff
@@ -0,0 +1,49 @@
...
+      name = generate_name
+      attempts += 1
+      raise NameRegistryError,
+        "Unique name cannot be generated." if attempts == max_attempts
+    end
+
+    name
+  end
+
+  def self.check_name_format(name)
```
paulh16: This method can be made private.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,49 @@
+class NameGenerator
+  # this should be the only class that cares about the registery
+  require_relative 'robot_registery'
+  require_relative 'robot_errors'
+
+  REGISTERY = RobotRegistery.new
+  ROBOT_NAME_FORMAT = /[[:alpha:]]{2}[[:digit:]]{3}/
+  UNIQUE_NAME_MAX_ATTEMPTS = 10
+
+  def self.generate_char
```
paulh16: This method can be made private, as can `generate_num` and `generate_name`.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,49 @@
...
+    name
+  end
+
+  def self.check_name_format(name)
+    unless name =~ ROBOT_NAME_FORMAT
+      raise NameFormatError, "The robot's name sucks! (#{name})"
+    end
+  end
+
+  def self.call
```
paulh16: Consider creating an instance of the Name Generator class with an `initialize` method. Using `call` confuses the `call` method on a Proc.

robot_name/robot_registery.rb
```diff
@@ -0,0 +1,20 @@
+class RobotRegistery
```
paulh16: Nice separation of classes to adhere to the Single Responsibility Principle!

robot_name/robot_registery.rb
```diff
@@ -0,0 +1,20 @@
...
+  attr_reader :names
+
+  def initialize(names = [])
+    @names = names
+  end
+
+  def add_name(name)
+    if !@names.include? name
```
paulh16: You can use `unless` here.

paulh16: This is very nit-picky, but you use two different styles: `method(parameter)` and `method parameter`. I'd suggest sticking with one.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,49 @@
...
+  def self.generate_num
+    rand(10)
+  end
+
+  def self.generate_name
+    "#{generate_char}#{generate_char}"+ 
+      "#{generate_num}#{generate_num}#{generate_num}"
+  end
+
+  def self.generate_unique_name(max_attempts)
```
paulh16: Is this method doing too much? Can you extract a separate method from it?

robot_name/name_generator.rb
```diff
@@ -0,0 +1,49 @@
...
+  def self.generate_char
+    ('A'..'Z').to_a.sample
+  end
+
+  def self.generate_num
+    rand(10)
+  end
+
+  def self.generate_name
+    "#{generate_char}#{generate_char}"+
```
djkotowski: Keep in mind also if you decide to continue using interpolation here, use `<<` to append instead of `+`.  `+` will create three strings (the two strings being combined, then the result), whereas `<<` modifies the original string.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,49 @@
...
+  # this should be the only class that cares about the registery
+  require_relative 'robot_registery'
+  require_relative 'robot_errors'
+
+  REGISTERY = RobotRegistery.new
+  ROBOT_NAME_FORMAT = /[[:alpha:]]{2}[[:digit:]]{3}/
+  UNIQUE_NAME_MAX_ATTEMPTS = 10
+
+  def self.generate_char
+    ('A'..'Z').to_a.sample
```
paulh16: See below; this can be a constant.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,49 @@
+class NameGenerator
+  # this should be the only class that cares about the registery
+  require_relative 'robot_registery'
+  require_relative 'robot_errors'
+
+  REGISTERY = RobotRegistery.new
```
djkotowski: Misspelling.  `REGISTRY`

robot_name/robot_registery.rb
```diff
@@ -0,0 +1,20 @@
+class RobotRegistery
+  attr_reader :names
+
+  def initialize(names = [])
+    @names = names
+  end
+
+  def add_name(name)
```
paulh16: Why is `add_name` returning true or false. It should add the name. When it returns false, it doesn't actually do anything.

robot_name/robot_registery.rb
```diff
@@ -0,0 +1,20 @@
...
+    if !@names.include? name
+      @names << name
+      true
+    else
+      false
+    end
+  end
+
+  def contains?(name)
+    @names.include? name
```
djkotowski: Wrap arg with parens.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,49 @@
...
+  ROBOT_NAME_FORMAT = /[[:alpha:]]{2}[[:digit:]]{3}/
```
paulh16: Consider restricting this regex further: `/^[[:alpha:]]{2}[[:digit:]]{3}$/`

robot_name/name_generator.rb
```diff
@@ -0,0 +1,49 @@
...
+    unless name =~ ROBOT_NAME_FORMAT
+      raise NameFormatError, "The robot's name sucks! (#{name})"
+    end
+  end
+
+  def self.call
+    name = generate_unique_name(UNIQUE_NAME_MAX_ATTEMPTS)
+    check_name_format(name)
+    REGISTERY.add_name(name)
+    name
```
djkotowski: You can avoid this step by using `#tap`.
```ruby
generate_unique_name(UNIQUE_NAME_MAX_ATTEMPTS).tap do |name|
  check_name_format(name)
  REGISTRY.add_name(name)
end
```

robot_name/robot_name.rb
```diff
@@ -1,32 +1,13 @@
...
 robot = Robot.new
-puts "My pet robot's name is #{robot.name}, but we usually call him sparky."
-
-# Errors!
-# generator = -> { 'AA111' }
-# Robot.new(name_generator: generator)
-# Robot.new(name_generator: generator)
+robot2 = Robot.new
+puts "My pet robots names are #{robot.name} and #{robot2.name}..." +
+  " but I never talk to them."
```
djkotowski: Use `<<` or `\` syntax for concatenating here.

paulh16:
Good start here; take a look at the comments and refactor the code. I liked the separation of classes!

robot_name/robot_registery.rb
avavita: Because adding the name can fail. I'm trying to keep the list of names distinct in the registry. Is there a more rubylike convention to follow here?

djkotowski: You're already ignoring the return value.  Just get rid of it.
```ruby
@names << name unless @names.include?(name)
```

robot_name/name_generator.rb
avavita: mmm I'm just gonna rename the method to 'robot_name' because that's more descriptive anyway. Unless there's a reason why you think I should be instantiating name generators? Currently it only knows constants... so each instantiation would be identical.

robot_name/robot_registery.rb
avavita: I disagree on the principle that the object should be responsible for the integrity of its own data. In this case, I'm more inclined to change the behavior of the generator rather than the behavior of the registry. It's doing work that really should belong to the registry. 

Thoughts...?

avavita: fwiw, anyone else that happens to look at this. We clarified this discussion in person and we're not actually at odds. We both agreed on the responsibilities of the classes, just had different ideas for implementation. I found that returning true/false greatly simplified the messages sent between generator and registry, so I'm going to keep it.

robot_name/name_generator.rb
avavita: I suppose I could do something like this. 

```ruby
    def self.generate_name_and_register
      name = generate_name
      REGISTRY.add_name(name) # returns true/false
    end

    def self.generate_unique_name(max_attempts)
      attempts = 0

      while !generate_name_and_register
        attempts += 1
        raise NameRegistryError,
          "Unique name cannot be generated." if attempts == max_attempts
      end

      name
    end
```

avavita: I can probably separate registration from the retry. Is this odd?

```ruby
    def self.generate_name_and_register
      name = generate_name
      REGISTRY.add_name(name) ? name : nil # is this weird?
    end

    def self.generate_unique_name(max_attempts)
      attempts = 0

      while attempts < max_attempts
        name = generate_name_and_register
        return name unless name.nil?
        attempts += 1
      end

      raise NameRegistryError,"Unique name cannot be generated."
    end
```

PR #589, opened 2015-04-02
Overlord redux

jmtai:
Added rspec tests for Bomb class


incazteca:
Rspec tests look good but be sure to use the right matchers. A resubmission is not needed.

djkotowski:
This looks good.  Take a look at the comments, but no resubmission needed.

PR #590, opened 2015-04-02
Overlord

wyaeiga:


overlord/bomb.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  ACTIVATION_CODE = '1234'
+  DEACTIVATION_CODE = '0000'
+
+  def initialize(activation_code = ACTIVATION_CODE, deactivation_code = DEACTIVATION_CODE)
+    @status = :inactive
+    @deactivation_attempts = 0
+
+    @activation_code = correct_format?(activation_code) ? activation_code : ACTIVATION_CODE
+    @deactivation_code = correct_format?(deactivation_code) ? deactivation_code : DEACTIVATION_CODE
```
jmmastey: /me ninjas in

I don't dig this line. If I provide a poorly formatted deactivation code (say I fat-finger and give you too many digits), you silently set the code to a default value without warning me. Since I thought I set it to something else, I will of course keep trying that code until the bomb explodes in my face.

Better to throw an error and refuse to start up, than to silently do something unexpected and explodey. :bomb: :cry:

wyaeiga: Ayyyyyyy, hey Joe!

Yeah, I see what you mean, currently I have it so that as long as the bomb is ever using default values, the front-end will display in big red letters "using default codes". I know that that's probably also a little bit subpar, but what do you think about it?

incazteca:
Fixes look good. No need for resubmission.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+  context 'when rearming' do
+    it 'resets the deactivation_attempt counter' do
+      bomb.activate(correct_activation_code)
+      bomb.deactivate(incorrect_deactivation_code)
+      bomb.deactivate(correct_deactivation_code)
+      bomb.activate(correct_activation_code)
+      expect(bomb.deactivation_attempts).to be(0)
+    end
+  end
+end
```
djkotowski: I like your use of contexts throughout.

djkotowski:
Great job.  You're done with this one.

PR #591, opened 2015-04-02
Avarobot

avavita:
don't bother commenting on the private method declarations I'm fixing them.


robot_name/name_generator.rb
```diff
@@ -0,0 +1,61 @@
...
+  VALID_NUMBERS = (1..9).to_a
+
+  def self.robot_name
+    generate_unique_name(UNIQUE_NAME_MAX_ATTEMPTS).tap do |name|
+      check_name_format(name)
+      REGISTRY.add_name(name)
+    end
+  end
+
+  private
```
avavita: does this include all methods below it? Or just the one directly below?

avavita: ...dan, hound doesn't like your "blank line below private" rule.

robot_name/robot_registery.rb
```diff
@@ -0,0 +1,20 @@
+class RobotRegistry
```
paulh16: Make sure you delete this file since you're now using `robot_registry.rb`.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,59 @@
+class NameGenerator
+  # this should be the only class that cares about the registery
+  require_relative 'robot_registry'
+  require_relative 'robot_errors'
+
+  REGISTRY = RobotRegistry.new
+  ROBOT_NAME_FORMAT = /^[[:alpha:]]{2}[[:digit:]]{3}$/
+  UNIQUE_NAME_MAX_ATTEMPTS = 10
+  VALID_LETTERS = ('A'..'Z').to_a
+  VALID_NUMBERS = (1..9).to_a
```
paulh16: Nice organization with your constants.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,13 @@
-class NameCollisionError < RuntimeError; end
-
 class Robot
   attr_accessor :name
+  require_relative 'name_generator'
```
paulh16: Good job separating the name generation from the Robot class.

robot_name/robot_registery.rb
avavita: whoops

jmmastey:
In case you didn't get the answer to your commit message question, you can in fact have private class methods using `private_class_method :foo` syntax. It's a bit of a code smell, though.

Overall, I usually bias towards having my classes contain mostly instance methods, plus maybe a few convenience class method wrappers. I've found my classes to be easier to reason about by structuring this way.

robot_name/name_generator.rb
jmmastey: Private as a keyword changes the "mode" for doing new `def` statements. So everything below it becomes a private. There's another way to invoke it `private :foo` which does just a single method. In fact, in Very Modern Ruby, `def` statements return their name (`:foo`) which means you can do `private def foo`. That feels icky to me, though.

Static methods (as mentioned elsewhere) aren't affected by private as a mode keyword.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,59 @@
...
+    generate_unique_name(UNIQUE_NAME_MAX_ATTEMPTS).tap do |name|
+      check_name_format(name)
+      REGISTRY.add_name(name)
+    end
+  end
+
+  def self.random_letter_sequence(number_of_characters)
+    number_of_characters.times.inject('') { |string, _iteration|
+      string << VALID_LETTERS.sample
+    }
```
paulh16: I'd recommend having this block over one line between curly braces. Your other option is to put it across three lines beween `do` and `end`. But you don't want to use curly braces for a multi-line block.

```number_of_characters.times.inject('') { |string, _iteration| string << VALID_LETTERS.sample }```

or 

```
number_of_characters.times.inject('') do |string, _iteration| 
  string << VALID_LETTERS.sample 
end```

robot_name/name_generator.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def self.random_number_sequence(number_of_digits)
+    number_of_digits.times.inject('') { |int, _iteration|
```
paulh16: Same thing as the above method: don't use curly braces for multi-line blocks.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def self.robot_name
+    generate_unique_name(UNIQUE_NAME_MAX_ATTEMPTS).tap do |name|
+      check_name_format(name)
+      REGISTRY.add_name(name)
+    end
+  end
+
+  def self.random_letter_sequence(number_of_characters)
+    number_of_characters.times.inject('') { |string, _iteration|
```
paulh16: Is there a reason you are placing an underscore before `iteration`?

robot_name/name_generator.rb
```diff
@@ -0,0 +1,59 @@
...
+      string << VALID_LETTERS.sample
+    }
+  end
+
+  def self.random_number_sequence(number_of_digits)
+    number_of_digits.times.inject('') { |int, _iteration|
+      int << VALID_NUMBERS.sample.to_s
```
paulh16: Can you rename `int` to be more meaningful?

robot_name/name_generator.rb
```diff
@@ -0,0 +1,59 @@
...
+    }
+  end
+
+  def self.generate_name
+    "#{random_letter_sequence(2)}#{random_number_sequence(3)}"
```
paulh16: I like this much more now. I agree with your comment on the magic numbers, but it is better than concatenating five method calls together.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,59 @@
...
+    # IB "Magic numbers." Cost of changing later is identical to now.
+  end
+
+  def self.generate_name_and_register
+    name = generate_name
+    REGISTRY.add_name(name) ? name : nil # is this weird?
```
paulh16: I do think we could refactor this. Perhaps you can have a method in `Registry` that returns true or false depending on whether the name was added?

robot_name/name_generator.rb
avavita: because it isn't used.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def self.generate_unique_name(max_attempts)
+    attempts = 0
+
+    while attempts < max_attempts
```
paulh16: As my comment in `generate_name_and_register` alludes to, see if refactoring the `Registry#generate_name_and_register`allows you to improve this method.

robot_name/robot_registry.rb
```diff
@@ -0,0 +1,20 @@
...
+  def add_name(name)
+    if !@names.include?(name)
+      @names << name
+      true
+    else
+      false
+    end
+  end
+
+  def contains?(name)
```
paulh16: Is this method used?

avavita: not anymore!

robot_name/name_generator.rb
avavita: yeah that makes more sense.

robot_name/name_generator.rb
paulh16: Gotcha

robot_name/name_generator.rb
avavita: oh. whoops. Yeah, that's confusing.

robot_name/robot_registry.rb
avavita: but it would make sense to use after reading your other comments.

paulh16: I agree.

robot_name/robot_name.rb
paulh16: The `require_relative` should go outside the class definition (it should be on line 1 in this file).

robot_name/name_generator.rb
```diff
@@ -0,0 +1,59 @@
+class NameGenerator
+  # this should be the only class that cares about the registery
+  require_relative 'robot_registry'
+  require_relative 'robot_errors'
```
paulh16: These `require_relatives` should be placed outside the class definition.

paulh16:
Very nice job on this refactoring. There are some simple changes you can make (the `require_relative`s and block syntax), but otherwise the only major thing I'd look at it revisiting the `generate_name_and_register` method.

djkotowski:
Thanks @jmmastey for taking a look.

Good job on this!  Take a look at the comments and refactor, but don't spend much more time on this one.  Paul is a great resource for feedback.

PR #594, opened 2015-04-03
Project Overlord

ckhadilkar:


overlord/spec/bomb_rspec.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  it 'returns true for valid code' do
+    expect(Bomb.valid_code?("1234")).to be true
+  end
+
+  it 'returns false for invalid code' do
+    expect(Bomb.valid_code?("abcd")).to be false
+  end
+
```
ckhadilkar: note to myself: use CORRECT_ACTIVATION_CODE instead of "1234"

overlord/bomb.rb
```diff
@@ -0,0 +1,66 @@
...
+  def retry_count
+    @retry += 1
+  end
+
+  def reset_retry
+    @retry = 0
+  end
+
+  def self.valid_code?(code)
+    code.match(/^\d{4}$/) ? true : false
```
zbright: You dont need the ternary here. Effectively what you are saying here is 
```
if true
  return true
else
  return false
end
```

so just `code.match()` will be fine

overlord/bomb.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def countdown
+    COUNTDOWN
+  end
+
+  private
+
+  def authenticate_activation_code?(code)
+    code.match(@activate_code) ? true : false
```
zbright: See above comment about ternarys

overlord/bomb.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def authenticate_deactivation_code?(code)
+    code.match(@deactivate_code) ? true : false
```
zbright: again

overlord/features/activate.feature
```diff
@@ -0,0 +1,24 @@
...
+  Scenario: Submit incorrect activation code
+    Given Super Villain is on Activate page
+    When Enters incorrect activation code
+    And Clicks Activate
+    Then Error message should be displayed
+    
+  Scenario: Submit empty activation code
+    Given Super Villain is on Activate page
+    When Clicks Activate
+    Then Error message should be displayed
```
zbright: You may want to add scenarios here to cover things like entering a code that is longer (or shorter that 4 characters), or entering a code with letters instead of numbers

overlord/features/step_definitions/configure_steps.rb
```diff
@@ -0,0 +1,52 @@
+Given(/^Bomb is not booted$/) do
+  true
```
zbright: This doesnt seem all that useful. What was the reason you decided to have a step simply return true?

overlord/features/step_definitions/deactivate_steps.rb
```diff
@@ -0,0 +1,47 @@
+Given(/^Bomb is active$/) do
+  visit('/activate')
+  fill_in('activate_code', with: '1234')
+  click_button('Activate')
+  visit('/deactivate')
+  expect(page).to have_content('Bomb is active')
```
zbright: You should not have any expects in a Given block

overlord/features/support/paths.rb
```diff
@@ -0,0 +1,29 @@
...
+  #
+  # step definition in web_steps.rb
+  #
+  def path_to(page_name)
+    case page_name
+
+    when /the home\s?page/
+      '/'
+
+    # Add more mappings here.
```
zbright: Make sure you delete anything in this file you arent using. Just adds clutter to the file

overlord/features/deactivate.feature
```diff
@@ -0,0 +1,29 @@
+Feature: Deactivate bomb
+    The plan has changed. There was a mole in the org. But he wont be an issue any more.
+    Super Villain wants to deactivate the bomb and deploy it at a different location.
+
+    Scenario: Display deactivate page
+      Given Bomb is configured with default codes
+      And Bomb is active
+      When Super Villain is on deactivate page
```
zbright: The wording here is a bit confusing. Make sure your whens are descriptive of an action. 
e.g.
`When Super Villain goes to the deactivation page` would be better here

overlord/features/deactivate.feature
```diff
@@ -0,0 +1,29 @@
...
+      And Bomb is active
+      When Super Villain enters incorrect deactivation code
+      And Clicks Deactivate button
+      Then Error message displayed
+    
+    Scenario: Enter incorrect deactivation code 3 times
+      Given Bomb is configured with default codes
+      And Bomb is active
+      When Super Villain enters incorrect deactivation code 3 times
+      Then He should see explode page
```
zbright: You should add a test here for invalid codes as you did in the config tests

overlord/bomb.rb
ckhadilkar: match returns the pattern if found, returns nil otherwise.
so its more like
if data
  return true
else
  return false
end

overlord/features/step_definitions/configure_steps.rb
ckhadilkar: It indicates bomb is not booted yet but, didn't exactly know how the satisfy it :)

overlord/features/deactivate.feature
ckhadilkar: agree

overlord/bomb.rb
zbright: Whoops, youre right thought match returned true/false. In any case, you can switch it to something like
`code ~= /^\d{4}$/`

overlord/overlord.rb
```diff
@@ -1,14 +1,79 @@
...
+    session[:bomb] = Bomb.new(activate_code, deactivate_code)
+    redirect('/activate')
+  else
+    @error_msg = "Invalid codes. Only numbers that are 4 characters long are allowed."
+  end
+  erb :configure
+end
+
+post '/activate' do
+  bomb = session[:bomb]
```
zbright: I would move this to a method so that you dont have to set it every time
```
def bomb
  session[:bomb]
end
```

overlord/features/activate.feature
ckhadilkar: yeah. i should have covered that.

overlord/overlord.rb
```diff
@@ -1,14 +1,79 @@
...
+  erb :activate
+end
+
+post '/deactivate' do
+  session[:countdown] = params[:d2].chomp('s').to_i
+  bomb = session[:bomb]
+  bomb.deactivate(params[:deactivate_code])
+  redirect('/activate') if bomb.status == "INACTIVE"
+  count = bomb.retry_count
+  tries = bomb.retry_limit - count
```
zbright: I would bring this logic into the bomb class. Just have a method for number of tries remaining

overlord/spec/bomb_rspec.rb
```diff
@@ -0,0 +1,62 @@
...
+  DEFAULT_DEACTIVATION_CODE = "0000"
+  CORRECT_ACTIVATION_CODE = "1234"
+  CORRECT_DEACTIVATION_CODE = "0000"
+  INCORRECT_ACTIVATION_CODE = "1111"
+  INCORRECT_DEACTIVATION_CODE = "2222"
+  RETRY_LIMIT = 3
+  let(:bomb) { Bomb.new }
+  
+  it 'returns default activation code' do
+    expect(Bomb.default_activation_code).to eq(DEFAULT_ACTIVATION_CODE)
```
zbright: This test would be tough to maintain. If someone were to go into the bomb class just to change the default code, they would also have to come in here and make the same change.

overlord/spec/bomb_rspec.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+  
+  it 'returns default deactivation code' do
+    expect(Bomb.default_deactivation_code).to eq(DEFAULT_DEACTIVATION_CODE)
```
zbright: see above comment

overlord/spec/bomb_rspec.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+  
+  it 'returns retry limit' do
+    expect(bomb.retry_limit).to eq(RETRY_LIMIT)
```
zbright: same as above

overlord/spec/bomb_rspec.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  it 'returns retry count' do
+    expect(bomb.retry_count).to eq(1)
+  end
+
+  it 'returns true for valid code' do
+    expect(Bomb.valid_code?("1234")).to be true
```
zbright: rspec has built in matchers for this. be_truthy would be preferable here

overlord/spec/bomb_rspec.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  it 'returns false for invalid code' do
+    expect(Bomb.valid_code?("abcd")).to be false
```
zbright: same as above, but be_falsey

overlord/spec/bomb_rspec.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+  
+  it 'returns bomb status as INACTIVE for incorrect activation code' do
+    bomb.activate(INCORRECT_ACTIVATION_CODE)
+    expect(bomb.status).to eq("INACTIVE")
+  end
+
+  it 'returns bomb status as INACTIVE for correct deactivation code' do
+    bomb.activate(CORRECT_DEACTIVATION_CODE)
+    expect(bomb.status).to eq("INACTIVE")
```
zbright: This test doesnt really cover what you are describing it to. I would have expected the test to activate with a correct activation code, then deactivate with a correct deactivation code, then check that the status is inactive

overlord/views/configure.erb
```diff
@@ -0,0 +1,18 @@
...
+  <font color="blue">Configure activation code:</font>
+  <input type="text" name="activation_code" id="activation_code">
+  <br />
+  <br />
+  <br />
+  <font color="blue">Configure deacivation code:</font>
+  <input type="text" name="deactivation_code" id="deactivation_code">
+  <br />
+  <br />
+  <input style="margin-top: 10px; background-color: green; font-size: 20px; font-weight: bold; height: 50px; width: 90px"
```
zbright: This is no good my friend. All of your styling should be in css files. Dont worry about it for now, but you will need to do it for cyoa

overlord/overlord.rb
```diff
@@ -1,14 +1,79 @@
...
+require 'pry'
 enable :sessions
 
 get '/' do
-  "Time to build an app around here. Start time: " + start_time
+  erb :home
+end
+
+get '/configure' do
+  erb :configure
```
zbright: Its not super obvious to me that you are using some of these gets. for anything other than rendering. If that is the case, you can just use the render keyword in your post methods instead of doing a full redirect

overlord/spec/bomb_rspec.rb
ckhadilkar: oh i c. i should have added the activation part. you'r right

overlord/views/configure.erb
ckhadilkar: still trying to get a hang of everything :). but yes I'll keep this in mind.

zbright:
Well done in general on this one. Some minor things to improve on in the future, but as long as @djkotowski thinks its good, I dont see a reason to resubmit. Only other comment I have is it would be better for you to commit more. You only have one commit which would make it harder for you to revert back if something broke during development. IMO you should commit early and often

ckhadilkar:
These are all valuable comments. I will definitely get more commits while
doing my cyoa.

Thanks

overlord/features/activate.feature
```diff
@@ -0,0 +1,24 @@
+Feature: Activate bomb
+  Super Villan is all syched-up! "Time to activate the bomb!", he says
+
```
djkotowski: I recommend looking at the `As a <role>, I want to <action>, in order to <reason>` format for feature descriptions.  Let your features and scenarios read like stories, and it will really help the format and organizations of your tests.  You can then write your tests in the first-person (e.g. `When I visit the activation page, I should see the bomb activation form`).

overlord/bomb.rb
```diff
@@ -0,0 +1,66 @@
...
+  def deactivate(code)
+    @status = BOMB_STATUS[:inactive] if authenticate_deactivation_code?(code)
+  end
+
+  def explode
+    @status = BOMB_STATUS[:explode]
+  end
+
+  def countdown
+    COUNTDOWN
```
djkotowski: No need for this method.  If you want to access constants from outside the class, use `Class::Constant`.

djkotowski:
This is looking good.  Take a look at the comments and refactor a bit.  No need for resubmission.

PR #595, opened 2015-04-07
Shopping cart

ckhadilkar:


magicalbanana:
Seems good. For formatting your features when it comes to nested Given, When and Then indent the And that is nested so it's much cleaner to see. 

So when I say:
```
Given
   And ...
When
   And ...
Then
   And ...
```

It's easier to note that the end here is still part of the Given or whatever prefix you are using.

magicalbanana:
I feel like you're missing more tests but I can't split screen right now because my dog ate my monitors. :crying_cat_face:

ckhadilkar:
Yes. I still have a few features remaining. I commit as I go. Continuous
Integration! :)

magicalbanana:
Tai nuostabu. :dancer: 

It means that's awesome. Let me know when you feel you're all done and I'll review it some more.

djkotowski:
An additional note:  never use the word "should" in your test descriptions.  It isn't confident.
```cucumber
# Not confident:
Then I should no longer see item 'x' in cart

# Confident:
Then I do not see item 'x' in cart
```

ckhadilkar:
What about in the Feature description? When I describe what the role is

magicalbanana:
Feature description should be a description of what it does, no scenarios or what not.

Like I mentioned above,
```
Feature: Authentication
Feature: Registration
Feature: Shipping Estimation
Feature: Creating an Address
Feature: Editing an Address
Feature: Deleting an Address
```
etc..

djkotowski:
Don't use it there.  Also, I recommend including the reason behind the desired feature.

```cucumber
As a <role>
I want to <feature>
In order to <reason>
```

It will help organize your features and, in production projects, help determine if the way you're implementing the feature is helpful and fulfilling the requirements.

PR #596, opened 2015-04-07
Avarobot

avavita:
I still don't love the generate_unique_name method, but at least now it's not responsible for the registry adds and I got to play with rescue.


robot_name/name_generator.rb
```diff
@@ -0,0 +1,65 @@
...
+      raise NameRegistryError, "#{name} is already in the registry."
+    end
+  end
+
+  def generate_unique_name(max_attempts = 10)
+    attempts = 0
+    while attempts < max_attempts
+      begin
+        attempts += 1
+        name = generate_name
```
avavita: I considered using #tap for this+next two statements but it felt a little "arrowhead" to me. Already have a begin within a while, didn't want to nest further.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,15 @@
-class NameCollisionError < RuntimeError; end
+require_relative 'name_generator'
 
 class Robot
```
ashtonthomas: the file name be `robot_name.rb` but the class name be `Robot`. Trickery?

ashtonthomas: Ah, I see where this class is also the 'runner' class. In that case, fine.. lol

robot_name/name_generator.rb
```diff
@@ -0,0 +1,65 @@
+require_relative 'robot_registry'
+require_relative 'robot_errors'
+
+class NameGenerator
+  # this should be the only class that cares about the registry
```
ashtonthomas: this is a comment

ashtonthomas:
@avavita, cool beans. LGTM

:+1:

PR #600, opened 2015-04-09
Dino catalog

schwerdt:
This is my code for the Dino Catalog level up exercise.


djkotowski:
@schwerdt will fix hound errors then resubmit

PR #603, opened 2015-04-09
Dino catalog

schwerdt:


dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+    #     @diet = nil
+    #     @weight = nil
+    #     @feet = nil
+    #     @description = nil
+
+    # Different csv files have different headers (dictionary keys).  We want
+    # to translate from the dictionary key into the proper variable.  We
+    # translate using the translation dictionary, which maps headers to
+    # instance variables.
+    translate = {}
```
incazteca: This entire method of translating the headers is best done in the data_reader class. Also a better way to create this hash would be the following:

```
translate = {
  genus: :name,
  period: :period,
  diet: :diet 
}
```

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
+class Dinos
```
incazteca: Change the filename to match the class name. Class name should be dinos.rb. However this class is dealing with a singular dinosaur so a more appropriate class name would be Dino with a file name of dinos.rb

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+    translate[:name] = :name
+    translate[:continent] = :continent
+    translate[:diet] = :diet
+    translate[:weight_in_lbs] = :weight
+    translate[:description] = :description
+
+    # Loop over the dino_data.  Map dino_data key to its corresponding
+    # variable name.  Set the corresponding
+    # variable equal to the value associated with the dino_data key.
+    dino_data.each do |dino_key, dino_value|
```
incazteca: This translation can be done in your read_data class.

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def print_dino
+    # This function prints out all of the data that was collected about this
+    # particular dino
+    dino_variables = %w(name continent period diet weight feet description)
+
+    # Build up the string of dino info using a loop
+    dino_string = "**************************************************\n"
+    for var in dino_variables
```
incazteca: You could loop over the instance variables already set with the method `instance_variables`. http://ruby-doc.org/core-2.2.1/Object.html#method-i-instance_variables

incazteca: You could create the loop with `self.instance_variables.map do |variable|`

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+    @continent = 'Africa' unless @continent
+
+    # Set @diet to false if all carnivorous diets are false
+    @diet = (@diet == 'Yes' || @diet == 'Carnivore' || \
+                                         diet == 'Insectivore')
+    # Convert weight into an integer
+    @weight = @weight.to_i unless @weight.nil?
+  end
+
+  def print_dino
```
incazteca: You can overload the `inspect` method of the class so that you can call Dino.to_s to print out the dinosaur.

dino_catalog/dino_filters.rb
```diff
@@ -0,0 +1,57 @@
...
+    @dino_objects = input_list
+  end
+
+  def find_bipeds
+    filtered_list = @dino_objects.select { |dino| dino.feet[/Biped/] == "Biped" }
+    DinoFilters.new(filtered_list)
+  end
+
+  def find_carnivores
+    filtered_list = @dino_objects.select { |dino| dino.diet == true }
```
incazteca: Could create method `carnivore?` to determine if dinosaur is carnivore or not. It's also much more easier to comprehend than having to figure out what a true diet is.

dino_catalog/dino_filters.rb
```diff
@@ -0,0 +1,57 @@
...
+  def find_small_dinos(size)
+    filtered_list = @dino_objects.select do |dino|
+      dino.weight < size unless dino.weight.nil?
+    end
+    # size should be an integer that is the cutoff for the weight
+    DinoFilters.new(filtered_list)
+  end
+
+  def print_list
+    for dino in @dino_objects
```
incazteca: change this to `@dino_objects.each do |dino|`

dino_catalog/read_data.rb
```diff
@@ -0,0 +1,26 @@
+load 'dino_class.rb'
+load 'dino_filters.rb'
+require 'csv'
+
+def read_data
```
incazteca: This should be made into a class. That way this code is more portable. In the initializer you can pass a list of files that need to be parsed. The class could also handle the translation of headers.

dino_catalog/read_data.rb
```diff
@@ -0,0 +1,26 @@
...
+  # Make an list of dino_class objects
+  dino_objects = []
+
+  # Loop over the csv files.  Create dino object and append them to a
+  # list of dino objects
+  CSV.foreach('dino_csv_list') do |new_file|
+    filename = new_file[0]
+    dino_data = CSV.read(filename, headers: true, header_converters: :symbol)
```
incazteca: In header converters you can pass a lambda to convert the different headers into one format.

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+    for var in dino_variables
+      dino_info = instance_variable_get("@#{var}")
+      if dino_info
+        dino_string << var + ": " + dino_string_helper(var, dino_info) + "\n"
+      end
+    end
+    puts dino_string
+  end
+
+  def dino_string_helper(var_name_string, value)
```
incazteca: You should probably have a method similar to this set the variable so that you no longer need to have the helper methods when printing the string.

dino_catalog/main.rb
```diff
@@ -0,0 +1,36 @@
+require 'csv'
+load 'dino_class.rb'
+load 'dino_filters.rb'
```
incazteca: use require instead of load.

incazteca: Also missing the loading of read_data in here.

incazteca:
Please close the PR and resubmit. There doesn't need to be many comments in the source code as there currently is so get rid a lot of those multi-line comments.

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+    #     @continent = nil
+    #     @diet = nil
+    #     @weight = nil
+    #     @feet = nil
+    #     @description = nil
+
+    # Different csv files have different headers (dictionary keys).  We want
+    # to translate from the dictionary key into the proper variable.  We
+    # translate using the translation dictionary, which maps headers to
+    # instance variables.
```
djkotowski: Avoid comments.  Comments are lies waiting to happen.

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+    # Loop over the dino_data.  Map dino_data key to its corresponding
+    # variable name.  Set the corresponding
+    # variable equal to the value associated with the dino_data key.
+    dino_data.each do |dino_key, dino_value|
+      # Look up dino_key in the translate dictionary. dino_var is currently a
+      # string.
+      dino_var = translate[dino_key]
+      # Convert dino_var to a variable (from its string name) and assign data to
+      # it.
+      instance_variable_set("@#{dino_var}", dino_value)
```
djkotowski: What happens if `translate[dino_key] == nil`?

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+    # Some instance fields need special treatment
+    @continent = 'Africa' unless @continent
```
djkotowski: This could also be written as
```ruby
@continent ||= "Africa"
```

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    # Set @diet to false if all carnivorous diets are false
+    @diet = (@diet == 'Yes' || @diet == 'Carnivore' || \
```
djkotowski: You're losing information here.  If the diet is "Piscivore" or "Insectivore", don't we want to hold onto that?

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+                                         diet == 'Insectivore')
+    # Convert weight into an integer
+    @weight = @weight.to_i unless @weight.nil?
+  end
+
+  def print_dino
+    # This function prints out all of the data that was collected about this
+    # particular dino
+    dino_variables = %w(name continent period diet weight feet description)
```
djkotowski: This should be a constant.

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    # Build up the string of dino info using a loop
+    dino_string = "**************************************************\n"
```
djkotowski: Should `Dinos` be responsible for console output?

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+    for var in dino_variables
+      dino_info = instance_variable_get("@#{var}")
+      if dino_info
+        dino_string << var + ": " + dino_string_helper(var, dino_info) + "\n"
```
djkotowski: It would make sense to use string interpolation here.  Also, every time you use `+`, you're creating new strings.  Use `<<` instead.

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def dino_string_helper(var_name_string, value)
+    # This function helps make the dino info string pretty when it is printed.
+    if var_name_string == "diet"
+      return dino_diet_helper(value)
+    elsif var_name_string == "weight"
+      return value.to_s + " lbs"
+    else
+      return value
+    end
```
djkotowski: You can shrink this using guard clauses.  Example:
```ruby
return <val1> if <condition1>
return <val2> if <condition2>
val3
```

dino_catalog/dino_class.rb
djkotowski: This isn't the clearest method naming.  The name implies the method returns a dino string helper.  If the method name is a noun, the method should return something represented by that noun.

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def dino_diet_helper(value)
```
djkotowski: Also unclear method naming.

dino_catalog/dino_filters.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def find_small_dinos(size)
+    filtered_list = @dino_objects.select do |dino|
+      dino.weight < size unless dino.weight.nil?
+    end
+    # size should be an integer that is the cutoff for the weight
+    DinoFilters.new(filtered_list)
+  end
+
+  def print_list
```
djkotowski: Does output really belong here?

dino_catalog/main.rb
```diff
@@ -0,0 +1,36 @@
...
+# Find small dinos
+puts "**********These are the dinos under 2 tons!*****"
+dino_filter_object.find_small_dinos(4000).print_list
+
+# Use multiple filters
+puts "**********These are the carnivorous bipeds!*******"
+dino_filter_object.find_carnivores.find_bipeds.print_list
+
+# Print one dino
+dino_filter_object.print_dino_by_index(0)
```
djkotowski: How about some interaction?

djkotowski:
Good start.  Please take a look at the comments and refactor where necessary.  Keep the single responsibility principle in mind.

PR #605, opened 2015-04-10
First attempt

travisv:


ashtonthomas:
Hey, @travisv. When you create your PR's can you:
- Create one PR for each exercise (one for dino, one for triangle, etc.)
- Edit the title of the PR to include the exercise name

This will just help us review your code ;-)

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
-# Killer facts about triangles AWW YEAH
+# killer facts about triangles AWW YEAH
```
ashtonthomas: Agreed, upper case Killer is so not cool :smile:

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  def from_triassic
+    triassic = []
+    dinos.map { |e| triassic << e if e['PERIOD'] == 'Triassic' }
+    triassic.join
+  end
+end
+dinos = African.new('african_dinosaur_export.csv', headers: true)
+puts dinos.from_triassic
```
ashtonthomas: @travisv, so you are packing a lot of stuff into this one file. These exercises are setup to bring out a lot of topics in writing clean, maintainable and easy to change code. Feel free to contest any of our comments, but I think we have a few higher-level topics to bring up here. This file contains code and logic around:
- parsing a file and converting
- a dino concept
- another dino concept/object
- logic around the catalog concept
- logic about running the file

Do you think it would make things easier to maintain if you had separate classes for each file (filenames should correspond to the class name too), and each class only worried about one thing? High cohesion, low coupling.

Each dino class here doesn't act like a portable ruby object. Each dino class knows a lot about ALL of the dinos. This kind of logic should only be in some DinoCatalog class and the Dino classes should encapsulate the functionality and properties of a single dino.

Also, you have uncovered something inherently ugly about real world requirements: parsing different types of files (our input isn't clean and perfect). A good strategy here is to extract out all the ugly logic into a separate file so you can keep your core app code clean and confident (see Avdi Grimm's Confident Ruby) 

Lastly, you may want to create a simple runner file 'Main', 'Runner', or something that is obvious and intention revealing so others will easily know how to interact with this code.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
-	end
-
-	def radians_to_degrees(rads)
-		return (rads * 180 / Math::PI).round
-	end
-end
+  attr_accessor :side1, :side2, :side3
+
+  def initialize(side1, side2, side3)
+    @side1, @side2, @side3 = side1, side2, side3
```
ashtonthomas: single line assignment is frowned upon in the style guides we use. We prefer to do this one on each line (Ruby allows for a lot of cute things. Some are good, but others don't add as much value to creating maintainable and changeable code)

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+  end
+
+  def equilateral
```
ashtonthomas: All of your methods that return a boolean should should end with a `?` because it just makes code awesome to read: http://pragmati.st/2012/03/24/the-elements-of-ruby-style-predicate-methods/

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+    side1 == side2 && side2 == side3
+  end
+
+  def isosceles
+    [side1, side2, side3].uniq.length == 2
+  end
+
+  def scalene
+    true if equilateral || isosceles
```
ashtonthomas: We don't really need that `true` in here. The current is equivalent to: 

```
if true
  return true
else
  return false
end
```

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+    puts 'This triangle is isosceles! Also, that word is hard to type.' if
+    isosceles
+    puts 'This triangle is scalene and mathematically boring.' if scalene
+    puts 'The angles of this Triangle are ' + angle_facts.join(',')
+    puts 'This triangle is also a right Triangle!' if angle_facts.include? 90
+    puts ''
+  end
+
+  def angle_facts
+    [angle_a(side1, side2, side3), angle_b(side1, side2, side3),
```
ashtonthomas: the format here is a little hard to digest at a glance.

ashtonthomas:
@travisv thanks for the PR. I hope the comments are helpful. I'm going to close this PR, but I recommend going through the comments and seeing if you can come up with improvements. Want to create two new PR's for each exercise, and we can take another look?

PR #608, opened 2015-04-10
Arrowhead

schwerdt:
Changed the if else nested block so that the function breaks at the point
of faulture and returns a print statement specifying the error for the case
when the key is not found in the dictionary.

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -17,18 +17,15 @@ class Arrowhead
 
   # FIXME: I don't have time to deal with this.
   def self.classify(region, shape)
-    if CLASSIFICATIONS.include? region
-      shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        puts "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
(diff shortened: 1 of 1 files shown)


arrowhead/arrowhead.rb
```diff
@@ -17,18 +17,15 @@ class Arrowhead
...
-        arrowhead = shapes[shape]
-        puts "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+
+    return 'Unknown region.' if not CLASSIFICATIONS.include?(region)
```
zbright: In this case (and on line 23), I think `unless` would be a better option than `if not`

arrowhead/arrowhead.rb
```diff
@@ -17,18 +17,15 @@ class Arrowhead
...
+    shapes = CLASSIFICATIONS[region]
+    return 'Unknown shape.' if not shapes.include?(shape)
```
zbright: Should keep the `raise errors` from the original file

zbright:
Good job on this. @djkotowski I dont see a reason to resubmit

djkotowski:
Agreed.  Great job, @schwerdt

PR #613, opened 2015-04-14
Dino catalog

schwerdt:
Refactored based on comments from first code review!


dino_catalog/data_reader.rb
```diff
@@ -0,0 +1,50 @@
+require_relative './dino.rb'
+require_relative './dino_filters.rb'
+require 'csv'
+
+class DataReader
+  
+  def initialize(csv_list_file)
+    #Input is a file containing a list of csv filenames
```
zsyed1: Generally you will want to fix hound errors.

dino_catalog/data_reader.rb
```diff
@@ -0,0 +1,50 @@
...
+    @csv_list_file = csv_list_file
+  end
```
zsyed1: Line 9: No need to store `csv_list_file` in a member variable since it is only used in your method `read_data`. Instead, just pass it through right here itself:

```
def initialize(csv_files)
  read_date(csv_files)
end

def read_date(csv_files)
...
```

zsyed1: Also, it is better to avoid putting the variable type in the variable name: `csv_files` is better than `csv_list_files`.

dino_catalog/data_reader.rb
```diff
@@ -0,0 +1,50 @@
...
+      name: :name,
+      continent: :continent,
+      diet: :diet,
+      weight_in_lbs: :weight,
+      description: :description,
+    }
+    if !translate[header_name.downcase.to_sym]
+      raise 'Key Error.  Ensure CSV headers are in the translate dictionary.'
+    else
+      return translate[header_name.downcase.to_sym]
```
zsyed1: Line 45: Don't need the else since the `raise` on line 44 will exit the method. Then you can remove the `return` keyword on line 46.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,64 @@
...
+  attr_reader :continent
+  attr_reader :diet
+  attr_reader :walking_style
+  attr_reader :weight
+  attr_reader :description
+
+  def initialize(dino_data)
+
+    # Each instance variable in Dino is a key in the dino_data dictionary.
+    dino_data.each do |dino_var, dino_value|
```
zsyed1: People here don't like comments. The understanding is that you should write your code in such a way that comments are incrementally less useful. I DO like comments however, especially in large code-bases.

zsyed1: Line 13: Better names for the key and value might be `attribute` and `value`. No need to prefix `dino_` on each of them.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,64 @@
...
+      instance_variable_set("@#{dino_var}", dino_value)
+    end
+
+    # Some instance fields need special treatment
+    @continent = 'Africa' unless @continent
+
+    @diet = 'Carnivore' if @diet == 'Yes'
+    @diet = 'Not carnivore' if @diet == 'No'
+
```
zsyed1: Maybe use a ternary operator here (line 20 - 21). What do you want to do if `@diet` is neither `Yes` nor `No`? Something like this might be more descriptive:
```
@diet = (@diet == 'Yes') ? 'Carnivore' : 'Not carnivore' if @diet.in?(['Yes', 'No'])
```
There's probably a better way to do this: @djkotowski Help please!

dino_catalog/dino.rb
```diff
@@ -0,0 +1,64 @@
...
+    @weight = @weight.to_i unless @weight.nil?
+  end
+
+  def print_dino
+    # Print out all of the data associated with particular dino
+    dino_values = self.instance_variables.map do |var|
+      dino_info = instance_variable_get("#{var}")
+      if dino_info != nil
```
zsyed1: Line 30: `if dino_info`

dino_catalog/dino.rb
```diff
@@ -0,0 +1,64 @@
...
+        var_string = var.to_s.slice(1..-1)
+        "#{var_string} : #{format_string(var_string, dino_info)} \n"
+      end
```
zsyed1: Line 31: Avoid using the variable type in the name. Instead of `var_string` you could use `value`

dino_catalog/dino.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+    dino_values = dino_values.select { |dino_string| not dino_string.nil? }
+    puts dino_values
+  end
+
+  def format_string(var_name_string, value)
+    return "#{value}  lbs" if var_name_string == "weight"
+    return value
```
zsyed1: Line 39: You can rename `var_name_string` to just `name`

dino_catalog/dino.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def carnivore?
+    ["Carnivore", "Insectivore", "Piscivore"].include?(self.diet)
+  end
+
+  def small_dino?(weight_cutoff)
+    self.weight < weight_cutoff unless self.weight.nil?
+  end
```
zsyed1: Line 49 and Line 53: No need for `.nil?` You can do `if self.weight`

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def initialize(dino_data)
+    # Each instance variable in Dino is a key in the dino_data dictionary.
+    dino_data.each do |attribute, value|
+      instance_variable_set("@#{attribute}", value)
+    end
+
+    # Some instance fields need special treatment
+    @continent = 'Africa' unless @continent
+
```
zsyed1: Line 17: `@continent ||= 'Africa'`

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+    value
+  end
+
+  def carnivore?
+    ["Carnivore", "Insectivore", "Piscivore"].include?(@diet)
+  end
+
+  def small_dino?(weight_cutoff)
+    @weight < weight_cutoff if @weight
+  end
```
zsyed1: Line 47: `@weight && @weight < weight_cutoff`

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def big_dino?(weight_cutoff)
+    @weight > weight_cutoff if @weight
+  end
```
zsyed1: Line 51: see comment on line 47.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def dino_in_era?(era_of_interest)
+    # ignore case of the text in era with /i in regular expression
+    !@era[/#{era_of_interest}/i].nil?
+  end
```
zsyed1: Line 56: `@era[/..../]`

dino_catalog/dino_filters.rb
```diff
@@ -0,0 +1,50 @@
...
+  def find_small_dinos(weight_cutoff)
+    filtered_list = @dino_objects.select { |dino| dino.small_dino?(weight_cutoff) }
+    DinoFilters.new(filtered_list)
+  end
+
+  def print_list
+    for dino in @dino_objects
+      dino.print_dino
+    end
+  end
```
zsyed1: Lines 41-43: `@dino_objects.each(&:print_dino)`

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+  def print_dino
+    dino_values = instance_variables.map do |var|
+      dino_info = instance_variable_get("#{var}")
+      if dino_info
+        attribute = var.to_s.slice(1..-1)
+        "#{attribute} : #{format_string(attribute, dino_info)} \n"
+      end
+    end
+    dino_values = dino_values.select { |dino_string| !dino_string.nil? }
+    puts dino_values
```
zsyed1: Line 33: `dino_values.reject! { |d| d.nil? }`

dino_catalog/main.rb
```diff
@@ -0,0 +1,31 @@
...
+puts "**********These are the dinos over 2 tons!******"
+dino_filter_object.find_big_dinos(4000).print_list
+
+puts "**********These are the dinos under 2 tons!*****"
+dino_filter_object.find_small_dinos(4000).print_list
+
+puts "**********These are the carnivorous bipeds!*******"
+dino_filter_object.find_carnivores.find_bipeds.print_list
+
+dino_filter_object.print_dino_by_index(0)
```
zsyed1: Nice work implementing the features. I would recommend making the changes and resubmitting a new PR only because it will help you remember some of the conventions that you'll be using throughout your Ruby experience. I enjoyed reading your code.

dino_catalog/dino.rb
schwerdt: Won't this return a string when I need to return true or false?  Is it okay to have:
if "matching_text" and let it evaluate to true?

zsyed1: Yes but that will evaluate to `true`. The technical lingo is `truthy`. So a `nil` value will evaluate to false.

PR #614, opened 2015-04-15
Robot name

schwerdt:


robot_name/robot.rb
```diff
@@ -0,0 +1,66 @@
+class NameGenerationError < RuntimeError; end
+
+class Robot
+  attr_accessor :robot_name
+
+  @@name_registry
+
+  def initialize(args = {})
```
paulh16: Instead of using an options hash, consider using a non-required keyword argument: 

This makes it clear what parameter `initialize` takes without needing to look at the method body.

robot_name/robot.rb
```diff
@@ -0,0 +1,66 @@
+class NameGenerationError < RuntimeError; end
+
+class Robot
+  attr_accessor :robot_name
+
+  @@name_registry
```
paulh16: Check to see if this is needed. I don't believe it is as it will be initialized when a Robot instance is created.

robot_name/robot.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def initialize(args = {})
+    @@name_registry ||= []
+    @name_generator = args[:name_generator]
+
+    attempts = 0
+    while attempts < 3
```
paulh16: There is a lot going on in the `initialize` method. I would extract this functionality into at least one method. Use the extract method refactoring.

robot_name/robot.rb
```diff
@@ -0,0 +1,66 @@
...
+      find_error
+    end
+  end
+
+  def generate_name
+    generate_char = -> { ('A'..'Z').to_a.sample } # convert range to array
+    generate_num = -> { rand(10) }
+    robot_name = (1..2).collect { generate_char.call }
+    robot_name += (1..3).collect { generate_num.call }
+    robot_name.join
```
paulh16: Instead of building the generated name in this fashion, see if you can use `inject`. It will make this simpler.

robot_name/robot.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    if valid_name?
+      @@name_registry << @robot_name
+    else
+      find_error
+    end
+  end
+
+  def generate_name
+    generate_char = -> { ('A'..'Z').to_a.sample } # convert range to array
```
paulh16: Consider not using Procs to generate characters and numbers. Instead, try some using a method, such as:

```
def random_letter_sequence(number_of_characters)
  number_of_characters.times.inject('') { |string, iteration| string << CHARACTERS.sample }
end
```

where `CHARACTERS` is simply a constant. You can do an equivalent idiom for generating numbers.

robot_name/robot.rb
```diff
@@ -0,0 +1,66 @@
...
+    robot_name += (1..3).collect { generate_num.call }
+    robot_name.join
+  end
+
+  def name_already_in_list?
+    @@name_registry.include?(@robot_name)
+  end
+
+  def wrong_name_format?
+    @robot_name[/^[[:alpha:]]{2}[[:digit:]]{3}$/].nil?
```
paulh16: I would personally alter the program to test whether the name is in the **correct** format. Then you can test whether the robot_name matches the regex:

`@robot_name` =~ [/^[[:alpha:]]{2}[[:digit:]]{3}$/]`

I think this is a better idiom than checking for nil.

robot_name/robot.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def generate_name
+    generate_char = -> { ('A'..'Z').to_a.sample } # convert range to array
+    generate_num = -> { rand(10) }
+    robot_name = (1..2).collect { generate_char.call }
+    robot_name += (1..3).collect { generate_num.call }
+    robot_name.join
+  end
+
+  def name_already_in_list?
```
paulh16: This is an example of a method that should be made private as it is only called from within the class. There are others as well. I'd recommend making all methods not called outside the class as private.

robot_name/robot.rb
```diff
@@ -0,0 +1,66 @@
...
+      if @name_generator
+        @robot_name = @name_generator.call
+      else
+        @robot_name = generate_name
+      end
+      break if valid_name?
+      attempts += 1
+    end
+
+    if valid_name?
```
paulh16: Instead of first checking whether the name is valid, consider first checking for, and if necessary, raising errors before adding the name to the registry. I think this will simplify your overall class and eliminate the need for this `if-else`.

robot_name/robot.rb
```diff
@@ -0,0 +1,66 @@
...
+    @robot_name[/^[[:alpha:]]{2}[[:digit:]]{3}$/].nil?
+  end
+
+  def valid_name?
+    !(name_already_in_list? || wrong_name_format?)
+  end
+
+  def find_error
+    if name_already_in_list?
+      raise NameGenerationError, 'Name already in list.'
```
paulh16: Instead of using the same Error for both conditions, consider having two separate Errors: a name collision, and an improperly formatted name. This will improve the clarity of the Error class names.

robot_name/robot.rb
```diff
@@ -0,0 +1,66 @@
+class NameGenerationError < RuntimeError; end
```
paulh16: An alternative idiom to consider is:

'NameGenerationError = Class.new(RuntimeError)'

robot_name/robot.rb
```diff
@@ -0,0 +1,66 @@
...
+
+class Robot
+  attr_accessor :robot_name
+
+  @@name_registry
+
+  def initialize(args = {})
+    @@name_registry ||= []
```
paulh16: Nice idiom!

robot_name/robot.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def name_already_in_list?
+    @@name_registry.include?(@robot_name)
+  end
+
+  def wrong_name_format?
+    @robot_name[/^[[:alpha:]]{2}[[:digit:]]{3}$/].nil?
+  end
+
+  def valid_name?
```
paulh16: After making some of the other changes recommended, I think this method can be written with improved clarity.

robot_name/robot.rb
```diff
@@ -0,0 +1,66 @@
...
+        @robot_name = @name_generator.call
+      else
+        @robot_name = generate_name
+      end
+      break if valid_name?
+      attempts += 1
+    end
+
+    if valid_name?
+      @@name_registry << @robot_name
```
paulh16: Good use of the `<<` method!

paulh16:
Nice job on this. Overall, taking another look while using shorter methods (using the extract method refactoring), along with refactoring the methods involved with error handling will greatly simplify the class. Finally, using methods instead of Procs to generate the names will really help.

Give it another go, and I think you'll be able to eliminate some code while improving the clarity of the overall program.

robot_name/robot.rb
djkotowski: I recommend instead
```ruby
robot_name = ""
2.times { robot_name << generate_char.call }
3.times { robot_name << generate_num.call }
```

PR #618, opened 2015-04-16
Dinos

travisv:
I apologize for my pull requests being a total disaster.  I think I finally have it figured out now.

### DIFF
--- a/dino_catalog/african_dino.rb
+++ b/dino_catalog/african_dino.rb
@@ -1,43 +1,42 @@
 require_relative 'dino'
 class African < Dino
-    def carnivores
-        carnivores = []
-        dinos.map do |e|
-            carnivores << e if e['CARNIVORE'] == 'Yes'
-        end
-        carnivores.join
+  def carnivores
+    carnivores = []
+    dinos.map do |e|
+      carnivores << e if e['CARNIVORE'] == 'Yes'
     end
+    carnivores.join
+  end
 
-    def big_or_small
-      big_or_small = []
-      dinos.map do |e|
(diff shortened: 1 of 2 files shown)


ashtonthomas:
@travisv lol yeah Hound can be merciless

dino_catalog/african_dino.rb
```diff
@@ -0,0 +1,42 @@
+require_relative 'dino'
+class African < Dino
```
ashtonthomas: do we ever use this class?

dino_catalog/dino.rb
```diff
@@ -0,0 +1,84 @@
+class Dino
+  require 'csv'
+  attr_accessor :dinos
+
+  def initialize(file, headers)
```
ashtonthomas: your dino class has a lot of logic around parsing. Also the parsing is a bit tricky. Wouldn't it be better to pull the parsing into a different file which could handle all the conversions stuff and hand back one type of dino (and not regular and african)?

dino_catalog/dino.rb
```diff
@@ -0,0 +1,84 @@
...
+    @dinos = CSV.read(file, headers)
+    #@dinos.headers.each(&:upcase!)
+  end
+
+  def bipeds
```
ashtonthomas: It's like you have a singular dino class which should represent a single dino, but is actually a class that handles parsing and filtering of a collection. Perhaps we need: DinoParser, DinoDex, Dino (?)

ashtonthomas:
@travisv, I only gave you a few comments, but I expect you to spend a considerable amount of time on this and robot name (but probably more on this dinodex exercise). Please see my comments and reach out with questions.

PR #619, opened 2015-04-16
Triangle facts

travisv:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,60 @@
...
-		return [angleA, angleB, angleC]
-	end
-
-	def radians_to_degrees(rads)
-		return (rads * 180 / Math::PI).round
-	end
+  attr_accessor :side1, :side2, :side3
+
+  def initialize(side1, side2, side3)
+    @side1, @side2, @side3 = side1, side2, side3
```
ashtonthomas: single line assignment is frowned upon in the style guides we recommend for these exercises

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,60 @@
...
+  end
+
+  def equilateral
```
ashtonthomas: any method that returns a boolean should end in `?` -> `equilateral?` this is idiomatic ruby and lets us write readable code like `if triangle.equilateral?` which is very intuitive.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,60 @@
...
+    puts 'This triangle is isosceles! Also, that word is hard to type.' if
+    isosceles
+    puts 'This triangle is scalene and mathematically boring.' if scalene
+    puts 'The angles of this Triangle are ' + angle_facts.join(',')
+    puts 'This triangle is also a right Triangle!' if angle_facts.include? 90
+    puts ''
+  end
+
+  def angle_facts
+    [angle_a(side1, side2, side3), angle_b(side1, side2, side3),
```
ashtonthomas: you can probably clean this up a little bit more by having the brackets on their own line and each angle on a separate line.

ashtonthomas:
@travisv, please see my comments and let me know what you think

PR #620, opened 2015-04-16
Dino catalog updated

schwerdt:


dino_catalog/data_reader.rb
```diff
@@ -0,0 +1,46 @@
...
+      period: :era,
+      carnivore: :diet,
+      weight: :weight,
+      walking: :walking_style,
+      name: :name,
+      continent: :continent,
+      diet: :diet,
+      weight_in_lbs: :weight,
+      description: :description,
+    }
```
djkotowski: This should be a constant.

dino_catalog/data_reader.rb
```diff
@@ -0,0 +1,46 @@
...
+    unless translate[header_name.downcase.to_sym]
+      raise 'Key Error.  Ensure CSV headers are in the translate dictionary.'
```
djkotowski: Why not make a `KeyError` for this and extract "Key Error." from the message?

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+    # Some instance fields need special treatment
+    @continent ||= 'Africa'
+
+    @diet = 'Carnivore' if @diet == 'Yes'
+    @diet = 'Not carnivore' if @diet == 'No'
+
+    @weight = @weight.to_i unless @weight.nil?
+  end
+
+  def print_dino
```
djkotowski: `Dino` shouldn't be handling console output.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+    puts dino_values
+  end
+
+  def format_string(attribute, value)
+    return "#{value}  lbs" if attribute == "weight"
+    value
+  end
+
+  def carnivore?
+    ["Carnivore", "Insectivore", "Piscivore"].include?(@diet)
```
djkotowski: This is good.  Simple.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+    @weight && @weight > weight_cutoff
+  end
+
+  def dino_in_era?(era_of_interest)
+    # ignore case of the text in era with /i in regular expression
+    !@era[/#{era_of_interest}/i].nil?
+  end
+
+  def biped?
+    @walking_style[/Biped/] == "Biped"
```
djkotowski: Prefer
```ruby
@walking_style ~= /Biped/
```

dino_catalog/dino_filters.rb
```diff
@@ -0,0 +1,48 @@
+require_relative './dino.rb'
+# The purpose of this class is to have methods that allow us to filter the
+# dinosaur list based on values in its instance variables.
+
+class DinoFilters
+  attr_reader :dino_objects
+
+  @dino_objects = nil
```
djkotowski: `@dinos` would be better here.

dino_catalog/dino_filters.rb
```diff
@@ -0,0 +1,48 @@
+require_relative './dino.rb'
+# The purpose of this class is to have methods that allow us to filter the
+# dinosaur list based on values in its instance variables.
+
+class DinoFilters
+  attr_reader :dino_objects
```
djkotowski: `:dinos` would be better here.

dino_catalog/dino_filters.rb
djkotowski: Unnecessary.

dino_catalog/dino_filters.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  @dino_objects = nil
+
+  def initialize(input_list)
+    @dino_objects = input_list
+  end
+
+  def find_bipeds
+    filtered_list = @dino_objects.select(&:biped?)
+    DinoFilters.new(filtered_list)
```
djkotowski: Good way to chain.

dino_catalog/dino_filters.rb
```diff
@@ -0,0 +1,48 @@
...
+    filtered_list = @dino_objects.select { |dino| dino.big_dino?(weight_cutoff) }
+    DinoFilters.new(filtered_list)
+  end
+
+  def find_small_dinos(weight_cutoff)
+    filtered_list = @dino_objects.select { |dino| dino.small_dino?(weight_cutoff) }
+    DinoFilters.new(filtered_list)
+  end
+
+  def print_list
```
djkotowski: `DinoFilters` shouldn't be responsible for console output.

dino_catalog/main.rb
```diff
@@ -0,0 +1,31 @@
...
+puts "**********These are the dinos over 2 tons!******"
+dino_filter.find_big_dinos(4000).print_list
+
+puts "**********These are the dinos under 2 tons!*****"
+dino_filter.find_small_dinos(4000).print_list
+
+puts "**********These are the carnivorous bipeds!*******"
+dino_filter.find_carnivores.find_bipeds.print_list
+
+dino_filter.print_dino_by_index(0)
```
djkotowski: _This_ file is where your output methods should be defined.

djkotowski: This also could use some user interaction.  Let the user enter commands to retrieve what they want.

djkotowski:
This is looking good.  Take a look at the comments for suggestions and improvements.  Add some command line functionality to your main runner.  Please resubmit when finished.

dino_catalog/dino.rb
schwerdt: Can we talk about why this is preferred? (In another piece of code someone else told me the same thing!)  I had thought it would be better to return a boolean value so that I don't have to rely on the caller to do the conversion for me.

djkotowski: In Ruby, `nil` and `false` are both "falsy".  Other values are "truthy".  Truthy values are treated as `true` in conditional statements and falsy values are treated as `false`.  Because of this, no conversion is needed.

If you really wanted to return true/false, you could alter it to be:
```ruby
@walking_style ~= /Biped/ ? true : false
```
though leaving the conditional out would be better IMO.

djkotowski: By suggestion of @bwthomas you can also use `!!` to convert truthy and falsy values into `true` and `false`.

PR #622, opened 2015-04-17
Robot name

travisv:


robot_name/robot_name.rb
```diff
@@ -1,25 +1,31 @@
-class NameCollisionError < RuntimeError; end
+class NameCollisionError < RuntimeError
+end
 
 class Robot
-  attr_accessor :name
+  attr_accessor :name, :registry
```
ashtonthomas: Now you may decide to leave `registry` accessible by the outside world, but I just want to bring up the topic of contracts/api of objects. Any public method is part of your *contract* (the objects api). Once you have it, you have to support it or things may break. `attr_accessor` creates two methods for you. Do you want others to use it? Do you want others to change it?

robot_name/robot_name.rb
```diff
@@ -1,25 +1,31 @@
...
+  NAME_ERROR = 'There was a problem generating the robot name!'
 
-  @@registry
+  def initialize(_args = {})
+    @registry ||= []
+    @name = name_generator
+    check_name
```
ashtonthomas: What happens if someone removes this line? Do you think we could improve the code by having check name somewhere else? What happens if you call name_generate somewhere else and don't call check_name? Perhaps we can move check_name in name_generate

robot_name/robot_name.rb
```diff
@@ -1,25 +1,31 @@
...
+class NameCollisionError < RuntimeError
+end
 
 class Robot
-  attr_accessor :name
+  attr_accessor :name, :registry
+  NAME_ERROR = 'There was a problem generating the robot name!'
 
-  @@registry
+  def initialize(_args = {})
```
ashtonthomas: good use of the underscore for the unused param. However, what if we want to pass in a name generator?

Now, to take this one step further, wouldn't it be nice if Robot only cared about things that a single Robot should care about? Maybe we can create a new Class that handles the name generation and worries about collisions as well as the `@@registry`. Your robot init could then look like: `def initialize(name_generator = RobotNameGenerator.new)`

robot_name/robot_name.rb
```diff
@@ -1,25 +1,31 @@
...
+    @registry ||= []
+    @name = name_generator
+    check_name
+  end
 
-  def initialize(args = {})
-    @@registry ||= []
-    @name_generator = args[:name_generator]
+  def name_generator
```
ashtonthomas: naming is an art form. It is easy to disapprove of names but hard to actually come up with good names. That being said, I think we can name this better (naming and design choice are very closely tied).

Perhaps we could create a method `ensure_name` which combines name generate and checking (but I think it would be better if you extract this logic into a different class as discuss above - and then you can have the `ensure_name` in that file

robot_name/robot_name.rb
```diff
@@ -1,25 +1,31 @@
...
+    (characters + numbers).join
+  end
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  def characters
```
ashtonthomas: Yeah, we really need a separate class for the name generator. Wouldn't it be better if your Robot class only cared about things immediately relevant to a robot and other things were extracted into more appropriate classes?

ashtonthomas:
@travisv take a look at these comments and see what you can do. I think this exercise brings up some great topics for us to go through

PR #624, opened 2015-04-17
Overlord for review

gguntupalli:


overlord/bomb.rb
```diff
@@ -0,0 +1,42 @@
...
+  def deactivate(deactivate_code)
+    if deactivate_code == @deactivation_code
+      @status = :inactive
+    else
+      @wrong_attempts += 1
+      explode if remaining_attempts == 0
+    end
+  end
+
+  def remaining_attempts
```
incazteca: why not just remove this method and instead when seeing if limit is reached check if `WRONG_ATTEMPTS_LIMIT <= @wrong_attempts`

overlord/features/activate.feature
```diff
@@ -0,0 +1,20 @@
...
+
+  Background:
+    Given a bomb is installed
+
+  Scenario: activate the bomb with correct code
+    When I enter the correct activation code
+    Then the bomb should be activated
+
+  Scenario: activate the bomb with correct code twice
+    When I enter the correct activation code
```
incazteca: The step can be changed to `I enter the correct activation code 2 time` so that you don't have to repeat the step twice.

overlord/features/deactivate.feature
```diff
@@ -0,0 +1,30 @@
...
+  I want to deactivate the bomb
+
+  Background:
+    Given a bomb is activated
+
+  Scenario: deactivate the bomb with correct code
+    When I enter the correct deactivation code
+    Then the bomb should be deactivated
+
+  Scenario: deactivate the bomb with correct code again
```
incazteca: This test is the same as the one before it. Remove it.

overlord/features/deactivate.feature
```diff
@@ -0,0 +1,30 @@
...
+    When I enter the correct deactivation code
+    Then the bomb should be deactivated
+
+  Scenario: deactivate the bomb with wrong code
+    When I enter the wrong deactivation code
+    Then the bomb should not be deactivated
```
incazteca: It would be easier to read this as `Then the bomb should be activated`. You could also make a generic `The bomb should be #{status}` type of statement so that you can have one statement that would check for the status you want.

overlord/overlord.rb
```diff
@@ -1,14 +1,56 @@
...
-  "Time to build an app around here. Start time: " + start_time
+  DEFAULT_ACTIVATION_CODE = Bomb::DEFAULT_ACTIVATION_CODE
+  DEFAULT_DEACTIVATION_CODE = Bomb::DEFAULT_DEACTIVATION_CODE
+  erb :boot
+end
+
+post '/' do
+  activation_code = params['activation_code']
+  deactivation_code = params['deactivation_code']
+  if Bomb.valid_code?(activation_code) && Bomb.valid_code?(deactivation_code)
```
incazteca: There is no need for this conditional statement. When creating the bomb in the statement below it should raise the error when it's initialized.

overlord/overlord.rb
```diff
@@ -1,14 +1,56 @@
...
+    session[:bomb] = Bomb.new(params['activation_code'], params['deactivation_code'])
+    redirect('/activate')
+  else
+    @error = "Error: Please enter a four digit number"
```
incazteca: You can rescue an error from the bomb initialization to set `@error`

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,54 @@
+require_relative "spec_helper.rb"
+WRONG_ACTIVATION_CODE = '4321'
+WRONG_DEACTIVATION_CODE = '1111'
+
+describe Bomb do
+  let(:bomb) { Bomb.new }
+
+  it "should check for valid code" do
+    expect(Bomb.valid_code?('1')).to be_falsey
```
incazteca: use predicate matchers. This can be:
`expect(Bomb).not_to be_valid_code('1)'`

More about predicate matchers can be read here: https://www.relishapp.com/rspec/rspec-expectations/v/3-2/docs/built-in-matchers/predicate-matchers

overlord/bomb.rb
```diff
@@ -0,0 +1,42 @@
...
+  WRONG_ATTEMPTS_LIMIT = 3
+
+  def initialize(activation_code = DEFAULT_ACTIVATION_CODE, deactivation_code = DEFAULT_DEACTIVATION_CODE)
+    @status = :inactive
+    @activation_code = activation_code
+    @deactivation_code = deactivation_code
+    @wrong_attempts = 0
+  end
+
+  def self.valid_code?(code)
```
incazteca: Have this code check if a code is valid when initializing and activating and deactivating. That way you don't have to have conditionals checking if the codes are valid in the overlord.rb file.

incazteca:
@gguntupalli Looks mostly good. Resubmit with changes based on comments.

overlord/bomb.rb
gguntupalli: I am using this method in overlord.rb to throw an error message with remaining attempts. I didn't want to do that calculation there, so I had this method and used it above.

overlord/features/boot.feature
```diff
@@ -0,0 +1,23 @@
...
+    And I enter valid deactivation code
+    When I boot the bomb
+    Then the bomb should be inactive
+
+  Scenario: configure the bomb with default codes
+    When I boot the bomb
+    Then the bomb should be inactive
+
+  Scenario: configure the bomb with invalid codes
+    Given I enter invalid activation code
```
djkotowski: Like @incazteca said above, the repeated steps could be simplified by specifying the number of times in the step.

overlord/features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,219 @@
+# Taken from the cucumber-rails project.
```
djkotowski: Don't use web steps.  They encourage poor and brittle feature-writing by including implementation details in your feature.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,54 @@
+require_relative "spec_helper.rb"
+WRONG_ACTIVATION_CODE = '4321'
+WRONG_DEACTIVATION_CODE = '1111'
+
+describe Bomb do
+  let(:bomb) { Bomb.new }
+
+  it "should check for valid code" do
```
djkotowski: Since this is describing `#valid_code?`, it would be beneficial to wrap this test in a `describe "#valid_code"` block.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    expect(Bomb.valid_code?('1')).to be_falsey
+    expect(Bomb.valid_code?('ssad')).to be_falsey
+    expect(Bomb.valid_code?('112fd')).to be_falsey
+    expect(Bomb.valid_code?('1234')).to be_truthy
+    expect(Bomb.valid_code?('123456')).to be_truthy
+  end
+
+  it "should be activated with default activation code" do
+    bomb.activate(Bomb::DEFAULT_ACTIVATION_CODE)
+    expect(bomb.status).to eq(:active)
```
djkotowski: This is good.  As an optional suggestion, you could consider including predicate methods in your bomb to check for statuses.

```ruby
expect(bomb.status).to eq(:active) # check the status manually as you are doing now
expect(bomb).to be_active # or include an active? predicate method to make the test read a little more clearly
```

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  it "should not be deactivated with wrong deactivation code" do
+    bomb.activate(Bomb::DEFAULT_ACTIVATION_CODE)
+    bomb.deactivate(WRONG_DEACTIVATION_CODE)
+    expect(bomb.status).not_to eq(:inactive)
+  end
+
+  it "should explode after reaching the wrong attempts limit" do
+    bomb.activate(Bomb::DEFAULT_ACTIVATION_CODE)
+    Bomb::WRONG_ATTEMPTS_LIMIT.times do
```
djkotowski: I like this use of constants.

djkotowski:
Looking good.  Please look at the comments, refactor and resubmit.

PR #625, opened 2015-04-17
Avascience

avavita:
"I used to think correlation implied causation. Then I took a statistics class. Now I don't."
"Sounds like the class helped."
"Well, maybe."


data_science/cohort.rb
```diff
@@ -0,0 +1,27 @@
+require 'pp'
+
+class Cohort
+attr_reader :name, :attempts, :successes
+
+  def initialize(attributes)
```
avavita: ...where the hell are you talking about hound?

data_science/cohort.rb
```diff
@@ -0,0 +1,27 @@
...
+    @name = attributes[:name].to_s
+    @attempts = attributes[:attempts].to_i
+    @successes = attributes[:successes].to_i
+  end
+
+  def conversion_rate
```
avavita: I switched to rubymine and I probably need to tweak the whitespace settings. Although it looks correct to me...?

zsyed1: Wierd ...

data_science/ab_runner.rb
```diff
@@ -0,0 +1,24 @@
+require 'pp'
+require_relative 'split_test'
+require_relative 'cohort_converter'
+
+abort('Please include one json file as an argument.') unless ARGV.count == 1
+
+converter = CohortConverter.new(ARGV[0])
+cohorts = converter.cohorts
+test = SplitTest.new(cohorts)
```
zsyed1: Line 7, 8: Since you never use `converter` other than to access its `cohorts`, you can just remove line 8 and do:
```
cohorts = CohortConverter.new(ARGV[0]).cohorts
```

data_science/cohort_converter.rb
```diff
@@ -0,0 +1,32 @@
...
+      map[row['cohort']].add_attempt
+      map[row['cohort']].add_success if row['result'].to_i == 1
+    end
+    map.values
+  end
+
+  private
+
+  def cohorts_map
+    cohort_collection = cohort_names.inject([]) { |cohorts, name| cohorts << Cohort.new(name: name) }
```
zsyed1: Line 21: Do you ever set the attributes `attempts` and `successes` when creating a `Cohort`? If not, you can change those two lines in `Cohort::initialize` to just set
```
@attempts = 0
@successes = 0
```
Though I suspect you put those there to facilitate testing.

data_science/cohort_converter.rb
```diff
@@ -0,0 +1,32 @@
...
+require_relative 'cohort'
+
+class CohortConverter
+  def initialize(json_file)
+    @data = JSON.parse(file_contents(json_file))
+  end
+
+  def cohorts
+    map = cohorts_map
+    @data.each do |row|
```
zsyed1: Line 10: Normally you want to use a more descriptive name than `map`. `cohorts` would have been nice, but that's your method name. You'll have to be creative!

data_science/cohort_converter.rb
```diff
@@ -0,0 +1,32 @@
...
+    map.values
+  end
+
+  private
+
+  def cohorts_map
+    cohort_collection = cohort_names.inject([]) { |cohorts, name| cohorts << Cohort.new(name: name) }
+    cohort_collection.each_with_object({}) { |cohort, map| map[cohort.name] = cohort }
+  end
+
```
zsyed1: Lines 21 - 24. Here you are trying to build a map from name to Cohort. I think it would be more clear to do this in one `each_with_object` block:
```
cohort_names.each_with_object({}) do |name, cohorts|
  cohorts[name] = Cohort.new(name: name)
end
```

data_science/split_test.rb
```diff
@@ -0,0 +1,34 @@
...
+  attr_reader :cohorts
+
+  def initialize(cohorts)
+    @cohorts = cohorts
+  end
+
+  def conversion_rates
+    @cohorts.each_with_object([]) { |cohort, rates| rates << cohort.conversion_rate }
+  end
+
```
zsyed1: Line 13: What you want here is the method `map`:
```
@cohorts.map { |cohort| cohort.conversion_rate }
```
A shortened form for this (since you are simply calling the method `conversion_rate` on each object) is:
```
@cohorts.map(&:conversion_rate)
```

data_science/split_test.rb
```diff
@@ -0,0 +1,34 @@
...
+  def confidence(level_of_confidence = 0.95)
+    @cohorts.each_with_object([]) do |cohort, levels|
+      levels << ABAnalyzer.confidence_interval(cohort.successes, cohort.attempts, level_of_confidence)
```
zsyed1: See comment above about `map`

data_science/spec/split_test_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  describe '#confidence' do
+    it 'finds min/max predictions for a given confidence level' do
+      expect(test.confidence[0]).to eq([0.968846875179678, 0.9978197914869885])
+    end
+  end
+
+  describe '#chi_square' do
+    it 'returns the p values for a chi_square test' do
+      expect(test.chi_square).to eq(0.0)
+    end
```
zsyed1: Chi-Square of 0.0 sounds fishy :smile:
It would be a better test if the value were non-zero. I like the fact that you generated a test-data set! I'm surprised the chi-square value is zero though,

data_science/cohort_converter.rb
```diff
@@ -0,0 +1,32 @@
...
+    @data = JSON.parse(file_contents(json_file))
+  end
+
+  def cohorts
+    map = cohorts_map
+    @data.each do |row|
+      map[row['cohort']].add_attempt
+      map[row['cohort']].add_success if row['result'].to_i == 1
+    end
+    map.values
```
zsyed1: Lines 10-14 would be a little cleaner with an `each_with_object`.

data_science/cohort_converter.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def cohorts_map
+    cohort_collection = cohort_names.inject([]) { |cohorts, name| cohorts << Cohort.new(name: name) }
+    cohort_collection.each_with_object({}) { |cohort, map| map[cohort.name] = cohort }
+  end
+
+  def cohort_names
+    @data.inject([]) { |names, row| names << row['cohort'] }.uniq
+  end
+
```
zsyed1: Line 26: Use `map`:
```
@data.map { |row| row['cohort'] }.uniq
```

data_science/split_test.rb
zsyed1: Nice work!
If this is your first PR for data-science that I would recommend making the changes and resubmitting.

data_science/spec/split_test_spec.rb
avavita: it is a very skewed set :)

data_science/cohort.rb
```diff
@@ -0,0 +1,27 @@
...
+    @successes.to_f / @attempts.to_f
+  end
+
+  def failures
+    @attempts - @successes
```
djkotowski: Prefer using the `attr_reader` when retrieving data (unless no reader exists for the instance variable).

data_science/cohort_converter.rb
djkotowski: `#map` could accomplish this line, though @zsyed1's suggestion below removes the need.

data_science/spec/cohort_converter_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'spec_helper'
+
+describe CohortConverter do
+  let(:data) { CohortConverter.new('spec/test_data.json') }
+
+  describe '#new' do
+    it 'takes a json object and returns a CohortConverter object' do
+      expect(data).instance_of? CohortConverter
```
djkotowski: `#new` always returns this.  Is this test necessary?  Instead, test initializing with valid and invalid data.

As a side note, when having to reference the class in the spec, use `described_class`.  That way, if you change the class name, you only have to change the first `describe`.

data_science/spec/cohort_converter_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+  end
+
+  describe '#cohorts' do
+    it 'returns a collection of cohorts.' do
+      data.cohorts.each do |cohort|
+        expect(cohort).instance_of? Cohort
```
djkotowski: Prefer
```ruby
expect(cohort).to be_a(Cohort)
```

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'spec_helper'
+
+describe Cohort do
+  attributes =
```
djkotowski: Wrap this in a `let`.

data_science/spec/split_test_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  describe '#cohorts' do
+    it 'returns a collection of cohorts' do
+      expect(test).instance_of? SplitTest
+    end
+  end
+
+  describe '#conversion_rates'
+    it 'returns a collection of conversion rates, one for each cohort' do
+      expect(test.conversion_rates[0]).to eq(0.9833333333333333)
```
djkotowski: You can also use the `be_within` expectation to handle precision.

https://www.relishapp.com/rspec/rspec-expectations/v/2-5/docs/built-in-matchers/be-within-matcher

djkotowski:
Good work!  Just a few comments throughout.  Take a look at them and refactor as necessary.  Make sure to consider all paths (bad and sad included) when testing.  If you have user input that should result in error, test that it does.  Resubmit when finished.

data_science/cohort.rb
avavita: disagree in this case because "failures" is a virtual attribute. If I used attr_reader for this one, I would have to recompute failures every time the consumer called #add_attempt or #add_success. Had I chosen not to "build up" each cohort, and instead chose to only instantiate one once I had all the data aggregated, I would use an attr_reader.

avavita: ...although I suppose that changes if the statement in the init runs each time the attr_reader version of "failures" is called, or if it only runs when the object is first instantiated.

data_science/spec/cohort_converter_spec.rb
avavita: but what if someone overrides the definition of #new?? (okay fine it's totally unnecessary)

data_science/cohort.rb
djkotowski: Sorry, I was unclear.

I mean use the `attr_reader`s for `attempts` and `successes` instead of their instance vars.

data_science/cohort_converter.rb
avavita: they already default to 0 because of the #to_i, and regardless of the fact that I do not use the object that way, I think it's better design to allow cohorts to be initialized with all their data should the need arise.

PR #626, opened 2015-04-17
Robot name with suggestions from Paul incorporated.

schwerdt:


robot_name/robot.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  private
+
+  def generate_name
+    if @name_generator
+      @name_generator.call
+    else
+      internal_name_generator
+    end
```
djkotowski: This could also be shortened to two lines if you wish.
```ruby
return @name_generator.call if @name_generator
internal_name_generator
```

robot_name/robot.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  private
+
+  def correct_name_format?
+    !!(@robot_name =~ /^[[:alpha:]]{2}[[:digit:]]{3}$/)
+  end
+
+  def valid_name?
+    (new_name? && correct_name_format?)
```
djkotowski: No need for parens here.

robot_name/robot.rb
```diff
@@ -0,0 +1,72 @@
...
+  private
+
+  def new_name?
+    !@@name_registry.include?(@robot_name)
+  end
+
+  private
+
+  def correct_name_format?
+    !!(@robot_name =~ /^[[:alpha:]]{2}[[:digit:]]{3}$/)
```
djkotowski: Good ol' double-negation hack.

robot_name/robot.rb
```diff
@@ -0,0 +1,72 @@
+NameCollisionError = Class.new(RuntimeError)
```
djkotowski: The syntax shown in the original would also be acceptable here.

djkotowski:
Great job!  Just a few comments, but you're done.

PR #627, opened 2015-04-17
Dino catalog with user interaction!

schwerdt:


incazteca:
Looks mostly good. Resubmit with the fix for the user input.

schwerdt:
Thanks Franco!  I fixed the user input to give a menu of options.  I also changed the code so that I loop over all possible filters and filter the list as I go (to avoid building up a string of filters and then using the eval statement).  I left the user_input class because I like how the setters shortened my parsing of the user input (to have the variable be nil unless I get a response that is meaningful).  Is there a way to do that without a using a class?

PR #633, opened 2015-04-23
Data science

schwerdt:


data_science/cohort.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def get_cohort_name(population_data)
+    current_cohort = population_data[0]["cohort"]
+    population_data.each do |entry|
+      raise CohortNameError if entry["cohort"] != current_cohort
+    end
+    @name = current_cohort.to_sym
+  end
+
+  private
```
incazteca: This is unnecessary.

data_science/cohort.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def get_num_conversions(population_data)
+    @num_conversions = 0
+    population_data.each do |entry|
+      @num_conversions += entry["result"]
+    end
+    @num_conversions
+  end
+
+  public
```
incazteca: Also unnecessary. By default any method is public. If you want a method to be private just have it below the private statement.

data_science/ABtest.rb
```diff
@@ -0,0 +1,34 @@
+require 'abanalyzer'
```
incazteca: File name should be ab_test.rb

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,51 @@
+require_relative '../cohort'
+
+require 'json'
+
+describe Cohort do
+  before :all do
+    file = File.read('smaller_data_set.json')
+    all_data = JSON.parse(file)
```
incazteca: This can be done in a let block: 
`let(:all_data) { JSON.parse(File.read('smaller_data_set.json')) }`

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    pop_data = all_data.select { |record| record["cohort"] == "A" }
+    @cohort = Cohort.new(pop_data)
```
incazteca: Same with cohort. This can be done in a let block as well.

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  describe("#new") do
+    it "takes a list of dictionaries and returns a Cohort object" do
+      expect(@cohort).to(be_an_instance_of(Cohort))
+    end
+  end
+
+  describe("#name") do
+    it "returns the correct cohort name" do
+      expect(@cohort.name).to(eql(:A))
```
incazteca: You could write the `to` methods without the parentheses.

incazteca: Also are you expecting the name to be a symbol?

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      expect(@cohort.num_conversions).to(eql(5))
+    end
+  end
+
+  it "returns the conversion rate for this cohort" do
+    expect(@cohort.compute_conversion_rate).to(eql(5/72.0))
+  end
+
+  it "returns the standard deviation for this cohort" do
+    expect(@cohort.compute_std_dev).to(eql(0.029958747929501817))
```
incazteca: Would probably be easier to read and update this if the number is defined in a let block:
`let(:expected_std_dev) { 0.029958747929501817 }`

data_science/spec/ABtest_spec.rb
```diff
@@ -0,0 +1,35 @@
+require_relative '../ABtest'
+
+require 'json'
+
+describe ABtest do
+  before :all do
+    file = File.read('smaller_data_set.json')
+    all_data = JSON.parse(file)
```
incazteca: This can be placed in a let block

data_science/spec/ABtest_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    cohort_b = Cohort.new(pop_b)
+    @ab_test = ABtest.new(cohort_a, cohort_b)
+  end
+
+  describe("#new") do
+    it "takes 2 cohorts and returns and ABtest object" do
+      expect(@ab_test).to(be_an_instance_of(ABtest))
+    end
+  end
+  describe("@new") do
```
incazteca: @djkotowski I'm not familiar with @new so correct me if I'm wrong with what follows. @schwerdt Is this supposed to be another test on the initialization method? If so you don't need to have this line and can just add the test under the other one in the describe('#new') block.

data_science/spec/ABtest_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      expect(@ab_test.cohort_b).to(be_an_instance_of(Cohort))
+    end
+  end
+
+  it "finds the leader in conversion rates" do
+    expect(@ab_test.leader).to(eql(@ab_test.cohort_b.name))
+  end
+
+  it "determines if there is a signifcant leader" do
+    expect(@ab_test.compute_leader_confidence_level).to(eql(0.3640661012729629))
```
incazteca: Place number in let block.

data_science/cohort.rb
```diff
@@ -0,0 +1,52 @@
...
+    current_cohort = population_data[0]["cohort"]
+    population_data.each do |entry|
+      raise CohortNameError if entry["cohort"] != current_cohort
+    end
+    @name = current_cohort.to_sym
+  end
+
+  private
+
+  def get_num_conversions(population_data)
```
incazteca: It would seem more efficient to retrieve the conversion numbers during the same loop where you split the cohorts.

incazteca:
@schwerdt Please close and resubmit based on comments.

data_science/cohort.rb
djkotowski: In other words, organize your classes to have public methods above `private` and private methods below.

data_science/spec/ABtest_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+describe ABtest do
+  before :all do
+    file = File.read('smaller_data_set.json')
+    all_data = JSON.parse(file)
+    pop_a = all_data.select { |record| record["cohort"] == "A" }
+    pop_b = all_data.select { |record| record["cohort"] == "B" }
+    cohort_a = Cohort.new(pop_a)
+    cohort_b = Cohort.new(pop_b)
+    @ab_test = ABtest.new(cohort_a, cohort_b)
```
djkotowski: This can also be placed in a `let` block.

data_science/spec/ABtest_spec.rb
djkotowski: Yep, this is supposed to be for `#new`.

data_science/spec/ABtest_spec.rb
djkotowski: You can also use the `be_within` matcher to avoid worrying about precision.

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  it "returns the 95% confidence level for this cohort" do
+    confidence_interval = []
+    confidence_interval.push(@cohort.conversion_rate - 1.96 * @cohort.std_dev)
+    confidence_interval.push(@cohort.conversion_rate + 1.96 * @cohort.std_dev)
+    expect(@cohort.compute_confidence_interval_95pct).to(eql(confidence_interval))
```
djkotowski: Find the values that you are expecting and declare them explicitly in a `let`.

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  before :all do
+    file = File.read('smaller_data_set.json')
+    all_data = JSON.parse(file)
+    pop_data = all_data.select { |record| record["cohort"] == "A" }
+    @cohort = Cohort.new(pop_data)
+  end
+
+  describe("#new") do
+    it "takes a list of dictionaries and returns a Cohort object" do
+      expect(@cohort).to(be_an_instance_of(Cohort))
```
djkotowski: This doesn't really provide much benefit.  `klass.new` will always* return an instance of `klass`.

( * Exceptions are too rare to concern yourself with.)

data_science/cohort.rb
djkotowski: Also, don't precede method names with `get_`.  It implies a Java-like reader and is un-Rubylike.

data_science/cohort.rb
```diff
@@ -0,0 +1,52 @@
...
+    population_data.each do |entry|
+      @num_conversions += entry["result"]
+    end
+    @num_conversions
+  end
+
+  public
+
+  def compute_conversion_rate
+    @conversion_rate = @num_conversions.to_f / @sample_size
```
djkotowski: Why not just have this memoized in the reader?

```ruby
def conversion_rate
  @conversion_rate ||= num_conversions.to_f / sample_size
end
```

data_science/cohort.rb
```diff
@@ -0,0 +1,52 @@
...
+    @name = get_cohort_name(population_data)
+    @sample_size = population_data.length
+    @num_conversions = get_num_conversions(population_data)
+    @conversion_rate = compute_conversion_rate
+    @std_dev = compute_std_dev
+  end
+
+  private
+
+  def get_cohort_name(population_data)
```
djkotowski: I'm not sure `Cohort` should be responsible for finding its own name out of the data it is provided.  I think it'd make more sense to have it passed in.

```ruby
def initialize(name, population_data)
  @name = name
  # ...
end

File: data_science/cohort.rb [hunk 17]
```diff
@@ -0,0 +1,52 @@
...
+  attr_accessor :name
+  attr_accessor :sample_size
+  attr_accessor :num_conversions
+  attr_accessor :conversion_rate
+  attr_accessor :std_dev
+
+  def initialize(population_data)
+    @name = get_cohort_name(population_data)
+    @sample_size = population_data.length
+    @num_conversions = get_num_conversions(population_data)
```
djkotowski: Why not just have these calculated when they're needed?  Define a memoizing reader for these.

data_science/cohort.rb
```diff
@@ -0,0 +1,52 @@
...
+class Cohort
+  attr_accessor :name
+  attr_accessor :sample_size
+  attr_accessor :num_conversions
+  attr_accessor :conversion_rate
+  attr_accessor :std_dev
+
+  def initialize(population_data)
+    @name = get_cohort_name(population_data)
+    @sample_size = population_data.length
```
djkotowski: You could also just define this as a method instead of a standard reader.

data_science/cohort.rb
```diff
@@ -0,0 +1,52 @@
+CohortNameError = Class.new(RuntimeError)
+
+class Cohort
+  attr_accessor :name
+  attr_accessor :sample_size
+  attr_accessor :num_conversions
+  attr_accessor :conversion_rate
+  attr_accessor :std_dev
```
djkotowski: Should we allow outside code to change these values when we're calculating them in this class?

data_science/ABtest.rb
```diff
@@ -0,0 +1,34 @@
...
+class ABtest
+  attr_accessor :cohort_a
+  attr_accessor :cohort_b
+  attr_accessor :leader
```
djkotowski: We're initializing with these values.  Do we need to allow outside code to write to these?

djkotowski:
Good first attempt.  You'll want to start using `let` blocks in your specs to clean them up.  Take a look at the comments, refactor and resubmit.  Ask me or Franco if you have any questions.

data_science/cohort.rb
schwerdt: Do you mean make the write my own function for the determination of the sample size?

djkotowski: Yes.  If you store the population data, you could have a method that does this.

```ruby
def sample_size
  population_data.length
end
```

data_science/cohort.rb
schwerdt: I kind of liked this way because then the user can't make the mistake of providing a cohort name that is inconsistent with that name (either data from 2 cohorts or data from cohort B when cohort A's data was passed in).  What are your thoughts on this?  Would it be preferred to take in the name of the cohort as an argument and then still check that it is consistent with each record passed in to the constructor instead?

djkotowski: I think you should take the user out of the equation.  Have a class responsible for splitting up and summarizing the data into individual cohorts.  In the end, I don't think `Cohort` should be responsible for parsing data to get its name, conversions, etc.  Something else should do that split, letting `Cohort` just hold the summarized data and make calculations off the summary as needed.

data_science/ABtest.rb
```diff
@@ -0,0 +1,34 @@
...
+    sample_data[@cohort_b.name] = { failures: compute_failures(@cohort_b),
+                                    successes: @cohort_b.num_conversions }
+    compare_cohorts = ABAnalyzer::ABTest.new(sample_data)
+    1.0 - compare_cohorts.chisquare_p
+  end
+
+  private
+
+  def compute_failures(cohort)
+    cohort.sample_size - cohort.num_conversions
```
djkotowski: One additional note:  This should be a method on Cohort, not ABTest.

data_science/ABtest.rb
```diff
@@ -0,0 +1,34 @@
+require 'abanalyzer'
+class ABtest
```
djkotowski: Class name should be ABTest.

PR #634, opened 2015-04-23
added fileto branch

Dana-G:
added empty file to branch


ashtonthomas:
@Dana-G, thanks for the PR! Looks like you are off to a good start. Please note that you created this pull request from your `master` branch which means that you now have a empty test file living on your master  branch. It is advised to do everything on a feature branch. `just_git` in this case. The reason for this is because we want to keep each exercise clean (we want to cut a new feature/topic branch from a clean master each time). I recommend the following to get your master back in a clean state:
```
git reset --hard 9929c41749239e2bdd71f5bbee4319b6edfce748
# only force push this once and avoid it from now own
git push -f origin master
git checkout -b new_topic_branch
# do some stuff
git add .
git commit -m "adding files for my exercise"
git push origin new_topic_branch
git checkout master
git checkout -b next_exercise_branch
# repeat...
```

This is just an example of how you may proceed. There are many different ways, so let me know if you have questions

- A

Dana-G:
Thanks. That is super helpful, and very appreciated!

PR #637, opened 2015-04-27
Dino catalog

schwerdt:
I fixed the user input to give a menu of options. I also changed the code so that I loop over all possible filters and filter the list as I go (to avoid building up a string of filters and then using the eval statement). I left the user_input class because I like how the setters shortened my parsing of the user input (to have the variable be nil unless I get a response that is meaningful). Is there a way to do that without a using a class?


dino_catalog/main.rb
```diff
@@ -0,0 +1,17 @@
+require_relative './data_reader.rb'
+require_relative './user_input.rb'
```
paulh16: Take a look at whether these needs to be `require_relative`. If you refer to `./file_name`, you should be able to use `require` instead.

dino_catalog/main.rb
```diff
@@ -0,0 +1,17 @@
...
+
+# If you want to run your own queries, you just need
+# to get the dino_filter object.
+dino_filter = DataReader.new('dino_csv_list').dino_filter
+
+user_input = UserInput.new
+user_input.query_user
+
+dino_summary =  user_input.perform_user_query(dino_filter)
```
paulh16: Consider structuring the `perform_user_query` to return nil if there are no dinos after filtering. Then you can structure the method below as:

```
if dino_summary.length > 0
  puts dino_summary
else
  puts "No dinos remain after filtering."
end
```

dino_catalog/data_reader.rb
```diff
@@ -0,0 +1,47 @@
...
+  def read_data(csv_files)  # csv_files = file containing list of csv filenames
+    dino_objects = []
+
+    # Each row in csv file is a separate dino object
+    CSV.foreach(csv_files) do |new_file|
+      filename = new_file[0]
+      dino_data = CSV.read(filename, headers: true, header_converters: lambda { |h| translate_header(h) })
+      dino_data.each do |row|
+        new_dino = Dino.new(row.to_hash)
+        dino_objects.push(new_dino)
```
paulh16: Consider using `dino_objects` << new_dino`

dino_catalog/data_reader.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def read_data(csv_files)  # csv_files = file containing list of csv filenames
+    dino_objects = []
+
+    # Each row in csv file is a separate dino object
+    CSV.foreach(csv_files) do |new_file|
+      filename = new_file[0]
+      dino_data = CSV.read(filename, headers: true, header_converters: lambda { |h| translate_header(h) })
+      dino_data.each do |row|
+        new_dino = Dino.new(row.to_hash)
```
paulh16: I'd prefer to have the desired object placed in the `Dino.new` method instead of calling `to_hash` here. In other words, convert `row` to a hash before creating a Dino.

dino_catalog/data_reader.rb
```diff
@@ -0,0 +1,47 @@
...
+    diet: :diet,
+    weight_in_lbs: :weight,
+    description: :description,
+  }
+
+  def initialize(csv_files)
+    read_data(csv_files)
+  end
+
+  def read_data(csv_files)  # csv_files = file containing list of csv filenames
```
paulh16: Consider making `read_data` and `translate_header` private methods.

paulh16: Is there an opportunity to extract a method to make sure the method does one thing? This method is rather long and doing several things.

dino_catalog/data_reader.rb
djkotowski: Prefer using `#map` here.

dino_catalog/data_reader.rb
```diff
@@ -0,0 +1,47 @@
...
+        new_dino = Dino.new(row.to_hash)
+        dino_objects.push(new_dino)
+      end
+    end
+    # Create DinoFilters object so that methods in dino_filters can be used
+    @dino_filter = DinoFilters.new(dino_objects)
+  end
+
+  def translate_header(header_name)
+    unless TRANSLATE[header_name.downcase.to_sym]
```
djkotowski: Take a look at `Hash#fetch` for this check.

dino_catalog/data_reader.rb
```diff
@@ -0,0 +1,47 @@
...
+  def initialize(csv_files)
+    read_data(csv_files)
+  end
+
+  def read_data(csv_files)  # csv_files = file containing list of csv filenames
+    dino_objects = []
+
+    # Each row in csv file is a separate dino object
+    CSV.foreach(csv_files) do |new_file|
+      filename = new_file[0]
```
djkotowski: This confuses me.  What are the contents of `new_file`?  Retrieving a filename from the first element of an array doesn't make sense to me.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+    # Each instance variable in Dino is a key in the dino_data dictionary.
+    dino_data.each do |attribute, value|
+      instance_variable_set("@#{attribute}", value)
+    end
+
+    # Some instance fields need special treatment
+    @continent ||= 'Africa'
+
+    @diet = 'Carnivore' if @diet == 'Yes'
+    @diet = 'Not carnivore' if @diet == 'No'
```
djkotowski: Herbivore

dino_catalog/user_input.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+    dino_filters.summarize_list
+  end
+
+  def select_filter(var, var_info, dino_filters)
+    return dino_filters.find_bipeds if var == "bipeds"
+    return dino_filters.find_carnivores if var == "carnivores"
+    return dino_filters.find_big_dinos(var_info) if var == "big_dinos"
+    return dino_filters.find_small_dinos(var_info) if var == "small_dinos"
+    return dino_filters.find_dinos_specific_era(var_info) if var == "era"
```
djkotowski: This could be accomplished using a hash to map command with filter.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+  attr_reader :description
+
+  def initialize(dino_data)
+    # Each instance variable in Dino is a key in the dino_data dictionary.
+    dino_data.each do |attribute, value|
+      instance_variable_set("@#{attribute}", value)
+    end
+
+    # Some instance fields need special treatment
+    @continent ||= 'Africa'
```
paulh16: Is there a different way to handle this special instance? What is we had another type of continent that required special treatment? This is a code smell.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    @diet = 'Carnivore' if @diet == 'Yes'
+    @diet = 'Not carnivore' if @diet == 'No'
+
+    @weight = @weight.to_i unless @weight.nil?
```
paulh16: Instead of `unless @weight.nil?`, you should be able to use `if @weight`.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+    dino_values.reject! { |dino_string| dino_string.nil? }
+    dino_values
+  end
+
+  def format_string(attribute, value)
+    return "#{value}  lbs" if attribute == "weight"
+    value
+  end
+
+  def carnivore?
```
paulh16: Nice method

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+    ["Carnivore", "Insectivore", "Piscivore"].include?(@diet)
+  end
+
+  def small_dino?(weight_cutoff)
```
paulh16: I'm not sure about the logic of this method and `small_dino`. If there is no weight, you are saying `small_dino?` and `big_dino?`, are false. By returning false, you are telling me that the dino is either big or small, but you don't really know for sure. Thus, I don't think this is the best way to handle these two methods.

dino_catalog/dino.rb
paulh16: Instead of placing this line in the `initialize` method, can you extract this into its own, well-named method? Perhaps something like `#set_weight`?

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+    @continent ||= 'Africa'
+
+    @diet = 'Carnivore' if @diet == 'Yes'
+    @diet = 'Not carnivore' if @diet == 'No'
+
+    @weight = @weight.to_i unless @weight.nil?
+  end
+
+  def summarize_dino
+    dino_values = instance_variables.map do |var|
```
paulh16: Is `var` well named?

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+      dino_info = instance_variable_get("#{var}")
+      if dino_info
+        attribute = var.to_s.slice(1..-1)
+        "#{attribute} : #{format_string(attribute, dino_info)} \n"
+      end
+    end
+    dino_values.reject! { |dino_string| dino_string.nil? }
```
paulh16: Consider using `#select` if possible. This should work:

`dino_values.select! { |dino_string| dino_string }`

Also, you shouldn't have to return `dino_values` in the next line as line 33 will return `dino_values`after `select` is called on it.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+    dino_values
+  end
+
+  def format_string(attribute, value)
```
paulh16: Does this method name clearly convey want the method does?

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+    @weight && @weight < weight_cutoff
+  end
+
+  def big_dino?(weight_cutoff)
+    @weight && @weight > weight_cutoff
+  end
+
+  def dino_in_era?(era_of_interest)
+    # ignore case of the text in era with /i in regular expression
+    !@era[/#{era_of_interest}/i].nil?
```
paulh16: See if there is a way to simply this method. We can discuss during our pairing session this week.

dino_catalog/user_input.rb
```diff
@@ -0,0 +1,74 @@
+require_relative './dino_filters.rb'
+
+ERAS_IN_DATA = ["Cretaceous (Early and Late)", "Jurassic", "Oxfordian",\
+                "Late Permian", "Triassic"]
+
+class UserInput
+  FILTERS = %w(bipeds carnivores era big_dinos small_dinos)
```
paulh16: Nice use of an alternative way to build an array of strings!

dino_catalog/user_input.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  attr_reader :bipeds
+  attr_reader :carnivores
+  attr_reader :era
+  attr_reader :big_dinos
+  attr_reader :small_dinos
+
+  def query_user
```
paulh16: An important task to do when asking for user data is sanitizing it and ensuring it is proper. What if the user enters nothing, or something that doesn't make sense. Will your code handle that and provide feedback to the user? Something to think about, but not necessary for this exercise.

dino_catalog/dino.rb
djkotowski: I'm fine with reject here, though pass the method as an argument as Hound suggests.

dino_catalog/dino.rb
djkotowski: I think that's exactly what we want here.  We _don't_ really know for sure.

dino_catalog/dino.rb
djkotowski:
```ruby
@era.downcase.include?(era_of_interest.downcase)
```
would do the same thing here

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def biped?
+    @walking_style[/Biped/] == "Biped"
```
djkotowski: `@walking_style.include?("Biped")`

dino_catalog/dino_filters.rb
```diff
@@ -0,0 +1,50 @@
...
+    filtered_list = @dinos.select { |dino| dino.small_dino?(weight_cutoff) }
+    DinoFilters.new(filtered_list)
+  end
+
+  def summarize_list
+    summarized_dinos = []
+    @dinos.each do |dino|
+      summarized_dinos += dino.summarize_dino
+    end
+    summarized_dinos
```
djkotowski: Use `#map` here.

dino_catalog/user_input.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    puts "filter for big dinos (enter the min weight cutoff or leave blank)?"
+    self.big_dinos = gets.chomp
+
+    puts "filter for small dinos (enter the max weight cutoff or leave blank)?"
+    self.small_dinos = gets.chomp
+  end
+
+  def bipeds=(value)
+    @bipeds = true if value =~ /^Y/i # otherise nil
```
paulh16: Consider adding `$` to constrain the end of line for the regex: `/^Y$/`.

dino_catalog/user_input.rb
```diff
@@ -0,0 +1,74 @@
...
+    self.era = gets.chomp
+
+    puts "filter for big dinos (enter the min weight cutoff or leave blank)?"
+    self.big_dinos = gets.chomp
+
+    puts "filter for small dinos (enter the max weight cutoff or leave blank)?"
+    self.small_dinos = gets.chomp
+  end
+
+  def bipeds=(value)
```
paulh16: Consider renaming `value` to something like `input` to be more meaningful.

dino_catalog/user_input.rb
```diff
@@ -0,0 +1,74 @@
...
+  def era=(value)
+    if value =~ /^[1]/
+      ERAS_IN_DATA.each { |option| puts option }
+      value = gets.chomp
+    end
+    @era = value if value =~ /^[A-Z]+/i
+  end
+
+  def big_dinos=(value)
+    @big_dinos = value.to_i if value.to_i != 0
```
paulh16: An alternative idiom to consider is `@big_dinos = value.to_i unless value.zero?`.

dino_catalog/user_input.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def small_dinos=(value)
+    @small_dinos = value.to_i if value.to_i != 0
+  end
+
+  def perform_user_query(dino_filters)
+    FILTERS.each do |var|
```
paulh16: Consider renaming `var` and `var_info` to be more meaningful.

dino_catalog/user_input.rb
```diff
@@ -0,0 +1,74 @@
...
+    dino_filters.summarize_list
+  end
+
+  def select_filter(var, var_info, dino_filters)
+    return dino_filters.find_bipeds if var == "bipeds"
+    return dino_filters.find_carnivores if var == "carnivores"
+    return dino_filters.find_big_dinos(var_info) if var == "big_dinos"
+    return dino_filters.find_small_dinos(var_info) if var == "small_dinos"
+    return dino_filters.find_dinos_specific_era(var_info) if var == "era"
+    raise "There was a problem."
```
paulh16: Good job adding an error check.

dino_catalog/dino_filters.rb
```diff
@@ -0,0 +1,50 @@
...
+
+class DinoFilters
+  attr_reader :dinos
+
+  def initialize(input_list)
+    @dinos = input_list
+  end
+
+  def find_bipeds
+    filtered_list = @dinos.select(&:biped?)
```
paulh16: Very nice use of the `&` idiom in the block!

dino_catalog/ChristineREADMEfile
```diff
@@ -0,0 +1,40 @@
+There are two classes in this repository.  One class, Dinos, (in the
```
paulh16: Great job on adding a well-documented `README`!

dino_catalog/data_reader.rb
```diff
@@ -0,0 +1,47 @@
...
+        dino_objects.push(new_dino)
+      end
+    end
+    # Create DinoFilters object so that methods in dino_filters can be used
+    @dino_filter = DinoFilters.new(dino_objects)
+  end
+
+  def translate_header(header_name)
+    unless TRANSLATE[header_name.downcase.to_sym]
+      raise KeyError,  'Ensure CSV headers are in the translate dictionary.'
```
paulh16: Good job thinking about error conditions.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,62 @@
...
+  attr_reader :era
+  attr_reader :continent
+  attr_reader :diet
+  attr_reader :walking_style
+  attr_reader :weight
+  attr_reader :description
+
+  def initialize(dino_data)
+    # Each instance variable in Dino is a key in the dino_data dictionary.
+    dino_data.each do |attribute, value|
```
paulh16: Consider extracting this method. We can do some of these refactorings during our pairing this week.

paulh16: But nice way to set the instance variables.

dino_catalog/dino_filters.rb
```diff
@@ -0,0 +1,50 @@
...
+    DinoFilters.new(filtered_list)
+  end
+
+  def find_carnivores
+    filtered_list = @dinos.select(&:carnivore?)
+    DinoFilters.new(filtered_list)
+  end
+
+  def find_dinos_specific_era(era_of_interest)
+    filtered_list = @dinos.select { |dino| dino.dino_in_era?(era_of_interest) }
```
paulh16: Nice method!

dino_catalog/dino_filters.rb
```diff
@@ -0,0 +1,50 @@
...
+    filtered_list = @dinos.select { |dino| dino.big_dino?(weight_cutoff) }
+    DinoFilters.new(filtered_list)
+  end
+
+  def find_small_dinos(weight_cutoff)
+    filtered_list = @dinos.select { |dino| dino.small_dino?(weight_cutoff) }
+    DinoFilters.new(filtered_list)
+  end
+
+  def summarize_list
```
paulh16: Consider using `inject` when trying to build an object while iterating over a collection. It's a very handy method. Let's do this when we pair.

dino_catalog/dino_filters.rb
schwerdt: Is this because it saves creating a new object in memory each iteration?  I ran this loop printing the object id of summarized_dino using the do loop and get a different object id each iteration.  I tried that again with map.  During the loop, summarized_dinos does not change its object id, but it looks like that is because it is not appended to during the loop.  Is there some temporary variable created that acts like summarized_dino and also is recopied each time something is appended, or am I avoiding recopying by using map?

paulh16: We can delve into this during our pairing session.

paulh16:
Good job overall on this! I liked your use of Ruby's idioms and ability to accept user input, as well as your separation of classes. We will delve into some of the comments during our pairing session.

I recommend this as done, @djkotowski

dino_catalog/dino.rb
djkotowski: Verify that the attribute is valid before setting the variable.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def dino_in_era?(era_of_interest)
+    # ignore case of the text in era with /i in regular expression
+    #!@era[/#{era_of_interest}/i].nil?
+    @era.downcase.include?(era_of_interest.downcase)
+  end
+
+  def biped?
+    @walking_style[/Biped/] == "Biped"
```
djkotowski: Preferred alternatives:
```ruby
@walking_style.include?("Biped")

!!(@walking_style =~ /Biped/)

!!(@walking_style.match(/Biped/))
```

dino_catalog/dino.rb
```diff
@@ -0,0 +1,64 @@
...
+  def summarize_dino
+    dino_values = instance_variables.map do |dino_attribute|
+      dino_info = instance_variable_get("#{dino_attribute}")
+      if dino_info
+        attribute = dino_attribute.to_s.slice(1..dino_attribute.length)
+        "#{attribute} : #{formatted_string(attribute, dino_info)} \n"
+      end
+    end
+    dino_values.select { |dino_string| dino_string }
+  end
```
djkotowski: Arrowhead code here.  Also, this seems like it is a console output concern, not a concern for Dino.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  private
+
+  def formatted_string(attribute, value)
+    return "#{value}  lbs" if attribute == "weight"
+    value
+  end
```
djkotowski: This also seems like an output concern.

dino_catalog/main.rb
```diff
@@ -0,0 +1,19 @@
...
+dino_files = ["dinodex.csv", "african_dinosaur_export.csv"]
+
+dino_filter = DataReader.new(dino_files).dino_filter
+
+user_input = UserInput.new
+user_input.query_user
+
+dino_summary =  user_input.perform_user_query(dino_filter)
+
+if dino_summary.length > 0
```
djkotowski:
```ruby
if dino_summary.any?
```

dino_catalog/user_input.rb
```diff
@@ -0,0 +1,81 @@
+require './dino_filters.rb'
+
+ERAS_IN_DATA = ["Cretaceous (Early and Late)", "Jurassic", "Oxfordian",\
```
djkotowski: Why is this constant outside a class?

dino_catalog/user_input.rb
djkotowski: No need to change this (it will take too much time), but most applications wouldn't ask for filters like this.  In many applications, commands accept flags or arguments identifying what you want filtered.

Possible Example:
```
filter --biped --big --diet=Herbivore
```

dino_catalog/user_input.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    puts "filter for big dinos (enter the min weight cutoff or leave blank)?"
+    self.big_dinos = gets.chomp
+
+    puts "filter for small dinos (enter the max weight cutoff or leave blank)?"
+    self.small_dinos = gets.chomp
+  end
+
+  def bipeds=(input)
+    @bipeds = true if input =~ /^Y/i # otherise nil
```
djkotowski: Why `nil` and not `false`?

dino_catalog/user_input.rb
```diff
@@ -0,0 +1,81 @@
...
+  def era=(input)
+    if input =~ /^[1]/
+      ERAS_IN_DATA.each { |era_option| puts era_option }
+      input = gets.chomp
+    end
+    @era = input if input =~ /^[A-Z]+/i
+  end
+
+  def big_dinos=(input)
+    @big_dinos = input.to_i if input.to_i != 0
```
djkotowski: Prefer
```ruby
@big_dinos = input.to_i unless input.to_i != 0
```

djkotowski:
Good job!  No need to refactor, but definitely take a look at the comments.  Pair with @paulh16 if you get a chance.

dino_catalog/dino_filters.rb
schwerdt: cool thanks!

PR #638, opened 2015-04-28
Data science

schwerdt:
I changed the cohort class so that it only has the responsibility of building Cohort objects.  Now there is a DataParser class that is responsible for reading the data from the file and extracting the needed quantities.


data_science/ab_test.rb
```diff
@@ -0,0 +1,34 @@
+require 'abanalyzer'
+class ABtest
```
djkotowski: Class should be ABTest.

data_science/ab_test.rb
```diff
@@ -0,0 +1,34 @@
...
+  attr_reader :cohort_a
+  attr_reader :cohort_b
+  attr_reader :leader
+
+  def initialize(cohort_a, cohort_b)
+    @cohort_a = cohort_a
+    @cohort_b = cohort_b
+    @leader = find_leader
```
djkotowski: This could just be a method called `leader`.

data_science/ab_test.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def compute_leader_confidence_level
+    sample_data = {}
+    sample_data[@cohort_a.name] = { failures: compute_failures(@cohort_a),
```
djkotowski: You're repeating hash definitions here.  You could just use `[@cohort_a, @cohort_b].each_with_object` to construct `sample_data`.

data_science/ab_test.rb
```diff
@@ -0,0 +1,34 @@
...
+  private
+
+  def find_leader
+    return @cohort_a.name if @cohort_a.conversion_rate > @cohort_b.conversion_rate
+    return @cohort_b.name if @cohort_b.conversion_rate > @cohort_a.conversion_rate
+    return nil if @cohort_a.conversion_rate == @cohort_b.conversion_rate
+  end
+
+  def compute_failures(cohort)
+    cohort.sample_size - cohort.num_conversions
```
djkotowski: This belongs on `Cohort`.

data_science/cohort.rb
```diff
@@ -0,0 +1,26 @@
...
+  attr_reader :num_conversions
+  attr_reader :conversion_rate
+  attr_reader :std_dev
+
+  def initialize(name, sample_size, num_conversions)
+    @name = name
+    @sample_size = sample_size
+    @num_conversions = num_conversions
+    @conversion_rate = @num_conversions.to_f / @sample_size
+    @std_dev = compute_std_dev
```
djkotowski: This is redundant (@std_dev is set in `compute_std_dev`).  Just change `compute_std_dev` to `std_dev` and don't set instance variables.  No need to when the value is calculated.

data_science/cohort.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def compute_confidence_interval_95pct
```
djkotowski: Should just be `confidence_interval`.

data_science/data_parser.rb
```diff
@@ -0,0 +1,30 @@
...
+    find_cohorts_and_conversions(input_data)
+  end
+
+  private
+
+  def read_data(input_file)
+    JSON.parse(File.read(input_file))
+  end
+
+  def find_cohorts_and_conversions(input_data)
```
djkotowski: When you have the word "and" in your method, it's probably a sign that you're doing too much.

data_science/data_parser.rb
```diff
@@ -0,0 +1,30 @@
...
+    unique_cohorts = Hash.new(0)
```
djkotowski: Use `{}` instead of `Hash.new`.

data_science/data_parser.rb
```diff
@@ -0,0 +1,30 @@
...
+    unique_cohorts_conversions = Hash.new(0)
+    input_data.each do |data_point|
+      cohort_name = data_point["cohort"].to_sym
+      unique_cohorts[cohort_name] += 1
+      unique_cohorts_conversions[cohort_name] += data_point["result"].to_i
+    end
+    @unique_cohorts = unique_cohorts
```
djkotowski: Why set instance variables equal to the scope variable instead of just using instance variables the entire time?

djkotowski: Also, since you're not running a "unique" operation anywhere, why not just name the variable `cohorts`?

data_science/data_parser.rb
```diff
@@ -0,0 +1,30 @@
...
+    @unique_cohorts_conversions =  unique_cohorts_conversions
```
djkotowski: Ditto above on naming.

data_science/main.rb
```diff
@@ -0,0 +1,40 @@
+require 'json'
+require_relative './cohort.rb'
+require_relative './ab_test.rb'
+require_relative './data_parser.rb'
+
+def print_cohort_info(cohort)
+  interval = cohort.compute_confidence_interval_95pct
+  conversion_rate = cohort.conversion_rate * 100
+  printf("Cohort %s sample size: %d and conversions: %d \n", cohort.name,
+                              cohort.sample_size, cohort.num_conversions)
```
djkotowski: When possible, prefer interpolation to `printf`.

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  it "returns the standard deviation for this cohort" do
+    expect(cohort.std_dev).to(eql(STD_DEV))
+  end
+
+  it "returns the 95% confidence level for this cohort" do
+    confidence_interval = []
+    confidence_interval.push(CONVERSION_RATE - CONFIDENCE_95PCT)
+    confidence_interval.push(CONVERSION_RATE + CONFIDENCE_95PCT)
```
djkotowski: You don't need to separate the array initialization from its value assignments.  This also belongs in a `let`.

```ruby
let(:expected_confidence_interval) do
  [CONVERSION_RATE - CONFIDENCE_95PCT,
   CONVERSION_RATE + CONFIDENCE_95PCT]
end
# ...
expect(cohort.confidence_interval).to eql(expected_confidence_interval)
```

data_science/spec/data_parser_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+LIST_LENGTH = 2
+SAMPLE_SIZE = 72
+NUM_CONVERSIONS = 5
+COHORT_NAME = :A
+
+describe DataParser do
+  let(:parsed_data) { DataParser.new('smaller_data_set.json') }
+
+  describe("#new") do 
+    it "has 2 instance variables which are dictionaries" do
```
djkotowski: This is a very vague description and doesn't describe `#new`.

data_science/spec/data_parser_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+     expect(parsed_data.unique_cohorts_conversions.keys.length).to(eql(LIST_LENGTH))
+   end
+
+  it "has an instance variable with key :A and  sample_size as the value" do 
+    expect(parsed_data.unique_cohorts[COHORT_NAME]).to(eql(SAMPLE_SIZE))
+  end
+
+  it "has an instance variable with key :A and num_conversions as the value" do 
+    expect(parsed_data.unique_cohorts_conversions[COHORT_NAME]).to(eql(NUM_CONVERSIONS))
+  end
```
djkotowski: Stating that a class has instance variables tells me nothing about the class.  Use `describe` to describe your methods with tests.

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+  end
+
+  describe("#num_conversions") do
+    it "returns the number of successes" do
+      expect(cohort.num_conversions).to(eql(NUM_CONVERSIONS))
+    end
+  end
+
+  it "returns the conversion rate for this cohort" do
```
djkotowski: This and the tests below look like they were meant to be wrapped in `describe` blocks.

data_science/spec/ab_test_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+LEADING_COHORT_NAME = :B
+CONFIDENCE_IN_LEADER = 0.3640661012729629
+
+describe ABtest do
+let(:cohort_a) { Cohort.new(COHORT1_NAME, COHORT1_SAMPLE_SIZE, COHORT1_NUM_CONVERSIONS) }
+let(:cohort_b) { Cohort.new(COHORT2_NAME, COHORT2_SAMPLE_SIZE, COHORT2_NUM_CONVERSIONS) }
+let(:ab_test) { ABtest.new(cohort_a, cohort_b) }
+
+  describe("#new") do
+    it "has 2 instance variables, cohort_A, cohort_B of type Cohort" do
```
djkotowski: `#new` doesn't have 2 instance variables.  Also, as I discuss partially below, we don't care if a class has instance variables.  That detail is not a concern beyond the class.  Instead, you could say an ABTest has a pair of cohorts, or you could describe their readers in `describe` blocks.

djkotowski:
Good iteration.  Take a look at the comments, refactor and resubmit.

Regarding your tests, think of them as documentation for users of your code.  Treat your classes as black boxes.  The user using your class doesn't care about how it works, they just want to know how to use it.  What do the methods' classes accept as arguments?  What is the expected output?  What if input is invalid?

data_science/ab_test.rb
schwerdt: Do you mean that I should make the function called 'leader' and then remove the instance variable @leader?

data_science/cohort.rb
schwerdt: Same thing with @leader in the ABTest class?

data_science/data_parser.rb
schwerdt: I am finding the number of trials and number of conversions in this routine because then I only have to loop over the data once. 

Once I have the number of trials for each cohort, is there a cleaner ruby way to do this than looping again?

data_science/data_parser.rb
schwerdt: I want to initialize the value of each key to zero the first time the key appears.  I did this to avoid having to check whether a key is new or has been seen before for each piece of data.  Is that a bad idea?

data_science/ab_test.rb
djkotowski: Yes.

data_science/cohort.rb
djkotowski: Exactly.

data_science/data_parser.rb
djkotowski: This is a representation of the cohort views.  You could merge this and conversions by creating a structure like:

```ruby
{ A: { conversions: 10, views: 25 },
  B: { conversions: 5, views: 20 }}
```

or just using mapping directly to Cohorts.

PR #642, opened 2015-05-04
DinoDex Exercise

dustinhansen:


aeberlin:
@dustinhansen I will finish reviewing this in the morning. Feel free to knock out all those pesky rubocop violations, though, in the mean time! Cheers.

dustinhansen:
I'll do my best, but there's a few of these "violations" I'm not understanding very well.

djkotowski:
Your editor should be set to use 2 spaces for indentation instead of tabs.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,102 @@
...
+		}
+	end
+
+	# Search for dinosaurs based on various criteria
+	
+	def search(search_filters)
+
+		results = []
+
+		search_filters.each { |key,value|
```
djkotowski: Multiline blocks use `do ... end` syntax instead of `{ ... }`.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,102 @@
...
+			'DIET' 				=> diet,
+			'WEIGHT_IN_LBS' 	=> row['Weight'],
+			'WALKING' 			=> row['Walking'],
+			'DESCRIPTION' 		=> nil
+		}
+	end
+
+	# Search for dinosaurs based on various criteria
+	
+	def search(search_filters)
```
djkotowski: This method follows the [Arrowhead Pattern](https://leveluprails.com/ruby.html#arrowhead).  Break the method and decision structures apart to make it more readable.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,102 @@
...
+
+		# return as JSON
+		puts results.to_json
+	end
+end
+
+dinodex = DinoDex.new
+
+# Grab all dinosaurs that were bipeds
+dinodex.search({ 'WALKING' => 'Biped' })
```
djkotowski: When passing a hash as a parameter, you don't need the braces.
```ruby
dinodex.search('WALKING' => 'BIPED')
```
works as well.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  # Read Pirate Bay CSV file and merge the results
+
+  CSV.foreach('african_dinosaur_export.csv', headers: true) do |row|
+    diet = row['Carnivore'].downcase == 'yes' ? 'Carnivore' : nil
+
+    @@dino_database << {
+      NAME: row['Genus'],
+      PERIOD: row['Period'],
+      CONTINENT: nil,
```
aeberlin: Hint: `african_dinosaur_export.csv`

I wonder what continent these dinosaurs are from ;)

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+end
+
+dinodex = DinoDex.new
+
+# Grab all dinosaurs that were bipeds
+dinodex.search(WALKING: 'Biped')
+
+# Grab all the dinosaurs that were carnivores (fish and insects count).
+dinodex.search('DIET' => %w('Carnivore', 'Insectivore', 'Piscivore'))
```
aeberlin: Any particular reason you switch between hash-colon and hash-rocket styles?

aeberlin:
Overall, I think that this is a good first iteration. Simple and lightweight. That being said, though, I do have some comments, but it's up to you whether you'd like to address them as they are beyond the required scope of this assignment (outside of point #1). A lot of the activities we're involved in, as developers, on a day-to-day basis leave requirements open to interpretation and subject to a sort of balancing act. Most of my comments will touch on this point.

1) I don't think that you're fully exercising the power of the CSV ruby library. I suggest taking a look at this particular bit of the library for help with parsing your `african_dinosaur_export.csv`, in a much faster fashion (additionally, look at header converters): http://ruby-doc.org/stdlib-2.0.0/libdoc/csv/rdoc/CSV.html#method-i-convert

2) Although this isn't a git workflow exercise, I highly suggest squashing your commits and giving your commits more meaningful messages. When you're assigned to a team, each commit on an active pull request will likely trigger a travis build, and you'll want to ask yourself whether or not it makes sense to test a build (on travis, test/audit with rubocop/rspec locally) for each of these commits individually, or if it makes more sense to conserve resources until your branch is in better standing. The additional question is whether or not it makes sense to unitize these commits for history -- is it more difficult for people to track what I did to this codebase when I squash commits? It is very much so a balancing act between maintaining relevant history, conserving build process resources, and being expressive about the code you write. Food for thought.

3) This is an ultra-lightweight, exact-to-specifications implementation. You can retrieve the specified query results, and display them. And, in that regard, it's excellent. I would emphasize that the output is not easy on the eyes, and would recommend either using a REPL interface such as `pry`, which would format console output of hashes and enable a query interface (although that's a bonus).

4) Sorry, I'll form this point via a pointed question: How would you handle an additional CSV data source, which has differing header names, similar to `african_dinosaur_export.csv` versus `dinodex.csv`? Would you write more code? Would it make more sense to define a configuration mechanism which can be updated more safely to include these additional data sources? Linear code is significantly more volatile than explicitly coded interfaces, which are interpreted based on a configuration set. I could make this more long-winded, and if you'd like to chat about it more in person, I'll be in the office again tomorrow. In the mean time, though, it's just more food for thought.

Anyway, that's all for now, cheers!

dustinhansen:
Hi Adam,

Thanks for your feedback. To address your points:

1.) I'll definitely have a look.

2.) I usually don't do commit storms like that but I had no way of knowing what houndci/rubocop was going to complain about without committing. That whole aspect of the process was unknown to me. I now have an addon for my text editor that will validate with rubocop locally so this should be avoidable in the future.

3.) I'll have a look at that. If this were for production I would have spent more time ensuring that proper headers were in place, which would help when displayed in a browser.

4.) The way I'd like to go about it is to create some definitions that map which column headers should act as aliases for which existing headers. I'd likely do this via an array of arrays and make those class variables. If any new column names encountered that aren't found in the definitions then we would have to throw an error. Does that make sense?

aeberlin:
Hey @dustinhansen,

Sorry, I was out sick yesterday. I would suggest installing rubocop, that is what hound uses.

Regarding (4), I would suggest using a YML file to configure your transforms, and reference the hash tree inside of your header and field converters..

aeberlin:
Reference these rule configs with rubocop:

- https://github.com/jmmastey/level_up_exercises/blob/master/.rubocop.yml (inherits from `.hound.yml`)
- https://github.com/jmmastey/level_up_exercises/blob/master/.hound.yml

PR #643, opened 2015-05-05
Data science updated

schwerdt:


data_science/ab_compare.rb
```diff
@@ -0,0 +1,26 @@
...
+                                  successes: cohort.num_conversions }
+    end
+    compare_cohorts = ABAnalyzer::ABTest.new(sample_data)
+    1.0 - compare_cohorts.chisquare_p
+  end
+
+  def leader
+    return @cohort_a.name if @cohort_a.conversion_rate > @cohort_b.conversion_rate
+    return @cohort_b.name if @cohort_b.conversion_rate > @cohort_a.conversion_rate
+    return nil if @cohort_a.conversion_rate == @cohort_b.conversion_rate
```
ashtonthomas: I think all you need to do here is have `nil` and ruby will return the last line evaluated

data_science/main.rb
```diff
@@ -0,0 +1,38 @@
...
+  conversion_rate = cohort.conversion_rate * 100
+  puts "Cohort #{cohort.name} sample size: #{"%d" % cohort.sample_size}" +
+       " and conversions: #{"%d" % cohort.num_conversions}"
+  puts "Cohort #{cohort.name} conversion rate: #{"%2.2f" % conversion_rate}% " +
+       "and range #{"%2.2f" % (interval[0]*100)} - " + 
+       "#{"%2.2f" % (interval[1]*100)}% (at 95% confidence)."
+end
+
+def print_cohort_comparison(compare)
+  puts "Leading cohort is #{compare.leader} with #{"%2.2f" %(compare.compute_leader_confidence_level*100)}% confidence."
```
ashtonthomas: Ideally we would say "there is not clear leader" if we don't have enough confidence but this is tangent to our primary goal so let's ignore it ;-)

ashtonthomas:
@schwerdt this looks good. I found it easy to understand (and therefore easy to change, build upon).

Good job!

:+1:

PR #644, opened 2015-05-06
Basic Ruby Exercises

dustinhansen:


djkotowski:
Be sure to use different branches for each exercise in the future.  Otherwise, it will become a pain for bucketeers to review.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,109 @@
+class DinoDex
```
djkotowski: *Note for reviewers:* @aeberlin is reviewing this file.

arrowhead/arrowhead.rb
```diff
@@ -15,16 +15,18 @@ class Arrowhead
...
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        puts "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
+      identify_shape(CLASSIFICATIONS[region], shape)
     else
       raise "Unknown region, please provide a valid region."
     end
```
djkotowski: This can be further simplified using guard clauses.  You don't even have to extract the logic in classify to another method that way.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,109 @@
...
+  require 'csv'
```
djkotowski: `require` statements go before the class definition.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,109 @@
...
+dinodex.search('DIET' => %w(Carnivore Insectivore Piscivore))
+
+# Grab dinosaurs for specific periods
+dinodex.search('PERIOD' => 'Cretaceous')
+
+# Grab only big (> 2 tons) or small dinosaurs.
+dinodex.search('min_weight' => 2000)
+
+# Print out details of a specific dinosaur
+dinodex.search('NAME' => 'Albertonykus')
```
djkotowski: What if you want to chain searches?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  # Search for dinosaurs based on various criteria
+
+  def search(search_filters)
+    results = []
+    search_filters.each do |key, value|
+      results.push(generate_results(key, value))
+    end
+    puts results.to_json
```
djkotowski: Should `Dinodex` be responsible for output?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,109 @@
...
+    @@dino_database << {
+      NAME: row['Genus'],
+      PERIOD: row['Period'],
+      CONTINENT: nil,
+      DIET: diet,
+      WEIGHT_IN_LBS: row['Weight'],
+      WALKING: row['Walking'],
+      DESCRIPTION: nil,
+    }
+  end
```
djkotowski: L9-44 should be separate from the Dinodex class in a runner script.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,57 @@
...
+  def isosceles
+    [side1, side2, side3].uniq.length == 2
+  end
+
+  def scalene
+    if equilateral || isosceles
+      true
+    else
+      false
+    end
```
djkotowski: This can be simplified.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,57 @@
...
-		return [angleA, angleB, angleC]
-	end
-
-	def radians_to_degrees(rads)
-		return (rads * 180 / Math::PI).round
-	end
+  attr_accessor :side1, :side2, :side3
+
+  def initialize(side1, side2, side3)
+    @side1, @side2, @side3 = side1, side2, side3
```
djkotowski: Prefer one assignment per line.

robot_name/robot_name.rb
```diff
@@ -5,22 +5,21 @@ class Robot
...
-
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+    @name = generate_name
+    if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
```
djkotowski: It may make sense here to extract the regex logic out to a method.

dino_catalog/dinodex.rb
dustinhansen: No. I wrote it that way because of how limited the usage is here. I'll change that.

arrowhead/arrowhead.rb
dustinhansen: I was trying to stick to the "7 lines or less" rule for methods.

dustinhansen: You mean something like this? `return false if (CLASSIFICATIONS.include? region || CLASSIFICATIONS[region].include? shape)`

This line is too long according to hound and fails anyway for some weird reason.

dustinhansen: Something like this violates hound's rule for lines that are too long? `raise "Unknown region, please provide a valid region." if !if CLASSIFICATIONS.include? region`

djkotowski:
Good job!  Take a look at the comments and refactor where necessary.  Try to resubmit the exercises in separate PRs when you get the chance.  Take a look at `tig` and cherry-picking to help build the new branches.

PR #645, opened 2015-05-06
My Cool File

sfoa6558:


ashtonthomas:
@sfoa6558, sweet! Good use of a topic branch 'just_git' this is important as you will want to:
```
checkout master
chechout -b new_exercise_branch
```
so you cut a clean branch from master and don't include any other exercises in your PR's (you will never want to commit directly to master and you will always want to cut from master - for the purposes of LevelUp)

:+1:

PR #648, opened 2015-05-07
Dino

dietpop:
working and robocopped


dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,21 @@
+class Dino
+  attr_reader :all_info,
+              :name,
+              :period,
+              :continent,
+              :diet,
+              :weight,
+              :walking,
+              :description
+
```
zsyed1: You can put all those attributes on one line, that's the preferred convention:
`attr_reader :all_info, :name, :period, :continent, ...`

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,21 @@
...
+    @all_info = dino_row.to_s
+    @name = dino_row[:name]
+    @period = dino_row[:period]
+    @continent = dino_row[:continent]
+    @diet = dino_row[:diet]
+    @weight = dino_row[:weight]
+    @walking = dino_row[:walking]
+    @description = dino_row[:description]
+  end
+end
```
zsyed1: Cosmetic: You can vertically align the `=` signs to make it prettier

zsyed1: Also the variable name `dino_row` may as be called `row`.

dino_catalog/dino_csv_import.rb
```diff
@@ -0,0 +1,32 @@
+require 'csv'
+require './dino_class.rb'
+
+class CsvImporter
+  def initialize(input_files)
+    @raw_dino_info = Array[]
+    input_files.each { |fname| @raw_dino_info += import_csv_file(fname) }
+    @raw_dino_info
+  end
```
zsyed1: Line 6 - No need for the `Array`, just `@raw_dino_info = []` is fine.
Line 8 - You can remove this line.

On that note, lines 7 and 8 can be achieved with a `reduce`:
```
@raw_dino_info = input_files.reduce([]) { |a, filename| a + import_csv_file(filename) } 
```

dino_catalog/dino_csv_import.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def import_csv_file(filename)
+    csv = Array[]
+    CSV.foreach(filename, headers: true, header_converters: :symbol) do |row|
+      csv << row
+    end
+    csv
+  end
+
```
zsyed1: Line 12  - `csv = []`

dino_catalog/dino_csv_import.rb
```diff
@@ -0,0 +1,32 @@
...
+    @dinos = Array[]
+    @raw_dino_info.each_with_index do |row, i|
+      row[:name] = row[:genus] unless row[:genus].nil?
+      row[:continent] = "Africa" if row[:continent].nil?
+      row[:diet] = "Herbivore" if row[:carnivore] == "No"
+      row[:diet] = "Carnivore" if row[:diet] != "Herbivore" || row[:carnivore] == "yes"
+      row[:weight] = row[:weight_in_lbs] if row[:weight].nil?
+      row[:weight] = row[:weight].to_i
+      @dinos[i] = Dino.new(row)
+    end
```
zsyed1: Lines 22 reads better as: `row[:name] = row[:genus] if row[:genus]`

zsyed1: Line 26 - `row[:weight] ||= row[:weight_in_lbs]`

dino_catalog/dino_csv_import.rb
```diff
@@ -0,0 +1,32 @@
...
+    csv
+  end
+
+  def create_dinos
+    @dinos = Array[]
+    @raw_dino_info.each_with_index do |row, i|
+      row[:name] = row[:genus] unless row[:genus].nil?
+      row[:continent] = "Africa" if row[:continent].nil?
+      row[:diet] = "Herbivore" if row[:carnivore] == "No"
+      row[:diet] = "Carnivore" if row[:diet] != "Herbivore" || row[:carnivore] == "yes"
```
zsyed1: Line 23: `row[:continent] ||= "Africa"`

dino_catalog/dino_csv_import.rb
```diff
@@ -0,0 +1,32 @@
...
+      row[:weight] = row[:weight_in_lbs] if row[:weight].nil?
+      row[:weight] = row[:weight].to_i
+      @dinos[i] = Dino.new(row)
+    end
+    @dinos
+  end
```
zsyed1: Line 28 - Not sure why you need the index `i`. You can just push them one by one `@dinos << Dino.new(row)`. Then you can replace the `each_with_index` with `each` in line 21.

dino_catalog/dino_main.rb
```diff
@@ -0,0 +1,41 @@
+require './dino_csv_import.rb'
+require './dinodex_class.rb'
+
+class UserInterface
+  def initialize
+    dino_input_files = ['./dinodex.csv', './african_dinosaur_export.csv']
+    @my_dinodex = DinoDex.new(CsvImporter.new(dino_input_files).create_dinos)
+  end
```
zsyed1: Line 7 - Would be more readable if you do
```
dinos = CsvImporter.new(dino_input_files).create_dinos
@my_dinodex = DinoDex.new(dinos)
```

dino_catalog/dino_main.rb
```diff
@@ -0,0 +1,41 @@
...
+    print_output(@my_dinodex.filter_dinos(@filter_query))
+  end
+
+  def prompt_user_for_inputs(filter_name)
+    puts "Would you like to view dinosaurs by #{filter_name}? Y/N"
+    if gets.chomp.downcase == 'y'
+      puts "Choose the number of the #{filter_name} you wish to view:"
+      @filter_query[filter_name.to_sym] = print_values(filter_name)[gets.chomp.to_i - 1]
+    end
+  end
```
zsyed1: Line 21 - More readable if you do:
```
choice = gets.chomp.to_i - 1
@filter_query[filter_name.to_sym] = print_values(filter_name)[choice]
```

dino_catalog/dinodex_class.rb
```diff
@@ -0,0 +1,27 @@
...
+    matching_dinos
+  end
+
+  def get_all_values(key)
+    values = Array[]
+    @dino_array.each do |i|
+      values << i.send(key)
+    end
+    values.uniq
+  end
```
zsyed1: Lines 21 - 24 can be replaced with
```
@dino_array.map { |i| t.send(key) }
```

dino_catalog/dinodex_class.rb
```diff
@@ -0,0 +1,27 @@
+require './dino_class.rb'
+class DinoDex
+  def initialize(dino_arry)
+    @dino_array = dino_arry
+  end
```
zsyed1: Lines 3 - 4: Like to avoid using the variable type in the name. It locks you down to that type. You can just rename `dino_arry` to `dinos`.
Also, there is no need to worry about shadowing here, the argument variable is distinct from the member variable because of the `@`. So you can call them both the same name and not have to worry about shadowing:
```
def initialize(dinos)
  @dinos = dinos
end
```
Is perfectly acceptable

dino_catalog/dinodex_class.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def filter_dinos(filter)
+    matching_dinos = @dino_array
+    filter.each do |k, v|
+      next if v == ""
+      if k.to_s == "weight"
```
zsyed1: line 10 - `next if v.empty?`

dino_catalog/dinodex_class.rb
zsyed1: Nice work! Your code is succinct and easy to read. I would recommend making the changes and resubmitting - only because it is good to practice these basic conventions early on.

dino_catalog/dino_csv_import.rb
dietpop: Excellent suggestions!

dino_catalog/dino_csv_import.rb
dietpop: done

dino_catalog/dino_csv_import.rb
dietpop: Good one.

dino_catalog/dino_csv_import.rb
dietpop: Yes. Changed in applicable assignments.

dino_catalog/dino_main.rb
dietpop: done

dino_catalog/dino_class.rb
```diff
@@ -0,0 +1,21 @@
+class Dino
+  attr_reader :all_info,
+              :name,
```
djkotowski: Thank you Hound for not recognizing correct alignment.

dietpop: Neither does rubocop.

dino_catalog/dino_csv_import.rb
djkotowski: And at that point, just use `#map`.

djkotowski:
Your filenames should match the snake case version of your class/module names.

> dinodex.rb
```ruby
class Dinodex
end
```

dino_catalog/dinodex_class.rb
```diff
@@ -0,0 +1,27 @@
...
+        matching_dinos.select! { |i| (i.weight >= filter[:weight].to_i) }
+      else
+        matching_dinos.select! { |i| i.send(k).include? v }
+      end
+    end
+    matching_dinos
```
djkotowski: You could make this method chainable by returning a new Dinodex with only the filtered dinosaurs.

dino_catalog/dinodex_class.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def get_all_values(key)
+    values = Array[]
+    @dino_array.each do |i|
+      values << i.send(key)
+    end
+    values.uniq
```
djkotowski: Can be replaced with map.
```ruby
def get_all_values(key)
  @dinos.map { |i| i.send(key) }.uniq
end
```

dino_catalog/dino_csv_import.rb
zsyed1: In this case - yes. The preferred convention for a one-line `reduce` is to use `|a, e|` so you could probably get away with `|a, f|`!

dino_catalog/dino_main.rb
dietpop: The user's choice has to come after the print_values method displays the options.  I think this code solves the readability issue by separating the `gets` onto it's own line, but is it now too verbose? 

'user_choices = print_values(filter_name)
choice = gets.chomp.to_i - 1
@filter_query[filter_name.to_sym] = user_choices[choice]`

dino_catalog/dino_csv_import.rb
dietpop: done

djkotowski:
Good first submission!  Take a look at the comments and refactor where needed.

dino_catalog/dinodex_class.rb
dietpop: Another awesome recommendation!

dino_catalog/dinodex_class.rb
dietpop: Good point to not use the object type in a variable name.  Note to self: I used `array` within the variable name to keep things straight while coding, and it should have been removed when finished.

dino_catalog/dinodex_class.rb
dietpop: .empty? is not an integer method.  `weight` is an integer and the following logic works because weight is also handled differently than the the other filters.   This is not clear to reader and should be handled differently so it's more obvious.  I'll fix it.

PR #651, opened 2015-05-13
arrowhead

dietpop:


arrowhead/fixed_arrowhead.rb
```diff
@@ -0,0 +1,30 @@
+class Arrowhead
+  # This seriously belongs in a database.
```
paulh16: You can delete the comments. :)

arrowhead/fixed_arrowhead.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  REGION_MESSAGE = "Unknown region, please provide a valid region."
+  SHAPE_MESSAGE = "Unknown shape value. What you talking about, Willis?"
+
+  # FIXME: I don't have time to deal with this.
+  def self.classify(region, shape)
+    raise REGION_MESSAGE unless CLASSIFICATIONS.include?(region)
+    shapes = CLASSIFICATIONS[region]
+    raise SHAPE_MESSAGE unless shapes.include?(shape)
+    puts "You have a(n) '#{shapes[shape]}' arrowhead. Probably priceless."
```
paulh16: You should only need one `puts` statement. It can either be here or on line 30. I prefer on line 30, but it is up to your discretion.

arrowhead/fixed_arrowhead.rb
```diff
@@ -0,0 +1,30 @@
...
+    },
+  }
+
+  REGION_MESSAGE = "Unknown region, please provide a valid region."
+  SHAPE_MESSAGE = "Unknown shape value. What you talking about, Willis?"
+
+  # FIXME: I don't have time to deal with this.
+  def self.classify(region, shape)
+    raise REGION_MESSAGE unless CLASSIFICATIONS.include?(region)
+    shapes = CLASSIFICATIONS[region]
```
paulh16: This is very nit-picky, but one of the concepts in Avdi Grimm's Confident Ruby is to structure your methods into distinct responsibilities: collect input; perform work; and handle failure. He advises that each responsibility should grouped together. The finished method should consist of stripes with these responsibilities in distinct sections.

With that said, if you move the to `raise` conditions next to each other, you'll achieve this:

``` ruby 
  shapes = CLASSIFICATIONS[region] (collect input)
  raise REGION_MESSAGE unless CLASSIFICATIONS.include?(region) (handle failure)
  raise SHAPE_MESSAGE unless shapes.include?(shape)
  "You have a(n) '#{shapes[shape]}' arrowhead. Probably priceless." (perform work)
end
```

paulh16:
@djkotowski

@dietpop: This is excellent work. There was nothing that stood out. Your style was on the mark, and your solution was simple and clear.

arrowhead/fixed_arrowhead.rb
dietpop: Yes, excellent point of grouping input, biz logic, and output.

PR #652, opened 2015-05-13
Dino

dietpop:
Incorporated all of the code review comments.  

Down to 2 rubocop offenses that can be ignored:
dino.rb:3:15: C: Align the parameters of a method call if they span more than one line.
              :weight, :walking, :description
              ^^^^^^^
dino.rb:5:3: C: Method has too many lines. [8/7]
  def initialize(row)
  ^^^


dino_catalog/user_interface.rb
```diff
@@ -0,0 +1,44 @@
...
+    @filter_query.each_key { |k| prompt_user_for_inputs(k) }
+    p @filter_query
+    print_output(@my_dinodex.filter_dinos(@filter_query))
+  end
+
+  def prompt_user_for_inputs(filter_name)
+    puts "Would you like to view dinosaurs by #{filter_name}? Y/N"
+    return unless gets.chomp.downcase == 'y'
+    puts "Choose the number of the #{filter_name} you wish to view:"
+    user_choices =  print_values(filter_name)
```
dietpop: There is just no making you happy.  You need to pick your battles, hound.

PR #654, opened 2015-05-13
Dino

dietpop:
Incorporated all of the code review comments.  

Down to 1 rubocop offense that we discussed and decided to ignore:
dino.rb:5:3: C: Method has too many lines. [8/7]
  def initialize(row)
  ^^^


danvayn:
hey! getting to this now

dino_catalog/dino.rb
```diff
@@ -0,0 +1,15 @@
+class Dino
+  attr_reader :all_info, :name, :period, :continent, :diet,
+    :weight, :walking, :description
+
+  def initialize(row)
```
danvayn: row is an odd variable name to use -- it locks it to the idea that it's reading rows from a csv, which is true here but may not be in the future for whatever silly reason

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,27 @@
+require './dino.rb'
+class DinoDex
+  def initialize(dino)
+    @dinos = dino
```
danvayn: might as well just have dino in your initialize -- this is a bit misleading otherwise

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def filter_dinos(filter)
+    @matching_dinos = @dinos
+    filter.each do |k, v|
```
danvayn: more clear variable names don't hurt here

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,27 @@
...
+      next if v.to_s.empty?
+      select_dinos_per_attribute(k, v)
+    end
+    @matching_dinos
+  end
+
+  def select_dinos_per_attribute(k, v)
+    if k.to_s == "weight"
+      @matching_dinos.select! { |i| (i.weight >= v.to_i) }
```
danvayn: try to avoid using bang!s. for example, you could instead do a `#<<`

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,27 @@
...
+  def filter_dinos(filter)
+    @matching_dinos = @dinos
+    filter.each do |k, v|
+      next if v.to_s.empty?
+      select_dinos_per_attribute(k, v)
+    end
+    @matching_dinos
+  end
+
+  def select_dinos_per_attribute(k, v)
```
danvayn: select_dinos should probably be a private method, yea?

dino_catalog/dino_dex.rb
danvayn: what do you think about using an attribute handler to deal with when weight is passed instead of checking key?

dino_catalog/user_interface.rb
```diff
@@ -0,0 +1,44 @@
+require './csv_importer.rb'
+require './dino_dex.rb'
+
+class UserInterface
```
danvayn: not sure if i like this file/class being called UserInterface. to interact with a dinodex, would you rather run dino_dex.rb or user_interface.rb?

dino_catalog/user_interface.rb
```diff
@@ -0,0 +1,44 @@
...
+  def initialize
+    dino_input_files = ['./dinodex.csv', './african_dinosaur_export.csv']
```
danvayn: i think initialize should take optional parameters for specific files -- and if any options aren't passed, maybe it should check its own directory for .csv files instead!

dino_catalog/user_interface.rb
```diff
@@ -0,0 +1,44 @@
...
+    dinos = CsvImporter.new(dino_input_files).create_dinos
+    @my_dinodex = DinoDex.new(dinos)
+  end
+
+  def run_user_interface
```
danvayn: i think this should be the only public method

dino_catalog/user_interface.rb
```diff
@@ -0,0 +1,44 @@
...
+    puts "Would you like to view dinosaurs by #{filter_name}? Y/N"
+    return unless gets.chomp.downcase == 'y'
+    puts "Choose the number of the #{filter_name} you wish to view:"
+    user_choices = print_values(filter_name)
+    choice = gets.chomp.to_i - 1
+    @filter_query[filter_name.to_sym] = user_choices[choice]
+  end
+
+  def print_values(filter_name)
+    values = Array[]
```
danvayn: you can just type `values = []`

dino_catalog/user_interface.rb
```diff
@@ -0,0 +1,44 @@
...
+  def run_user_interface
+    @filter_query = { name: "", period: "", continent: "", diet: "", weight: 0,
+                      walking: "" }
+    @filter_query.each_key { |k| prompt_user_for_inputs(k) }
+    p @filter_query
+    print_output(@my_dinodex.filter_dinos(@filter_query))
+  end
+
+  def prompt_user_for_inputs(filter_name)
+    puts "Would you like to view dinosaurs by #{filter_name}? Y/N"
```
danvayn: it would be better to display the entire dino table at the start, as it is a dino index

dino_catalog/user_interface.rb
```diff
@@ -0,0 +1,44 @@
...
+    return unless gets.chomp.downcase == 'y'
+    puts "Choose the number of the #{filter_name} you wish to view:"
+    user_choices = print_values(filter_name)
```
danvayn: no extra credit? :cry:

dino_catalog/csv_importer.rb
```diff
@@ -0,0 +1,38 @@
...
+  def import_csv_file(filename)
+    csv = []
+    CSV.foreach(filename, headers: true, header_converters: :symbol) do |row|
+      csv << row
+    end
+    csv
+  end
+
+  def create_dinos
+    @raw_dino_info.map do |row|
```
danvayn: dino might be a better name than row

dino_catalog/csv_importer.rb
```diff
@@ -0,0 +1,38 @@
...
+require './dino.rb'
+
+class CsvImporter
+  def initialize(input_files)
+    @raw_dino_info = input_files.reduce([]) { |a, f| a + import_csv_file(f) }
+  end
+
+  def import_csv_file(filename)
+    csv = []
+    CSV.foreach(filename, headers: true, header_converters: :symbol) do |row|
```
danvayn: you can write your own header converters, and that'll get rid of 19 and 21 and the method in 27

dino_catalog/csv_importer.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+    csv
+  end
+
+  def create_dinos
+    @raw_dino_info.map do |row|
+      row[:name] ||= row[:genus]
+      row[:continent] ||= "Africa"
+      row = process_weight_field(row)
+      row = process_diet_field(row)
```
danvayn: i think you only need to do `row[:diet] =` here

dino_catalog/csv_importer.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def process_weight_field(row)
+    row[:weight] ||= row[:weight_in_lbs]
+    row[:weight] = row[:weight].to_i
+    row
+  end
+
+  def process_diet_field(row)
+    row[:diet] = (row[:carnivore] == "No") ? "Herbivore" : "Carnivore"
```
danvayn: row.delete(:carnivore) does this logic

djkotowski:
Good submission.  Take a look at the comments for improvements.  No need to resubmit.

PR #655, opened 2015-05-14
Overlord

schwerdt:


katieschilling:
Quick note re commit messages: 

first line should be a quick summary, 80 characters or less. Skip a line, then a brief description. Your messages are getting cut off, and some of them are a bit long. However, you are awesome at having descriptive commit messages, so keep that up.

Check out this blog post, but know I still use the -m flag, so it's all opinion: https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message

overlord/bomb.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def initialize(activation_code, deactivation_code)
+    @activation_code = set_code(activation_code, "1234")
+    @deactivation_code = set_code(deactivation_code, "0000")
+    @time_remaining = BOMB_DURATION # seconds
+    @active = false
+    @exploded = false
+    @num_deactivation_attempts = 3
+    @time_remaining = BOMB_DURATION # seconds
+    @explosion_time = nil  # This shouldn't be set until the bomb is activated
```
katieschilling: In order to make the initialize method neater: 
- You don't have to pass in default codes. They'd fit better as constants.
- Why are `@active` and `@exploded` their own instance variables? Why not a state variable with those as symbols?
- Having both`@time_remaining` and `@explosion_time` is redundant. BOMB_DURATION could also just be stored as the initial value for `@time_remaining`.
- All the attr_readers can live on one line; also you could use attr_accessor for :status and :number_of_attempts_remaining
- You only need to set ` @num_deactivation_attempts = 3` once in the initialize (not again on line 31), and to update it you can write if you use attr_accessor

`#update_remaining_time` contains two responsibilites, updating the remaining time and checking if the bomb should explode. Consider separating these into two methods and adding `#explode?` Such a method could contain the logic from line 49 as well.

overlord/overlord.rb
```diff
@@ -1,14 +1,86 @@
...
+  @user_bomb.update_remaining_time
+  update_bomb(@user_bomb)
+  return (@user_bomb.time_remaining).to_s if @user_bomb.time_remaining > 0
+end
+
+get '/bombexploded' do
+  @user_bomb = retrieve_bomb
+  locals = { bomb: @user_bomb }
+  return erb :boot_status unless @user_bomb
+  return erb :boot_status if @user_bomb && !@user_bomb.active && !@user_bomb.exploded
```
katieschilling: The goal of `Bomb` is to simplify what is stored in the session and in `Overlord`. By storing the bomb object in the session, you eliminate a major benefit of this structure. Storing objects in the session can let them get out of sync with the db. The only time I've seen it done well is when the object was composed over several actions then saved.

It's a small object here, but you don't want to make a habit of it.

overlord/bomb.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def set_code(code, default)
+    return code if code =~ /^\d{4}$/
+    return default if code == ""
+    raise BombCodeError, "Your code, #{code}, is invalid. Choose a 4 digit code."
```
katieschilling: - use immutable symbols instead of mutable strings for your state
- use attr_reader when you're going to read and write the values
- no need to pass in default codes since they're constants
- `#set_code` has three concerns right now: validating the code, returning the code, and raising an error. `#set_code` can be simplified by extracting an `#is_valid?` that is just `/^\d{4}$/`. I would `raise error unless act_code.is_valid?` and then set the code in the next line directly without a method call `@act_code = act_code.nil? DEFAULT_ACT_CODE : act_code`.
- Having both`@time_remaining` and `@explosion_time` is redundant. BOMB_DURATION could also just be stored as the initial value for `@time_remaining`.
- `#update_remaining_time` contains two responsibilities, updating the remaining time and checking if the bomb should explode. Consider separating these into two methods and adding `#explode?` Such a method could contain the logic from line 41 as well, and it would dry up the last 15 lines of the class.

overlord/overlord.rb
```diff
@@ -1,14 +1,68 @@
...
+
+post '/startbomb' do
+  @user_bomb = retrieve_bomb
+  return erb :boot_page unless @user_bomb
+  @user_bomb.start_bomb(params[:activation_code])
+  update_bomb(@user_bomb)
+  if @user_bomb.state == "active"
+    erb :countdown_page
+  else
+    erb :start_bomb_page_failed
```
katieschilling: This pattern of having a template for success and a template for failure will lead to a ton of templates in a bigger project. 

Look into using flash notices. They show at the top of your template and have nice built in styling. It's also what we use IRL. https://github.com/SFEley/sinatra-flash

overlord/overlord.rb
```diff
@@ -1,14 +1,68 @@
...
+  end
+end
+
+post '/attemptdeactivation' do
+  @user_bomb = retrieve_bomb
```
katieschilling: Give the same names to the templates and their corresponding states, and these guard statements can become a one-liner. Whenever you see repetitive code, DRY it up.

I like how you're separating the deactivation and updating bomb.

overlord/overlord.rb
```diff
@@ -1,14 +1,68 @@
...
 require 'sinatra'
+require './bomb.rb'
+require 'pry'
 
 enable :sessions
 
 get '/' do
-  "Time to build an app around here. Start time: " + start_time
+  session[:user_bomb] = nil
+  erb :boot_page
```
katieschilling: In line with my comment about naming the templates and states, you could put that template selection in your index and simply redirect to the index at the end of each route with an `after do` block. This will DRY out the rest of your routes.

overlord/overlord.rb
```diff
@@ -1,14 +1,68 @@
...
+get '/getremainingtime' do
+  @user_bomb = retrieve_bomb
+  @user_bomb.update_remaining_time
+  update_bomb(@user_bomb)
+  return (@user_bomb.time_remaining).to_s if @user_bomb.time_remaining > 0
+end
+
+get '/bombexploded' do
+  @user_bomb = retrieve_bomb
+  return erb :boot_page unless @user_bomb
```
katieschilling: This will be DRYed out if you do the template & state naming thing that we talked about.

overlord/public/update_time.js
```diff
@@ -0,0 +1,25 @@
...
+        clearInterval(status_checker);   
+        window.location.replace('/bombexploded'); }
+      } 
+    }
+    xmlhttp.open("GET","/getremainingtime",true);
+    xmlhttp.send();
+}
+
+function callUpdateTime() {
+  status_checker = setInterval(updateTime,500);
```
katieschilling: Instead of hitting your `/getremainingtime` endpoint every 500ms, see if you can only check the time on state change or when you get to the explosion time. This could result in storing the explosion time in the session so that your JS knows when to explode based on timeout. You could still send a message to explode or deactivate based on state.

Think about messages! Sandi Metz!

I can help with this. :)

katieschilling:
Using the flash notices for errors will help cut down on the logic within the erbs and reduce the number. I can comment on individual erbs if needed, but I expect there will be changes based on the other comments.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+VALID_DEACT_CODE = "1357"
+INVALID_ACT_CODE = "abcd"
+INVALID_DEACT_CODE = "hello"
+TIME_REMAINING = 10
+
+describe Bomb do
+  let(:newbomb_valid) { Bomb.new(VALID_ACT_CODE, VALID_DEACT_CODE) }
+  let(:newbomb_invalid) { Bomb.new(INVALID_ACT_CODE, INVALID_DEACT_CODE) }
+
+  it "is in an inactive state before start_bomb is called" do
```
katieschilling: phrase these tests as a description of what the method does, not the bomb.
`describe "#start_bomb" do
  context "with invalid activation code" do
    it "sets state to :started" do`

overlord/features/activate_bomb.feature
```diff
@@ -0,0 +1,21 @@
...
+
+  Scenario: enter the correct activation code to start the bomb
+    Given I successfully boot the bomb
+    When I submit the correct activation code
+    Then I see the bomb timer start
+
+  Scenario: enter the incorrect activation code and do not start the bomb
+    Given I successfully boot the bomb
+    When I submit incorrect activation code
+    Then I see an error message and a prompt to activate
```
djkotowski: This sounds like two steps.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,89 @@
...
+When(/^I submit an invalid deactivation code$/) do
+  fill_in('deactivation_code', with: "2468")
+  click_button('Submit')
+end
+
+Then(/^I see the bomb has not deactivated$/) do
+  expect(page).to have_content("attempt(s) remain")
+end
+
+When(/^I submit (\d+) invalid activation codes$/) do |arg1|
```
djkotowski: Use a better name than `arg1` here.

overlord/views/bomb_status.erb
```diff
@@ -0,0 +1,26 @@
...
+       height: 25px; 
+       width: 500px; 
+       }
+  </style>
+
+  <% if bomb.active %>
+     <h2> Time remaining: </h2> 
+     <h3 id="time_remaining"> </h3>
+     <form action="/attemptdeactivation" method = "POST">
+       Deactivation Code: <input type="text" name = "deactivation_code" class="deactivation_code"> <br>
```
djkotowski: If you wrap "Deactivation Code:" with a `<label>` tag linked to your text field, clicking on it will bring focus to the text field.  Also, avoid `<br />` tags.

_Note: This is not the point of this exercise, so don't worry about refactoring.  Just keep it in mind for future UIs._

overlord/views/countdown_page.erb
```diff
@@ -0,0 +1,17 @@
+<html>
+  <style>
```
djkotowski: Avoid mixing styles with HTML.

djkotowski:
Good submission.  Take a look at the comments and refactor where needed.  Resubmit when ready.

overlord/features/step_definitions/bomb_steps.rb
schwerdt: When I ran my cucumber tests without any steps, it gave me the regular expressions.  

Can we talk about what arg1 is doing?

overlord/views/countdown_page.erb
schwerdt: use a css style sheet instead?

PR #656, opened 2015-05-14
Robot

dietpop:


robot_name/robot_name.rb
```diff
@@ -1,32 +1,49 @@
...
-  @@registry
+  NAME_ERROR = "There was a problem generating the robot name!"
+  REG_EX = '[[:alpha:]]{2}[[:digit:]]{3}'
 
   def initialize(args = {})
-    @@registry ||= []
+    @registry ||= args[:used_names]
     @name_generator = args[:name_generator]
+    assign_name
+    bad_name = !(@name =~ /#{REG_EX}/) || @registry.include?(@name)
```
zsyed1: In line 18 you are calling `assign_name` that will eventually set the member variable `@name`. The code will be easier to follow if instead you did something line `@name = generate_name`. Also, you are setting `@name_generator` in line 17 and invoking it within `assign_name`. Instead of storing the generator, you could pass it into `generate_name`. Confused yet? :smile: 

Here's what I'm suggesting:

```
@registry ||= args[:used_names]
generator = args[:name_generator]
@name = create_name(generator)
```
That way you don't need to store the generator in a member variable (i.e. `@name_generator`).

zsyed1: In line 19, there are really two different errors that result in a bad name. But line 20 assumes it is a `NameCollisionError`. So you should probably split this check into to checks:

```
raise NameCollisionError, COLLISION_MSG if @registry/include?(@name)
raise MalformedNameError, MALFORMED_MSG unless @name =~ /#{REG_EX}/
```
Or something like this.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,49 @@
...
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def generate_name
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
+    @name = "#{generate_char.call}#{generate_char.call}"
+    @name += "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
   end
```
zsyed1: You can do some thing like `(1..3).to_a.map { generate_num.call }.join` instead of pasting it three times.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,49 @@
...
 end
 
-robot = Robot.new
+used_names = RobotNames.new
+robot = Robot.new(used_names: used_names.names)
+used_names.names << robot
 puts "My pet robot's name is #{robot.name}, but we usually call him sparky."
```
zsyed1: I like what you did here, creating a `RobotNames` instance to hold names already used. However I think that is the intent of the class member `@@registry`, i.e. to hold all names of all robots.

zsyed1: You shouldn't have to manually append `RobotNames` each time you create a new robot. It should happen automatically, hence the `@@registry` member.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,49 @@
...
 
 # Errors!
-# generator = -> { 'AA111' }
-# Robot.new(name_generator: generator)
-# Robot.new(name_generator: generator)
+generator = -> { 'AA111' }
+robot = Robot.new(name_generator: generator, used_names: used_names.names)
+puts "My pet robot's name is #{robot.name}, but we usually call him sparky."
+robot = Robot.new(name_generator: generator, used_names: used_names.names)
```
zsyed1: Please let me know if you'd like to sit down and go through this. PR comments can sometimes make it difficult to see the big picture!

djkotowski:
Good first iteration.  Take a look at the comments and refactor where necessary.

robot_name/robot_name.rb
dietpop: Rubocop didn't like the use of the class variable @@registry so I added a class that holds the list of names.  Is it good or bad practice to use class variables?

zsyed1: Got it. Just remove the double @. A single @ in the class scope should work. So before the initialize putting:
```
@registry = []

def initialize(....
```

_should_ work.

PR #659, opened 2015-05-18
Triangle facts

sfoa6558:
Worked on this exercise


myCoolFile
```diff
@@ -0,0 +1 @@
+sdfasdfasdfsdfdfgdgfddfgdfgdfdfgdfgdfg
```
zsyed1: This doesn't look like Triangle Facts :smile:

PR #663, opened 2015-05-18
Overlord

dustinhansen:
Issues:
 * Had issues getting hidden elements to work well in Capybara
 * Could probably DRY the test suite a bit more but not sure what’s the
correct/best procedures for doing so.


overlord/bomb.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def deactivate(code)
+    return false if @status == 'exploded'
+    if code == @deactivation_code
+      @status = 'inactive'
+      reset_errors
+    else
+      log_error
+    end
+  end
```
jmmastey: This returns `false`, `nil`, or `0` depending on what action actually takes place. Potentially this is okay if we're actually ignoring the return value of the method, but in that case we should probably just `return if @status == :exploded` on #19. Or alternatively just return `self` in every case.

If you want to be able to use the return value, consider returning a little nicer value.

overlord/bomb.rb
```diff
@@ -0,0 +1,43 @@
...
+    @deactivation_code = deactivation_code
+    @status = 'inactive'
+    reset_errors
+  end
+
+  def activate(code)
+    return false if code != @activation_code || @status == 'exploded'
+    @status = 'active'
+    reset_errors
+  end
```
jmmastey: So I can technically get infinite retries as long as I keep activating the bomb over and over again?

overlord/overlord.rb
```diff
@@ -1,14 +1,37 @@
...
+class Overlord < Sinatra::Base
+  enable :sessions
 
-get '/' do
-  "Time to build an app around here. Start time: " + start_time
-end
+  get '/?:activation_code?/?:deactivation_code?' do
+    if params.key?(:activation_code) && params.key?(:deactivation_code)
+      session[:bomb] = Bomb.new(params[:activation_code], params[:deactivation_code])
+    else
```
jmmastey: Given this is a `get` method, it shouldn't modify the state of the program. Browsers are allowed to cache these requests and thereby not hit the app, which would violate the logic of the program if I requested different bombs multiple times.

Technically, also, I can reboot an exploded bomb this way, but that's more of a limitation of how the web works versus bombs.

overlord/overlord.rb
```diff
@@ -1,14 +1,37 @@
...
+      session[:bomb] = Bomb.new
+    end
```
jmmastey: Move this entire conditional into its own method, as it stands it obscures much of the method.

jmmastey: Also, if I specify activation but not deactivation this code appears to break. And cannot specify deactivation without activation, though that seems implicitly possible (based on the fact that you allow activation as a parameter without deactivation).

overlord/overlord.rb
```diff
@@ -1,14 +1,37 @@
...
+      bomb_status: session[:bomb].status,
+    }
+  end
+
+  post '/security' do
+    submit_code(params[:code])
+    response = { bomb_status: session[:bomb].status }
+    if session[:bomb].error_count > 0
+      response[:error_count] = session[:bomb].error_count
+    end
```
jmmastey: Just send `error_count` no matter what. That way I can rely on it in the response.

overlord/overlord.rb
```diff
@@ -1,14 +1,37 @@
...
+    JSON.generate(response)
+  end
```
jmmastey: `response.to_json` should suffice

overlord/public/assets/app/js/main.js
```diff
@@ -0,0 +1,26 @@
...
+      if (typeof response.error_count !== 'undefined' && response.error_count > 0) {
+        error_text = 'Failed attempts: ' + response.error_count;
+      }
+      $('.bomb-errors').html(error_text);
+
+      // destroy bomb
+      if (response.bomb_status == 'exploded') {
+        $('form').remove();
+      }
+    });
```
jmmastey: Split this callback method into a named function. As is the case with ruby, we want to keep different concerns (initialization vs. event handling) separated.

overlord/public/assets/app/js/main.js
```diff
@@ -0,0 +1,26 @@
...
+      response = $.parseJSON(data);
+
+      // status
+      $('.bomb-status').html(response.bomb_status)
+
+      // display failed attempts to deactivate
+      error_text = '';
+      if (typeof response.error_count !== 'undefined' && response.error_count > 0) {
+        error_text = 'Failed attempts: ' + response.error_count;
+      }
```
jmmastey: Here's where you could simplify a conditional if you just always included `error_count` :)

overlord/public/assets/lib/bootstrap/js/bootstrap.js
```diff
@@ -0,0 +1,2317 @@
...
+
+
++function ($) {
+  'use strict';
+
+  // AFFIX CLASS DEFINITION
+  // ======================
+
+  var Affix = function (element, options) {
+    this.options = $.extend({}, Affix.DEFAULTS,
```
jmmastey: This is the thing that annoys me about these libraries. 90% of the code in your codebase is part of a UI library. In rails this is fixable by just using a bootstrap gem that separately manages the bootstrap code. Not sure how well that works for Sinatra.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      expect(bomb.status).to eq('exploded')
+    end
+  end
+
+  before(:all) do
+    @security_codes = {
+      activate: '1234', # default
+      deactivate: '0000', # default
+    }
+  end
```
jmmastey: Put `before` blocks at the top. I had to search for @security_codes.

That said, just use `let(:activation_code)` etc to create these kinds of vars. Instance vars are a smell in idiomatic rspec.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    subject(:bomb) { Bomb.new(@security_codes[:activate], @security_codes[:deactivate]) }
+    subject(:activation_code) { @security_codes[:activate] }
+    subject(:deactivation_code) { @security_codes[:deactivate] }
+
+    it 'state should be in "inactive" state' do
+      expect(bomb.status).to eq('inactive')
+    end
+
+    it 'should be in "active" state when we enter the correct activation code' do
+      activate_bomb
```
jmmastey: It's a little deceptive to put the action itself into a helper method, even if you use it elsewhere to activate or deactivate the bomb. I might consider the repetition of the activation code here to be acceptable so whomever reads the test can grok the code.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    end
+
+    it 'should be in "exploded" state when in "active" state and an incorrect deactivation code was entered 3 times consecutively' do
+      activate_bomb
+      explode_bomb
+      expect(bomb.status).to eq('exploded')
+    end
+
+    it 'should remain in "inactive" state when we enter the incorrect activation code' do
+      bomb.activate(random_code)
```
jmmastey: The `random_code` is unnecessary. Using a `let(:wrong_code)` value would be easier to read and cause less confusion about the generation of random codes. Also, this test will flap once every 10,000 runs, though that's probably okay :)

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      expect(bomb.status).to eq('inactive')
+    end
+
+    it 'should remain in "active" state when we enter the incorrect deactivation code' do
+      activate_bomb
+      bomb.deactivate(random_code)
+      expect(bomb.status).to eq('active')
+    end
```
jmmastey: Should this code include checks for the number of incorrect attempts? Either that or a separate test for the same.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    it 'should remain in "exploded" state when we enter the correct activation code' do
+      activate_bomb
+      explode_bomb
+      activate_bomb
+      expect(bomb.status).to eq('exploded')
+    end
+
+    it 'should remain in "exploded" state when we enter the correct deactivation code' do
+      activate_bomb
+      explode_bomb
```
jmmastey: Since we have several `activate && explode` tests, this would be better as:

`
describe Bomb do
  ...
  context "while exploded" do
    subject(:bomb) do
      Bomb.new(...).tap |bomb|
        3.times { bomb.deactivate(" :( ") }
      end

      # tests here
  end
end

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+  context 'using custom security codes' do
+    it_behaves_like 'mutable bomb' do
+      before(:all) do
+        @security_codes = {
+          activate: random_code,
+          deactivate: random_code,
+        }
+      end
+    end
+  end
```
jmmastey: Clever :) Possibly too clever at this level of bomb complexity (all we're testing is two default parameters atm). Does the job, tho.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,91 @@
...
+    submit_security_form
+  end
+
+  def submit_security_form
+    find('form[name="security"] input[type="submit"]').click
+  end
+end
+World(SessionStepsHarness)
+
+Given(/^I am viewing an (.*) bomb$/) do |bomb_status|
```
jmmastey: It's normal to just include `def` statements in the `step_definition` files like this one. Module is unnecessary.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,91 @@
...
+  def submit_random_code(blacklist = [])
+    new_code = ''
+    loop do
+      new_code = ''
+      4.times { new_code += rand(9).to_s }
+      break unless blacklist.include? new_code
+    end
+    code_field.set new_code
+    submit_security_form
+  end
```
jmmastey: Much like in the rspecs, randomness here is overrated and just adds complexity. Better off:

```
def submit_code(code = "this is clearly the wrong code")
  code_field.set(code)
  submit_security_form
end

File: overlord/features/step_definitions/bomb_steps.rb [hunk 18]
```diff
@@ -0,0 +1,91 @@
...
+
+  def default_deactivation_code
+    '0000'
+  end
+
+  def explode_bomb
+    activate_bomb
+    3.times do
+      submit_random_code([default_activation_code])
```
jmmastey: The below comment about randomness being overrated is because of this thing. I was surprised to see a single-element array containing one element, and even more surprised when that was a "list" of codes _not_ to submit.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def code_field
+    find_field('code')
+  end
+
+  def deactivate_bomb
+    code_field.set default_deactivation_code
+    submit_security_form
+  end
```
jmmastey: If you refactor `submit_random_code` to just be `submit_code`, these get a little DRYer.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,91 @@
...
+  visit '/'
+  case bomb_status
+  when 'active'
+    activate_bomb
+  when 'inactive'
+    deactivate_bomb
+  when 'exploded'
+    explode_bomb
+  end
+end
```
jmmastey: This is not a great re-use of code. Better to have separate steps for each of the statuses if you have to invoke methods for each one. If there was a form of `Bomb.new(starting_status)` (there shouldn't be, but for other use cases something like that could happen, then this form of dynamic matcher is fine.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,91 @@
...
+When(/^I enter any code that is not the (.*) code (\d+) times?$/) do |code_type, attempts|
+  attempts.to_i.times do
+    case code_type
+    when 'activation'
+      submit_random_code([default_activation_code])
+    when 'deactivation'
+      submit_random_code([default_deactivation_code])
+    end
+  end
+end
```
jmmastey: See above refactor to `submit_random_code`.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,91 @@
...
+
+When(/^I enter the (.*) code$/) do |code_type|
+  case code_type
+  when 'activation'
+    code_field.set default_activation_code
+  when 'deactivation'
+    code_field.set default_deactivation_code
+  end
+  submit_security_form
+end
```
jmmastey: Case statements are really not fun for grokking code. In the features themselves, I might have:

```
Background: I have a bomb with activation code '0000' and deactivation code '9999'

  ... Later ...
  When I submit code '0000'
```

This kind of literalism is okay. The tests need to be documentation as much as verification.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,91 @@
...
+
+Then(/^the bomb is (.*)$/) do |bomb_status|
+  fail unless bomb_status == find('span.bomb-status').text
```
jmmastey: Tests should use rspec syntax:

`expect(find('span.bomb-status')).to have_text(bomb_status)`

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,91 @@
...
+end
+
+Then(/^the security code form should be gone$/) do
+  fail unless first('form').nil?
+end
+
+Then(/^nothing happens$/) do
+  # do nothing
+end
```
jmmastey: Good use of a comment :)

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,91 @@
...
+  submit_security_form
+end
+
+Then(/^the bomb is (.*)$/) do |bomb_status|
+  fail unless bomb_status == find('span.bomb-status').text
+end
+
+Then(/^the security code form should be gone$/) do
+  fail unless first('form').nil?
+end
```
jmmastey: `expect(page).not_to have_element("form")`

overlord/features/security_codes.feature
```diff
@@ -0,0 +1,39 @@
+Feature: Security Codes
+  In order to securely arm or disarm the bomb
+  As a bomber
+  I want to be able to enter security codes that change the status of the bomb
+
+  @javascript
```
jmmastey: If they're all javascript (your app only works with JS in this case), just set the global driver to poltergeist and you can delete all these.

overlord/features/security_codes.feature
```diff
@@ -0,0 +1,39 @@
+Feature: Security Codes
+  In order to securely arm or disarm the bomb
+  As a bomber
+  I want to be able to enter security codes that change the status of the bomb
```
jmmastey: I dig it :)

overlord/features/security_codes.feature
```diff
@@ -0,0 +1,39 @@
...
+
+  @javascript
+  Scenario: Deactivate an inactive bomb
+    Given I am viewing an inactive bomb
+    When I enter any code that is not the activation code 1 time
```
jmmastey: The `1 time` thing is awkward.

overlord/features/security_codes.feature
```diff
@@ -0,0 +1,39 @@
...
+  @javascript
+  Scenario: Unsuccessfully deactivate an active bomb
+    Given I am viewing an active bomb
+    When I enter any code that is not the activation code 3 times
+    Then the bomb is exploded
+
+  @javascript
+  Scenario: Interacting with an exploded bomb
+    Given I am viewing an exploded bomb
+    Then the security code form should be gone
```
jmmastey: This bomb never explodes on its own, which is a little funny :) As long as we touch it, it'll be fine.

More of unexploded ordinance.

overlord/features/status_indicator.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Status Indicator
+  In order to determine the current status of the bomb
+  As a bomber
+  I want to be able to view the current status of the bomb
+
+  @javascript
+  Scenario: View the status of an inactive bomb
+    Given I am viewing an inactive bomb
+    Then the bomb is inactive
```
jmmastey: The structure here is a little where. There's no `When` step in sight, which is a little weird (I get why it happens w/ the one-page structure of the app). I might convert the `Given`s here into `When`s for idiomatic reasons.

Also, the `Then` descriptions don't read like Status Indications. In fact, they're identical to the assertions for the _actual_ bomb status (as opposed to its indicator). They should read something like `Then the bomb's status is shown as inactive`.

jmmastey:
Hey Dustin, 90% there. I added some notes on refactors for better idiomatic usage of the suite, which I think will answer your questions about how to make the code more DRY.

I don't see the hidden elements you're talking about, but feel free to hit me up on Slack. Generally, Capybara behaves a lot like a user, so if you try to click on things that aren't visible it'll throw a bit of a conniption.

jmmastey:
Also, fix yer hound violations :)

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,91 @@
...
+  def submit_security_form
+    find('form[name="security"] input[type="submit"]').click
+  end
+end
+World(SessionStepsHarness)
+
+Given(/^I am viewing an (.*) bomb$/) do |bomb_status|
+  visit '/'
+  case bomb_status
+  when 'active'
```
dustinhansen: Every linter/beautifier I've found favors the indentation template I've used here. I'd have to disable my tools in order to do what hound wants here. How important is this convention?

overlord/bomb.rb
dustinhansen: Correct. The way I read the requirements, the bomb should only explode after 3 failed attempts to deactivate an active bomb. If I'm wrong and need to change this let me know.

overlord/overlord.rb
dustinhansen: I simply needed some way for the end user to create a bomb with their own custom codes. I could auto-generate some random codes and then show them to the user. Is that more of what you're looking for? In a real world scenario these codes would have been implemented when the bomb was manufactured and the user would have these in whatever documentation came with the bomb. Given this is just a coding exercise I decided to let the user create their own bombs using a custom URL.

overlord/overlord.rb
dustinhansen: How did it break for you? If you don't specify a custom activation and deactivation code, then the bomb simply uses the default security codes. This is all mocked up functionality because you would never manufacture a bomb from scratch in a web browser. If I get a better idea for exactly what the requirements are for bomb creation I can better address this.

overlord/public/assets/lib/bootstrap/js/bootstrap.js
dustinhansen: It wasn't really necessary for this project but I just dropped it in as a force of habit. I would typically utilize server-side processing to combine/minify everything neatly like you're saying.

overlord/spec/bomb_spec.rb
dustinhansen: In this instance we are setting the default security codes and then overwriting them in a certain context. Is `let` still the right thing to use in this case? I'm having a tough time coming up with a way to do it correctly...

overlord/spec/bomb_spec.rb
dustinhansen: Should I be using `subject` or `let`? Where should these be placed? I haven't asked for any specific direction thus far in all of these exercises but being DRY with rspec beyond what I've done so far is proving to be pretty confusing.

overlord/spec/bomb_spec.rb
dustinhansen: How would the test flap? I am checking to make sure the random code does not match the activation or deactivation codes. I would think that generating random codes would prove to be a better test than testing the same code over and over, no?

overlord/spec/bomb_spec.rb
dustinhansen: Good call. You're correct.

overlord/spec/bomb_spec.rb
dustinhansen: Is there another way to do it if I want to run the same suite of tests twice (once with the default security codes and one with random security codes)?

PR #664, opened 2015-05-18
Initial Attempt Is Complete

dustinhansen:


data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
+require 'abanalyzer'
+require 'json'
+
+class SplitTest
+  attr_reader :cohorts
+
+  @@confidence_interval = 0.95
```
jmtai: Since these values are simply constants, the Ruby convention is to indicate this by naming your variable in all caps, e.g. `CONFIDENCE_INTERVAL`

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+  @@expected_keys = %w(date cohort result)
```
jmtai: Same thing applies here: `EXPECTED_KEYS`

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def initialize(file)
+    get_data(file)
```
jmtai: There's a **lot** going on here. `initialize` is used to setup an object, but now we are reading and parsing data files, computing results, creating other objects, etc. If you had no knowledge of the internals of this class, would you expect all of this to be done after simply calling `SplitTest.new`?

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+    tally_results
+    create_tester
+    add_totals
+    check_results
+    analyze_results
+  end
+
+  private
+
+  def get_data(file)
```
jmtai: So naming is an art form, and a lot of it comes down to personal preference but I wonder if this could be made clearer with a more descriptive name? This method is actually doing 2 things: reading the data file from disk and parsing the data into a hash. Ideally, methods should have a single responsibility (i.e. one method for reading the data, and another for parsing it). However, for something this simply maybe it's ok to just rename the method to `get_and_parse_data`?

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+    @data = JSON.parse(File.read(file))
+  rescue
+    raise 'file not found'
```
jmtai: Be careful here. If you're going to catch exceptions, you'll want to make sure your error message actually describes what is going wrong. Since you're reading the file **and** parsing it, this message is going to be inaccurate if something goes during the reading of the file (not just opening it) or during the parsing of the data.

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  def check_data_format
+    @@expected_keys.each do |expected_key|
+      raise 'invalid data format' unless @data.first.key?(expected_key)
```
jmtai: 'invalid data format' isn't super descriptive. It might be nice to include the name of the expected_key that is missing.

jmtai: `@data.first` will grab the first entry from your data array. What happens if there is an error somewhere else in the file?

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+  end
+
+  def tally_results
```
jmtai: This error can be annoying, but it's usually an indicator that you're trying to do too much within a single method.

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+    @cohorts = {}
+    @data.each do |sample|
+      cohort = sample['cohort'].to_sym
+      @cohorts[cohort] = { no: 0.0, yes: 0.0 } unless @cohorts.key?(cohort)
```
jmtai: Does it make sense to use floats for the counts here?

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+      if [0, 1].includes?(sample[:result])
```
jmtai: Here you're checking to make sure the result value is 0 or 1 but `check_data_format` has been run at this point. Perhaps we'd want to include verification of these result values there, instead?

If we can make it a precondition that all the values are valid, we can eliminate this check.

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+        @cohorts[cohort][:no] += 1 unless sample[:result] == 1
+        @cohorts[cohort][:yes] += 1 unless sample[:result] == 0
```
jmtai: The reverse logic here is a little confusing at first glance. It might be clearer if it were something like this:
```
@cohorts[cohort][:no] += 1 if sample[:result] == 0
@cohorts[cohort][:yes] += 1 if sample[:result] == 0
```

data_science/splittest.rb
dustinhansen: The thought was that check_data_format checks the headers only (so maybe check_headers would be a better name). I thought about adding data validation in that method but it seemed like a wasted process. We're going to have to go through the data line by line later anyways, so why not just validate as we go? For not I'm letting bad data just get ignored but we could choose to throw an error instead.

data_science/splittest.rb
dustinhansen: I agree but rubocop barks at me to always use the "unless" syntax..?

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+  def check_data_format
+    @@expected_keys.each do |expected_key|
+      raise 'invalid data format' unless @data.first.key?(expected_key)
+    end
+  end
+
+  def tally_results
+    @cohorts = {}
+    @data.each do |sample|
+      cohort = sample['cohort'].to_sym
```
jmtai: What would happen if the input data had a mix of upper case and lower case cohort labels?

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def create_tester
+    @abtester = ABAnalyzer::ABTest.new @cohorts
+  rescue ABAnalyzer::InsufficientDataError
+    raise 'insufficient data'
+  end
+
+  def add_totals
+    @cohorts.each do |cohort, cohort_data|
+      @cohorts[cohort][:total] = cohort_data[:yes] + cohort_data[:no]
```
jmtai: So you've got your data nicely separated into separate cohorts, but we've got all of these brackets needed to access the data. Wouldn't it be easier to read if you could do something like this instead?

`cohort.total = cohort.yes + cohort.no`

Maybe it would be cleaner if you created a `cohort` class and used that to encapsulate the individual values and calculations that operate per-cohort?

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+  end
+
+  def check_results
+    @cohorts.each do |cohort, cohort_data|
+      @cohorts[cohort][:total] = @cohorts[cohort][:yes] + @cohorts[cohort][:no]
+      raise 'insufficient data' unless cohort_data[:total] >= 20
```
jmtai: Hardcoded magic numbers are something you want to avoid. The 20 here should be assigned to a constant, and given a meaningful name.  Perhaps `MIN_ALLOWABLE_SAMPLES` or something along those lines?

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  def analyze_results
+    @cohorts.each do |cohort, cohort_data|
+      @cohorts[cohort][:error_bars] = ABAnalyzer.confidence_interval(cohort_data[:yes], cohort_data[:total], @@confidence_interval)
+      @cohorts[cohort][:conversion_rate] = cohort_data[:yes] / cohort_data[:total]
+      @cohorts[cohort][:confidence] = @@confidence_interval
+    end
+  end
+end
```
jmtai: You've got the conversion rate and error bars but you're missing the confidence level that the current leader is better than random (2.iii from the requirements).

data_science/splittest_spec.rb
```diff
@@ -0,0 +1,55 @@
+require './splittest'
+
+describe SplitTest do
+  data_source = 'data/test_data.json'
```
jmtai: Make this a constant: `DATA_SOURCE`

data_science/splittest_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  it 'should return a positive number to represent the total number of samples per cohort' do
```
jmtai: So this test checks to make sure your class returns a number, but do we know it's the correct result? Since we're using a predefined test data set, we also know what the correct results should be. If someone changes the code and it returns a different number for the total samples, would this test fail?

Tests should be designed to check that the behavior of the class is correct - not just that it returns a result.

data_science/splittest_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    split_test = SplitTest.new(data_source)
+    split_test.cohorts.each do |_cohort, cohort_data|
+      expect(cohort_data[:total]).to be >= 0
+    end
+  end
+
+  it 'should return a positive number (or zero) to represent the total number of conversions per cohort' do
```
jmtai: Same comment here. Make sure your tests check for correct behavior.

data_science/splittest_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  it 'should return a number between 0 and 1 to represent the confidence level' do
+    split_test = SplitTest.new(data_source)
+    split_test.cohorts.each do |_cohort, cohort_data|
+      expect(cohort_data[:confidence]).to be >= 0.0
+      expect(cohort_data[:confidence]).to be <= 1.0
+    end
+  end
+
+  it 'should return a "file not found" error if a data set could not be found' do
```
jmtai: This test is brittle because your source code is combining multiple exceptions into a single error message. There are other errors that could result in this error message but they aren't being tested.

data_science/splittest.rb
jmtai: For a situation like this, unless actually makes the logic harder to understand, so I'd ignore rubocop. The typical place where unless is preferred is for negative conditions.  Something like this:

`[do something] unless condition`
vs
`[do something] if !condition`

That being said, there are times where I think `unless` actually makes the logic harder to understand, and in those situations I prefer using `if`.

jmtai:
This is a nice first pass. Take a look at the comments, add in the missing functionality, and work on improving your tests.  I'd also strongly suggest creating a `cohort` class and compare with your existing implementation to see if you think that improves the structure and readability of your solution.

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def check_data_format
+    @@expected_keys.each do |expected_key|
+      raise 'invalid data format' unless @data.first.key?(expected_key)
+    end
+  end
+
+  def tally_results
+    @cohorts = {}
+    @data.each do |sample|
```
djkotowski: When you have syntax like
```ruby
variable = {}
array.each do |element|
  # modify variable here
end
```
it usually can be simplified to
```ruby
variable = array.each_with_object({}) do |element, memo|
  # modify memo here
end
```

data_science/splittest.rb
djkotowski: Are you ever expecting a result other than 0 or 1?

data_science/splittest.rb
```diff
@@ -0,0 +1,72 @@
...
+      @cohorts[cohort] = { no: 0.0, yes: 0.0 } unless @cohorts.key?(cohort)
+      if [0, 1].includes?(sample[:result])
+        @cohorts[cohort][:no] += 1 unless sample[:result] == 1
+        @cohorts[cohort][:yes] += 1 unless sample[:result] == 0
+      end
+    end
+  end
+
+  def create_tester
+    @abtester = ABAnalyzer::ABTest.new @cohorts
```
djkotowski: Wrap params with parens.  There are a few exceptions, but it's better to err with parens.

data_science/splittest_spec.rb
```diff
@@ -0,0 +1,55 @@
+require './splittest'
+
+describe SplitTest do
+  data_source = 'data/test_data.json'
+
+  it 'should return a positive number to represent the total number of samples per cohort' do
+    split_test = SplitTest.new(data_source)
```
djkotowski: Extract this to a `let`.

data_science/splittest_spec.rb
djkotowski: If you want to ensure that it raises a specific error, create a custom error class to encapsulate it, then test against that class.

```ruby
# file_not_found_error.rb
class FileNotFoundError < StandardError; end

# split_test_spec.rb
expect { SplitTest.new("data/file_does_not_exist.json") }.to raise_error(FileNotFound)
```

djkotowski:
Good first iteration.  Use `let` and `subject` statements to extract repeated logic in specs.  Also watch your naming.  A class named `SplitTest` should be in `split_test.rb`.

Take a look at the comments and refactor where needed.  Resubmit when ready.

data_science/splittest.rb
dustinhansen: Yes, because we need to perform division on them later to get the conversion rate.

jmtai: While it's true that you need a float to calculate the conversion rate, perhaps it would be better to simply convert these values to floats when they are actually needed in that format with `.to_f`?

If you were looking at this code for the first time, what would an initialization value of `0.0` suggest to you? Would you assume that these variables could have fractional values?

dustinhansen: I see your point. I was trying to avoid unnecessary overhead by not having to convert the integer to a float later (who knows how many times), but I take it sounds like readability and terseness should usually win out over performance when it comes to Ruby development? I'm almost done with my refactoring of this one and will submit later today.

jmtai: Yep, the primary goal is to write clear, maintainable code. Performance is not unimportant, but for something like this storing the counts as floats would probably fall into the category of "premature optimization".

data_science/splittest.rb
dustinhansen: Good catch. For some reason I thought to_sym() downcased everything automagically.

data_science/splittest.rb
dustinhansen: 0 or 1 are the only expected values. Anything else gets ignored and not included. I could throw an error here but I'm letting it silently ignore the issue for now. I suppose I will add an exception in the next iteration.

Are there any guidelines as to what we should and should not ignore when it comes to rubocop/hound?

PR #665, opened 2015-05-18
refactored the code in triangle_facts

sfoa6558:
finally got it right.

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -1,58 +1,72 @@
 # Killer facts about triangles AWW YEAH
 class Triangle
-	attr_accessor :side1,:side2,:side3
+  attr_accessor :side1, :side2, :side3
 
-	def initialize(side1,side2,side3)
-		@side1,@side2,@side3 = side1,side2,side3
-	end
+  def initialize(side1, side2, side3)
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+  end
 
-	def equilateral()
-		return side1 == side2 && side2 == side3
-	end
+  def equilateral
+    return true if side1 == side2 && side2 == side3
+    puts 'This triangle is equilateral!'
+  end
 
-	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
-	end
+  def isosceles
+    return true if [side1, side2, side3].uniq.length == 2
+    puts 'This triangle is isosceles! Also, that word is hard to type.'
+  end
 
-	def scalene()
-		unless ! (equilateral || isosceles)
-	    false
-	  else
-	    true
-	  end
-	end
+  def scalene
+    return false if equilateral || isosceles == false
+    puts 'This triangle is scalene and mathematically boring.'
+  end
 
-	def recite_facts
-		puts 'This triangle is equilateral!' if equilateral
-		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+  def recite_facts
+    equilateral
 
-		angles = self.calculate_angles(side1,side2,side3)
-		puts 'The angles of this triangle are ' + angles.join(',')
+    isosceles
 
-		puts 'This triangle is also a right triangle!' if angles.include? 90
-		puts ''
-	end
+    scalene
 
-	def calculate_angles(a,b,c)
-		angleA = radians_to_degrees(Math.acos((b**2 + c**2 - a**2) / (2.0 * b * c)))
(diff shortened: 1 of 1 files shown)


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,72 @@
...
+  end
+
+  def equilateral
+    return true if side1 == side2 && side2 == side3
+    puts 'This triangle is equilateral!'
+  end
+
+  def isosceles
+    return true if [side1, side2, side3].uniq.length == 2
+    puts 'This triangle is isosceles! Also, that word is hard to type.'
```
zsyed1: Line 17 - you mean `return false unless [side1...] == 2` right?

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,72 @@
...
+  end
+
+  def scalene
+    return false if equilateral || isosceles == false
+    puts 'This triangle is scalene and mathematically boring.'
```
zsyed1: Line 22: Use unary operators to check if something is false e.g. `if !running` as opposed to `if running == false`. Also, the logic here should be:
```
return false if equlateral || isosceles
```

zsyed1: For methods that return `true` or `false` the convention is to name them with a trailing question mark: `def isosceles?`

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,72 @@
...
+    puts 'The angles of this triangle are ' + angles.join(',')
 
+    puts 'This triangle is also a right triangle!' if angles.include? 90
+
+    puts ''
+  end
+
+  def calculate_angelsa(a, b, c)
+    radians_to_degrees(Math.acos((b**2 + c**2 - a**2) / (2.0 * b * c)))
+  end
```
zsyed1: Lines 43, 47, 51 are almost identical except for the order of the arguments. You should prbably put them into a single method: `calculate_included_angle(legA, legB, opp)`. Then you can change `calculate_angles` do:
```
def calculate_angles(a, b, c)
  [calculate_included_angle(a, b, c),
   calculate_included_angle(a, c, b),
   calculate_included_angle(b, c, a),
end
```
And remove the three methods `calculate_angelsX`.
(note misspelling of `angles` as `angels` :smile: )

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,72 @@
...
 ]
-triangles.each { |sides|
-	tri = Triangle.new(*sides)
-	tri.recite_facts
-}
+
+triangles.each do |sides|
+  tri = Triangle.new(*sides)
+  tri.recite_facts
+end
```
zsyed1: Lines 70 - 71: Since you're just reciting the facts you can do this in one line:
```
Triangle.new(*sides).recite_facts
```

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,72 @@
...
+    scalene
+
+    angles = calculate_angles(side1, side2, side3)
+
+    puts 'The angles of this triangle are ' + angles.join(',')
 
+    puts 'This triangle is also a right triangle!' if angles.include? 90
+
+    puts ''
+  end
```
zsyed1: If you change the methods `isosceles`, `scalene` and `equilateral` to boolean predicate methods, you can move all the printing logic out of those three methods and into this method. It makes the code cleaner by reducing the responsibility of those methods. Besides `recite_facts` should do the _reciting_.

So for example, if you remove line 13 and move it here it would look something like this:
```
def equilateral?
  return @side1 == @side2 && @side2 == @side3
end

def recite_facts
  ...
  ...
  puts 'This triangle is equilateral' if equilateral?
  ...
  ...
end
```

triangle_facts/triangle_facts.rb
djkotowski: As a side note, I'd prefer using a predicate method here and separating the puts logic.

triangle_facts/triangle_facts.rb
djkotowski: And ditto my above statement.

djkotowski:
Good first iteration.  Take a look at the comments and refactor where needed.  Resubmit when ready.

triangle_facts/triangle_facts.rb
sfoa6558: Line 17 Two questions. Why would it be return false? Is it because that express evaluates to true already, so there is no point returning true.  A predicate method has a ? mark and returns a boolean right?  How would you seperate the puts logic?

sfoa6558: I read the rest of the comments I see.

PR #674, opened 2015-05-20
Refactoring based on Dan's notes

dustinhansen:


robot_name/robot_name.rb
```diff
@@ -3,30 +3,26 @@ class NameCollisionError < RuntimeError; end
...
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+    @name = generate_name(2, 3)
+    @@registry << @name
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def generate_name(number_of_letters, number_of_digits)
+    new_name = ''
+    (1..number_of_letters).each { new_name += ('A'..'Z').to_a.sample }
```
djkotowski: Prefer `number_of_letters.times`

djkotowski: Also, prefer `<<` over `+=`.
`+=` combines two strings into one new string
`<<` appends the second string to the first, altering the original

robot_name/robot_name.rb
```diff
@@ -3,30 +3,26 @@ class NameCollisionError < RuntimeError; end
...
+    if name_exists?(new_name)
+      raise NameCollisionError, "There was a problem generating the robot name!"
     end
+    new_name
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def name_exists?(name)
+    name =~ /[[:alpha:]]{2}[[:digit:]]{3}/ || @@registry.include?(name)
```
djkotowski: This method is checking if the name is invalid or exists, even though it implies it's only checking the latter.

djkotowski:
Good work!  Take a look at the comments.  No need for resubmission.

PR #678, opened 2015-05-20
Shopping cart

schwerdt:


shopping_cart/add_remove_items.feature
```diff
@@ -0,0 +1,35 @@
...
+    Given I am on the JellyBelly item page
+    When I enter -5 in the Quantity field
+    And I click Add Item
+    Then I see an error message indicating only positive integers quantities can be added to the cart
+
+  Scenario: remove an item from the shopping cart
+    Given I am on the cart page
+    When I click the Remove Item button for the fish figurine item (quantity 3)
+    Then I see a message indicating all 3 fish figurine items were removed
+    And The number of items in the cart decreased by 3
```
jmmastey: These same quantity changes would be nice to have for `add a negative number of items to the shopping cart` and `add an item to the shopping cart`.

Also, as a minor nit, don't capitalize `The` :)

shopping_cart/add_remove_items.feature
```diff
@@ -0,0 +1,35 @@
...
+
+  Scenario: decrement an item quantity from shopping cart
+    Given I am on the cart page
```
jmmastey: Realistically, you'd probably need another `Given`, or possibly a `Background` step, that specifies that you currently have some garden hoses in your cart.

shopping_cart/cart_item_navigation.feature
```diff
@@ -0,0 +1,20 @@
+Feature: Navigate to/from shopping cart
```
jmmastey: I like the way you split these tests. This feels like a natural organization.

shopping_cart/cart_item_navigation.feature
```diff
@@ -0,0 +1,20 @@
...
+
+  In order to navigate from the shopping cart
+  As the customer
+  I should click on products or shopping cart
```
jmmastey: This description is a little weak, fwiw.

```
In order to see more about the items in my cart
As a customer
I want to navigate between the product pages and my cart
```

shopping_cart/cart_item_navigation.feature
```diff
@@ -0,0 +1,20 @@
...
+
+  Scenario: navigate to specific item using name link
+    Given I am on the cart page
+    When I click on JellyBelly link in the cart
+    Then I see the item page
```
jmmastey: This might be one case where a longer scenario becomes reasonable. Not a requirement at all, but consider the virtues of:

```
Scenario: Navigate to an item by way of the cart
Given I have 3 JellyBelly items in my cart
When I visit the cart page
Then I should see a link to the JellyBelly item
When I click on the JellyBelly link
Then I am on the item page
```

It's something that should only ever be used sparingly, but it tells a nice understandable story to someone who's trying to understand the system behavior.

shopping_cart/checkout.feature
```diff
@@ -0,0 +1,50 @@
+Feature: Finalize order
+ 
+  In order to checkout
+  As the customer
+  I should enter address, payment, and coupon information
+ 
+  Scenario: begin checkout process
+    Given I am on the cart page
+    When I click the checkout button
+    Then I am prompted to enter my address
```
jmmastey: I'm not sure whether I like this or not. It's specific, but possibly too specific. If we changed to a one-page checkout flow, or something like it, this test would break. I think the `begin checkout process` requirements haven't "changed", so it probably shouldn't fail in that case.

Consider possibly changing to `Then I am on the first step of the checkout process.`

shopping_cart/checkout.feature
```diff
@@ -0,0 +1,50 @@
...
+
+  Scenario: enter a invalid credit card number
+    Given I am on the checkout page
+    When I enter an invalid credit card number
+    Then I see an error stating that the card number was not accepted
+  
+  Scenario: enter a valid credit card number
+    Given I am on the checkout page
+    When I enter a valid credit card number
+    Then I see an error stating that the card number was accepted
```
jmmastey: So for `zip`, `coupon`, `credit card number`, you'd usually have a _bunch_ more cases. It's okay that they're omitted here as long as you know that. :)

shopping_cart/checkout.feature
```diff
@@ -0,0 +1,50 @@
...
+
+  Scenario: submit order
+    Given I correctly entered shipping and billing information
+    When I click Submit Order
+    Then I see a page stating that the order has been accepted
```
jmmastey: To re-use steps, you'd probably have:

```
Then I am on the purchase confirmation page
And I see a message that I have completed my order
```

This is nice too, though.

shopping_cart/login_behavior.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Allow user to maintain shopping cart after logging in (from anonymous user state)
+
+  In order to create an order
+  As the customer
+  I fill my shopping cart and log in
+
```
jmmastey: Definitely want something like:

```
Background: I am not logged in
  And I have a working user account
```

jmmastey:
Good job. Minor nits and stylistic stuff, but this is solid. Good organization of the tests, and none of them are fifteen steps long. :+1:

jmmastey:
Oh, final nitpick: in ruby, we usually use `snake_case` for branch names. Makes it easier to collaborate on the same repo.

PR #679, opened 2015-05-21
Dino catalog

sfoa6558:
Refactored Dino_Dex


dinosaur.rb
```diff
@@ -0,0 +1,30 @@
...
+    @data[:weight] = attributes['weight_in_lbs']
+    @data[:walking] = attributes['walking']
+    @data[:description] = attributes['description']
+  end
+
+  def simple_filter?(attrb, value)
+    @data[attrb] == value
+  end
+
+  def carnivore?(attrb, value)
```
zbright: Why pass in the attribute here? If you are checking whether its a carnivore, you are always going to be checking the diet attribute, right?

dino_dex.rb
```diff
@@ -0,0 +1,95 @@
...
+  def find_big
+    result = @dinosaurs.select { |dino| dino.big? }
+    puts result.to_s
+    Dinosaur.new(result.to_s)
+  end
+
+  def find_small
+    result = @dinosaurs.select { |dino| dino.small? }
+    puts result.to_s
+    Dinosaur.new(result.to_s)
```
zbright: it seems like the function of this method is to print out the small dinos, so why are you creating a new Dinosaur after completing that functionality?

dino_dex.rb
```diff
@@ -0,0 +1,95 @@
...
+               )do |row|
+      a = Dinosaur.new(row.to_hash)
+      @dinosaurs << a
+    end
+  end
+
+  def find_simple(attrb, value)
+    result = @dinosaurs.select { |dino| dino.simple_filter?(attrb, value) }
+    puts result.to_s
+    Dinosaur.new(result.to_s)
```
zbright: see other note about creating new Dinos in methods like this

dino_dex.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def find_carnivore(attrb, value)
+    result = @dinosaurs.select { |dino| dino.carnivore?(attrb, value) }
+    puts result.to_s
+    Dinosaur.new(result.to_s)
```
zbright: see other note about creating new Dinos in methods like this

dino_dex.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def find_big
+    result = @dinosaurs.select { |dino| dino.big? }
+    puts result.to_s
+    Dinosaur.new(result.to_s)
```
zbright: see other note about creating new Dinos in methods like this

dino_dex.rb
```diff
@@ -0,0 +1,95 @@
+# DinoDex is the main class
+require 'csv'
+require_relative 'dinosaur.rb'
+# DinoDex is the main class
+class DinoDex
+  SIZE_VALUE = 4000
+  NAME_CONVERTER = lambda do |field|
```
zbright: These lambdas are cool :+1:

dino_dex.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def find_small
+    result = @dinosaurs.select { |dino| dino.small? }
+    puts result.to_s
+    Dinosaur.new(result.to_s)
+  end
+end
+# Just uncomment any condition/filter you want to test.
+dino_dex = DinoDex.new('dinodex.csv')
```
zbright: Everything from here to the end of the file should be in a different class and possibly be in a method called `run`

dino_dex.rb
```diff
@@ -0,0 +1,95 @@
...
+    Dinosaur.new(result.to_s)
+  end
+
+  def find_small
+    result = @dinosaurs.select { |dino| dino.small? }
+    puts result.to_s
+    Dinosaur.new(result.to_s)
+  end
+end
+# Just uncomment any condition/filter you want to test.
```
zbright: Everything from here to the end of the file should be in a different class and possibly be in a method called run

dino_dex.rb
zbright: There is one major flaw in your implementation. You are supposed to have a program that accepts the name of the csv as a command line parameter and you should allow the user to select which filters they want through the command line

zbright: For example: once the user starts the program using something like `ruby main.rb dinodex.csv` you should prompt them with a menu such as 
```
  def menu
    ["Enter your choice(s): ",
    "Bipeds - Display all bipeds",
    "Carnivores - Display all carnivores",
    "Time Period - Display all from a specified time period",
    "Big - Display all big dinos",
    "Small - Display all small dinos",
    "Select - Select one dion to view",
    "Reset - Remove all filter",
    "Exit - quit program",
    "You can also hit enter to view current list again"].join("\n")
  end
```

zbright: As mentioned earlier, you would also need a `run` method that could look similar to this:
```
def run
    while true
      puts menu

      user_input = $stdin.gets.chomp.downcase
      break if user_input == 'exit'

      process_filters(user_input)
      puts dino_list.pretty_print

    end
  end
```

zbright:
@sfoa6558 Theres a lot of good stuff going on in here, just need to change how the program is actually run. There's a lot of info in the comments, but it might be easier to go over them in person if youre not exactly sure what I mean. Happy to talk anytime if you'd like.

dino_dex.rb
djkotowski: No need to create a run method.  Just pull it out into a runner file.

dinosaur.rb
djkotowski: Also, `carnivore?` should know about `%w(Carnivore Insectivore Piscivore)`, not your runner.

myCoolFile
```diff
@@ -0,0 +1 @@
+sdfasdfasdfsdfdfgdgfddfgdfgdfdfgdfgdfg
```
djkotowski: Be sure to remove this in the dino catalog branch.

DocumentationForRuby
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/ruby
```
djkotowski: Remove this.

triangle_facts.rb
```diff
@@ -0,0 +1,72 @@
+# Killer facts about triangles AWW YEAH
```
djkotowski: Please clean this branch.

dinosaur.rb
```diff
@@ -0,0 +1,30 @@
+# Dinosaure provides the filter functionality
+class Dinosaur
+  SIZE_VALUE = 4000
+  def initialize(attributes)
+    @data = {}
```
djkotowski: Why use a hash and not attributes?

dino_dex.rb
```diff
@@ -0,0 +1,95 @@
...
+    field
+  end
+
+  CONVERTERS = [DIET_CONVERTER, WEIGHT_CONVERTER, NAME_CONVERTER, :downcase]
+  def initialize(name_csv)
+    @dinosaurs = []
+    CSV.foreach(name_csv,
+                headers: true,
+                header_converters: CONVERTERS
+               )do |row|
```
djkotowski: End paren should be at the end of the L27, not on its own line.

djkotowski:
This is looking good.  You'll want to use `attr_reader` in your Dinosaur class though.  Just storing the attributes in a hash makes the list of possible attributes vague.

Please take a look at the other comments, refactor and resubmit when ready.

dinosaur.rb
```diff
@@ -0,0 +1,30 @@
...
+    @data[:name]  = attributes['name']
+    @data[:period] = attributes['period']
+    @data[:continent] = attributes['continent']
+    @data[:diet] = attributes['diet']
+    @data[:weight] = attributes['weight_in_lbs']
+    @data[:walking] = attributes['walking']
+    @data[:description] = attributes['description']
+  end
+
+  def simple_filter?(attrb, value)
```
djkotowski: This is a confusing name.

dino_dex.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  def find_carnivore(attrb, value)
+    result = @dinosaurs.select { |dino| dino.carnivore?(attrb, value) }
+    puts result.to_s
+    Dinosaur.new(result.to_s)
+  end
+
+  def find_big
+    result = @dinosaurs.select { |dino| dino.big? }
```
djkotowski: Fix Hound errors too.

dino_dex.rb
sfoa6558: Hi Zach! I am creating a new dinosaur at the end of almost all the methods to be able to chain methods together. I pass in the changed results to the dino, and return it. So, the next method will that Dino. Since the requirements call for you to be able to chain methods like be able to search for a small dinosaur who is also a carnivore.

PR #680, opened 2015-05-21
Command Line Version of Dinodex

dustinhansen:
To run: ruby “start.rb”

I've only included two questions for now. It's been built so you could pretty easily add more questions.


dino_catalog/dinodexdata.rb
```diff
@@ -0,0 +1,62 @@
...
+  def get_main_data
+    CSV.open(MAIN_DATA_SOURCE,
+             headers: true,
+             header_converters: :symbol
+    ) do |csv|
+      csv.to_a.map(&:to_hash)
+    end
+  end
+
+  def header_converters
```
djkotowski: This method name implies it is only returning header converters, not assigning them to CSV's global scope.

dino_catalog/dinodexdata.rb
```diff
@@ -0,0 +1,62 @@
...
+    CSV::HeaderConverters[:map_to_main] = lambda do |header|
+      header = header.downcase.to_sym
+      return FIELD_MAPPING[header] if FIELD_MAPPING.key?(header)
+      header
+    end
+  end
+
+  def converters
```
djkotowski: Ditto above.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,62 @@
+require 'json'
+require './dinodexdata'
+
+class DinoDex
+  def initialize
+    @dinodexdata ||= DinoDexData.new
```
djkotowski: DinoDexData is a bit vague and confusing.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def search(search_filters, format = '')
+    results = []
+    search_filters.each do |key, value|
+      results << generate_results(key, value)
+    end
```
djkotowski: L10-13 could be replaced by
```ruby
results = search_filters.map do |key, value|
  generate_results(key, value)
end
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def get_results_from_string(key, value)
+    @dinodexdata.data.select do |dino|
+      dino[key] && dino[key].downcase.include?(value.downcase)
+    end
+  end
+
+  # Look for dinos that exist across all results
+
+  def parse_results(results)
```
djkotowski: It would be more efficient to filter the same set of data each time than to filter the global set and find the intersection.

djkotowski:
Look into building a Dinosaur class that knows its own attributes (is it big?  what is its name?  etc.).  An array of Dinosaurs is easier to filter than an array of hashes.  At the moment, Dinodex has too many responsibilities.

dino_catalog/dinointerface.rb
```diff
@@ -0,0 +1,85 @@
...
+      p PROMPT_MESSAGES[:results] + list_results(results)
+    end
+  end
+
+  def list_results(results)
+    output = ''
+    results.each_with_index do |dino, i|
+      output += dino[:name]
+      output += ', ' if i < results.length - 1
+    end
```
djkotowski: Look at `Array#join`

dino_catalog/dinointerface.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  def exit_session
+    p PROMPT_MESSAGES[:exit]
+    exit
+  end
+
+  def ask_questions
+    QUESTIONS.each_with_index do |data, i|
+      sleep(1) while @search_criteria.length < i - 1
```
djkotowski: Why sleep?

dustinhansen: I didn't know of a good alternative to say "pause on this line until a certain condition is met". That line is meant to hold processing of the next question until the previous one is complete (trying to be DRY).

djkotowski: You're blocking here.  If the condition isn't met when you reach this point, it will never be met.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,62 @@
...
+    if value.is_a?(Enumerable)
+      get_results(key, value)
+    elsif key == 'min_weight'
+      get_results_from_min_weight(value)
+    else
+      get_results_from_string(key, value)
+    end
+  end
+
+  def get_results(key, values = [])
```
djkotowski: Avoid using the `get_` prefix in method names.

dino_catalog/dinointerface.rb
```diff
@@ -0,0 +1,81 @@
...
+      question: "Which period did the dinosaur live in? (1=Cretaceous,2=Jurassic,3=Triassic)",
+      answers: {
+        '1' => 'Cretaceous',
+        '2' => 'Jurassic',
+        '3' => 'Triassic',
+      },
+      search_key: :period,
+    },
+  ]
+
```
djkotowski: I prefer a command line interface over a menu-driven interface.  Questions take a while for a user to answer, whereas using commands can be quicker.

Example:
```
DINODEX
========
>> load dinodex.csv
Imported dinosaurs.
>> load african_dinosaur_export.csv
Imported dinosaurs.
>> list --bipeds --big --period=Cretaceous
( list of dinos here)
```

Don't worry about developing this now, but think about how you would go about doing it.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,55 @@
...
+  def retrieve_results(key, values = [])
+    values = values.map(&:downcase)
+    @dino_database.data.select do |dino|
+      dino[key] && values.include?(dino[key].downcase)
+    end
+  end
+
+  def retrieve_results_from_min_weight(min_weight)
+    @dino_database.data.select do |dino|
+      dino['weight_in_lbs'] && dino['weight_in_lbs'].to_i >= min_weight
```
djkotowski: If you used a Dinosaur object, you could extract some of this logic to it.  Should the index be responsible for comparing weight, period, etc., or does that responsibility belong with the individual dino?

djkotowski:
Each refactor shows some improvements, but there are some aspects (e.g. the lack of a Dinosaur class) that could have been made.  At this point, don't concern yourself with resubmitting the exercise. Spend some time thinking about what classes have which responsibilities, and whether those responsibilities belong elsewhere.

dustinhansen:
OK. I thought about a Dinosaur class, but when I saw Ruby had built-in methods for filtering arrays of hashes, I decided to go with something that would be less code but ultimately a less flexible model.

PR #681, opened 2015-05-22
Overlord

schwerdt:


incazteca:
Looks good to me. Just be sure to fix some of the hound issues. But a pull request is not required.

overlord/overlord.rb
```diff
@@ -1,14 +1,73 @@
 # run `ruby overlord.rb` to run a webserver for this app
 
 require 'sinatra'
+require './bomb.rb'
```
djkotowski: Use `require_relative` here.

djkotowski:
I agree with @incazteca.  Nice job!

PR #682, opened 2015-05-22
arrowhead.rb

sfoa6558:


exceptions.rb
```diff
@@ -0,0 +1,6 @@
+# This class is used to define errors
+class UnknownRegionError < RuntimeError
+end
+
+class UnknownShapeError < RuntimeError
+end
```
jmmastey: When errors are only used within one class (I think that's case to say here), we can just include them at the top or bottom of the main file (`arrowhead.rb`).

arrowhead.rb
```diff
@@ -0,0 +1,27 @@
+require_relative 'classifications.rb'
+# This is the main class for the exercise
+class Arrowhead
+  @classify_class = Classify.new
+  # This seriously belongs in a database.
+  CLASSIFICATIONS = {
```
jmmastey: It's odd that the `CLASSIFICATIONS` and `classify` live here, while `classifications.rb` file doesn't contain classifications, but _does_ classify.

classifications.rb
```diff
@@ -0,0 +1,21 @@
+require_relative 'exceptions.rb'
+# This class is used to define classfications
+class Classify
+  REGION_ERROR = 'Unknown region, please provide a valid region.'
+  UNKNOWN_SHAPE = "Unknown shape value.
+   Are you sure you know what you're talking about?"
```
jmmastey: These can actually be defined as part of the error classes if you really want.

classifications.rb
```diff
@@ -0,0 +1,21 @@
...
+  def get_shapes(classifications, region, shape)
+    shapes = classifications[region] if classifications.include? region
+    raise UnknownRegionError,
+          REGION_ERROR unless classifications.include? region
+    get_arrowhead(shapes, shape) if shapes
+  end
+
+  def get_arrowhead(shapes, shape)
+    arrowhead = shapes[shape] if shapes.include? shape
+    puts "You have a(n) '#{arrowhead}' arrowhead. Probably priceless." if shapes.include? shape
```
jmmastey: This code is a bit messy. Same with the above method. The old `if/else` has changed into three repetitions of `if shapes.include? shape`. As you read the method, for each line you have to carve out a mental exception of "only if shapes include shape".

Instead, try adding a single guard to the beginning of the method, to take care of the exception case (which is hopefully not the normal case):

```
def get_arrowhead(shapes, shape)
  raise UnknownShapeError, UNKNOWN_SHAPE unless shapes.include? shape
  
  arrowhead = shapes[shape]
  puts "You have a(n) #{arrowhead} arrowhead. Probably priceless."
end
```

Now the code is easier to follow and more concise. We only worry about the error state once, at the beginning. Sadly, this does repeat the logic for fetching a particular type of arrowhead. We can DRY that up with a little re-arrangement:

```
def get_arrowhead(shapes, shape)
  arrowhead = shapes[shape] or raise UnknownShapeError
  puts "You have a(n) #{arrowhead} arrowhead. Probably priceless."
end

I'm not positive this is better than the last one. I think it is, but there's a lot of logic on that one line. Such is life.

jmmastey:
I think you ended up with too much indirection by splitting the code apart. Take a look at the comments on the previously-arrowheaded methods themselves and see if you feel like it's better as two classes or if it's okay as one. Your call.

Good work so far!

PR #683, opened 2015-05-22
Add rlabuonora.txt

rlabuonora:
Hello there! This was pretty hard so I wasn't very considerate of the message in the file, so I'll say it here: thanks man!


jmmastey:
Good. Make sure it's not in your master branch anymore. We talked briefly on slack about it. :)

PR #687, opened 2015-05-26
Yadda - part 3 and 4

ryancowan:
Relevant files for parts 3 and 4:
https://github.com/ryancowan/level_up_exercises/blob/yadda_20150107_rc/yadda/yadda-views.sql
https://github.com/ryancowan/level_up_exercises/blob/yadda_20150107_rc/yadda/yadda-indices.sql
https://github.com/ryancowan/level_up_exercises/blob/yadda_20150107_rc/yadda/yadda-explains.sql

Here are the `EXPLAIN`s for the views before adding any indices:
Top Beers: http://explain.depesz.com/s/kIh
Recent Score: http://explain.depesz.com/s/ku4h
Also Enjoy: http://explain.depesz.com/s/VJE4

The big take away is that we are spending a lot of time in the `ratings` table, so that needs some decent indices supporting it. The other big issue is that we're spending a lot of time gathering up the data for the view, which can be reduced with more indices.

Here are the `EXPLAIN`s after adding the indices:
Top Beers: http://explain.depesz.com/s/RXu
Recent Score: http://explain.depesz.com/s/iUYJ
Also Enjoy: http://explain.depesz.com/s/8xUl

The Top Beers and Recent Score views now take nearly no time to get results. These two views went from 250/300 milliseconds down to less than 1. The Also Enjoy view also shows improvement, although not as drastic. In this case, we're going from 230 milliseconds down to around 50. The `ratings` table is still an issue with this view, specifically the Index Scan on `ratings`.


yadda/yadda-indices.sql
```diff
@@ -0,0 +1,13 @@
+-- Usage: psql -d yadda -f yadda/yadda-indices.sql
+
+-- indices
+CREATE UNIQUE INDEX breweries_id_index ON breweries(brewery_id);
+CREATE INDEX persons_id_index ON persons(person_id);
+CREATE UNIQUE INDEX ratings_id_index ON ratings(rating_id);
+CREATE INDEX ratings_person_id_index ON ratings(person_id);
+CREATE INDEX ratings_beer_id_index ON ratings(beer_id);
```
djkotowski: No need to write indices for primary keys.  They are implicit unique indices.

djkotowski:
Great job on these!

PR #691, opened 2015-05-27
Arrowhead

hebelken:


arrowhead/arrowhead.rb
```diff
@@ -15,18 +14,29 @@ class Arrowhead
...
-        puts "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
+      true
+    else
+      puts "Unknown region, please provide a valid region."
+      false
+    end
+  end
```
zsyed1: Lines 18 - 23: The convention for `if-else` is:

```
return true if CLASSIFICATIONS.include?(region)
puts "Unknown region, please provide a valid region"
return false
```

That reduces 6 lines to 3. but there is another issue ... see next comment!

zsyed1: Line 21: A method with a name like `known_region?` is called a `predicate method`, a fancy name for saying it returns `true/false`. So that should be its _only_ responsibility. Line 21 innocuously prints a string, but that should really live outside the method and in your `classify` method below. Your `classify` method is the one responsible for displaying messages.

Now, once you remove line 21, you will see that you can reduce your method to just one line!

arrowhead/arrowhead.rb
```diff
@@ -15,18 +14,29 @@ class Arrowhead
...
+  def self.known_shape?(region, shape)
+    if CLASSIFICATIONS[region].include? shape
+      true
     else
-      raise "Unknown region, please provide a valid region."
+      puts "Unknown shape value. Are you sure you know what you're"\
+         " talking about?"
+      false
+    end
+  end
```
zsyed1: Lines 27 - 33, please see comment on method `classify`.

arrowhead/arrowhead.rb
```diff
@@ -15,18 +14,29 @@ class Arrowhead
...
+
+  def self.classify(region, shape)
+    if known_region?(region) && known_shape?(region, shape)
+      arrowhead = CLASSIFICATIONS[region][shape]
+      puts "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
     end
```
zsyed1: Lines 37 - 40: As I mentioned above you can move the display messages into this method - something line:

```
  raise INVALID_REGION_MSG unless known_region?(region)
  raise INVALID_SHAPE_MSG unless known_shape?(region)
  arrowhead = ...
  puts ...
end
```

And store the two messages as constants in your class definition.

zsyed1: This is a good start! You can reduce the arrowheads even further if you follow the suggestions. I would recommend another round of refactor to really nail this down.

arrowhead/arrowhead.rb
```diff
@@ -1,5 +1,11 @@
 class Arrowhead
-  # This seriously belongs in a database.
+  INVALID_REGION_MSG = "Unknown region, please provide a valid region."
+  INVALID_SHAPE_MSG = "Unknown shape value. Are you sure you know what you're"\
+                      " talking about?"
+  SUCCESS_SHAPE = lambda do |var|
```
djkotowski: Why not just make this a method?

djkotowski:
Looks good.  You're done.

PR #694, opened 2015-05-27
Shopping Cart Test Plan v2

dustinhansen:
* Broke things out into separate features
* Eliminated any mentions of back end or behind the scenes operations
* Simplified the wording of things (more like plain English)


shopping_cart/features/cart.feature
```diff
@@ -0,0 +1,56 @@
...
+Scenario: Viewing an empty Cart (new user)
+  Given I am a User with no previous site sessions
+    And I am viewing the Cart page
+  Then I see messaging telling me that the Cart is currently empty
+
+Scenario: Viewing an empty Cart (no items)
+  Given I am a User
+    And I am viewing the Cart page
+    And there are 0 items in the Cart
+  Then I see messaging telling me that the Cart is currently empty
```
jmmastey: If there's only `Given` and `Then` in a test, it feels better to just make it a `When`. For instance, I think in this test the thing you're "doing" is `When I view the cart page`. This is very much a stylistic point, YMMV.

shopping_cart/features/cart.feature
```diff
@@ -0,0 +1,56 @@
...
+
+Scenario: View order totals
+  Given I am an User
+    And I am viewing the Cart page
+    And I have 1 or more Items in my Cart
+  Then the subtotal displayed is a total of all Items in my Cart (before tax and minus discounts)
```
jmmastey: This is perhaps too verbose :) I find it hard to parse this entire sentence.

shopping_cart/features/cart.feature
```diff
@@ -0,0 +1,56 @@
...
+    And the total displayed is a total of all Items in my Cart (after tax)
+
+Scenario: Revisiting a previous Cart
+  Given I am a User with a previous site session
+    And I have a Items in my Cart from a previous session
```
jmmastey: I'm not sure how I would write the assertions in this test. Any ideas?

shopping_cart/features/cart.feature
```diff
@@ -0,0 +1,56 @@
...
+  Then the contents of my previous Cart remain as part of my current Cart
+    And the order totals are updated accordingly
```
jmmastey: Not sure where the updating happens. I'm just checking my cart (same as `View order totals`)

shopping_cart/features/cart.feature
```diff
@@ -0,0 +1,56 @@
...
+
+Scenario: User adds Items to the Cart
+  Given I am a User
+    And I am viewing an Item page
+  When I click "Add To Cart"
+  Then the quantity of the corresponding Item in my Cart increases by the number selected in the "Quantity" dropdown
```
jmmastey: Definitely too verbose. One way to do this would be:

```
Background: There exists a Widget item that costs $10

Scenario: User adds items to the cart
  Given I am a user with an empty shopping cart
  When I visit the Widget product page
    And I click on Add to Cart
  Then I have one Widget in my cart
    And my subtotal is $10.00
```

jmmastey: Basically, it's okay to be really literal with these tests. They don't have to read as a completely generic case (and in fact they're harder to read when they _are_ written that way)

shopping_cart/features/coupon_codes.feature
```diff
@@ -0,0 +1,15 @@
...
+    And I have 1 or more Items in my Cart
+  When I enter a valid coupon code
+    Then the corresponding amount is subtracted from my subtotal
+
+Scenario: Apply invalid coupon/discount code to Cart
+  Given I am an User
+    And I am viewing the Cart page
+    And I have 1 or more Items in my Cart
+  When I enter an invalid coupon code
+    Then I see error messaging telling me the coupon code is invalid
```
jmmastey: Real-world, I'd definitely expand this to cover more cases. (expired coupon, wrong item, etc)

shopping_cart/shopping_cart.sublime-project
```diff
@@ -0,0 +1,8 @@
+{
+	"folders":
+	[
+		{
+			"path": "."
+		}
+	]
+}
```
jmmastey: `.gitignore` ;)

shopping_cart/test_plan.md
```diff
@@ -0,0 +1,72 @@
...
+  And I have a items in my cart from a previous session
+  And my previoous session was under the same user account
+  Then the contents of my previous cart remain as part of my current cart
+
+## Scenario: Revisiting a previous cart (unauthenticated users)
+  Given I am a Prior User
+  And I am not logged in
+  And I have a items in my cart from a previous session
+  And my previous session happened on the same device I am currently using
+  Then the contents of my previous cart remain as part of my current cart
```
jmmastey: Ooh a test plan! I dig it!

jmmastey:
Looking pretty good. Long and short, the only thing that's really worth changing is to make the tests Even More English. This is a really good start, past where most people are, but they read so generically that it's hard in some cases to tell what the real intent of the code is.

Suggestion: Have one of the interns read the tests and see how confused they get. :)

shopping_cart/features/cart.feature
dustinhansen: I figured "Given" was a mandatory part of each rule. It sounds like there are no hard rules surrounding required clauses when writing Cucumber rules?

dustinhansen: I like it. Makes sense.

shopping_cart/features/cart.feature
dustinhansen: I am guessing that for the purposes of writing Cucumber specs, we don't care what constitutes an accurate subtotal? Is that typically something you would recommend get handled in the step definitions instead?

shopping_cart/features/cart.feature
dustinhansen: I figured we could just write some test data to the user's cookie. We would of course destroy the existing cookie before doing so. Any problems with that?

shopping_cart/features/cart.feature
dustinhansen: But then, should I write tests similar to what you've written put for adding TWO items to cart? THREE? FOUR? I was trying to write one rules that would cover all of those scenarios. Am I getting too specific? Thoughts are welcome.

shopping_cart/shopping_cart.sublime-project
dustinhansen: If anyone else uses Sublime, this file will help them by loading up the project the way I had it loaded. But if it's company policy to take it out I will do so.

PR #695, opened 2015-05-27
Robot name

hebelken:


robot_name/robot_name.rb
```diff
@@ -1,32 +1,41 @@
...
 class Robot
-  attr_accessor :name
-
-  @@registry
+  attr_reader :name
 
   def initialize(args = {})
     @@registry ||= []
-    @name_generator = args[:name_generator]
+    self.name = args[:name_generator] ? args[:name_generator] : generate_name
```
jmmastey: The idea behind the `name_generator` (look at old line 12) is that it's a Proc, as opposed to a string. This new handler removes that functionality.

Can you think of a way to uniformly handle a passed proc and also the default `generate_name`?

robot_name/robot_name.rb
```diff
@@ -1,32 +1,41 @@
...
+  end
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  def name=(value)
+    @name = value if check_duplicate?(value)
+  end
```
jmmastey: If I do the following:

```
robot = Robot.new()
robot.name = "bob"
```

I can cause invalid results. This method should probably own the `registry`, as opposed to the constructor.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,41 @@
...
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def self.registry
+    @@registry
+  end
```
jmmastey: This code seems more or less unused? Oh, I see it at the bottom. Sure, it's a short method. :)

robot_name/robot_name.rb
```diff
@@ -1,32 +1,41 @@
...
+
+  def check_duplicate?(str)
+    if !@@registry.include?(str)
+      true
+    else
+      puts "A robot named '#{str}' already exists. Skipping robot."
+      return false
     end
+  end
```
jmmastey: There are several antipatterns happening in this method, it could use some revision. I'm going to nitpick here just so you see what I'm seeing:

1. The param name `str` is really unhelpful. `name` would be a better choice, or something else connoting the significance of the parameter.
2. The method name is a verb (`check`), which in the command/query pattern connotes that the method will mutate state. It does not, nor should it given it is a predicate method (ends in a question mark). A better method name would be something like `name_duplicated?`.
3. The method does `puts`, making it impossible to use this library in an app that doesn't print to the command line. 
4. If the name _is_  a duplicate, it fails to set a name at all but allows the program to continue to execute. This will lead to invalid objects down the road unless name is optional (which in this case it probably shouldn't be).
5. Don't do explicit return statements. It's actually unnecessary where it's used right now.
6. `if/else` statements that return `true/false` should be refactored to simply be the condition itself:

```
def name_duplicated?
  !registry.include?(str)
end
```

This makes the method _super_ easy to understand.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,41 @@
...
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def generate_name
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
+
+    2.times.map { generate_char.call }.join +
+      3.times.map { generate_num.call }.join
```
jmmastey: The linebreak here is a _little_ sketchy. More on that opinion: https://medium.com/@jmmastey/blunt-tools-line-length-and-complexity-ce540fb5b4bc

For `generate_char` and `generate_num`, is there any reason they need to be procs within this method? Can we move them somewhere else outside this method and still use them?

jmmastey:
Mostly small nitpicks, the method that checks for duplicate names is where I'd really focus on refactoring the code a bit. Let me know if you have any questions (or think I'm full of shit, which happens).

I see you're on the level up Slack, hit me up there (@joe) and we can talk more. Thanks!

PR #696, opened 2015-05-27
Triangle facts

hebelken:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,82 @@
...
-	def radians_to_degrees(rads)
-		return (rads * 180 / Math::PI).round
-	end
-end
+  attr_accessor :side1
+  attr_accessor :side2
+  attr_accessor :side3
+  attr_accessor :angle_a
+  attr_accessor :angle_b
+  attr_accessor :angle_c
```
jmmastey: If you do want accessors, keep the one-line style. Maybe two if it's too long:

```
attr_accessor :side1, :side2, :side3
attr_accessor :angle_a, :angle_b, :angle_c
```

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,82 @@
...
+
+  protected :side1
+  protected :side2
+  protected :side3
+  protected :angle_a
+  protected :angle_b
+  protected :angle_c
```
jmmastey: Interestingly you've found an unstylish rubyism that wasn't in the original :)

Protected accessors are a moderate anti-pattern. It's better to just use `@var` style if you really want them to stay internal. Also, from what I can tell, this makes it impossible to _read_ the vars, but not to _write_ them, which is confusing.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,82 @@
...
+
+  def initialize(side1, side2, side3)
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+    @angle_a = determine_angle(@side2, @side3, @side1)
+    @angle_b = determine_angle(@side1, @side3, @side2)
+    @angle_c = determine_angle(@side1, @side2, @side3)
```
jmmastey: If they're accessors, you need to calculate the angles on the fly. As it stands I can set a new side length and the angles become invalid. Really, this calculation isn't that bad (a few divisions take microseconds) so it'd almost certainly be acceptable to just re-calculate the values every time.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,82 @@
...
+  end
+
+  def law_of_cosines(a, b, c)
+    (a**2 + b**2 - c**2) / (2.0 * a * b)
+  end
```
jmmastey: Honest question: how do you feel about this method? I've tried it this way, versus the other one, and I can't decide whether it's a better refactor or not.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,82 @@
...
+    [@side1, @side2, @side3].uniq.length == 2
+  end
+
+  def scalene?
+    if !(equilateral? || isosceles?)
+      true
+    else
+      false
+    end
+  end
```
jmmastey: I mentioned the same thing in robot name, but since this is an `if/else` returning `true/false`, you can collapse the entire method body to just `!(equilateral? || isosceles?)`

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,82 @@
...
+
+  def right?
+    if (@angle_a || @angle_b || @angle_c) == 90
+      true
+    else
+      false
+    end
+  end
```
jmmastey: See above about conditionals. I also think this code is technically broken:

```
irb(main):003:0* 45 || 45 || 90                                                                  │
=> 45
```

Better off just using `.include?`

jmmastey:
Looks pretty good. Read the feedback and make sure you understand the reasoning, then go ahead and move on. Thanks!

triangle_facts/triangle_facts.rb
hebelken: I'm going by the thoughtbot style guide for this.
"Avoid multiple assignments per line (one, two = 1, 2)."

should I only apply this rule when assigning value to variables?

triangle_facts/triangle_facts.rb
hebelken: I'm not thrilled with anything I've come up with for this yet...

```ruby
  def determine_angle(a, b, c)
    law_of_cosines = (a**2 + b**2 - c**2) / (2.0 * a * b)
    rads = Math.acos(law_of_cosines)
    radians_to_degrees(rads)
  end
```

On one hand I feel like I am contradicting my own style by keeping ```radians_to_degrees()``` as its own method and removing ```law_of_cosines```, I'm attempting to justify it because ```law_of_cosines``` needs to take the exact same parameters as ```determine_angle```, and I don't like the idea of passing through the same params for the method. 

This still doesn't feel quite right though.

PR #699, opened 2015-05-28
Overlord v2 - Refactored based on feedback

dustinhansen:
bomb_spec.rb - Joe mentioned to stop using an instance variable for
@security_codes, but I need that variable to persist to the random_code
method. Using let(), it would lose scope every time. An instance
variable is the only way I’ve seen to do this and have found lots of
examples of people using it. If there’s a more elegant way I’d love for
someone to show me.


overlord/bomb.rb
```diff
@@ -0,0 +1,43 @@
+class Bomb
+  attr_reader :error_count
+  attr_reader :status
+
+  def initialize(activation_code = '1234', deactivation_code = '0000')
+    @activation_code = activation_code
+    @deactivation_code = deactivation_code
+    @status = 'inactive'
```
incazteca: Use symbols for setting statuses.

overlord/overlord.rb
```diff
@@ -1,14 +1,39 @@
...
+require './bomb'
 
-enable :sessions
+class Overlord < Sinatra::Base
+  enable :sessions
 
-get '/' do
-  "Time to build an app around here. Start time: " + start_time
-end
+  get '/?:activation_code?/?:deactivation_code?' do
```
incazteca: Make your routes meaningful. This should be a POST to setup with the activation and deactivation codes that a user will want to set up.

overlord/overlord.rb
```diff
@@ -1,14 +1,39 @@
...
+  end
+
+  post '/security' do
+    submit_code(params[:code])
+    response = { bomb_status: session[:bomb].status }
+    response[:error_count] = session[:bomb].error_count
+    response.to_json
+  end
+
+  def create_bomb(params)
```
incazteca: This entire method is not really necessary. The bomb can be initialized when a user enters the overlord page. You can initially create the bomb `session[:bomb] = Bomb.new` and then have a setup route to set the activation codes with a `set_activation_code` method. Or you could check if the parameters given are present and then pass them in when creating the bomb.

overlord/views/index.erb
```diff
@@ -0,0 +1,23 @@
...
+<html>
+	<head>
+		<title>Overlord</title>
+		<link href="/assets/lib/bootstrap/css/bootstrap.min.css" rel="stylesheet">
+		<link href="/assets/app/css/main.css" rel="stylesheet">
+	</head>
+	<body>
+		<h1>Overlord Example</h1>
+		<form name="security" action="/" method="post">
+			<input type="number" name="code" max="9999">
```
incazteca: Provide a box for the activation code and deactivation code. The current page is very confusing. Also include a label for the input.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+  def deactivate_bomb
+    subject.deactivate(@security_codes[:deactivate])
+  end
+
+  def explode_bomb
+    activate_bomb
+    3.times { subject.deactivate(random_code) }
+  end
+
+  before(:all) do
```
incazteca: Use let blocks for the code

```
let(good_activation_code) {'1234'}
let(bad_code) {'abcd'}
etc....
```

overlord/features/security_codes.feature
```diff
@@ -0,0 +1,39 @@
+Feature: Security Codes
```
incazteca: Your feature files should be grouped by a type of activity. Such as activation or deactivation of a bomb. An activation feature file will then include steps that are related to activating a bomb.

overlord/features/status_indicator.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Status Indicator
```
incazteca: You should just use the status indicator in the Then steps in the other feature files. There's no real need for these tests.

incazteca:
The bomb spec needs some work in using the let blocks as well as removing the loop at the bottom. Also the features files should be better organized. Also the routing for this needs to be worked on as well so that they are meaningful. Please resubmit.

overlord/public/assets/lib/bootstrap/js/bootstrap.js
```diff
@@ -0,0 +1,2317 @@
+/*!
```
djkotowski: This should be minified.

overlord/features/security_codes.feature
```diff
@@ -0,0 +1,39 @@
...
+  In order to securely arm or disarm the bomb
+  As a bomber
+  I want to be able to enter security codes that change the status of the bomb
+
+  @javascript
+  Scenario: Deactivate an inactive bomb
+    Given I am viewing an inactive bomb
+    When I enter any code that is not the activation code 1 time
```
djkotowski: This is kind of a long way of just saying "When I enter an invalid activation code _n_ times".  (I'm being very nitpicky, here.  Not a big deal.)

overlord/features/security_codes.feature
```diff
@@ -0,0 +1,39 @@
...
+  @javascript
+  Scenario: Unsuccessfully deactivate an active bomb
+    Given I am viewing an active bomb
+    When I enter any code that is not the activation code 3 times
+    Then the bomb is exploded
+
+  @javascript
+  Scenario: Interacting with an exploded bomb
+    Given I am viewing an exploded bomb
+    Then the security code form should be gone
```
djkotowski: You generally will not want a `Given`, `Then` scenario without a `When`.  In this case, the action `When` is viewing the exploded bomb.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,94 @@
...
+    submit_security_form
+  end
+
+  def default_activation_code
+    '1234'
+  end
+
+  def default_deactivation_code
+    '0000'
+  end
```
djkotowski: The defaults should be constants.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,94 @@
...
+      4.times { code += rand(9).to_s }
+      return code unless blacklist.include?(code)
+    end
+  end
+
+  def submit_security_form
+    find('form[name="security"] input[type="submit"]').click
+  end
+end
+World(SessionStepsHarness)
```
djkotowski: Why define these in a module only to bring the module into the world?  You can just extract these methods out.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,122 @@
+require 'pry'
+require './bomb.rb'
+
+describe Bomb do
+  def random_code(num = 4)
+    loop do
+      code = ''
+      num.times { code += rand(9).to_s }
+      return code unless @security_codes.include? code
+    end
```
djkotowski: This is flappy, as you'll be using a different code each time.  Think of the edge cases and explicitly define the codes you want entered.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+      it 'should remain in "exploded" state when we enter the correct deactivation code' do
+        subject.deactivate(random_code)
+        expect(subject.status).to eq('exploded')
+      end
+    end
+  end
+
+  context 'using default security codes' do
+    it_behaves_like 'mutable bomb' do
+    end
```
djkotowski: Remove the empty code block.

djkotowski:
Good work!  Take a look at the comments and refactor.  Resubmit when ready.

dustinhansen:
I am still having a really tough time understanding what the ask is for creating Bombs. We're not using a database so I have no way of storing information securely. My only option is to store the activation/deactivation codes in a cookie. Is that acceptable? I'm dying for a database right now.

PR #700, opened 2015-05-28
Arrowhead exercise

chichiwang:
* Flatten arrowhead code

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -15,19 +15,18 @@ class Arrowhead
     },
   }
 
-  # FIXME: I don't have time to deal with this.
+  ERR = {
+    region: "Unknown region, please provide a valid region.",
+    shape: "Unknown shape value. "\
+           "Are you sure you know what you're talking about?",
+  }
+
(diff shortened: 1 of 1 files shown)


djkotowski:
Looks good to me.  Well done!

arrowhead/arrowhead.rb
```diff
@@ -15,19 +15,18 @@ class Arrowhead
...
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        puts "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+    raise ERR[:region] if CLASSIFICATIONS[region].nil?
```
zbright: In this case, I prefer to use `unless CLASSIFICATIONS[region]`

Not sure what the style guide says on it, but either way is fine

djkotowski: I agree with Zach on using `unless` here.

PR #701, opened 2015-05-28
Triangle facts

sfoa6558:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+    side1 == side2 && side2 == side3
+  end
+
+  def isosceles?
+    [side1, side2, side3].uniq.length == 2
+  end
+
+  def scalene?
+    false if equilateral? || isosceles?
+  end
```
jmmastey: This method now returns `false` or `nil`, which is also falsy. Drop the `false if` and it'll all work.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+
+  def recite_facts
+    type_of_triangle
+
+    angles = calculate_angles(side1, side2, side3)
+
+    puts 'The angles of this triangle are ' + angles.join(',')
+
+    puts 'This triangle is also a right triangle!' if angles.include? 90
```
jmmastey: Add parens around the `90`.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+    puts 'This triangles is scalene'  if scalene?
+  end
+
+  def calculate_included_angle(a, b, c)
+    radians_to_degrees(Math.acos((b**2 + c**2 - a**2) / (2.0 * b * c)))
+  end
+
+  def calculate_angles(a, b, c)
+    [calculate_included_angle(a, b, c), calculate_included_angle(a, c, b),
+     calculate_included_angle(b, c, a)]
```
jmmastey: If you use this form (which is not bad), just add a newline to the second line and line them all up.

jmmastey:
Added one or two nits, but it's solid. Good work, move along :)

PR #702, opened 2015-05-28
Robot name

sfoa6558:


robot_name/robot_name.rb
```diff
@@ -1,29 +1,41 @@
...
 
-  @@registry
-
-  def initialize(args = {})
-    @@registry ||= []
+  def check_name(args)
     @name_generator = args[:name_generator]
+    @name = @name_generator.call if @name_generator
+    check_error if @name_generator
+  end
```
jmmastey: this method, if it were named more accurately, would be `maybe_generate_and_then_check_name`, which is what we're trying to avoid. also, the default `generate_name` happens WAAAAY down into the file. Coalesce those into one place so that the name generating logic is all in one place.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,41 @@
...
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  def generate_name
+    prefix = ('A'..'Z').to_a.sample(2).join
```
jmmastey: There is a tiny tiny tiny logic change here: I cannot create `ZZ123`. I think this is probably acceptable.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,41 @@
...
+    suffix = (100..999).to_a.sample(1).join
```
jmmastey: This, on the other hand, is clever, and should be fine. If you're taking one item, you can just do `(100..999).to_a.sample` and drop a little complexity.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,41 @@
...
+    @name =  prefix + suffix
```
jmmastey: Better to set the name whatever method owns "name setting", _not_ the one that owns "default name generating".

jmmastey: Ditty for checking errors. This method should do only one thing: generate a plausible name.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,41 @@
...
+    check_error
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
+  def check_error
+    puts @name
```
jmmastey: If you `puts` the name we can't use this library for an app (like Rails) other than command line. Just remove this.

jmmastey:
Added a few comments. You should move the logic for the name generation, but that's the only real issue. Keep it up :)

PR #703, opened 2015-05-28
Robot Name

chichiwang:
* Break the logic in the class `Robot` up into smaller methods

**Note:**
Rubocop did not like the use of the class variable `@@registry`, so after consulting with Robert Nubel I created a [Class Instance Variable](http://thoughts.codegram.com/understanding-class-instance-variables-in-ruby/) called `registry`.

Rubocop has 0 problems with this.

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -29,6 +29,11 @@ class Robot
     rname.join
   end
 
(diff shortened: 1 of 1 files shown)


djkotowski:
Looks good to me.

PR #705, opened 2015-06-01
Third Stab at Cucumber Rules

dustinhansen:
It would be great if someone could sit with me for 30 minutes or so to
help me understand what level of detail to include in these specs. In
some cases it seems like I’m too detailed and in other cases I’m not
specific enough. Does Enova have any published guidelines on this?


shopping_cart/features/cart.feature
```diff
@@ -0,0 +1,54 @@
...
+    And my Cart is empty
+  When I view the Cart page
+  Then I see messaging telling me that the Cart is currently empty
+
+Scenario: View order totals
+  Given I am a User
+    And I have 2 Widgets in my Cart
+  When I view the Cart page
+  Then I my subtotal is $20.00
+    And my total is $25.00
```
djkotowski: These features are too detailed.  Going from "2 widgets in cart" to "Subtotal is $20.00 and total is $25" doesn't make much sense.  If you reduce the detail, it will be easier to understand what you're testing.

shopping_cart/features/cart.feature
```diff
@@ -0,0 +1,54 @@
...
+
+Scenario: Revisiting a previous Cart
+  Given I am a User
+    And I have previous Cart data in my cookie
+  When I view the Cart page
+  Then I see the contents of my previous Cart
+
```
djkotowski: What about this scenario?

1.  User X has a cart with some items
2.  Anonymous (not logged in) user adds items to their cart
3.  Anonymous user signs in as User X

What does the cart look like?

shopping_cart/features/cart.feature
```diff
@@ -0,0 +1,54 @@
...
+  Then I have one Widget in my Cart
+    And my subtotal is $10.00
+    And my total is $15.00
+
+Scenario: User removes Item from the Cart
+  Given I am a User
+    And I have 2 Widgets in my Cart
+  When I view the Cart page
+    And I click "Remove From Cart"
+  Then I have no Widgets in my Cart
```
djkotowski: Change this to "0 Widgets" and you have reusability

shopping_cart/features/cart.feature
```diff
@@ -0,0 +1,54 @@
...
+
+Scenario: User adds Items to the Cart
+  Given I am a User
+    And my Cart is empty
+  When I am view the Widget product page
+    And I click "Add To Cart"
+  Then I have one Widget in my Cart
+    And my subtotal is $10.00
+    And my total is $15.00
+
```
djkotowski: What about adding an out-of-stock item or exceeding the current stock?

shopping_cart/features/shipping_estimates.feature
```diff
@@ -0,0 +1,15 @@
+Feature: Shipping Estimates
+
+Scenario: Get estimated shipping costs via ZIP code
+  Given I am a User
+    And I have 2 Widgets in my Cart
+  When I view the Cart page
+    And I enter "60647" in the "ZIP Code" field
+  Then I see shipping estimates for the "60647" ZIP code
+
+Scenario: Failed attempt to get estimated shipping costs
```
djkotowski: Good look at bad path

shopping_cart/features/coupon_codes.feature
```diff
@@ -0,0 +1,31 @@
...
+  When I view the Cart page
+    And I enter "AUGUSTSPECIAL" in the "Coupon Code" field
+  Then I see an "Expired Coupon Code" error
+
+Scenario: Apply an Item-specific coupon/discount code to a Cart where the Item doesn't exist
+  Given I am a User
+    And I have 2 Widgets in my Cart
+  When I view the Cart page
+    And I enter "LAWN15" in the "Coupon Code" field
+  Then I see an "Invalid Coupon Code" error
```
djkotowski: Would this be a useful error in this situation?

djkotowski:
This looks really good.  Just a few comments throughout.  No need for a resubmission.

shopping_cart/features/cart.feature
dustinhansen: There definitely should be a rule about exceeding current stock but if an Item is out of stock then then the user wouldn't see an "Add to Cart" button. I didn't define that because I thought we were trying to limit this exercise to the Cart page only. If I should expand this to include product inventory, account logins, product pages, etc. let me know?

shopping_cart/features/cart.feature
dustinhansen: I originally wrote my Cucumber rules to be extremely reusable but the rules began to resemble actual English less and less as I pursued that. Now it seems I've swung too far in the other direction so I'll have to find the happy medium.

shopping_cart/features/coupon_codes.feature
dustinhansen: I don't think so but I'm not sure how to test for "nothing happens". I'll give it a try.

djkotowski: Maybe a different error (e.g. relevant item not in cart)

PR #706, opened 2015-06-01
Data science v3

dustinhansen:
* Heavy rspec refactoring based on Dan's notes
 * Simplified options for generating reports (was creating a test suite nightmare)


data_science/cohort.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+    @no += 1 if result == 0
+    @yes += 1 if result == 1
+    generate_results
+  end
+
+  private
+
+  def generate_results
+    total = @no + @yes
```
djkotowski: When fetching instance vars that have readers, use the reader method rather than the instance variable.

data_science/cohort.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  CONFIDENCE_INTERVAL = 0.95
+
+  class InvalidResultError < StandardError
+    def initialize(result = nil)
+      @result = result
+    end
+
+    def message
+      "Unexpected result: " + value + " (0 or 1 are the accepted values)"
```
djkotowski: Prefer using interpolation here.
```ruby
"Unexpected result: #{value} (0 or 1 are the accepted values)"
```

djkotowski: Also, is `value` supposed to be `result` here?

data_science/cohort.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def initialize
+    @no = 0
+    @yes = 0
+  end
+
+  def add_sample_result(result)
+    unless [0, 1].include?(result)
+      error = InvalidResultError.new(result)
+      raise error
```
djkotowski: No need to split this into two lines.

```ruby
raise InvalidResultError.new(result)
```

zbright:
This looks really good. Job well done!

data_science/spec/splittest_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  subject(:test) { described_class.new }
+  let(:report) { test.generate_report(DATA_SOURCES[:test]) }
+
+  describe '#generate_report' do
+
+    context 'invalid data' do
+      it 'should raise a InvalidDataFormatError exception' do
+        expect do
+          test.generate_report(DATA_SOURCES[:invalid])
```
djkotowski: I still would prefer this be put in a `let` statement rather than making the `expect` block multiline.

```ruby
context "with invalid data" do
  let(:generated_report) { test.generate_report(DATA_SOURCES[:invalid]) }

  it "should raise an InvalidDataFormatError exception" do
    expect { generated_report }.to raise_error(Exceptions::InvalidDataFormatError)
  end
end
```

djkotowski: There's also another way you could do this to DRY it up even more, though it approaches a readability concern.

```ruby
describe "#generate_report" do
  let(:generated_report) { test.generate_report(data_source) }

  context "with invalid data" do
    let(:data_source) { DATA_SOURCES[:invalid] }
    # tests here
  end

  # more contexts
end

File: data_science/splittest.rb [hunk 5]
```diff
@@ -0,0 +1,71 @@
...
+    self
+  end
+
+  def confidence
+    raise Exceptions::NoDataSourceError if @cohorts.length < 1
+    tester = ABAnalyzer::ABTest.new(formatted_ab_values)
+    tester.chisquare_p
+  rescue ABAnalyzer::InsufficientDataError => e
+    error = Exceptions::InsufficientDataError.new(e.message)
+    raise error
```
djkotowski: Prefer
```ruby
raise Exceptions::InsufficientDataError, e.message
```

data_science/splittest.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+    end
+  end
+
+  def formatted_ab_values
+    values = {}
+    @cohorts.each do |name, data|
+      values[name] = { yes: data.yes, no: data.no }
+    end
+    values
```
djkotowski: L65-69 can be shortened with `#each_with_object`

```ruby
def formatted_ab_values
  @cohorts.each_with_object({}) do |(name, data), values|
    values[name] = { yes: data.yes, no: data.no }
  end
end
```

data_science/splittest_exceptions.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  class InsufficientDataError < StandardError
+    def initialize(message = '')
+      @message = message
+    end
+
+    def message
+      @message
+    end
```
djkotowski: `@message` should be included already as a reader.  If not, just define it with `attr_reader`.

djkotowski:
Great job on this!  I've added a few nitpicks, but no need for a resubmission.

PR #709, opened 2015-06-02
Dino catalog

hebelken:
I'm now using the CSV library, and utilizing the headers that I can customize fairly easily.
After reading some of Sandi Metz, I think its important that Dinosaur can answer questions about itself, so I've moved the main filter logic out of the catalog, and into the dinosaur class.

given the open inputs that I am accepting through the REPL, I was really struggling with cleanly branching based on specific keywords.. I settled on making each keyword accept the same kind of input. This works alright as long as users follow the rules...

I think if I were to go about another refactor I would interpret "chain" requirement less literally.

There is a lot more code in this version, but I think overall this is easier to read. Let me know what you think.


dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,33 @@
...
+    @dinosaur_catalog = DinosaurCatalog.new(dinos)
+  end
+
+  def run_user_interface
+    loop do
+      print GREETING
+      query = gets.chomp.split('.')
+      abort if query == ["exit"]
+      @dinosaur_catalog = @dinosaur_catalog.process_query(query)
+      @dinosaur_catalog.facts
```
zsyed1: This is a cool design. Did you consider making this method mutating `def process_query!` instead of passing back a new object each time? The `DinosaurCatalog` initializer simply sets the value of `@dinosaurs`.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,44 @@
...
+    @continent = row[:continent]
+    @diet = row[:diet]
+    @carnivore = row[:carnivore]
+    @weight = row[:weight].to_i if row[:weight]
+    @locomotion = row[:locomotion]
+    @description = row[:description]
+  end
+
+  def locomotion?(search_val)
+    locomotion.downcase.chomp.include? search_val if locomotion
```
zsyed1: A method-name like `def has_locomotion?(search_val)` makes a little more sense to me, but this is good as well.

zsyed1: Put parentheses around the arguments list.

```
locomotion && locomotion.downcase.chomp.include?(search_val)
```

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def period?(search_val)
+    period.downcase.chomp.include? search_val if period
+  end
+
+  def size?(search_val)
+    return big? if search_val == "big"
+    return small? if search_val == "small"
+  end
```
zsyed1: Another way to do this is:
```
return false unless ['big', 'small'].include?(search_val)
send(search_val + '?')
```
Yours is probably more readable.

dino_catalog/dinodex.rb
hebelken: That makes more sense... there is no reason to return a new object.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,52 @@
...
+    @dinosaurs = @dinosaurs.select do |dino|
+      dino.send(function.to_sym, param)
+    end
+    @dinosaurs.uniq
+  end
+
+  def facts
+    @dinosaurs.each do |dinosaur|
+      puts ""
+      puts instance_variable_collection(dinosaur)
```
zsyed1: Lines 37 and 39, don't need the empty string, just `puts` alone prints a new-line. You could also interpolate this on line 38:

```
puts "\n#{instance_variable_collection(dinosaur)}\n"
```

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def instance_variable_collection(dinosaur)
+    value = []
+    dinosaur.instance_variables.each do |var|
+      if dinosaur.instance_variable_get(var)
+        value << var.to_s + ": " + dinosaur.instance_variable_get(var).to_s
+      end
+    end
+    value
+  end
```
zsyed1: Lines 43 - 51: Your method logic is very readable the way you've done it. An alternative way that the ruby community might prefer is to use a `select` first over line 46, then use a `map` to create the array in 47.

Also, in line 47 you can use string interpolation:
```
value << "#{var}: #{dinosaur.instance_variable_get(var)}"
```
The calls to `to_s` are implicitly made in the interpolation, i.e. `#{var}` amounts to `#{var.to_s}`.

dino_catalog/dinosaur_loader.rb
```diff
@@ -0,0 +1,40 @@
...
+  def initialize(directory)
+    @directory = directory
+    @dinosaurs = []
+    get_files(directory)
+  end
+
+  def get_files(directory)
+    Dir.foreach(directory) do |item|
+      next if item == '.' || item == '..' || item == '.DS_Store'
+      load_file(item)
```
zsyed1: Line 16: :smile:

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,44 @@
...
+  def initialize(row)
+    @name = row[:name]
+    @period = row[:period]
+    @continent = row[:continent]
+    @diet = row[:diet]
+    @carnivore = row[:carnivore]
+    @weight = row[:weight].to_i if row[:weight]
+    @locomotion = row[:locomotion]
+    @description = row[:description]
+  end
```
zsyed1: You can vertically align the `=` equal signs to make this look prettier - totally optional.

dino_catalog/instructions.txt
```diff
@@ -0,0 +1,20 @@
...
+  weight: Returns all dinosaurs weight this exact number
+  exit: exits DinoDex
+
+Each of your queries will build off of eachother, to reset your query please relaod your dinodex
+
+Example 1: This will print all information for small bipedal dinosaurs that you know about
+  locomotion biped
+  size small
+Example 2: This will print all information for small bipedal dinosaurs that you know about
+  locomotion biped.size small
```
zsyed1: I enjoyed reading your code. I don't think the current comments warrant another PR though once @djkotowski kotowski goes through this there may very well be more comments!

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def print_instructions
+    File.readlines(INSTRUCTIONS).each do |line|
+      puts line
+    end
+  end
+end
+
+dino_dex = DinoDex.new('catalogs/')
```
djkotowski: It often isn't considered "Rubylike" to have a runner _class_.  Instead, extract the runner methods from the class.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,44 @@
...
+    period.downcase.chomp.include? search_val if period
+  end
+
+  def size?(search_val)
+    return big? if search_val == "big"
+    return small? if search_val == "small"
+  end
+
+  def weight?(search_val)
+    @weight == search_val.to_i if @weight
```
djkotowski: Use the attr_readers for these.  Only use the instance variable when setting.

dino_catalog/dinosaur_catalog.rb
djkotowski: If you choose not to use the select/map combo @zsyed1 suggests above, you can also use `#each_with_object` to eliminate L44 and L50.

```ruby
def instance_variable_collection(dinosaur)
  dinosaur.instance_variables.each_with_object([]) do |var, values|
    if dinosaur.instance_variable_get(var)
      values << "#{var}: #{dinosaur.instance_variable_get(var)}"
    end
  end
end

File: dino_catalog/dinosaur_loader.rb [hunk 11]
```diff
@@ -0,0 +1,40 @@
...
+
+  def initialize(directory)
+    @directory = directory
+    @dinosaurs = []
+    get_files(directory)
+  end
+
+  def get_files(directory)
+    Dir.foreach(directory) do |item|
+      next if item == '.' || item == '..' || item == '.DS_Store'
```
djkotowski: If you just ignore anything matching `/^\./`, that should cover these.  Otherwise, it may be simpler to use `next if ['.', '..', '.DS_STORE'].include?(item)`

dino_catalog/dinosaur_loader.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def load_file(filename)
+    filename = @directory + filename
+    CSV.foreach(
+      filename,
+      headers: true,
+      header_converters: header_converter) do |row|
+      @dinosaurs << Dinosaur.new(row)
+    end
```
djkotowski: With this indentation pattern, it's difficult to see at a glance where the argument list ends and the code block begins.

dino_catalog/dinosaur_loader.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def header_converter
```
djkotowski: This could just be a constant.

djkotowski:
Well done!  I've added some nitpicky comments throughout, but you're done with this one.

PR #711, opened 2015-06-03
Dino catalog

sfoa6558:
Refactored Dino Dex


dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,58 @@
...
+    @walking == 'Biped'
+  end
+
+  def late_cretaceous?
+    @period == 'Late Cretaceous'
+  end
+
+  def early_cretaceous?
+    @period == 'Early Cretaceous'
+  end
```
djkotowski: What about all Cretaceous?

For periods, it makes sense to use a more generalized period matcher (maybe using a regex).

dino_catalog/dino_start.rb
```diff
@@ -0,0 +1,26 @@
...
+loop do
+  begin
+    user_input = gets.chomp.downcase
+    dino_parse = DinoParse.new
+    dino =  DinoMenu.new(dino_parse.parse_csv(CSV_OPTIONS[user_input.to_sym]))
+    dino.run
+    break
+  rescue TypeError
+    puts 'Please only enter these options' + CSV_OPTIONS.keys.to_s.upcase
+    redo
```
djkotowski: Since the `begin` block takes up the entire loop, a `redo` is not necessary here.

dino_catalog/dino_menu.rb
```diff
@@ -0,0 +1,68 @@
...
+    puts 'Please only enter these options' + FILTER_OPTIONS.keys.to_s.upcase
+  end
+
+  def chain_methods(methods)
+    individual_method   =  methods.to_s.split(' ')
+    result = @dinodex
+    individual_method.each do |meth|
+      meth_sym = meth.to_sym
+      result = result.send(FILTER_OPTIONS[meth_sym])
+    end
```
djkotowski: This can be simplified by using `#inject` to get a new result each time.

dino_catalog/dino_menu.rb
```diff
@@ -0,0 +1,68 @@
...
+DINOPTIONS
+
+  CHAINED_OPTIONS = [:bic, :bib, :cs]
+
+  def initialize(dinos)
+    @dinodex = DinoDex.new(dinos)
+  end
+
+  def filter(user_input)
+    print_val(@dinodex.send FILTER_OPTIONS[user_input])
```
djkotowski: Wrap arguments with parentheses.

dino_catalog/dino_menu.rb
```diff
@@ -0,0 +1,68 @@
...
+  rescue TypeError
+    puts 'Please only enter these options' + FILTER_OPTIONS.keys.to_s.upcase
```
djkotowski: When is `TypeError` raised?  It's often better to validate input than to rely on rescuing errors.

djkotowski:
Much improved!  A few comments throughout, but no need for a resubmission.  Good job!

dino_catalog/dino_menu.rb
djkotowski: Also, when concatenating strings like this, use `<<` to alter the original (left-most) string rather than `+` which creates a new string.

PR #716, opened 2015-06-08
added all the files for the data science project

sfoa6558:
This is the first version of data_science


data_science/load_data.rb
```diff
@@ -0,0 +1,15 @@
+require 'json'
+
+class LoadData
```
jmtai: Naming is a tricky thing and sometime more art than science, but there are definitely some things we can do to make our code clearer and easier to understand.

Here you've named your class "LoadData", which is an **action**, not a "thing".  Something like "DataLoader" might be more appropriate here.

data_science/load_data.rb
```diff
@@ -0,0 +1,15 @@
...
+  attr_accessor :json_data
+
+  def initialize(json_file)
```
jmtai: Again, the naming of the input parameter makes it harder to understand what's going on. `json_file` would lead me to believe that the parameter here is some sort of JSON file object, not a filename. Perhaps `json_filename', instead?

data_science/load_data.rb
```diff
@@ -0,0 +1,15 @@
...
+    raise "This is an empty file" unless json_file
```
jmtai: Whoops! Is this error message accurate? The code isn't actually examining the **contents** of the file, it's just checking to see if a filename has been provided.

data_science/split_test.rb
```diff
@@ -0,0 +1,74 @@
+require_relative 'load_data.rb'
+require 'abanalyzer'
+
+# This class is used to define errors for invalid cohorts
+class InvalidCohortError < RuntimeError
```
jmtai: Nice job defining a descriptive error class. But is this the only way the cohort data can be invalid? What if the cohort label isn't 'A' or 'B'? What if there's no date provided?

data_science/split_test.rb
```diff
@@ -0,0 +1,74 @@
...
+  attr_accessor :visitors
+
+  def initialize(data)
+    @json_data = data
+    @values = {}
+    @conversiona = 0
+    @conversionb = 0
+    @nonconversiona = 0
+    @nonconversionb = 0
+    @visitors = []
```
jmtai: `visitors` is storing all of the result values from the JSON data, but it's only being used to calculate `cohorts_count`. Perhaps it would be better to simply store the count directly instead?

data_science/split_test.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def cohorts_count
+    @visitors.size
+  end
+
+  def calc_conversions
+    @json_data.each do |element|
+      calc_conversiona(element) if element['cohort'] == 'A'
+      calc_conversionb(element) if element['cohort'] == 'B'
```
jmtai: What if the data labels are lowercase 'a' and 'b'?

Also, take a look at these two lines.  They look almost identical, right? Perhaps there's a way to make a more generic method that can be used for either cohort.

data_science/split_test.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+  end
+
+  def calc_conversiona(element)
+    raise InvalidCohortError unless element['result'] == 0 || element['result'] == 1
+    @conversiona += 1 if element['result'] == 1
+    @nonconversiona += 1 if element['result'] == 0
+  end
+
+  def calc_conversionb(element)
```
jmtai: `calc_conversiona` and `calc_conversionb` are almost identical. We want our code to be DRY (Don't Repeat Yourself). Here we have a couple of functions that are doing exactly the same thing, just operating on differently named output  variables.

What if we had a `Cohort` class that would include all the operations that can be applied per-cohort. Do you think that would help you remove a lot of code duplication?

data_science/split_test.rb
```diff
@@ -0,0 +1,74 @@
...
+    raise InvalidCohortError unless element['result'] == 0 || element['result'] == 1
+    @conversionb += 1 if element['result'] == 1
+    @nonconversionb += 1 if element['result'] == 0
+  end
+
+  def conversion_rate(conversion, nonconversion)
+    total__conversion = conversion + nonconversion
```
jmtai: Any particular reason you used two underscores here in your variable name?

data_science/split_test.rb
```diff
@@ -0,0 +1,74 @@
...
+    ABAnalyzer.confidence_interval(conversion, total__conversion, CONV_FACTOR)
+  end
+
+  def create_converhash(conversion, nonconversion)
```
jmtai: The convention is to **not** include return data types in the method name. Instead, try to come up with a more descriptive name.  Since this is returning data that is being passed into the ABAnalyzer, perhaps something like `analyzer_cohort_data` (or whatever better name you come up with).

data_science/split_test.rb
```diff
@@ -0,0 +1,74 @@
...
+    convert_and_non_converta = {}
```
jmtai: Any particular reason this has an `a` suffix? The name implies that this data only belongs to cohort `a` but this is actually a generic method.

data_science/spec/split_tester_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  describe 'Number of conversions for each cohort' do
+    it 'returns the number of conversions when given a cohort' do
+      expect(ab_test_split.conversiona).to eq(13)
+      expect(ab_test_split.conversionb).to eq(10)
+    end
+  end
+
+  describe 'Get conversion rate for cohort' do
```
jmtai: You've got the error bars, but what is the actual conversion rate per cohort?

data_science/spec/split_tester_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    it 'returns conversion rate for each cohort' do
+      expect(ab_test_split.conversion_rate(ab_test_split.conversiona,ab_test_split.nonconversiona)).to \
+       eq([0.26727411833986714, 0.6292776057980639])
+      expect(ab_test_split.conversion_rate(ab_test_split.conversionb,ab_test_split.nonconversionb)).to \
+       eq([0.11245469104354977, 0.3753501870052307])
+    end
+  end
+
+  describe 'Accept or reject Null Hypothesis that Current leader is better' do
```
jmtai: current leader is better than what?

jmtai:
Good start! There's a fair bit of code duplication here that could be reduced by creating a separate `Cohort` class.

You'll also want tests for your `LoadData` class and make sure you're not just testing the happy path. We want our code to be robust so we should make sure we test the sad path and error path, as well.

data_science/spec/split_tester_spec.rb
```diff
@@ -0,0 +1,34 @@
+require_relative 'spec_helper'
+
+RSpec.describe SplitTestAB do
+  let(:data) { LoadData.new('../testabdata_sample.json') }
+  let(:ab_test_split) { SplitTestAB.new(data.json_data) }
+
+  describe 'Total Sample Size' do
```
djkotowski: Write the method name for these describes.

```ruby
describe "#total_sample_size" do
```

data_science/spec/split_tester_spec.rb
```diff
@@ -0,0 +1,34 @@
+require_relative 'spec_helper'
+
+RSpec.describe SplitTestAB do
```
djkotowski: This file should be `split_test_ab_spec.rb`

data_science/split_test.rb
```diff
@@ -0,0 +1,74 @@
+require_relative 'load_data.rb'
```
djkotowski: This file should be `split_test_ab.rb`

djkotowski:
Good first attempt.  Take a look at the comments, refactor and resubmit when ready.

PR #717, opened 2015-06-08
Data science

hebelken:
TDD try 1..
Not using english tests.


data_science/lib/conversion_calculator.rb
```diff
@@ -0,0 +1,44 @@
+require 'import_data'
+require 'abanalyzer'
+
+class ConversionCalculator
+  GROUP_NAME = 'cohort'
```
jmtai: Nice use of constants. However, what happens if the labels in the data set were spelled *Cohort* or *COHORT*?

data_science/lib/conversion_calculator.rb
```diff
@@ -0,0 +1,44 @@
...
+  RESULT_NAME = 'result'
+  attr_reader :data
+
+  def initialize(data)
+    @data = data
+    values = {}
+    values[:agroup] =
+      { converted: conversion_num('A'), unconverted: non_conversion_num('A') }
```
jmtai: This is a personal preference, but if you need to split a hash definition onto multiple lines, this form makes it easier to read and edit:

```
values[:agroup] = {
  converted: conversion_num('A'),
  unconverted: non_conversion_num('A')
}
```

Now it's easy to comment out a particular entry without having to modify anything else.

data_science/lib/conversion_calculator.rb
```diff
@@ -0,0 +1,44 @@
...
+    values[:bgroup] =
+      { converted: conversion_num('B'), unconverted: non_conversion_num('B') }
```
jmtai: Similar to your `GROUP_NAME` and `RESULT_NAME` constants, what happens if the data set contains lower case cohort labels?

data_science/lib/conversion_calculator.rb
```diff
@@ -0,0 +1,44 @@
...
+    @analyzer = ABAnalyzer::ABTest.new(values)
+  end
+
+  def sample_size(group)
+    data.count { |item| item[GROUP_NAME] == group }
```
jmtai: Do you think it's a good idea to recompute the sample size every time you access it? Do you expect the sample size to change after the initial import?

data_science/lib/conversion_calculator.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def conversion_num(group)
+    group_data = data.select { |item| item[GROUP_NAME] == group }
+    group_data.count { |item| item[RESULT_NAME] == 1 }
+  end
```
jmtai: Here the data is being parsed **and** counted. Ideally, we'd want our methods to be single responsibility. Can you think of a way to make that happen?

jmtai: Additionally, as with the sample size, does it make sense to filter and count the conversions from the raw data each time you want to read the value?

data_science/lib/conversion_calculator.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def non_conversion_num(group)
+    group_data = data.select { |item| item[GROUP_NAME] == group }
+    group_data.count { |item| item[RESULT_NAME] == 0 }
+  end
+
+  def conversion_rate(group)
+    ABAnalyzer.confidence_interval(conversion_num(group), sample_size(group), 0.95)
```
jmtai: The method name is `conversion_rate` but it's calling `ABAnalyzer.confidence_interval`. Are you sure this method is accurately named?

data_science/spec/conversion_calculator_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+require 'import_data'
+require 'abanalyzer'
+
+RSpec.describe ConversionCalculator do
+  before(:each) do
+    @confidence = 0.95
+    @a_conversion = 340
+    @a_sample = 660
+    @b_conversion = 380
+    @b_sample = 728
```
jmtai: All these variables are just constants, so you don't need to put them in a `before(:each)` block.  Just define them at the top level. e.g.

```
RSpec.describe ConversionCalculator do
  CONFIDENCE = 0.95
  A_CONVERSION = 340
  etc.
```

Also, the convention would be to put your `require` statements in `spec_helper.rb` and just require `spec_helper.rb` in each of your spec files.

data_science/spec/conversion_calculator_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'conversion_calculator'
+require 'import_data'
+require 'abanalyzer'
+
+RSpec.describe ConversionCalculator do
```
jmtai: No need for the explicit `RSpec` here. You can just write

```ruby
describe ConversionCalculator do
```

data_science/spec/conversion_calculator_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  it "#conversion_num" do
+    expect(@calc.conversion_num('A')).to eq @a_conversion
+    expect(@calc.conversion_num('B')).to eq @b_conversion
+  end
+
+  it "#non_conversion_num" do
+    expect(@calc.non_conversion_num('A')).to eq @a_sample - @a_conversion
+    expect(@calc.non_conversion_num('B')).to eq @b_sample - @b_conversion
```
jmtai: It's best not to calculate the expected result here... you run the danger of creating tautological tests. For your particular implementation this works ok, but imagine if your `non_conversion_num` method was implemented like so:

```ruby
def non_conversion_num(group)
  sample_size(group) - conversion_num(group)
end
```

Now the test mirrors the implementation, so there's no way for it to fail.

data_science/spec/conversion_calculator_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  it "#conversion_rate" do
```
jmtai: Is this test actually checking the conversion rate or is it actually the confidence interval?

data_science/spec/conversion_calculator_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    expect(@calc.conversion_rate('A')).to eq [0.47702328622933254, 0.5532797440736977]
+    expect(@calc.conversion_rate('B')).to eq [0.4856925781577534, 0.5582634657982906]
+  end
+
+  it "#chi_square" do
+    expect(@calc.chi_square).to eq 0.0646179059352707
```
jmtai: This is the chi_square score, but how do we get (from 2.iii in the Requirements) `Confidence level that the current leader is in fact better than random`?

data_science/lib/import_data.rb
```diff
@@ -0,0 +1,10 @@
+require 'json'
+
+class ImportData
```
jmtai: Classes are **things** - that is, a collection of variables and methods that are used to perform actions. Grammatically, classes should be **nouns** and methods **verbs**.  Here you've got a class named `ImportData`, which is a verb.

Perhaps something like `DataImporter` would be better? That would make it more consistent with the naming of your other class: `ConversionCalculator`

data_science/lib/import_data.rb
```diff
@@ -0,0 +1,10 @@
...
+  attr_reader :data
+
+  def initialize(filename)
+    file = File.read(filename)
```
jmtai: What happens if this is called without a filename (i.e. filename = nil). What if a filename is passed in for a file that doesn't exist? What if the data in the file is bad? You see where I'm going with this? :)

jmtai:
Good start! The implementation is compact and easy to read, but what if you had a `Cohort` class that could encapsulate the cohort-related data and methods. Maybe it would be cleaner and even easier to comprehend if you were able to write something like this:

```ruby
cohort.conversions
cohort.sample_size
```
vs
```
@calc.conversion_num('A')
@calc.sample_size('A')

Take a look at the comments, add some more tests so you're not just testing the happy path, and resubmit. Let me know if you have any questions.

File: data_science/spec/conversion_calculator_spec.rb (hunk 9 above)
Comment (hebelken) [inline data_science/spec/conversion_calculator_spec.rb]: I understand what you are getting at. I don't want to be using methods that I define as a result of my test. I suppose I shouldn't create a noticeable relationship between my methods in the test code even if it is subtracting two constants.

File: data_science/lib/import_data.rb (hunk 13 above)
Comment (hebelken) [inline data_science/lib/import_data.rb]: I'm adding an exception to the initialize method that simply checks for a filename to exist, and changing the library name to be JsonImporter. This kind of error checking always confuses me though. 

The way it stands now, a user will be able to eventually figure out that they need to pass a file to the constructor, and that the file needs to exist, and be JSON. The built in ruby exceptions, along with the File, and json libraries will get them to a place where they can use my library.. After my change, it makes it a tad easier for them to get to the answer. I find it difficult to figure out where to draw this line though.

File: data_science/lib/conversion_calculator.rb (hunk 2 above)
Comment (djkotowski) [inline data_science/lib/conversion_calculator.rb]: And why not go a step further and define values in one line?

File: data_science/spec/conversion_calculator_spec.rb (hunk 7 above)
Comment (djkotowski) [inline data_science/spec/conversion_calculator_spec.rb]: Also, `let` statements can be a benefit here, especially if you want to let them be overridden by contexts.

File: data_science/spec/conversion_calculator_spec.rb (hunk 9 above)
Comment (djkotowski) [inline data_science/spec/conversion_calculator_spec.rb]: Another good scenario for using `let` statements.

File: data_science/spec/conversion_calculator_spec.rb (hunk 10 above)
Comment (djkotowski) [inline data_science/spec/conversion_calculator_spec.rb]: Also, this should be a `describe` and `it` pairing.

File: data_science/spec/conversion_calculator_spec.rb (hunk 11 above)
Comment (djkotowski) [inline data_science/spec/conversion_calculator_spec.rb]: Ditto above on `describe`.

File: data_science/spec/spec_helper.rb [hunk 14]
```diff
@@ -0,0 +1,91 @@
+# This file was generated by the `rspec --init` command. Conventionally, all
```
djkotowski: Delete generated comments.

djkotowski:
Good first attempt.  Take a look at the comments, refactor and resubmit.  Also look at [betterspecs.org](http://betterspecs.org) for a reference.

Note:  Better Specs has had some server issues lately, so don't be surprised if the link fails on the first try.

PR #720, opened 2015-06-09
Cleaned up the code

vbhat:
First attempt.

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -1,32 +1,51 @@
-class NameCollisionError < RuntimeError; end
+class NameCollisionError < RuntimeError
+end
 
 class Robot
   attr_accessor :name
 
-  @@registry
+  class << self
+    attr_accessor :registry
+  end
+  @registry = []
 
   def initialize(args = {})
-    @@registry ||= []
-    @name_generator = args[:name_generator]
+    @name_generator = args[:name_generator] || method(:generate_random_name)
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+    @name = @name_generator.call
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
-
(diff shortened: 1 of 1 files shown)


robot_name/robot_name.rb
```diff
@@ -1,32 +1,51 @@
...
+    @name = "#{generate_char.call}#{generate_char.call}"
+    @name += "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
   end
 end
 
+def name_generation_error
+  puts 'There was a problem generating the robot name!'
+end
+
+def name_wrong_format?(name)
```
djkotowski: Maybe just `name_invalid?`?

robot_name/robot_name.rb
```diff
@@ -1,32 +1,51 @@
...
+    self.class.registry << name
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def generate_random_name
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
+    @name = "#{generate_char.call}#{generate_char.call}"
+    @name += "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
```
djkotowski: Since you're modifying the existing string, use `<<` instead of `+=`.  `+=` creates a new string and reassigns `@name`, whereas `<<` alters the original string.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,51 @@
...
-    end
+    raise NameCollisionError, name_generation_error if name_invalid?(name)
+    self.class.registry << name
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def generate_random_name
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
```
djkotowski: If L22-23 were their own methods, this method would be simplified.

djkotowski:
Small methods everywhere!  Great work.  You're done.

PR #721, opened 2015-06-09
Edited code to make it look cleaner

vbhat:
First attempt.

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -1,34 +1,45 @@
 class Arrowhead
-  # This seriously belongs in a database.
-  CLASSIFICATIONS = {
-    far_west: {
-      notched: "Archaic Side Notch",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Agate Basin",
-      bifurcated: "Cody",
-    },
-    northern_plains: {
-      notched: "Besant",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Humboldt Constricted Base",
-      bifurcated: "Oxbow",
-    },
-  }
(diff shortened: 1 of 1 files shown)


arrowhead/arrowhead.rb
```diff
@@ -1,34 +1,45 @@
...
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
+  def initialize(classifications)
+    @classifications = classifications
+  end
+
+  def region_check(region)
+    raise "Unknown region, please provide a valid region." \
+    unless classifications.include? region
```
djkotowski: Wrap args in parentheses for clarity.  It's readable here, but it easily gets more difficult as you add complexity.

djkotowski: Also, avoid using `\` to split lines.  Find ways to simplify the expressions.

arrowhead/arrowhead.rb
```diff
@@ -1,34 +1,45 @@
...
+
+  def shape_check(region, input_shape)
+    if classifications[region].include? input_shape
+      arrowhead = classifications[region][input_shape]
+      puts "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
     else
-      raise "Unknown region, please provide a valid region."
+      raise "Unknown shape value. Are you sure you know \
+      what you're talking about?"
     end
```
djkotowski: This is still a bit arrowhead-y.  If you move your `raise` to a guard clause, you get rid of the `if`/`else` statement.

djkotowski:
Good work.  Take a look at the comments; no need for a resubmission.

PR #722, opened 2015-06-09
First working version of Dino Catalog

vbhat:


dino_catalog/dino_parse.rb
```diff
@@ -0,0 +1,38 @@
+require 'csv'
+require 'pry'
+require './dinosaur'
+
+class CSVParser
+  def self.read(*filenames)
+    dino_array = filenames.each_with_object([]) do |filename, dinos|
+      CSV.foreach(filename, headers: true, header_converters: :downcase) do |row|
```
incazteca: A lambda can be passed in here so that you can call `fix_each_dino` method.

dino_catalog/dino_parse.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def self.dinos_convert(dinos)
+    dinos.each do |dino|
+      fix_each_dino(dino) if dino.key?("genus")
+    end
+    dinos
+  end
+
+  def self.dino_objects(dinos)
```
incazteca: You can make this method private. Also I don't think the `self` is needed here.

dino_catalog/dino_parse.rb
```diff
@@ -0,0 +1,38 @@
+require 'csv'
+require 'pry'
+require './dinosaur'
+
+class CSVParser
+  def self.read(*filenames)
+    dino_array = filenames.each_with_object([]) do |filename, dinos|
+      CSV.foreach(filename, headers: true, header_converters: :downcase) do |row|
+        dinos << row.to_hash
```
incazteca: Instead of saving the dinosaur hash to an array just have the Dinosaur object created and have that in the array. You'll be able to get rid of the `dinos_convert` and `dino_objects` methods.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,27 @@
...
+require 'pry'
+require './dino_parse'
+require './dinodex'
+
+dinos = CSVParser.read('dinodex.csv', 'african_dinosaur_export.csv')
+
+dinodex = Dinodex.new(dinos)
+original_dinodex = dinodex
+
+puts "Please enter desired properties (Biped, Carnivore, Big or Small, Period) : "
```
incazteca: Is there an option for listing all the dinosaurs?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,43 @@
...
+    Dinodex.new(big_dinos)
+  end
+
+  def small
+    small_dinos = dino_array.select { |dino| dino.weight.to_i < 4000 }
+    Dinodex.new(small_dinos)
+  end
+
+  def carnivore
+    diet_array = %w(Yes Carnivore Insectivore Piscivore)
```
incazteca: In the CSV parsing you should have the `Yes` become Carnivore.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,43 @@
...
+    carnivore_dinos = dino_array.select { |dino| diet_array.include? dino.diet }
```
incazteca: The Dino object should have a `carnivore?` method which contains the logic deciding if it's a carnivore or not. This method in Dinodex will just then be:
```
def carnivore
  Dinodex.new(dino_array.select { |dino| dino.carnivore? } )
end
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,43 @@
...
+    Dinodex.new(carnivore_dinos)
+  end
+
+  def biped
+    biped_dinos = dino_array.select { |dino| dino.walking == "Biped" }
```
incazteca: Again, Dino should have a `walking?` method.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,27 @@
...
+properties.each do |property|
+  method = "#{property}".to_sym
+
+  dinodex = dinodex.send(method) if dinodex.methods.include?(method)
+end
+
+puts dinodex.dino_array.map(&:name)
+puts "\n"
+
+puts "Please enter a dinosaur you'd like to learn more about:"
```
incazteca: You could just have the script loop until the user types in `exit`. At the top of each loop you create a new DinoDex object and basically allow the user to perform another query once they're done with the one they just did.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,52 @@
...
+    @period = dinosaur["period"]
+    @diet = dinosaur["diet"]
+    @weight = dinosaur["weight_in_lbs"]
+    @walking = dinosaur["walking"]
+    @continent = dinosaur["continent"]
+    @description = dinosaur["description"]
+  end
+
+  def to_s
+    print_name
```
incazteca: You could probably simplify this by making this a HEREDOC or having the attributes printed out in a loop.

incazteca:
@vbhat Take a look at the comments and resubmit. By using the header converters in the CSV parsing you'll be able to streamline the CSVParser class. Also be sure to have the logic regarding the Dinosaurs walk, size, diet... etc be in the Dino class.

dino_catalog/dinodex.rb
djkotowski: Which then becomes
```ruby
def carnivore
  Dinodex.new(dino_array.select(&:carnivore?))
end
```

dino_catalog/dinodex.rb
djkotowski: Don't name variables after their type.

dino_catalog/dino_catalog.rb
vbhat: The way it works right now, if I don't enter any properties, it lists all the dinosaurs. I'll go ahead and make it clearer.

vbhat:
Thanks for the feedback guys. I'll get started on refactoring this soon.

PR #723, opened 2015-06-10
Overlord v3 - Optimized w/ AJAX

dustinhansen:
* Bomb “activate” and “deactivate” functions - I know it can be more
DRY but not sure how to go about it.


dustinhansen:
Closing this because I noticed there are Cucumber tests yet to be written.

PR #728, opened 2015-06-16
Data science

hebelken:
Attempting to fix bugs form the initial review.

overall new design
conversion rate now correctly named
new method for confidence interval
users can now specify the group, and result names
key metrics no longer calculated at the time of access
going for more single-responsibility. (CohortImporter, might not fall into this)
rspec tests now using let and describe-it pairing
using #different to tell if the current leader is better than random.
nil filenames now raising custom error


data_science/spec/cohort_importer_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+describe CohortImporter do
+  describe "#initialize" do
+    it "raises an error when nil filename is passed to it" do
+      expect { CohortImporter.new(nil) }.to raise_error(InvalidFile)
+    end
+  end
+
+  let(:cohort_importer) { CohortImporter.new('data_export_2015_06-3.json') }
+  let(:parsed_group) { { "B" => { conversions: 380, non_conversions: 348 }, "A" => { conversions: 340, non_conversions: 320 } } }
+  let(:group_a) { { "A" => { conversions: 340, non_conversions: 320 } } }
```
zbright: Lines 11-15 are not used, please delete them

data_science/spec/cohort_importer_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  let(:group_b) { { "B" => { conversions: 380, non_conversions: 348 } } }
+  let(:cohort_a) { Cohort.new(group_a) }
+  let(:cohort_b) { Cohort.new(group_b) }
+
+  describe "#groups" do
+    it "sets up the groups for the cohorts" do
+      expect(cohort_importer.groups).to eq parsed_group
+    end
+  end
```
zbright: You might want to consider adding tests for poorly formatted json files and for missing file names. I saw in the other files that you have nice errors being thrown, use em!

data_science/lib/cohort.rb
```diff
@@ -0,0 +1,37 @@
...
+  def initialize(args)
+    @data = args
+    parse_args(args)
+    @sample_size = sample_size
+    @conversion_rate = calculate_conversion_rate
+    @confidence_interval = calculate_confidence_interval
+  end
+
+  def parse_args(args)
+    args.each do |key, _value|
```
zbright: Not sure I understand this loop. These variables would get overwritten each time

data_science/lib/cohort_importer.rb
```diff
@@ -0,0 +1,43 @@
...
+  def setup_groups(groups = {})
+    @data.each do |record|
+      groups[record[@group]] = { conversions: 0, non_conversions: 0 }
+    end
+    groups
+  end
+
+  def transform_conversions(groups)
+    @data.each do |record|
+      groups[record[@group]][:conversions] += 1 if record[@result] == 1
```
zbright: This seems like a good opportunity to use a ternary. Something like 
```
type = record[@result] == 1 ? :conversions ": :non_converstions
```

zbright: This also might be a good spot to use `.each_with_object` where the object would be the groups hash

data_science/lib/cohort_importer.rb
```diff
@@ -0,0 +1,43 @@
...
+      groups[record[@group]][:non_conversions] += 1 if record[@result] == 0
+    end
+    groups
+  end
+
+  def make_cohorts(groups)
+    cohorts = []
```
zbright: This is definitely a good spot for `.each_with_object` using the cohorts array as the object. It even gets rid of the return on line 41!

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  let(:group_a) { { "A" => { conversions: 340, non_conversions: 320 } } }
+  let(:group_b) { { "B" => { conversions: 380, non_conversions: 348 } } }
+  let(:group_c) { { "C" => { conversions: 1000, non_conversions: 6 } } }
+  let(:cohort_a) { Cohort.new(group_a) }
+  let(:cohort_b) { Cohort.new(group_b) }
+  let(:cohort_c) { Cohort.new(group_c) }
+
+  describe "#name" do
+    it "knows its name" do
+      expect(cohort_a.name).to eq "A"
```
zbright: Dont think you need to test all three of these in each of these blocks. If one fails they will all fail

data_science/spec/ab_tester_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  let(:group_c) { { "C" => { conversions: 1000, non_conversions: 6 } } }
+  let(:cohort_a) { Cohort.new(group_a) }
+  let(:cohort_b) { Cohort.new(group_b) }
+  let(:cohort_c) { Cohort.new(group_c) }
+  let(:abtest_a) { ABTester.new([cohort_a, cohort_b]) }
+  let(:abtest_b) { ABTester.new([cohort_a, cohort_b, cohort_c]) }
+
+  describe "#calculate_better_than_random" do
+    it "knows when the leader is better than random" do
+      expect(abtest_a.calculate_better_than_random).to eq false
```
zbright: There are rspec matchers for this kinda stuff. Take a look at 'be_truthy` and `be_falsy` for future reference

data_science/lib/cohort_importer.rb
```diff
@@ -0,0 +1,43 @@
...
+    @result = result
+    file = File.read(filename)
+    @data = JSON.parse(file)
+    @groups = transform_conversions(setup_groups)
+    @cohorts = make_cohorts(@groups)
+  end
+
+  private
+
+  def setup_groups(groups = {})
```
zbright: You dont ever pass anything into this method. In general, setting the default here isnt wrong but I try to avoid it. You open yourself up to a whole bunch of other problems as soon as another developer starts working on the codebase. But maybe thats just me. Im curious what @djkotowski thinks about this

hebelken: I agree. There is no reason to include the default like that when I can do the same thing on the first line. I think that makes a lot more sense. I made the method private to make sure the groups can't be changed after they have been setup after all...

data_science/spec/ab_tester_spec.rb
hebelken: It doesn't sit well with me that this test could pass if the method returned 7. 

I suppose following convention here by making the this method ```calculate_better_than_random?``` could make it more clear to the next developer on the project, and that might decrease that chance, but I still have a bit in my stomach about the truthy and falsy test.

Am I just paranoid?

hebelken:
didn't realize pushing more changes would impact this pull request, sorry for cluttering things up.

data_science/spec/cohort_importer_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+describe CohortImporter do
+  describe "#initialize" do
+    it "raises an error when nil filename is passed to it" do
+      expect { CohortImporter.new(nil) }.to raise_error(InvalidFile)
+    end
+  end
+
+  let(:cohort_importer) { CohortImporter.new('data_export_2015_06-3.json') }
+  let(:parsed_group) { { "B" => { conversions: 380, non_conversions: 348 }, "A" => { conversions: 340, non_conversions: 320 } } }
```
djkotowski: You can split this into several lines.

data_science/lib/ab_tester.rb
```diff
@@ -0,0 +1,25 @@
...
+
+class ABTester
+  attr_reader :groups
+
+  def initialize(args)
+    groups = make_groups(args)
+    @groups = groups.reduce({}, :merge)
+  end
+
+  def calculate_better_than_random
```
djkotowski: Why not just `better_than_random?`?

data_science/spec/ab_tester_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  let(:group_c) { { "C" => { conversions: 1000, non_conversions: 6 } } }
+  let(:cohort_a) { Cohort.new(group_a) }
+  let(:cohort_b) { Cohort.new(group_b) }
+  let(:cohort_c) { Cohort.new(group_c) }
+  let(:abtest_a) { ABTester.new([cohort_a, cohort_b]) }
+  let(:abtest_b) { ABTester.new([cohort_a, cohort_b, cohort_c]) }
+
+  describe "#calculate_better_than_random" do
+    it "knows when the leader is better than random" do
+      expect(abtest_a.calculate_better_than_random).to be_falsy
```
djkotowski: If you follow my naming suggestion above, you could use predicate matchers here.

```ruby
expect(abtest_a).not_to be_better_than_random
expect(abtest_b).to be_better_than_random
```

djkotowski:
Looks good!  A few suggestions throughout, but no need for a resubmission.

PR #730, opened 2015-06-19
Overlord

dietpop:


overlord/bomb.rb
```diff
@@ -0,0 +1,56 @@
+class Bomb
+  ACTIVATION_CODE = '1234'
+  DEACTIVATION_CODE = '0000'
+  COUNTDOWN = 30
+  ALLOWED_RETRIES = 3
+  attr_accessor :activation_code, :deactivation_code, :state, :retries
```
paulh16: I'd have a space here to separate your constants from your accessor methods.

overlord/bomb.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def initialize
```
paulh16: Consider refactoring this method to accept user defined codes.

overlord/bomb.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def bomb_state
+    @state
+  end
+
+  def boot_bomb
+    @state = :booted
+  end
+
+  def activate(code)
```
paulh16: This and the `deactivate` method return booleans, so I'd prefer them to end in a question mark. But the bigger question is the naming of the method. I'd expect active to only change the bomb's state to active, not to return true or false as well. I'd like to talk about this in person.

overlord/bomb.rb
```diff
@@ -0,0 +1,56 @@
...
+    if valid_deactivation_code?(code)
+      @state = :deactivated
+      true
+    else
+      false
+    end
+  end
+
+  def valid_activation_code?(code)
+    if code == ACTIVATION_CODE
```
paulh16: Consider having using `@activation_code` instead of `ACTIVATION_CODE`. This will make your program more flexible, and extends on the comment I made in your initialize method.

overlord/features/activate.feature
```diff
@@ -0,0 +1,17 @@
+
```
paulh16: Overall comment: really nice job splitting your feature files into small, manageable files. This makes it easy to identify specific features.

overlord/features/activate.feature
```diff
@@ -0,0 +1,17 @@
...
+Feature: Activate the bomb
+  In order potentially blow myself up
+  As the user
+  I want to activate a booted bomb
+
+  Scenario: Activate bomb
+    Given I am on the activate page
+    #And the bomb is booted
+    When I enter 1234
```
paulh16: Consider phrasing this without the specifics of the code. Such as:

`When I enter a correct activation code`

I think that makes it easy to understand what this step accomplishes at first glance.

overlord/features/activate.feature
```diff
@@ -0,0 +1,17 @@
...
+
+  Scenario: Activate bomb
+    Given I am on the activate page
+    #And the bomb is booted
+    When I enter 1234 
+    Then I see the bomb is active
+
+  Scenario: Fail to activate bomb
+    Given I am on the activate page
+    When I submit the wrong activation code 1111
```
paulh16: Same thing here. Consider changing this to:

`When I enter an incorrect activation code`

overlord/features/activate.feature
```diff
@@ -0,0 +1,17 @@
+
+Feature: Activate the bomb
+  In order potentially blow myself up
+  As the user
+  I want to activate a booted bomb
+
+  Scenario: Activate bomb
+    Given I am on the activate page
```
paulh16: Since this is comment across the two scenarios in this feature, you can extract it into a background step:

https://github.com/cucumber/cucumber/wiki/Background

overlord/features/deactivate.feature
```diff
@@ -0,0 +1,22 @@
+
+Feature: Deactivate an active bomb
+  In order to not get blown up
+  As the user
+  I want to deactivate the active bomb
+
+  Scenario: I deactivate the bomb
+    Given I am on the activated bomb page
+    When I input the correct deactivation code 0000
```
paulh16: Same comment as in `activate.feature`. There is no need to be super specific on the code.

overlord/features/deactivate.feature
```diff
@@ -0,0 +1,22 @@
...
+    #And the countdown is greater than 0
+    Then I see the bomb deactivated message
+
+  Scenario: I fail to deactivate the bomb
+    Given I am on the activated bomb page
+    When I input the wrong code 1111
+    Then I see a bomb active message
+
+  Scenario: I do nothing 
+    Given I am on the activated bomb page
```
paulh16: Extract the Given steps for these three scenarios into a background step.

overlord/features/deactivate.feature
```diff
@@ -0,0 +1,22 @@
...
+    When I wait 31 seconds
```
paulh16: Here is another instance of specificity. Unless I look in the code, I don't know what 31 seconds implies. But if you phrase this step as `When the counter expires`, then I immediately understand.

overlord/features/step_definitions/activate_steps.rb
```diff
@@ -0,0 +1,29 @@
+Given(/^I am on the activate page$/) do
```
paulh16: I really like your step definitions here. Consider organizing them in order by Given, When and Then. That way, if I'm looking through the file, they are easier to find.

overlord/views/activate.erb
```diff
@@ -0,0 +1,41 @@
+<style>
```
paulh16: Really nice work using layouts!

overlord/public/main.js
```diff
@@ -0,0 +1,48 @@
+
+var timer = document.getElementById("timer");
```
paulh16: Wow! Impressive using straight JavaScript instead of jQuery! :)

overlord/my_overlord.rb
```diff
@@ -0,0 +1,76 @@
...
+
+get '/detonate' do
+  erb :detonate
+end
+
+post '/state' do
+  code = params['input']
+  code_is_valid = false
+  my_bomb = session[:my_bomb]
+  if (my_bomb.bomb_state == :active) && my_bomb.deactivate(code)
```
paulh16: Consider extracting `my_bomb.bomb_state == :active) && my_bomb.deactivate(code)` into its own well-named method.

overlord/features/boot.feature
```diff
@@ -0,0 +1,11 @@
+Feature: Boot the bomb
```
paulh16: Nice feature here. Can't think of anything else needed.

overlord/features/activate.feature
```diff
@@ -0,0 +1,17 @@
+
+Feature: Activate the bomb
+  In order potentially blow myself up
+  As the user
+  I want to activate a booted bomb
+
+  Scenario: Activate bomb
```
paulh16: Consider the name of this scenario and the next. Would the following be more clear?

`Enter correct activation code`
`Enter incorrect activation code`

Just a question I'd like to pose. I think the above may be more clear about the purpose of the two scenarios.

overlord/features/deactivate.feature
```diff
@@ -0,0 +1,22 @@
...
+    When I input the correct deactivation code 0000
+    #And the countdown is greater than 0
+    Then I see the bomb deactivated message
+
+  Scenario: I fail to deactivate the bomb
+    Given I am on the activated bomb page
+    When I input the wrong code 1111
+    Then I see a bomb active message
+
+  Scenario: I do nothing
```
paulh16: I'd rename this to: 

`The counter expires`

overlord/features/deactivate.feature
```diff
@@ -0,0 +1,22 @@
+
+Feature: Deactivate an active bomb
+  In order to not get blown up
+  As the user
+  I want to deactivate the active bomb
+
+  Scenario: I deactivate the bomb
```
paulh16: See the comment on naming in `activate.feature`.

overlord/features/deactivate.feature
```diff
@@ -0,0 +1,22 @@
...
+    Given I am on the activated bomb page
+    When I input the correct deactivation code 0000
+    #And the countdown is greater than 0
+    Then I see the bomb deactivated message
+
+  Scenario: I fail to deactivate the bomb
```
paulh16: Good job considering the various paths in this feature. You'll expand on this later in the Test Engineering module with Happy, Sad and Bad paths.

overlord/features/step_definitions/activate_steps.rb
```diff
@@ -0,0 +1,29 @@
...
+  fill_in('code', with: code)
+  click_button('Activate')
+end
+
+Then(/^I see the bomb is active$/) do
+  expect(page).to have_content('active')
+end
+
+Given(/^The bomb is booted$/) do
+  visit "/"
```
paulh16: Be on the lookout for duplication and opportunities to DRY out your code with the extract method refactoring. See lines 2-3 and 17-18.

paulh16:
@dietpop @djkotowski Overall, this is very nice work. Here is some general comments:

1. Nice job putting most of the logic in the bomb class and not cluttering up your Sinatra file. You'll see this again in Rails with the "Skinny Controller; Fat Model" concept.

2. Good use of views and layouts.

3. Really nice organization of your Cucumber tests. They were split up into individual files quite nicely. Just take a look at naming for clarity.

4. Good job on not adding too many steps to each scenario, and keeping much of the small actions in the step definitions. You don't want "Click button X" in the feature file. You do a nice job of this.

5. Look for opportunities to DRY up your code, whether through the use of Background steps in the feature file or support methods in your step definitions.

The only thing missing is unit tests for your bomb class. This should be rather quick to resubmit.

overlord/bomb.rb
djkotowski: I'm not sure making these predicate methods would be ideal.  Predicate methods tend to imply that they only return information about the object, not that they're modifying it.

overlord/bomb.rb
djkotowski: Simplify
```ruby
if condition
  true
else
  false
end
```

to

```ruby
condition
```

overlord/bomb.rb
```diff
@@ -0,0 +1,56 @@
...
+      false
+    end
+  end
+
+  def valid_deactivation_code?(code)
+    if code == DEACTIVATION_CODE
+      true
+    else
+      false
+    end
```
djkotowski: Ditto above.

overlord/features/boot.feature
djkotowski: Maybe:

```cucumber
Scenario:  Boot bomb unsuccessfully
  Given I have an unbooted bomb
  And Dan is standing behind me
  When I press the boot bomb button
  Then everything breaks
```

overlord/my_overlord.rb
```diff
@@ -0,0 +1,76 @@
...
+  session[:my_bomb] = my_bomb
+  redirect('/activate')
+  # add feature that accepts act and deact codes from Villian
+end
+
+get '/activate' do
+  my_bomb = session[:my_bomb]
+  state = my_bomb.bomb_state
+  if state
+    @state_message = "bomb is #{state.to_s}"
```
djkotowski: Yep.  Interpolation always calls `#to_s`

djkotowski:
Nice work!  Just a few comments throughout.  Add some specs, resubmit, and you should be good to go!

overlord/features/activate.feature
dietpop: Good tip on the background.  Thanks for the link.

overlord/features/deactivate.feature
dietpop: Yes, good point.

overlord/public/main.js
dietpop: :)

overlord/my_overlord.rb
dietpop: Yes

overlord/my_overlord.rb
```diff
@@ -0,0 +1,76 @@
...
+  state = my_bomb.bomb_state
+  if state == :active
+    redirect('/deactivate')
+  end
+  if my_bomb.activate(params[:code])
+    @state_message = "Bomb is #{state.to_s}"
+    session[:my_bomb] = my_bomb
+    redirect('/deactivate')
+  else
+    @state_message = "Wrong code. Bomb is #{state.to_s}."
```
dietpop: Noted.

overlord/bomb.rb
dietpop: Doh!  A 5 line if statement on a boolean.

PR #735, opened 2015-06-22
arrowhead refactoring

yuvalroth:
code cleanup

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -2,36 +2,34 @@ class Arrowhead
   # This seriously belongs in a database.
   CLASSIFICATIONS = {
     far_west: {
-      notched: "Archaic Side Notch",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Agate Basin",
-      bifurcated: "Cody",
+      notched: 'Archaic Side Notch',
+      stemmed: 'Archaic Stemmed',
+      lanceolate: 'Agate Basin',
+      bifurcated: 'Cody',
     },
     northern_plains: {
-      notched: "Besant",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Humboldt Constricted Base",
-      bifurcated: "Oxbow",
+      notched: 'Besant',
+      stemmed: 'Archaic Stemmed',
+      lanceolate: 'Humboldt Constricted Base',
+      bifurcated: 'Oxbow',
     },
   }
 
   def self.classify(region, shape)
-    if CLASSIFICATIONS.include? region
-      classify_shape_for_specfic_region region, shape
-    else
-      raise "Unknown region, please provide a valid region."
+    unless CLASSIFICATIONS.include? region
+      raise 'Unknown region, please provide a valid region.'
     end
+    classify_shape_for_specific_region region, shape
   end
 
   private_class_method :new
-  def self.classify_shape_for_specfic_region(region, shape)
+  def self.classify_shape_for_specific_region(region, shape)
     shapes = CLASSIFICATIONS[region]
-    if shapes.include? shape
-      "You have a(n) '#{shapes[shape]}' arrowhead. Probably priceless."
-    else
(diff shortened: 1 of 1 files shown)


arrowhead/arrowhead.rb
```diff
@@ -15,20 +15,24 @@ class Arrowhead
...
   def self.classify(region, shape)
     if CLASSIFICATIONS.include? region
-      shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
+      classify_shape_for_specfic_region region, shape
```
jmtai: Extracting this check to a new method is a good way to reduce the overall size of the `classify` method as well as reduce the overall arrowheadedness. However, in Ruby the convention is to prefer guard clauses over if-else constructs.

For example, you can write the first line of the function like this:
```ruby
raise "Unknown region, please provide a valid region." if CLASSIFICATIONS.include? region
```

arrowhead/arrowhead.rb
```diff
@@ -15,20 +15,24 @@ class Arrowhead
...
     else
       raise "Unknown region, please provide a valid region."
     end
   end
+
+  private_class_method :new
+  def self.classify_shape_for_specfic_region(region, shape)
```
jmtai: Oops! Typo here: `specfic`  :)

arrowhead/arrowhead.rb
```diff
@@ -15,20 +15,24 @@ class Arrowhead
...
+    shapes = CLASSIFICATIONS[region]
+    if shapes.include? shape
+      "You have a(n) '#{shapes[shape]}' arrowhead. Probably priceless."
+    else
+      raise "Unknown shape value."\
```
jmtai: Add a space before the `\` character.

jmtai:
Nice job! Try refactoring the code using guard clauses and resubmit when you're done.

djkotowski:
Guard clauses help you break execution in a single line.  They typically follow these formats:

```ruby
raise YourException, "message" if condition
# or
return value unless condition
# or
break if condition
```

arrowhead/arrowhead.rb
```diff
@@ -2,32 +2,34 @@ class Arrowhead
   # This seriously belongs in a database.
   CLASSIFICATIONS = {
     far_west: {
-      notched: "Archaic Side Notch",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Agate Basin",
-      bifurcated: "Cody",
+      notched: 'Archaic Side Notch',
```
djkotowski: Sometimes rubocop complains about this, but we generally don't care if you use single or double-quotes, as long as they make sense for what you're trying to accomplish.

arrowhead/arrowhead.rb
```diff
@@ -2,32 +2,34 @@ class Arrowhead
...
-      shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
+    unless CLASSIFICATIONS.include? region
```
djkotowski: Wrap arguments in parentheses.

arrowhead/arrowhead.rb
```diff
@@ -2,32 +2,34 @@ class Arrowhead
...
+      raise 'Unknown region, please provide a valid region.'
     end
+    classify_shape_for_specific_region region, shape
```
djkotowski: Ditto above.

djkotowski:
Looks good.  Take a look at the comments for future exercises.  No need for a resubmission.

PR #736, opened 2015-06-22
cleanup triangle_facts

yuvalroth:
code cleanup

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -1,58 +1,73 @@
 # Killer facts about triangles AWW YEAH
 class Triangle
-	attr_accessor :side1,:side2,:side3
+  attr_accessor :side1, :side2, :side3
 
-	def initialize(side1,side2,side3)
-		@side1,@side2,@side3 = side1,side2,side3
-	end
+  def initialize(side1, side2, side3)
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+  end
 
-	def equilateral()
-		return side1 == side2 && side2 == side3
-	end
+  def equilateral
+    side1 == side2 && side2 == side3
+  end
 
-	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
-	end
+  def isosceles
+    [side1, side2, side3].uniq.length == 2
+  end
 
-	def scalene()
-		unless ! (equilateral || isosceles)
-	    false
-	  else
-	    true
-	  end
-	end
+  def scalene
+    if equilateral || isosceles
+      true
+    else
+      false
+    end
+  end
 
-	def recite_facts
-		puts 'This triangle is equilateral!' if equilateral
-		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+  def recite_facts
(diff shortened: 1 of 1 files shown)


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,73 @@
...
-	end
+  attr_accessor :side1, :side2, :side3
+
+  def initialize(side1, side2, side3)
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+  end
+
+  def equilateral
```
jmtai: In Ruby, the convention is for predicated methods (methods which simply return `true` or `false`) to end with a question mark. So this method would be named `equilateral?`. This helps the code read more naturally, since something like
```ruby
triangle.equilateral
```
looks like a statement, but
```ruby
triangle.equilateral?
```
makes it clear that we're asking if the triangle is equilateral

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,73 @@
...
+    side1 == side2 && side2 == side3
+  end
+
+  def isosceles
```
jmtai: Same comment about naming here.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,73 @@
...
+    [side1, side2, side3].uniq.length == 2
+  end
+
+  def scalene
+    if equilateral || isosceles
```
jmtai: Just as you've done with the `equilateral`and `isosceles` methods, you can eliminate the if-else and just use the conditional statement since Ruby will return the value of the evaluated conditional.

jmtai: Also, check the logic here - the triangle is being classified as scalene if it's either equilateral or isosceles. Does that sound right?

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,73 @@
...
+    angles = calculate_angles(side1, side2, side3)
+    puts 'The angles of this triangle are ' + angles.join(',')
+    puts 'This triangle is also a right triangle!' if angles.include? 90
+    puts ''
+  end
+
+  def calculate_angles(a, b, c)
+    angle_a = calculate_angle_a(a, b, c)
+    angle_b = calculate_angle_b(a, b, c)
+    angle_c = calculate_angle_a(a, b, c)
```
jmtai: Whoops! Does this look right?

jmtai:
Looks good! Take a look at the comments and refactor a bit but no need to resubmit this one.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,73 @@
...
+    end
+    puts 'This triangle is scalene and mathematically boring.' if scalene
+    recite_facts_based_on_angles
+  end
+
+  def recite_facts_based_on_angles
+    angles = calculate_angles(side1, side2, side3)
+    puts 'The angles of this triangle are ' + angles.join(',')
+    puts 'This triangle is also a right triangle!' if angles.include? 90
+    puts ''
```
djkotowski: No need for the empty string here.  Just `puts` will take care of that for you.

djkotowski:
Good work!  Take a look at the comments, refactor and resubmit.  You're almost done!

PR #737, opened 2015-06-22
Exercise: Creates short methods and short lines of code

yuvalroth:
Exercise: Create short methods and short lines of code

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -3,23 +3,41 @@ class NameCollisionError < RuntimeError; end
 class Robot
   attr_accessor :name
 
-  @@registry
+  @registry ||= []
 
   def initialize(args = {})
-    @@registry ||= []
     @name_generator = args[:name_generator]
+    generate_name
+    check_name_collision
+    @registry << @name
+  end
 
+  def check_name_collision
+    return unless invalid_generated_name
+    raise NameCollisionError, 'There was a problem generating the robot name!'
+  end
+
+  def generate_name
     if @name_generator
       @name = @name_generator.call
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+      @name = defualt_name_generator
     end
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def defualt_name_generator
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
+    "#{generate_char.call}#{generate_char.call}"\
+    "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  end
+
+  def invalid_generated_name
+    !does_generated_name_follow_pattern || @registry.include?(name)
+  end
+
+  def does_generated_name_follow_pattern
+    (name =~ /[[:alpha:]]{2}[[:digit:]]{3}/)
   end
 end
 


robot_name/robot_name.rb
```diff
@@ -3,23 +3,41 @@ class NameCollisionError < RuntimeError; end
 class Robot
   attr_accessor :name
 
-  @@registry
+  @registry ||= []
```
jmtai: Careful - this probably isn't doing what you want. The original code clearly defines the registry as a **class** variable by using the '@@' notation. Here you've changed the notation so that it **looks** like an instance variable but since it's defined outside of any method body, it's actually still being defined on the class itself.

The problem now is that the first time you reference `@registry` from within one of your methods, it won't actually be defined, since that is an **instance** variable.

Try running the code and see what happens.

robot_name/robot_name.rb
```diff
@@ -3,23 +3,41 @@ class NameCollisionError < RuntimeError; end
...
+  end
+
+  def defualt_name_generator
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
+    "#{generate_char.call}#{generate_char.call}"\
+    "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  end
+
+  def invalid_generated_name
```
jmtai: Predicate methods should end with a '?'. You'll also typically see names with the condition being tested as the last word. e.g. `generated_name_invalid?`

robot_name/robot_name.rb
```diff
@@ -3,23 +3,41 @@ class NameCollisionError < RuntimeError; end
...
-    @@registry ||= []
     @name_generator = args[:name_generator]
+    generate_name
+    check_name_collision
+    @registry << @name
+  end
+
+  def check_name_collision
+    return unless invalid_generated_name
+    raise NameCollisionError, 'There was a problem generating the robot name!'
```
jmtai: You can combine these two lines into a single raise by using the form
```ruby
raise NameCollisionError, 'There was a problem generating the robot name!' if invalid_generated_name
```

If you want to shorten up the length of the line, you can define the error message in the NameCollisionError class itself like so:
```ruby
class NameCollisionError < RuntimeError
  def message
    'There was a problem generating the robot name!'
  end
end
```

robot_name/robot_name.rb
```diff
@@ -3,23 +3,41 @@ class NameCollisionError < RuntimeError; end
 class Robot
```
jmtai: Your file name should match the name of the class. It doesn't matter for these exercises but when you start working with Rails it will cause issues with the autoloader.

robot_name/robot_name.rb
```diff
@@ -3,23 +3,41 @@ class NameCollisionError < RuntimeError; end
...
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+      @name = defualt_name_generator
     end
+  end
+
+  def defualt_name_generator
```
jmtai: Watch your spelling. :smile:

robot_name/robot_name.rb
```diff
@@ -3,23 +3,41 @@ class NameCollisionError < RuntimeError; end
...
+    "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  end
+
+  def invalid_generated_name
+    !does_generated_name_follow_pattern || @registry.include?(name)
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def does_generated_name_follow_pattern
```
jmtai: Naming is tricky and sometimes comes down to personal preference but I'd argue that this one is unnecessarily long. It's also really a predicate method since we're looking for a true or false result.

Perhaps something like `name_is_valid?`

robot_name/robot_name.rb
```diff
@@ -3,23 +3,41 @@ class NameCollisionError < RuntimeError; end
...
+    (name =~ /[[:alpha:]]{2}[[:digit:]]{3}/)
```
jmtai: The parens aren't necessary here.

robot_name/robot_name.rb
```diff
@@ -3,23 +3,41 @@ class NameCollisionError < RuntimeError; end
...
+    @registry << @name
+  end
+
+  def check_name_collision
+    return unless invalid_generated_name
+    raise NameCollisionError, 'There was a problem generating the robot name!'
+  end
 
+  def generate_name
     if @name_generator
```
jmtai: Here you've got an if-else to select between a custom or default name generator. Wouldn't it be nicer if you could just specify a default generator when the instance is created so you could always just use `@name_generator.call`?

Hint: Your initialize method sets the default value for the args parameter to an empty hash. Perhaps we could put something else there instead?

robot_name/robot_name.rb
```diff
@@ -3,23 +3,41 @@ class NameCollisionError < RuntimeError; end
...
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+      @name = defualt_name_generator
     end
+  end
+
+  def defualt_name_generator
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
```
jmtai: Using lambdas here make the code a little harder to read, since you're now forced to use `.call` to get a result. Perhaps it would be cleaner to use methods so you could just use something like `random_char` and `random_num` instead?

robot_name/robot_name.rb
```diff
@@ -3,23 +3,41 @@ class NameCollisionError < RuntimeError; end
...
+    "#{generate_char.call}#{generate_char.call}"\
+    "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
```
jmtai: For repeated invocations it might make the code easier to read if you use the `times` Integer method:
```ruby
name = ""
2.times { name << generate_char.call }
3.times { name << generate_num.call.to_s }
```

jmtai:
Good start! Take a look at the comments, refactor and resubmit.

robot_name/robot_name.rb
djkotowski: You can gain access to the class instance registry with a class method or accessor.

Example:
```ruby
class << self
  attr_reader :registry
end

@registry ||= []

def initialize(args = {})
  # ...
  self.class.registry << @name
end
```

There's also a decent reference [here](http://www.railstips.org/blog/archives/2006/11/18/class-and-instance-variables-in-ruby/).

djkotowski:
Nice work!  [Ditto](https://github.com/jmmastey/level_up_exercises/pull/737#issuecomment-115285498) what Josh said.

PR #740, opened 2015-06-23
Simplify overlord

djkotowski:
For review by @paulh16

### DIFF
--- a/overlord/spec/wire_bundle_spec.rb
+++ b/overlord/spec/wire_bundle_spec.rb
@@ -1,0 +1,62 @@
+require_relative "spec_helper"
+
+describe WireBundle do
+  let(:disarming_wire_count) { 3 }
+  let(:detonating_wire_count) { 2 }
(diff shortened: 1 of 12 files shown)


overlord/spec/wire_bundle_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  let(:disarming_wire_count) { 3 }
+  let(:detonating_wire_count) { 2 }
+  let(:total_wire_count) { disarming_wire_count + detonating_wire_count }
+  subject(:bundle) do
+    described_class.new(disarming_wire_count, detonating_wire_count)
+  end
+
+  describe "#new" do
+    it "requires a number of disarm and detonating wires" do
+      expect(bundle).to have(disarming_wire_count).disarming_wires
```
bwthomas: I'd prefer to leave ```rspec-collection_matchers``` out of our dependencies. If we do that, then we would just have something like this:

```expect(bundle.disarming_wires.length).to be >= disarming_wire_count```

PR #747, opened 2015-06-25
Arrowhead jnyman

jnyman:
This is my refactor attempt at the Arrowhead code example.

My goal was to tighten up the classify() method, which would allow me to most efficiently raise the error condition when the test condition was checked. I wanted that logic to be as compressed as possible but still expressive as to intent.


djkotowski:
Awesome!  Nice job.

PR #749, opened 2015-06-26
Just git

jklemen:
Completed Basic Ruby Section.


djkotowski:
Make sure every Ruby exercise is in its own clean branch.

jklemen:
Should I go back and change it?

PR #752, opened 2015-06-29
Robot Name jnyman

jnyman:
I split this into multiple files to make it easier to see what I was doing.

I'm not sure why Procs were used for the generator argument since those names were not being "generated"; they were simply provided. Adding a call to a proc seemed a bit too much complexity to me just to provide a string, so I changed that bit, hopefully still in the spirit of the exercise.

I tried to keep each class from knowing as much about each other as possible. RobotFactory violates that with Robot since it creates a new Robot class. I'm not sure of the best way to avoid that or even if I should. That seems like the logical place to create a new robot instance, particularly because a robot should not be created if the name is invalid and the RobotFactory is the best place to know if that occurred.

Theoretically any name generator or name validator could be passed in to the factory, which is why I made the separate classes. I tried to make sure there was a small surface area to each class's interface.

I didn't want exceptions to stop up the program. I should be able to create a series of robots, some that may be invalid. An invalid robot should not stop the program from executing, hence the rescue. That's the part I'm the least certain of with my design.

### DIFF
--- a/robot_name/name_generator.rb
+++ b/robot_name/name_generator.rb
@@ -1,0 +1,10 @@
+class NameGenerator
+  def random_name
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
+
+    alpha = "#{generate_char.call}#{generate_char.call}"
+    numeric = "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+    alpha + numeric
+  end
+end
--- a/robot_name/robot_factory.rb
+++ b/robot_name/robot_factory.rb
@@ -28,13 +28,11 @@ class RobotFactory
   end
 
   def check_robot_name(name)
-    begin
-      validator.validate_name(name, registry)
-    rescue StandardError => e
-      report_creation_error(e, name)
-    else
-      create_new_robot_in_registry(name)
-    end
+    validator.validate_name(name, registry)
+  rescue StandardError => e
+    report_creation_error(e, name)
+  else
+    create_new_robot_in_registry(name)
   end
 
   def report_robot_creation(name)
--- a/robot_name/robot_factory.rb
+++ b/robot_name/robot_factory.rb
@@ -1,0 +1,47 @@
+class RobotFactory
+  attr_accessor :generator, :validator, :registry
+
+  def initialize(generator, validator, registry)
+    @generator = generator
+    @validator = validator
+    @registry = registry
+  end
+
+  def create_robot(options = {})
+    name = give_robot_name(options)
+    check_robot_name(name)
+  end
+
+  private
+
+  def create_new_robot_in_registry(name)
+    robot = Robot.new(name)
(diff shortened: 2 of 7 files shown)


robot_name/name_generator.rb
```diff
@@ -0,0 +1,10 @@
+class NameGenerator
+  def random_name
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
+
+    alpha = "#{generate_char.call}#{generate_char.call}"
+    numeric = "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+    alpha + numeric
```
djkotowski: If you use `<<` instead of `+` to concatenate, you get a slight efficiency boost, as you'll be modifying the leftmost string rather than creating a new one.

robot_name/name_generator.rb
```diff
@@ -0,0 +1,10 @@
+class NameGenerator
+  def random_name
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
```
djkotowski: By moving these two lambdas out to methods, you can simplify the calls below.  Also, since they don't change, you'll only be defining them once.

robot_name/robot_factory.rb
```diff
@@ -0,0 +1,45 @@
...
+    robot = Robot.new(name)
+    registry.add_name(robot.name)
+    report_robot_creation(robot.name)
+    robot
+  end
+
+  def give_robot_name(options)
+    name = options[:name] if options[:name]
+    name = generator.random_name unless options[:name]
+    name
```
djkotowski: You can simplify this method with guard clauses.

```ruby
return options[:name] if options[:name]
generator.random_name
```

robot_name/robot_factory.rb
```diff
@@ -0,0 +1,45 @@
...
+  private
+
+  def create_new_robot_in_registry(name)
+    robot = Robot.new(name)
+    registry.add_name(robot.name)
+    report_robot_creation(robot.name)
+    robot
+  end
+
+  def give_robot_name(options)
```
djkotowski: Naming on this method is a bit confusing.  It implies that you're giving a robot a name, rather than simply returning a name given a set of options.

robot_name/robot_factory.rb
```diff
@@ -0,0 +1,45 @@
...
+    name = generator.random_name unless options[:name]
+    name
+  end
+
+  def check_robot_name(name)
+    validator.validate_name(name, registry)
+  rescue StandardError => e
+    report_creation_error(e, name)
+  else
+    create_new_robot_in_registry(name)
```
djkotowski: Why are you creating a robot in `#check_robot_name`?

djkotowski:
Nice job!  The division of responsibility is a bit confusing, especially where some methods are named one thing yet do another.  Try to clear it up a bit and resubmit.

PR #753, opened 2015-06-29
Robot Name jnyman

jnyman:
Second attempt, cleaning up the code smells indicated in my first attempt.

The NameGenerator in particular is a bit more robust now. I feel that the RobotFactory has really been slimmed down and it feels a lot more cohesive, by streamlining actions into fewer methods, which in turn forced me to name those methods more intelligently.


djkotowski:
Looks good!

PR #758, opened 2015-06-29
Robot name

jklemen:


johnmcconnell:
nice job :+1:

robot_name/robot_name.rb
```diff
@@ -1,32 +1,72 @@
...
+class RobotRegistry
+  attr_reader :registry
+
+  def initialize
+    @registry = Set.new
+  end
+
+  def insert(robot)
+    cautious_insert(robot)
+  end
```
djkotowski: Why have `insert` and `cautious_insert`?

robot_name/robot_name.rb
```diff
@@ -1,32 +1,72 @@
...
+
+  def cautious_insert(robot)
+    unless valid_name?(robot)
+      raise NameCollisionError, 'That robot name is invalid!'
```
djkotowski: Is this really a collision error?

robot_name/robot_name.rb
```diff
@@ -1,32 +1,72 @@
...
+    end
+
+    if registered?(robot)
+      raise NameCollisionError, 'That robot name is already registered!'
+    end
+
```
djkotowski: Don't forget Ruby also lets you use this syntax:
```ruby
raise error if condition
```

robot_name/robot_name.rb
```diff
@@ -1,32 +1,72 @@
...
+    @registry.member?(robot.name)
+  end
+
+  def valid_name?(robot)
+    valid_name = /[[:alpha:]]{2}[[:digit:]]{3}/
+    robot.name =~ valid_name
+  end
+end
+
 class Robot
```
djkotowski: Move this to a separate file.  All files should be named the same as their classes, only using snake case.

djkotowski:
Nice job!  This is the first time I've seen `Set` used instead of an array.

I have a few comments throughout.  No need for a resubmission, though.

robot_name/robot_name.rb
jklemen: Nope, that should be a different type of error. Good call.

robot_name/robot_name.rb
jklemen: I wan't certain when I started writing it if I wanted a 'blind insert' that simply added to the registry without performing any checks. So I made it a wrapper but in the end forgot to change it.

jklemen:
Thanks!

PR #762, opened 2015-06-29
Triangle

lmenova:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,63 @@
...
+  def initialize(side1, side2, side3)
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+  end
+
+  def recite_facts
+    recite_triangle_type
+    angles = calculate_angles(side1, side2, side3)
+    puts 'The angles of this triangle are ' << angles.join(',')
```
zsyed1: You can use interpolation:
```
puts "The angles of this triangle are #{angles.join(',')}"
```

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,63 @@
...
+
+  def scalene?
+    !(equilateral? || isosceles?)
+  end
+
+  def calculate_angles(a, b, c)
+    [
+      radians_to_degrees(get_radians(a, b, c)),
+      radians_to_degrees(get_radians(b, c, a)),
+      radians_to_degrees(get_radians(c, b, a)),
```
zsyed1: Lines 40 - 42: Instead of repeating these three lines in their entirety, it makes sense to move the `radians_to_degrees` call down into the the function that computes the angle. Something like this:

```
def calculate_angles(a, b, c)
  [get_angle(a, b, c), get_angle(b, c, a), get_angle(c, b, a)]
end
```

and then

def get_angle(hyp, legA, legB)
  radians = Math.acos((legA**2 + legB**2 - hyp**2) / (2 * legA * legB))
  (radians * 180 / Math::PI).round
end
```

File: triangle_facts/triangle_facts.rb [hunk 3]
```diff
@@ -1,58 +1,63 @@
...
+
+  def isosceles?
+    [side1, side2, side3].uniq.length == 2
+  end
+
+  def scalene?
+    !(equilateral? || isosceles?)
+  end
+
+  def calculate_angles(a, b, c)
```
zsyed1: Since the sides of the triangle are member variables, there is no need to pass them into this function. Instead, just rename `def calculate_angles` to `def angles` and replace `(a, b, c)` with `(side1, side2, side3)`. Then you can erase line 13 entirely and merely calling `angles` will return the array you desire.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,63 @@
...
+    puts 'This triangle is also a right triangle!' if angles.include?(90)
+    puts ''
+  end
+
+  def recite_triangle_type
+    puts "This triangle is equilateral!" if equilateral?
+    puts "This triangle is isosceles! "\
+    "Also, that word is hard to type." if isosceles?
+    puts "This triangle is scalene and mathematically boring." if scalene?
+  end
```
zsyed1: Lines 19 - 24: Not sure if you need these in a separate method. You can put these conditional `puts` in the `recite_facts` method itself. Once you remove line 13 (see comments below about `angles`), you are within the 5-line limit!

zsyed1:
Nice work! I would recommend resubmitting only because it would be nice for you to see some of these style changes early on in Level-Up.

PR #765, opened 2015-06-30
Dino catalog

jklemen:


dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,227 @@
+require 'csv'
+require 'json'
+require 'set'
+
+class Dino
+  attr_accessor :attrs
+
+  def initialize(data)
+    keys = %w(name period continent diet weight walking description)
+    @attrs = Hash[*keys.zip([nil] * keys.count).flatten]
```
johnmcconnell: I find it extremely odd you are explicitly setting the keys to a nil value. I'm guessing you have a good reason.

johnmcconnell: Also you can just make this Hash[keys.zip([nil] * keys.count)]. No need to flatten and then splat

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,227 @@
...
+    @attrs.merge!(data)
+  end
+
+  def dino_facts
+    @attrs.each do |attribute, value|
+      fattr = "#{attribute.capitalize}:".ljust(20)
```
johnmcconnell: https://github.com/ahoward/fattr

johnmcconnell: https://github.com/johnmcconnell/attr_init

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,227 @@
...
+  def lte(field, value)
+    Dinodex.new(
+      @dinos.select do |dino|
+        next if (actual = dino.attrs[field]).nil?
+        compare(field, actual.downcase, value.downcase) <= 0
+      end,
+    )
+  end
+
+  def gte(field, value)
```
johnmcconnell: completely fine you did it this way but you only need to implement one of these methods and the rest will fall into place.

```ruby
def lt(v1, v2)
end
```
and you get all these for free
```
def gte(v1, v2)
   !lt(v1, v2)
end
def lte(v1, v2)
   gte(v2, v1)
end
def gt(v1, v2)
   gte(v1, v2) && !lt(v2, v1)
end
```

newfile.txt
```diff
@@ -0,0 +1 @@
+hello, world
```
johnmcconnell: Might want to remove this.

johnmcconnell:
:+1:

dino_catalog/dinodex.rb
jklemen: Here, fattr = "formatted attribute". But it appears my over-abbreviation might have been confusing, given your knowledge of the existence of these repos. I'm assuming you want me to change it.

dino_catalog/dinodex.rb
jklemen: Good point!

dino_catalog/dinodex.rb
jklemen: The reason for nils has to do with this line in the dino_catalog problem:

> "If there are facts missing, please don't print empty values, just skip that heading."

I needed a 'default values' hash, and decided to explicitly set the values to nil instead of some other random value for initialization purposes.

jklemen: Also didn't know that about Hash[...]. I thought the arg had to be of the form

> [key, value, key, value,...]

But

> [[key, value], [key, value], ...]

Also being a possibility is good to know.

dino_catalog/dinodex.rb
johnmcconnell: Nah don't worry about it. Just giving you some stuff to look at.

dino_catalog/dinodex.rb
djkotowski: By default, `hash[non_existent_key] == nil`.  No need to initialize it as such.  Just skip over missing headings.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,227 @@
+require 'csv'
+require 'json'
+require 'set'
+
+class Dino
+  attr_accessor :attrs
```
djkotowski: Wouldn't it make more sense to include each attribute separately?  Take advantage of using a PORO for Dino instead of a hash.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,227 @@
...
+    @attrs.each do |attribute, value|
+      fattr = "#{attribute.capitalize}:".ljust(20)
+      fval = value.nil? ? '---' : value
+      puts "#{fattr}#{fval}"
+    end
+    puts
+  end
+end
+
+class Dinodex
```
djkotowski: Please separate each class into their own file.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,227 @@
...
+    if val1.nil? || val2.nil?
+      val1.__id__ <=> val2.__id__
+    elsif field == 'weight'
+      val1.to_i <=> val2.to_i
+    else
+      val1 <=> val2
+    end
+  end
+
+  def lt(field, value)
```
djkotowski: Avoid acronyms.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,227 @@
...
+    Dinodex.new(
+      @dinos.select do |dino|
+        next if (actual = dino.attrs[field]).nil?
+        compare(field, actual.downcase, value.downcase) < 0
+      end,
```
djkotowski: Unnecessary comma.

dino_catalog/dinodex.rb
djkotowski: Also, by putting these checks as predicates into Dino, you'll be able to simplify their implementation in Dinodex by just calling the predicate methods.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,227 @@
...
+    d_attrs['name'] = d_attrs['name'] || d_attrs['genus']
+
+    new_diet = (d_attrs['carnivore'] ? 'Carnivore' : "Non-Carnivore")
+    d_attrs['diet'] = d_attrs['diet'] || new_diet
+
+    filter(d_attrs)
+  end
+
+  def parse(dinofile)
+    dinos = CSV.parse(open(dinofile).read)
```
djkotowski: By providing a block to `#open`, you can ensure that the IO closes once the file is read.

```ruby
dinos = CSV.parse(open(dinofile, &:read))
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,227 @@
...
+      val1.__id__ <=> val2.__id__
+    elsif field == 'weight'
+      val1.to_i <=> val2.to_i
+    else
+      val1 <=> val2
+    end
+  end
+
+  def lt(field, value)
+    Dinodex.new(
```
djkotowski: Returning a new Dinodex each time to support chaining without being destructive.  Nice.

djkotowski:
Nice first iteration.  Make sure to separate your classes into different files.  Also, avoid hard-to-read acronyms.  They may be shorter to write, but they're harder to read.  Your code should _be_ your documentation.

Think about where responsibilities should lie.  Should your Dinodex be responsible for comparing Dinos, or should that responsibility lie with Dino?

Some sort of UI would be useful too.  Don't go overboard, though.  A CLI works fine for this project.

Please make changes where needed and resubmit when ready.  Let me know if you have any questions.

dino_catalog/dinodex.rb
jklemen: I considered it, and you'll see that I implemented it that way in an earlier version. I ended up going with the hash as my design choice.

dino_catalog/dinodex.rb
jklemen: Right, right. Keep forgetting about that.

dino_catalog/dinodex.rb
jklemen: That's my fault, I thought the 'lt, gt, eq' etc syntax was fairly universal.

dino_catalog/dinodex.rb
jklemen: Rubocop says otherwise?

dino_catalog/dinodex.rb
jklemen: Thanks (:

dino_catalog/dinodex.rb
jklemen: Cool! I assumed the immediate call to read flushed the contents of the file and closed it. Will change.

PR #766, opened 2015-06-30
Arrowhead

rcaileff:
solution for LevelUp exercise Arrowhead

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -15,19 +15,17 @@ class Arrowhead
     },
   }
 
-  # FIXME: I don't have time to deal with this.
   def self.classify(region, shape)
-    if CLASSIFICATIONS.include? region
-      shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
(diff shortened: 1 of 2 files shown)


arrowhead/arrowhead.rb
```diff
@@ -15,19 +15,17 @@ class Arrowhead
...
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
+    shapes = get_if_included(:region, region, CLASSIFICATIONS)
+    arrowhead = get_if_included(:shape, shape, shapes)
+    "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
+  end
+
+  def self.get_if_included(category, key, collection)
```
djkotowski: This does almost exactly what `Hash#fetch` does.

arrowhead/arrowhead.rb
```diff
@@ -15,19 +15,17 @@ class Arrowhead
...
+    unless collection.include? key
+      raise "Unknown #{category}.  Please provide a valid #{category}."
     end
+    collection[key]
```
djkotowski: A shorter solution would be to use `Hash#fetch` and rescue the error.

djkotowski:
Nice job!  You're done with this exercise.

PR #768, opened 2015-06-30
Robot

lmenova:


robot_name/robot_name.rb
```diff
@@ -1,32 +1,56 @@
...
+  @registry = []
+  DUP_NAME_ERR_MSG = "Robot name already exists in Robot Registry"
+
+  def self.add_name_to_registry(robot_name)
+    @robot_name = robot_name
+    raise NameCollisionError, DUP_NAME_ERR_MSG if @registry.include?(robot_name)
+    @registry << robot_name
+    puts "Added #{robot_name} to registry: #{@registry}"
+    rescue NameCollisionError => e
+      puts "Error: #{robot_name}.  #{e}"
```
djkotowski: The `rescue` statement should be indented at the same level as `def self.add_name_to_registry`.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,56 @@
...
   def initialize(args = {})
-    @@registry ||= []
-    @name_generator = args[:name_generator]
-
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+    @name = args[:name_generator] || generate_name
```
djkotowski: By altering the code to provide the name, not a generator, to the class, the key `:name_generator` is no longer accurate.

Also, since the generator is the only thing being provided, you don't need to take in an argument hash.  Just have a single `name` parameter.  Argument hashes are primarily used when you have multiple parameters you want to accept, with one or more of those parameters being optional.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,56 @@
...
+    RobotRegistry.add_name_to_registry(name) if name_valid?
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
+  def generate_name
+    @name = ('A'..'Z').to_a.sample(2).join + rand(0..999).to_s.rjust(3, "0")
```
djkotowski: Interesting way to handle the digits.  I recommend moving both parts of the name generator to separate methods to clarify what is going on.

Also, use `<<` instead of `+` when the leftmost string can be modified. `<<` appends to the original string object, while `+` creates a new string entirely.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,56 @@
...
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def name_valid?
+    return true if name =~ /[[:alpha:]]{2}[[:digit:]]{3}/
+    raise InvalidNameError, INVALID_NAME_MSG
+    rescue InvalidNameError => e
+      puts "Error: #{name}. #{e}"
```
djkotowski: Why raise an error only to rescue it in the next line?

robot_name/robot_name.rb
djkotowski: Also, don't rescue an error you're throwing in the same method.  Just break the execution of the method when the condition is met.

djkotowski:
Good first iteration.  Take a look at the comments and refactor where needed.

PR #773, opened 2015-07-02
Robot

lmenova:


robot_name/robot_name.rb
```diff
@@ -1,32 +1,66 @@
-class NameCollisionError < RuntimeError; end
+class RegistryError < StandardError; end
+
+class RobotRegistry
+  attr_accessor :robot_name
+
+  @registry = []
+  DUP_NAME_ERR_MSG = "Robot name already exists in Robot Registry"
+
+  def self.add_name_to_registry(robot_name)
```
djkotowski: `add_name` would be fine for a method name, since this is the `RobotRegistry` class.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,66 @@
...
+    @robot_name = robot_name
```
djkotowski: Unnecessary instance variable assignment.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,66 @@
...
+  end
+end
 
 class Robot
   attr_accessor :name
+  INVALID_NAME_MSG = "The name is invalid"
 
-  @@registry
+  def initialize(args = nil)
+    args ? @name = args.call : @name = generate_name
```
djkotowski: Don't use two assignments in a ternary.

If you choose to use a ternary operator, avoid redundancy with:

```ruby
@name = args ? args.call : generate_name
```

Also, please pick a clearer name for `args`.

djkotowski:
Nice work.  A few comments throughout, but no need for a resubmission.

PR #774, opened 2015-07-04
Dino Catalog jnyman

jnyman:
For this application, I ended up using Hirb, which I found through some Rails discussions where it was mentioned that Hirb can do nice things for console output. For me, this gem helped format the table information from the CSV files in a nice way.

My goal was to keep certain functionality as separate as possible. For example, the Rolodex and the Importer could technically be used with any sort of business entity. It just happens that with DinoDex, that the entity is a Dinosaur.

There are some things I could do to make the application more robust. Rethinking some aspects of the CLI interface would be front and center. Also the interface is not as case-insensitive as I would like. I'm willing to work on these things, as well as the extra credit, but this exercise is already taking way too long (on my part) so I want to make sure to get something out there and get some comments on it.


dino_catalog/USAGE.md
```diff
@@ -0,0 +1,65 @@
...
+
+    DinoDex <Query Mode> $ "Biped"
+
+*Grab all the dinosaurs that were carnivores (fish and insects count).*
+
+    DinoDex <Query Mode> $ diet=Carnivore
+
+    DinoDex <Query Mode> $ "Carnivore"
+
+???????????????????
```
jmtai: ?

dino_catalog/USAGE.md
```diff
@@ -0,0 +1,65 @@
+Running the application:
```
jmtai: :+1:  for the documentation.

dino_catalog/converter.rb
```diff
@@ -0,0 +1,31 @@
+class Converter
+  def self.dino_data_convert
+    carnivore_convert
```
jmtai: Since you're only calling `carnivore_convert` from this method, it isn't really necessary. No need to add extra indirection just for the sake of renaming a method.

dino_catalog/converter.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def self.carnivore_convert
+    lambda do |value, field|
+      field[:header].to_s == 'diet' ? diet_type(value) || value : value
```
jmtai: Using the ternary operator in combination with logical operators in the return expressions makes it a bit hard to parse at first glance. Wrapping the expression in parens will make it easier to see what is going on:
```ruby
field[:header].to_s == 'diet' ? (diet_type(value) || value) : value
```
However, is the `||` here really necessary? Should the `diet_type` method ever return a nil value?

dino_catalog/converter.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+  end
+
+  def self.diet_type(diet_value)
+    diet_value = diet_value.to_s.downcase
+    diet_value == 'yes' ? 'Carnivore' : '' if diet_value =~ /yes|no/
```
jmtai: This one is even harder to parse. Mixing the ternary operator along with an if expression means you're cramming too much logic into a single line.

Also, `diet_value` is initially checked to see if it equals `yes` but if not, the regex in the if expression checks to see if `diet_value` matches `yes` or `no`. There's no need to check for `yes` again.

dino_catalog/converter.rb
jnyman: Excellent suggestion on the parentheses. I agree that makes things very clear. Also, good question on the use of the `||`. In looking back at that, I'm not entirely sure why I did that, which I suppose in itself doesn't speak well of the code. I will make a point of going back and checking that  code today, removing the `||` part. To answer your question, no -- `diet_type` should never return a nil value.

dino_catalog/converter.rb
jmtai: If a dinosaur is not a Carnivore, should the diet be blank? If it doesn't eat meat, then it only eats plants, which would make it an Herbivore.

jnyman: Hmm. Yeah, I see your point. Just thinking out loud here: I'm checking the header value (in `carnivore_convert`). If that value is "diet" I call `diet_type` with the value in that field. If the value in that field is "yes", I change the value of that field to "Carnivore" instead. Otherwise, I simply leave it blank. So -- yeah -- that last `if` check seems useless.

I believe what I did -- and I'm going to double check -- is that I refactored this area quite a bit. I think in doing so, I didn't take well enough account of how I was distributing the logic.

jnyman: Although, that being said, technically insectivores and piscivores are considered carnivores, just with a limited diet. :) But from a data standpoint, just calling all of them "Carnivore" would not be accurate if you wanted to query on specifics, particularly since one data table provided did make those distinctions.

This is a perfect example of where a customer discussion would come in and discuss the implementation, particularly with different data sets.

dino_catalog/converter.rb
```diff
@@ -0,0 +1,31 @@
...
+  def self.header_convert
+    lambda do |header|
+      header.downcase!
+      rename_columns(header)
+      header.to_sym
+    end
+  end
+
+  def self.rename_columns(header)
+    header.gsub!('genus', 'name')
```
jmtai: `gsub!` using regular expressions, which are not super efficient. Here you've got 4 `gsub!` calls per header, even though you'll only ever have a single match.

Since you're doing a simple substitution something like
```ruby
return 'name' if header == 'genus'
```
would be more efficient, since it will bypass remaining checks as soon as the first match is found.

jmtai: Also, since you've got 4 different values to check, you could simply use a case statement, although @djkotowski will probably tell you case statements are evil. :smile:

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env ruby
+
+require 'hirb'
+
+require_relative 'importer'
+require_relative 'converter'
+require_relative 'rolodex'
+require_relative 'dinosaur'
+
+class DinoDex
```
jmtai: DinoDex is doing quite a bit here. It's handling command line input, searching,  checking dinosaur size, and printing output. Ideally, we want our classes (and methods) to be single responsibility. Having too many responsibilities makes comprehending and modifying a class more difficult.

dino_catalog/converter.rb
jnyman: Excellent point on the efficiency consideration. I've seen that the notion of a guard clause is looked upon favorably and it seems if I followed that approach, as you suggest, that would be a better implementation. Thanks for the suggestion; this will be a good one to help me build up my intuition on how to structure code.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,124 @@
...
+
+  def initialize(config = {})
+    extend Hirb::Console
+    setup_options(config)
+    start_dinodex
+  end
+
+  def big(data_set)
+    data_set.select do |_name, dinosaur|
+      dinosaur.weight.to_i > 2000
```
jmtai: Here the DinoDex class is checking specific dinosaur values. Perhaps this should be done in the Dinosaur class, instead?

What if you could write something like this:
```ruby
data_set.select { |_name, dinosaur| dinosaur.big? }
```
Do you think that's cleaner?

jmtai: Also, definitely try to avoid having magic values in your code. If you have a constant, declare it as a constant and give it a meaningful name. Maybe something like
```ruby
MAX_SMALL_DINO_WEIGHT = 2000
```

dino_catalog/dinodex.rb
jnyman: Good point. In my case, as I think on it, perhaps the immediate thing that stands out is the command line aspects coupled with the display of those commands. Taking this to one level I could see having DinoDexCommand and DinoDexDisplay. (<-- Not the names I would use! Just trying to be illustrative.)

In this case, though, now someone reading my code who wants to know how a command works has to look at two classes. I could combine the command execution and the command output display into one class -- but then that's one class doing two things. (Technically. Some would consider execution of a command and its output to be a single responsibility, I suppose.)

This is the part I struggle with the most, I'll be the first to admit. What is a "responsibility"? What's too granular? What's not granular enough? It's something I still struggle with.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,124 @@
...
+      path: config[:path] || 'data/*.csv',
+      headers: true,
+      converters: config[:converters] || Converter.dino_data_convert,
+      header_converters: config[:header_converters] || Converter.header_convert,
+    }
+  end
+
+  def start_dinodex
+    create_file_path
+    @data = Importer.new(options, Dinosaur)
```
jmtai: Do you ever expect `DinoDex` to import anything other than `Dinosaur` objects? Abstraction is nice, but premature abstraction can lead to technical debt. If you don't foresee a need for the abstraction in the near future, it's best to not put it in.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,124 @@
...
+    loop do
+      print 'DinoDex $ '
+      command = gets.chomp
+      exit if command == 'quit'
+      execute_command(command)
+    end
+  end
+
+  def execute_command(command)
+    list if command == 'list'
```
jmtai: Since your commands have the same names as your methods, you can do something like this instead of individually processing each item:
```ruby
send command.to_sym
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,124 @@
...
+    details if command == 'details'
+    query if command == 'query'
+    help if command == 'help'
+  end
+
+  def help
+    puts
```
jmtai: For long, multiline strings, take a look at Ruby HEREDOCs.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,124 @@
...
+    item = @rolodex.show_details(@data.data_set, *list)
+    table(item.values, fields: Dinosaur.headers)
+  end
+
+  def query
+    conditions = ''
+    loop do
+      print 'DinoDex <Query Mode> $ '
+      conditions = gets.chomp
+      break if conditions == 'exit'
```
jmtai: Minor nit: Nothing in the onscreen prompt or help indicates that `exit` is the command to exit query mode.

dino_catalog/dinodex.rb
jnyman: I don't expect DinoDex to import anything other than Dinosaur, but I would expect Importer to. So my thinking here was that the Importer could (technically) be replaced entirely and DinoDex would still continue to function.

This being said, I wrote this just coming off of reading "Practical Object-Oriented Design in Ruby" which drills in that you should inject dependencies, including classes. I was also coming off reading all the debate about classes that end in "er" and how those should similarly not assume that what they deal with now is what they will always deal with.

Ironically, they gave the same reason you stated: technical debt. Meaning, you make an importer too specific to what it's importing and -- even if you never import another type of thing other than the Dinosaur -- you have still made one thing (Importer) know too much about another (Dinosaur, in this case).

So ... I don't know. These ideas are all still swirling around in my head. I appreciate the comment because this is an area that I am really trying to get a handle on.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,124 @@
...
+    details if command == 'details'
+    query if command == 'query'
+    help if command == 'help'
+  end
+
+  def help
+    puts
+    puts "The following commands are available:\n\n"
+    puts "list:\t\tDisplay a list of all data."
+    puts "details:\tDisplasy a list that you can select data from."
```
jmtai: Typo: `Displasy` :smile:

jnyman: What -- you never heard of the Displasy action ?!?

Kidding, of course. You are right. That is a ridiculous thing for me to have missed.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,124 @@
...
+      print 'DinoDex <Query Mode> $ '
+      conditions = gets.chomp
+      break if conditions == 'exit'
+      present_results(conditions)
+    end
+  end
+
+  def present_results(conditions)
+    results = @rolodex.query(@data.data_set, conditions)
+    if results.empty?
```
jmtai: You can use the guard clause format here to eliminate the if-else-end structure:
```ruby
return puts 'No results found with your criteria.' if results.empty?
table(results.values, fields: Dinosaur.headers)
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,124 @@
...
+      puts 'No results found with your criteria.'
+    else
+      table(results.values, fields: Dinosaur.headers)
+    end
+  end
+
+  def create_file_path
+    return unless File.extname(options[:path]).empty?
```
jmtai: Nice use of the guard clause.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,124 @@
...
+    @options[:path] += '/' unless ends_with_slash?(options[:path])
+    @options[:path] += '*.csv' if no_extension?(options[:path])
+  end
+
+  def no_extension?(path)
```
jmtai: Nice use of predicate methods.

dino_catalog/dinodex.rb
jnyman: I do think that what you suggest is clearer. In fact, I think I intended that application. Currently Rolodex does not assume the type of entity it is dealing with and so that is passed into it. Here I'm passing in self as the second parameter, which meant the "big" and "small" methods were on DinoDex.

In reality, I could just pass in Dinosaur.

So why didn't I? Here is probably a good illustration of what you mentioned in another comment. I think DinoDex is doing a bit too much. So I was in a position where it was being press-ganged into handling various aspects, rather than farming those aspects out appropriately. That said, I'm a little surprised I made this particular decision since `big` and `small` clearly apply to the Dinosaur.

I think part of me was concerned that `big` and `small` don't refer to an instance of the Dinosaur but rather as a query qualifier that is applied to a set of Dinosaurs. A Dinosaur could reasonably expected to have `weight`, let's say. But what is "big" and "small" may change based on what the user wants the implementation to do.

So I think I was planning too far ahead: I was assuming that big and small may need to be modified as parameters on the DinoDex (perhaps passed in as a hash). That way the notion of "big" and "small" could be modified without having to modify a class.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,25 @@
+class Dinosaur
+  %i(
+    name
+    period
+    continent
+    diet
+    weight
+    movement
+    description
+  ).each { |s| attr_accessor s }
```
jmtai: Clever, but definitely not the preferred format. Just explicitly declare the accessors. You can combine multiple arguments on a single line:
```ruby
attr_accessor :name, :period, :continent, etc.
```

jnyman: The reason I didn't do this is that the line was too long for Rubocop. And others here have said that having multiple attr_accessor lines is not good. A few of my cohorts got dinged on that. So the implementation I did here was what some Ruby folks on StackOverflow mentioned to do in cases where Rubocop doesn't like your long lines.

Not suggesting that means you are incorrect, mind you, about this not being a preferred format. It's just that this seemed to be something a few folks here and there were suggesting. Personally, I don't know why having multiple attr_accessor rows is a bad thing (assuming your one attr_accessor line goes past the Rubocop limit. That would actually be my preferred implementation.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  HEADERS = %w(Name Period Continent Diet Weight Movement Description)
+
+  def self.headers
+    columns = []
+    HEADERS.each { |header| columns << header.downcase }
+    columns
```
jmtai: Ruby's `each_with_object` method is designed to make this pattern more compact:
```ruby
HEADERS.each_with_object([]) { |header, columns| columns << header.downcase }
```

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def initialize(data)
+    data.each do |header, value|
+      method("#{header}=").call(value)
```
jmtai: You could also do this:
```ruby
instance_variable_set("@#{header}", value)
```
It might be clearer what's going on with that form, but your way is fine, too.

dino_catalog/importer.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    create_data_set(CSV.read(file, options))
+  end
+
+  def create_data_set(file_contents)
+    file_contents.by_row.each do |row|
+      data = {}
+      row.headers.each do |header|
+        data[header] = row[header]
+      end
```
jmtai: You can use `each_with_object` here, as well.

dino_catalog/importer.rb
```diff
@@ -0,0 +1,43 @@
...
+
+class Importer
+  attr_reader :entity, :options
+  attr_accessor :data_set
+
+  def initialize(options, entity)
+    self.data_set = {}
+    @entity = entity
+    @options = options
+    get_data_files(options[:path])
```
jmtai: I generally try to avoid using a generic word like `get` when naming methods, since it usually doesn't fully describe what is actually happening. `get_data_files` is actually reading AND parsing the data files. Perhaps something like `load_and_parse_files` would be more appropriate?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,124 @@
...
+
+  def big(data_set)
+    data_set.select do |_name, dinosaur|
+      dinosaur.weight.to_i > 2000
+    end
+  end
+
+  def small(data_set)
+    data_set.select do |_name, dinosaur|
+      dinosaur.weight.to_i <= 2000
```
jmtai: What happens if weight is nil?

dino_catalog/importer.rb
jnyman: Good point. Agreed. In fact, as I look at it, my `get_data_files` could probably be called `load_files` and then my `load_file` could be called `parse_file`. That way the methods are kept distinct in terms of the SRP concept. This seems accurate to me since my `load_file` method is where the CSV read action takes place, which is sort of "parsing" the file to read it into a format. Then again, my `create_data_set` is sort of doing that, depending on how you define "parsing."

So I could also see perhaps moving the logic currently in `load_file` up to what I would call `load_data_files` (formerly: `get_data_files`).

dino_catalog/dinodex.rb
jnyman: Good question. Currently it would print out the item. For example, with the data set as it is, the following exists:

    "Afrovenator"=>#<Dinosaur:0x007ff9b2044148 @name="Afrovenator",
    @period="Jurassic", @diet="Carnivore", @weight=nil,@movement="Biped">

Notice the weight is nil. In this case, doing the command "small" from the query interface would return this dinosaur. The command "big" would not.

I should have some handling in there for this. As I think on it, what I could do is map any nil values to empty strings. That way I could make checks for empty? or something like that.

dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,117 @@
...
+    return false unless condition.to_s.match(/[^'"].*['"]/)
+    search_term = []
+    terms = condition.to_s.delete('"').delete("'")
+    search_term << terms.strip
+    add_to_conditions(search_term, index)
+    true
+  end
+
+  def qualifier_condition(condition, index)
+    return unless condition.match(/big|small/)
```
jmtai: What happens if I type something like `"Jurassic",  big` (i.e. there is whitespace between the comma and the `big`)?

dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,117 @@
...
+  def check_conditions(condition, data)
+    data = process_condition_array(condition, data) if condition.class == Array
+    data = process_condition_hash(condition, data) if condition.class == Hash
+    data
+  end
+
+  def process_condition_array(condition, data)
+    condition.each do |term|
+      data = @entity.send term, data if term.class == Symbol
+      data = search_data_text(data, term) unless term.class == Symbol
```
jmtai: I'd argue here that using a combination of `if` and `unless` is less readable than an `if-else` block.

dino_catalog/rolodex.rb
jnyman: Yep, that's a bug. It will fail because the method that is being called is not " big" -- with the space!

At bare minimum, I need to do this:

    qualifiers << condition.strip.to_sym

dino_catalog/rolodex.rb
jnyman: I agree. The reason I didn't is Rubocop then says the method is too long.

dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,117 @@
...
+  def parse_conditions(conditions)
+    conditions.split(',').each_with_index do |condition, index|
+      valid_text = text_condition(condition, index)
+      valid_qualifier = qualifier_condition(condition, index) unless valid_text
+      column_condition(condition, index) unless valid_text || valid_qualifier
+    end
+    @options
+  end
+
+  def text_condition(condition, index)
```
jmtai: These are a bit confusing because you're returning `true/false` but the method itself has side effects (i.e. it's modifying the options variable via `add_to_conditions`).

Consider refactoring these into a predicate method and a modifier method to make the methods single responsibility.

dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,117 @@
...
+    true
+  end
+
+  def add_to_conditions(condition, index)
+    @options[index] = condition
+  end
+
+  def search_data_text(data_set, term)
+    data_set.select do |key, value|
+      result = (key == term)
```
jmtai: Since `result` is only used by the following method call, you can just pass `key == term` directly instead of creating a temporary variable. Either way, `result` isn't super descriptive.

dino_catalog/rolodex.rb
jnyman: Good points. I originally did have things a little different than here, and a bit more broken apart, but Rubocop was saying my class was too long at that point so I was trying to condense things down. That said, I didn't do a predicate method approach with those and that's a great suggestion.

The more I look at this, the more I realize I was probably not being wise when using Hirb. Because as I look at my choices, I realize a lot of them are predicated upon how I felt I had to pass the data around such that it was always queryable (via the class methods) so that it was always displayable by Hirb. Mind, I'm not saying Hirb was the cause of my problems since a lot of the design choices are mine regardless. But I do feel I was a bit too tied to an implementation in some cases and that impacted how I thought about how my data is traveling through the application.

dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,117 @@
...
+      search_by_header_column(value, term, result)
+    end
+  end
+
+  def search_by_header_column(value, term, result)
+    value.instance_variables.each do |header_variable|
+      values = value.instance_variable_get(header_variable)
+      result = values.to_s.include?(term) unless result
```
jmtai: Variable shadowing is confusing. It's hard to tell what the correct operation should be when local variable names are the same as passed in parameter names.

dino_catalog/rolodex.rb
jnyman: Interesting. I can see your point. That said, as a code reader, if I see a method called "search_data_text" and I see that what gets returned is placed in a `result` variable, that would seem to be descriptive enough given the context, no?

Even given that, though, I see your point. It's interesting because this:

    search_by_header_column(value, term, (key == term)

Is actually harder for me to read. I don't know that "key == term" is a result, at least by looking at it necessarily if I hadn't wrote it. But if I argue that way, then why is my approach different:

    result = (key == term)
    search_by_header_column(value, term, result)

Don't know. :) I guess sometimes I like when a particular construct -- like `key == term` -- is stored in a variable just so I have no ambiguity about what the person meant.

jnyman: And, on reflection, yeah I agree entirely. I'm really storing a truth value there. So, yeah, `result` is really asking "was there a result? True or false?" and in that case, it's not really the result of the search but rather a result of whether the search found anything.

So, yes: entirely agreed. Not descriptive and in fact potentially confusing.

dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,117 @@
...
+    end
+    result
+  end
+
+  def create_header_methods
+    @headers.each do |header|
+      self.class.class_eval do
+        define_method "query_#{header.downcase}" do |data_set, *data|
+          query_by_header(header, data_set, data)
```
jmtai: `data_set` and `data` are pretty similar sounding. It makes it a little difficult to figure out what the code is trying to do.

dino_catalog/converter.rb
djkotowski: Definitely avoid ternary operators in cases like these.  Ternary operators should _only_ be used for very simple assignments, and often make code more confusing.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,124 @@
...
+  def start_dinodex
+    create_file_path
+    @data = Importer.new(options, Dinosaur)
+    @rolodex = Rolodex.new(Dinosaur::HEADERS, self)
+    display_title
+    execute_command_loop
+  end
+
+  def display_title
+    puts '-' * 30 << "\nDINODEX\n" << '-' * 30
```
djkotowski: I recommend wrapping `'-' * 30` in parentheses for better readability.

dino_catalog/dinodex.rb
djkotowski: Just be sure to provide a way of limiting what command can be sent.

dino_catalog/dinodex.rb
djkotowski: I think I knew someone who had displasy.  Didn't end well.

dino_catalog/dinodex.rb
djkotowski: :+1:

dino_catalog/dinosaur.rb
djkotowski: I have no issue with multiple attr_accessor lines.  The only issue is Rubocop's preferred styling.
```ruby
# My preferred spacing
attr_accessor :attr1, :attr2, :attr3,
              :attr4, :attr5

# Rubocop's preferred spacing
attr_accessor :attr1, :attr2, :attr3
  :attr4, :attr5
```

dino_catalog/dinosaur.rb
djkotowski: Actually, map would be the best option here.

```ruby
HEADERS.map { |header| header.downcase }
# which is simplified to
HEADERS.map(&:downcase)
```

dino_catalog/dinosaur.rb
jnyman: Yeah, I'm with you on your preferred spacing. There's a lot of Ruby code that I think looks nice when you line things up like this, but Rubocop definitely has issues with a lot of that, at least from what I've seen. (Lest it sound like I'm bad mouthing Rubocop, it has helped me quite a bit. I just wish it wasn't so picky about what I would consider visual styling, as opposed to code styling.)

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,25 @@
...
+    continent
+    diet
+    weight
+    movement
+    description
+  ).each { |s| attr_accessor s }
+
+  HEADERS = %w(Name Period Continent Diet Weight Movement Description)
+
+  def self.headers
```
djkotowski: Given that you're altering `HEADERS`, I would name this something to indicate the difference, especially since anyone looking at this will see that `Dinosaur.headers` and `Dinosaur::HEADERS` both exist.

dino_catalog/dinosaur.rb
jnyman: And just when I think I'm building up my Ruby intuition to better code constructs .... there's always something else. That last simplification is incredibly concise; almost functional in appearance. I really like that.

dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,117 @@
...
+        end
+      end
+    end
+  end
+
+  def query_by_header(header, data_set, data)
+    data_set.select do |_key, value|
+      valid_data = false
+      values = value.method("#{header.downcase}").call
```
jmtai: `values = value...` makes it sound like this should contain multiple items, but it's actually just returning a string.

dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,117 @@
...
+      data.each { |item| valid_data = values.to_s.include?(item) }
```
jmtai: Instead of creating an external `valid_data` flag, you can use the `any?` method:
```ruby
data.any? { |item| values.to_s.include?(item) }
```

jmtai:
Good first pass! Take a look at the comments, refactor a bit and resubmit when ready. Feel free to stop by if you'd like to discuss any of the comments further.

dino_catalog/rolodex.rb
jnyman: Good point. Yeah, in looking at this, I was clearly thinking in terms of the loop -- i.e., ultimately multiple values are being returned. But you are correct -- at any given iteration there is only going to be a single string in `values`.

dino_catalog/dinosaur.rb
djkotowski: The one thing `method` provides that `instance_variable_set` doesn't is header validation.  If there is no `"#{header}="` method, it fails, whereas `instance_variable_set` doesn't care.

dino_catalog/importer.rb
djkotowski: Is this any different than `CSV::Row#to_hash`?

dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  def check_conditions(condition, data)
+    data = process_condition_array(condition, data) if condition.class == Array
+    data = process_condition_hash(condition, data) if condition.class == Hash
+    data
+  end
+
+  def process_condition_array(condition, data)
+    condition.each do |term|
+      data = @entity.send term, data if term.class == Symbol
```
djkotowski: Wrap args in parens.

dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,117 @@
...
+      data = search_data_text(data, term) unless term.class == Symbol
+    end
+    data
+  end
+
+  def process_condition_hash(condition, data)
+    condition.each do |key, value|
+      data = method("query_#{key}").call(data, value)
+    end
```
djkotowski: Look at using `#inject`/`#reduce` here.

dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,117 @@
...
+    terms = condition.to_s.delete('"').delete("'")
+    search_term << terms.strip
+    add_to_conditions(search_term, index)
+    true
+  end
+
+  def qualifier_condition(condition, index)
+    return unless condition.match(/big|small/)
+    qualifiers = []
+    qualifiers << condition.to_sym
```
djkotowski: Couldn't L65-66 just be
```ruby
qualifiers = [condition.strip.to_sym]
```

dino_catalog/rolodex.rb
djkotowski: Also see if you can mitigate the arrowhead pattern here.

djkotowski:
As Josh said, nice first iteration.  Look at the comments, refactor and resubmit when ready, and let us know if you have questions.

jnyman:
Cool. Will do. I'm contemplating changing the implementation quite a bit, though. I don't think I did myself any favors with the Hirb approach. I think it looks really, really nice when you run it -- but at the cost of a lot of complexity in the code, at least given my current knowledge.

I don't mind taking time to figure out how to get it to work with Hirb, but I'm extremely worried about the amount of time this exercise already took me. I'll take a first stab today at reimplementing it as it is and then, depending on the amount of tears, will figure out next steps.

Thanks for the useful feedback. This was all extremely helpful.

PR #778, opened 2015-07-07
Data science

jklemen:
Added the tests file. I realize this should have been done beforehand, but I overlooked that line in the exercise. I corrected this in the overlord exercise by writing the tests first, as instructed.

### DIFF
--- a/data_science/data_science.rb
+++ b/data_science/data_science.rb
@@ -1,45 +1,88 @@
 require 'json'
 
-json = JSON.parse(open('data.json').read).to_a
+class DataScience
+  attr_accessor :json, :cohorts, :sample
 
-# 1. Total sample size and number of conversions for each cohort.
-$n = json.count
-puts "Sample size: #{$n}"
-cohorts = json.partition{ |item|
-	item['cohort'] == 'A'
-}
+  def initialize(json_file)
+    json_data = open(json_file).read
+    @json = JSON.parse(json_data).to_a
+    @sample = @json.count
 
-def stats(cohort, name)
-	count = cohort.inject(0){ |sum, data| sum += data['result']}
-	puts "Cohort #{name}: #{count} / #{cohort.count}"
+    parts = json.partition { |item| item['cohort'] == 'A' }
+    @cohorts = Hash[*%w(A B).zip(parts).flatten(1)]
+  end
+
+  def total_sample_size
+    puts "Sample size: #{@sample}"
+  end
+
+  def cohort_conversions(cohort_name)
+    cohort = @cohorts[cohort_name]
+    cohort.inject(0) { |sum, data| sum + data['result'] }
+  end
+
+  def cohort_stats(cohort_name)
+    cconv = cohort_conversions(cohort_name)
+    csample = @cohorts[cohort_name].count
+
+    [cconv, csample]
+  end
+
+  # 1. Total sample size and number of conversions for each cohort.
+  def report_sample_sizes
+    @cohorts.keys.each do |cname|
+      cstats = cohort_stats(cname)
+
+      puts "Cohort #{cname}: #{cstats[0]} / #{cstats[1]}"
+    end
+  end
+
+  def calc_mean_error(cohort_name)
+    cstats = cohort_stats(cohort_name)
+
+    p = (cstats[0].to_f / cstats[1].to_f).round(5)
+    se = (2 * Math.sqrt(p * (1 - p) / @sample.to_f)).round(5)
+
+    [p, se]
+  end
+
+  # 2. Conversion rate (w/ error bars) for each cohort with a 95% confidence.
+  # Square root of (p * (1-p) / n)
+  def report_mean_error
+    @cohorts.keys.each do |cname|
+      p, se = calc_mean_error(cname)
+
+      puts "Conversion Rate #{cname} = #{p}"
+      puts "Standard Error #{cname}  = #{se}"
+    end
+  end
+
+  def calc_confidence(cohort_name)
+    p, se = calc_mean_error(cohort_name)
+
+    low = (p - se).round(3)
+    high = (p + se).round(3)
+
+    [low, p, high]
+  end
+
+  # 3. Confidence level that the current leader is in fact better than random.
+  # (done in confidence)
+  def report_confidence
+    @cohorts.keys.each do |cname|
+      low, p, high = calc_confidence(cname)
+
+      puts "95% Confidence Interval for Cohort #{cname}: "
+      puts "|(#{low})--- (#{p}) ---(#{high})|"
+    end
+  end
+
+  def report_all
+    total_sample_size
+    report_sample_sizes
+    report_mean_error
+    report_confidence
+  end
 end
 
-puts
-stats(cohorts[0], 'A')
-stats(cohorts[1], 'B')
-puts
-
-# 2. Conversion rate (including error bars) for each cohort with a 95% confidence.
-# Square root of (p * (1-p) / n)
-
-def confidence(cohort, name)
-	count = cohort.inject(0){ |sum, data| sum += data['result']}
-
-	p = (count.to_f / cohort.count.to_f).round(5)
-	se = (2 * Math.sqrt(p * (1 - p) / $n.to_f)).round(5)
-	puts "Conversion Rate = #{p}"
-	puts "Standard Error = #{se}"
-
-	low = (p - se).round(2)
-	high = (p + se).round(2)
-	puts "95% Confidence Interval for Cohort #{name}: "
-	puts "|(#{low})--- (#{p}) ---(#{high})|"
-end
-
-confidence(cohorts[0], 'A')
-puts
-confidence(cohorts[1], 'B')
-puts
-
-# 3. Confidence level that the current leader is in fact better than random.
-# (done in confidence)
+DataScience.new('data_export_2014_06_20_15_59_02.json').report_all
--- a/data_science/data_science.rb
+++ b/data_science/data_science.rb
@@ -1,0 +1,44 @@
+require 'json'
+
+json = JSON.parse(open('data.json').read).to_a
+
+# 1. Total sample size and number of conversions for each cohort.
+$n = json.count
+puts "Sample size: #{$n}"
+cohorts = json.partition{ |item|
+	item['cohort'] == 'A'
+}
+
+def stats(cohort, name)
+	count = cohort.inject(0){ |sum, data| sum += data['result']}
+	puts "Cohort #{name}: #{count} / #{cohort.count}"
+end
+
+puts
+stats(cohorts[0], 'A')
+stats(cohorts[1], 'B')
+puts
+
+# 2. Conversion rate (including error bars) for each cohort with a 95% confidence.
+# Square root of (p * (1-p) / n)
+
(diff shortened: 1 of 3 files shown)


data_science/data_science.rb
```diff
@@ -0,0 +1,81 @@
+require 'json'
+
+class DataScience
+  attr_accessor :cohorts, :sample_size
+
+  def initialize(json_file)
+    json_data = open(json_file).read
```
jmtai: While this form allows you to concisely read the data from a file, it doesn't allow you to actually close the file after you're done reading from it.  You can use the block form to ensure the file is closed after you're done reading it:
```ruby
open(json_file) { |file| file.read }
```

data_science/data_science.rb
```diff
@@ -0,0 +1,81 @@
...
+    json = JSON.parse(json_data).to_a
+    @sample_size = json.count
```
jmtai: Hmmm... this is the total sample size of *all* cohorts. Why is this needed?

data_science/data_science.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    parts = json.partition { |item| item['cohort'] == 'A' }
```
jmtai: What happens if the cohort label is lowercase?

data_science/data_science.rb
```diff
@@ -0,0 +1,81 @@
...
+    @cohorts = Hash[*%w(A B).zip(parts).flatten(1)]
```
jmtai: This is certainly concise, but would someone looking at this the first time be able to understand it in 5 seconds?
Do you think something like this is more readable?
```ruby
@cohorts = {
  "A" => parts[0],
  "B" => parts[1]
}
```

data_science/data_science.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def cohort_conversions(cohort_name)
```
jmtai: Any time you see yourself naming methods with and identical prefix, it can be a hint that maybe these belong in a separate class. In this case, there are several things that are specific to a single cohort (e.g. sample size, conversions, error bars, etc) and some things that aren't (e.g. chi-square test, printing results to the console).

The principle of single responsibility tells us that we'd ideally like our classes to do just one thing (or a group of closely related things), so perhaps it would be better to create a `Cohort` class with a cohort-specific variables and methods?

data_science/data_science.rb
```diff
@@ -0,0 +1,81 @@
...
+    cohort = @cohorts[cohort_name]
+    cohort.inject(0) { |sum, data| sum + data['result'] }
```
jmtai: Since you're storing the raw data in `@cohorts`, every time you want to get the number of conversions you have to iterate through the entire array to count the results. Do you ever expect this number to change? If not, wouldn't it be more efficient to simple do this once and store the result so the next time you access it, it's just a simply lookup?

data_science/data_science.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def cohort_stats(cohort_name)
+    conv = cohort_conversions(cohort_name)
+    sample = @cohorts[cohort_name].count
```
jmtai: Naming can be pretty subjective, but we should definitely try to use the most descriptive names possible for our variables. `conv` and `sample` are pretty terse and somewhat cryptic. What about something like `num_conversions` and `total_samples`?

data_science/data_science.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    [conv, sample]
+  end
+
+  # 1. Total sample size and number of conversions for each cohort.
+  def report_sample_sizes
+    @cohorts.keys.each do |name|
+      stats = cohort_stats(name)
+
+      puts "Cohort #{name}: #{stats[0]} / #{stats[1]}"
```
jmtai: Raw indexed arrays are another indication that maybe there's a better way to represent your data. Ruby is an object oriented language, so why not use that to our advantage? Do you think it would make your code more readable if you could do something like this?
```ruby
@cohorts.each { |cohort| puts "Cohort #{cohort.name}: #{cohort.num_conversions} / #{cohort.total_samples}"
```

data_science/data_science.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+  end
+
+  def calc_mean_error(cohort_name)
+    stats = cohort_stats(cohort_name)
+
+    p = (stats[0].to_f / stats[1].to_f)
```
jmtai: `p` is definitely not a great variable name. What you're calculating here is the conversion rate, so why not just call it that? Also, the Chi-square test calculates a p-value, so one could get a little confused by the use of the name `p` here.

data_science/data_science.rb
```diff
@@ -0,0 +1,81 @@
...
+    se = (2 * Math.sqrt(p * (1 - p) / @sample_size.to_f))
```
jmtai: You're calculating the error bars per cohort, but `@sample_size` has been defined to be the size of *all* items in the data set.

data_science/data_science.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      puts "Conversion Rate #{name} = #{p}"
+      puts "Standard Error #{name}  = #{se}"
+    end
+  end
+
+  def calc_confidence(cohort_name)
+    p, se = calc_mean_error(cohort_name)
+
+    [p - se, p, p + se].map { |num| num.round(3) }
```
jmtai: FYI, the calculate the confidence interval w/95% confidence, you actually need a scaling factor on the standard error value: http://onlinestatbook.com/2/estimation/mean.html

For normal distributions, this ends up being 1.96

data_science/data_science.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  # 3. Confidence level that the current leader is in fact better than random.
+  # (done in confidence)
+  def report_confidence
```
jmtai: This is the confidence interval for the conversion rate, but **not** the confidence level that the current leader is in fact better than random. For that you'll need to use the Chi-square test. You can use the ABAnalyzer gem mentioned in the problem description to do this for you: https://github.com/bmuller/abanalyzer

data_science/rspec/test_spec.rb
```diff
@@ -0,0 +1,54 @@
+require '../data_science.rb'
+
+describe DataScience do
+  before :all do
```
jmtai: `let` is usually preferred here because let blocks are lazily evaluated and also memoized so they are more efficient that `before` blocks.

data_science/rspec/test_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    data_file = '../data_export_2014_06_20_15_59_02.json'
+    @ds = DataScience.new(data_file)
+  end
+
+  it "should have the correct sample size" do
+    @ds.sample_size.should == 2892
```
jmtai: The `should` syntax is actually deprecated - you'll want to use the newer `expect` syntax for your tests.

data_science/rspec/test_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  it "should have two cohorts" do
+    @ds.cohorts.count.should == 2
+  end
+
+  it "should report the correct cohort sample sizes" do
+    a_stats = @ds.cohort_stats('A')
+    a_stats[1].should == 1349
```
jmtai: Just as we saw in the class implementation, using raw arrays to represent your data makes it less understandable. It'd be nice if the test read more like english - something like
```ruby
expect(cohort.sample_size).to eq(1349)
```

data_science/rspec/test_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    (a_stats[0] - 0.028).abs.should <= 0.01
+    (a_stats[1] - 0.035).abs.should <= 0.01
+    (a_stats[2] - 0.042).abs.should <= 0.01
+
+    b_stats = @ds.calc_confidence('B')
+    (b_stats[0] - 0.043).abs.should <= 0.01
+    (b_stats[1] - 0.051).abs.should <= 0.01
+    (b_stats[2] - 0.059).abs.should <= 0.01
+  end
+end
```
jmtai: Once you implement the Chi-square test, you'll need a test for that here, as well.

jmtai: You've got tests for the happy path (i.e. the data file contains no errors). What if the data set is corrupted, though?

data_science/data_science.rb
jklemen: It won't be? My implementation is domain-specific, given two cohorts with names 'A' and 'B'. Should I extend it to accept any general name?

jmtai:
Good start! Try refactoring the code a bit (and add in the missing functionality) and resubmit when you're ready.

data_science/data_science.rb
jklemen: Good point. Definite violation of the 5 second rule.

data_science/data_science.rb
jklemen: I think you're onto something.

data_science/data_science.rb
jklemen: I will be refactoring everything into a 'Cohort' class that will address this problem

data_science/data_science.rb
jklemen: It's difficult to balance descriptive snake-cased names with the 80-character line length rule, but you're right.

data_science/data_science.rb
jklemen: Oh, so I divide by just the sample size of the cohort for the standard error?

data_science/data_science.rb
jklemen: I *believe* this is handled prematurely in calc_mean_error, by mistake. I will propagate the constant down to the correct method.

data_science/rspec/test_spec.rb
jklemen: Just learned about this yesterday, will change!

data_science/data_science.rb
jmtai: Correct, that calculation is per-cohort.

data_science/data_science.rb
djkotowski: And therefore can be shortened to.
```ruby
open(json_file, &:read)
```

data_science/data_science.rb
djkotowski: It would be useful to support case-insensitivity.

data_science/data_science.rb
djkotowski: Just be sure not to name your variables anything like `num_conversions_of_cohort_formerly_known_as_a`

data_science/data_science.rb
djkotowski: It may even be ideal to redefine `to_s` on Cohort to provide this functionality.

data_science/data_science.rb
jklemen: noted.

data_science/data_science.rb
jklemen: Great suggestion!

data_science/rspec/test_spec.rb
jklemen: Interesting. I will change this.

PR #779, opened 2015-07-07
Dino catalog

jklemen:


dino_catalog/dino.rb
```diff
@@ -0,0 +1,50 @@
+class Dino
+  attr_accessor :attrs
+
+  def initialize(data)
+    keys = %w(name period continent diet weight walking description)
```
incazteca: Is there a reason for not just setting these with attr_accessor? The default value for these attributes would be nil.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,50 @@
...
+    @attrs = Hash[keys.zip([nil] * keys.count)]
+    @attrs.merge!(data)
+  end
+
+  def dino_facts
+    @attrs.each do |attribute, value|
```
incazteca: If you were to have the attributes defined through `attr_accessor

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,93 @@
...
+    }.to_json
+  end
+
+  def inspect
+    dinos.inspect
+  end
+
+  private
+
+  def _lt(field, value)
```
incazteca: There isn't a need in ruby to prepend private methods with `_`

incazteca:
Dino Catalog looks good to me. A resubmission isn't needed. However, I would like to know why you just didn't set the attributes in the dinosaur model with `attr_accessor`

dino_catalog/dino.rb
jklemen: But then I would need a method to convert instance variable strings into accessor method calls, no?

dino_catalog/dinodex.rb
jklemen: Personal preference. Lets me know at a glance that it is private when used elsewhere in the class. If not doing it is a globally acknowledged rule, then I will discontinue this practice in Ruby.

jklemen:
No attr_accessor for 3 reasons, in my opinion:
1. Easier initialization.  With attr_accessor, I would need to call all *n* setter methods. I would also still be using a param hash in this instance, so why not just merge that hash with a default one stored internally?
2. Allows for more generic processing with my queries without having to convert strings to accessor method calls.
3. Easier and more secure external method chaining. Example:
> dinodex.less_than('weight', 10000).equals('diet', 'carnivore')

would require implementing a method (or 2) in dinodex to convert #param1 to a method in a Dino. Or, if I were feeling particularly dirty, I could use send(#param1) on a dino object to call the getter method. But that would allow my user to enter anything they want (like 'weight=') and have it sent to my dino objects to be executed. Which is obviously bad.

dino_catalog/dino.rb
jklemen: See below

dino_catalog/dinodex.rb
djkotowski: It's bad practice, as you'll then have to change the method name if you decide to make it public.

dino_catalog/dino.rb
djkotowski: Not really.  You can use `#send` to help out there.

dino_catalog/dinoparse.rb
```diff
@@ -0,0 +1,35 @@
...
+    d_attrs['name'] = d_attrs['name'] || d_attrs['genus']
+
+    new_diet = (d_attrs['carnivore'] ? 'Carnivore' : "Non-Carnivore")
+    d_attrs['diet'] = d_attrs['diet'] || new_diet
+
+    filter(d_attrs)
+  end
+
+  def parse(dinofile)
+    dinos = CSV.parse(open(dinofile).read)
```
djkotowski: Prefer
```ruby
dinos = CSV.parse(open(dinofile, &:read))
```

This ensures the stream is closed after reading.

dino_catalog/dinoparse.rb
```diff
@@ -0,0 +1,35 @@
...
+    keys = dinos.shift.map(&:downcase)
+    dinos.map do |dino_attrs|
+      raw_hash = Hash[keys.zip(dino_attrs)]
```
djkotowski: Generally try to avoid using data types in variable names.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  def sort(field)
+    Dinodex.new(
+      @dinos.sort do |a, b|
+        a.compare(field, b.attrs[field])
+      end,
+    )
+  end
+
+  def json
```
djkotowski: This should be named `to_json` to be consistent with convention.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,93 @@
...
+    hash.inject(self) do |ddex, pair|
+      ddex.equal(pair[0], pair[1])
+    end
+  end
+
+  def sort(field)
+    Dinodex.new(
+      @dinos.sort do |a, b|
+        a.compare(field, b.attrs[field])
+      end,
```
djkotowski: Remove the trailing comma.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,93 @@
...
+    )
```
djkotowski: This is a bit arrowhead-y.  I'd break `@dinos.sort` out of the parameter list, then pass the result into `Dinodex.new` separately.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,93 @@
...
+  def initialize(dinos)
+    @dinos = dinos || []
+  end
+
+  def index
+    @dinos.each(&:dino_facts)
+  end
+
+  def less_than(field, value)
+    Dinodex.new(_lt(field, value))
```
djkotowski: This is a bit too much of an abstraction.  It's fine including the body of `_lt` in this method.  It would also be consistent with what you did with `#sort`.  The same goes with the methods below.

dino_catalog/tests.rb
```diff
@@ -0,0 +1,63 @@
...
+file2 = 'african_dinosaur_export.csv'
+
+dinos1 = Dinoparse.new(file1).dinos
+dinos2 = Dinoparse.new(file2).dinos
+
+dinodex = Dinodex.new(dinos1 + dinos2)
+
+# QUERIES
+puts "Grab all dinosaurs that were bipeds"
+puts "-" * 20
```
djkotowski: To avoid repeating yourself, I'd store this in a constant (e.g. `SEPARATOR`).

dino_catalog/dino.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def equal?(field, value)
+    compare(field, value) == 0
+  end
+
+  def like?(field, value)
+    myval = attrs[field]
+    if myval.nil? || value.nil?
+      (myval.nil? && value.nil?)
```
djkotowski: Redundant parentheses.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,93 @@
...
+    Dinodex.new(_eq(field, value))
+  end
+
+  def not_equal(field, value)
+    Dinodex.new(@dinos - _eq(field, value))
+  end
+
+  def less_or_equal(field, value)
+    Dinodex.new(
+      _lt(field, value) + _eq(field, value),
```
djkotowski: Unnecessary trailing comma.  See below methods too.

djkotowski:
I highly recommend using accessors for your Dino class, and using `#send` to dynamically access them.  If you need to dynamically access the reader, you just pass the field name to `#send`.
Example:
```ruby
 def compare(field, with)
  # TODO: check to ensure `field` is accessible
  myval = send(field)
  if !(myval && with)
    myval.__id__ <=> with.__id__
  elsif field == 'weight'
    myval.to_i <=> with.to_i
  else
    myval.downcase <=> with.downcase
  end
end
```

The setter can be accessed similarly:
```ruby
send("#{field}=", new_value)
```

djkotowski:
Good first iteration.  Your querying implementation is interesting.

Take a look at the comments, refactor and resubmit when ready.  Talk to Franco or me if you have any questions.

dino_catalog/dinodex.rb
jklemen: Makes sense I guess. I'll change it to conform.

dino_catalog/dinoparse.rb
jklemen: Right, forgot to change that.

dino_catalog/dinoparse.rb
jklemen: Why?

dino_catalog/dinodex.rb
jklemen: Good call.

dino_catalog/dinodex.rb
jklemen: But...rubocop says so D: I guess I'll change this at the cost of the extra violation.

dino_catalog/dinodex.rb
jklemen: Sure

dino_catalog/dinodex.rb
jklemen: Can you elaborate? I call _lt in other functions and simply wanted to DRY it out.

dino_catalog/tests.rb
jklemen: I'll refactor this entire file.

dino_catalog/dinoparse.rb
djkotowski: Because you shouldn't be as concerned about the data type.  Only the methods available to you.  It is related to the "Comments lie" principle.

dino_catalog/dinodex.rb
djkotowski: Just in `less_than_or_equal`.  Instead of taking the union of the two filters (requiring you to iterate through the collection twice), add a `less_than_or_equal?` method to Dino and use that.  That way, you only have to filter the collection once.  With this change applied to the other methods, you reduce the size of Dinodex.

dino_catalog/dinodex.rb
kotp: It is saying to because of the do..end, it is tied to (syntax/precedence-wise) to the argument list (and represents as an array) rather than to the sort method.   So it is right, as it is referring to what it thinks is a continued array.  But wrong in your intentions.  Convert it to the stronger precedence {} for, or perhaps better yet, assign it to a variable and pass that variable in as the argument to Dinodex.new.  This will also help with that 5 second rule, as the name of the variable will help determine the idea that you are passing as the argument.

PR #781, opened 2015-07-07
Initial Dino Catalog

Tornquist:


dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,84 @@
...
+    dinos = DinoParse.new(files)
+    DinoDex.new(dinos)
+  end
+
+  def initialize(dinos)
+    @dinos = dinos
+  end
+
+  def bipeds
+    DinoDex.new(dinos.select { |dino| dino.walking.downcase == "biped" })
```
djkotowski: If you move the decision logic out to `Dinosaur` via predicate methods, you'll be able to shorten this.

Example:
```ruby
def bipeds
  DinoDex.new(dinos.select(&:biped?))
end
```

The same goes for the below methods.

dino_catalog/dinoparse.rb
```diff
@@ -0,0 +1,55 @@
...
+    standardized = []
+    dinosaurs.each do |dinosaur|
+      temp = {}
+      temp[:genus] = dinosaur[:name] || dinosaur[:genus]
+      temp[:weight] = dinosaur[:weight_in_lbs] || dinosaur[:weight]
+      temp[:walking] = dinosaur[:walking]
+      temp[:continent] = dinosaur[:continent]
+      temp[:description] = dinosaur[:description]
+      temp[:diet_details] = dinosaur[:diet]
+      temp[:carnivore] = dinosaur[:carnivore] ||
```
djkotowski: Since this is essentially a boolean value, why not make it one?  Save "Yes" and "No" for formatted output.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,84 @@
...
+class DinoDex
+  attr_accessor :dinos
+
+  def self.new_from_files(files)
+    dinos = DinoParse.new(files)
+    DinoDex.new(dinos)
+  end
+
+  def initialize(dinos)
+    @dinos = dinos
```
djkotowski: Great for supporting chaining.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def triassic
+    periods("triassic")
+  end
+
+  def filter(*filters)
+    individual_results = []
+    filters.each do |filter|
+      individual_results << send(filter)
+    end
```
djkotowski: This can be replaced with `#map`.

```ruby
def filter(*filters)
  filters.map { |filter| send(filter) }
end
```

This is also a dangerous method, as it doesn't verify that the filter exists or is allowed.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,84 @@
...
+    DinoDex.new(individual_results.map(&:dinos).inject(&:&))
+  end
+
+  def to_s
```
djkotowski: This should _return_ a string, not output it.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,84 @@
...
+      puts dino
+    end
+  end
+
+  private
+
+  def dino_period_search(search)
+    dinos.select do |dino|
+      matches = dino_periods_to_match_list(dino, search)
+      matches.inject { |a, b| a || b }
```
djkotowski: This statement is very hard to understand.  What are `a` and `b`?

dino_catalog/dinoparse.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+      results << temp
+    end
+    results
+  end
+
+  def self.read(files)
+    raw_dinosaurs = []
+    files.each do |file|
+      CSV.foreach(file, headers: true, header_converters: :downcase) do |row|
```
djkotowski: Sadly, CSV doesn't have a `#map`.  At least not without loading the entire file first.

dino_catalog/dinoparse.rb
```diff
@@ -0,0 +1,55 @@
...
+        temp[:modifier] = parts.first
+      end
+      results << temp
+    end
+    results
+  end
+
+  def self.read(files)
+    raw_dinosaurs = []
+    files.each do |file|
```
djkotowski: You can remove L47 and L53 by using `#each_with_object`.

```ruby
def self.read(files)
  files.each_with_object([]) do |file, raw_dinosaurs|
    # Your CSV code here
  end
end
```

dino_catalog/dinodex.rb
Tornquist: I think this line reads clearer when used with the statement reduce.  It essentially takes all of the matches (an array of boolean values) and or's them together. You end up with a single success or failure statement.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,48 @@
+class Dinosaur
+  attr_accessor :genus, :weight, :walking, :continent, :description,
+    :carnivore, :diet_details, :periods
```
djkotowski: Curse Hound for preferring this alignment instead of aligning the symbols.

dino_catalog/sample.md
```diff
@@ -0,0 +1,32 @@
+# Dino Catalog Sample Program
+
+### Dependencies
+```sh
+require 'csv'
+require 'rails'
```
djkotowski: This isn't a Rails app? Why require it?

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,48 @@
...
+    @description = dinosaur[:description]
+    @diet_details = dinosaur[:diet_details]
+    @carnivore = dinosaur[:carnivore]
+    @periods = dinosaur[:periods]
+  end
+
+  def to_s
+    result = "Dino Details:"
+
+    print_order = [:genus, :periods, :description, :weight, :walking,
```
djkotowski: This should be a constant.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,48 @@
...
+    print_order.each do |field|
+      if field
+        result << "\n#{field.to_s.titleize}: "
+        if (field != :periods)
+          result << send(field).to_s
+        else
+          result << generate_period_string
+        end
+      end
+    end
```
djkotowski: Avoid the extra indentation of L22-29 with a guard clause.

```ruby
print_order.each do |field|
  next unless field

  result << "\n#{field.to_s.titleize}: "
  if (field == :periods)
    result << generate_period_string
  else
    result << send(field).to_s
  end
end
```

dino_catalog/dinoparse.rb
```diff
@@ -0,0 +1,55 @@
...
+      temp[:periods] = dino_periods_processor(dinosaur[:period])
+      standardized << temp
+    end
+    standardized
+  end
+
+  def self.dino_periods_processor(periods)
+    results = []
+    periods.split(" or ").each do |segment|
+      temp = {}
```
djkotowski: What is `temp`?  Ensure the naming makes sense to a reader glancing at this for the first time.

dino_catalog/dinoparse.rb
```diff
@@ -0,0 +1,55 @@
...
+                         (dinosaur[:diet] == "Carnivore" ? "Yes" : "No")
+      temp[:periods] = dino_periods_processor(dinosaur[:period])
+      standardized << temp
+    end
+    standardized
+  end
+
+  def self.dino_periods_processor(periods)
+    results = []
+    periods.split(" or ").each do |segment|
```
djkotowski: Another place where `#map` would be beneficial.

dino_catalog/dinoparse.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def self.convert_to_object(dinosaurs)
+    dinosaurs.map{ |dinosaur| Dinosaur.new(dinosaur) }
+  end
+
+  def self.standardize(dinosaurs)
+    standardized = []
+    dinosaurs.each do |dinosaur|
+      temp = {}
```
djkotowski: Why `temp`?  This is dino information.  Be specific with naming.

djkotowski:
Good first iteration.

Be sure to use a Gemfile for dependencies and to only include what is necessary for the application to run.  `#titleize` is not a big enough method to warrant including Rails.  If you absolutely need ActiveSupport core extensions for String, include that instead.

Your Dinodex is also responsible for knowing facts about each dinosaur (What makes a dino a biped?  Is it from the Jurassic period?  Et cetera.)  Dinosaur should handle that responsibility.

Take a look at the comments, refactor and resubmit when ready.  Let me know if you have any questions.

dino_catalog/sample.md
Tornquist: I'm using some rails support methods to make things a little cleaner.  I'll pull it into a Gemfile though.

dino_catalog/dinodex.rb
djkotowski: `Enumerable#any?` would make more sense in this case.

dino_catalog/sample.md
djkotowski: It's specifically just `active_support`

dino_catalog/dinoparse.rb
Tornquist: Made this change.  I'll be submitting the new iteration soon.  I think the workflow got worse with this change.  I'm sure there is a clean way to do it, but I'm not happy with the one that I chose.

PR #784, opened 2015-07-07
Arrowhead v2

Tornquist:
I'm not happy with the error messages.  I put them on a separate line so that Hound wouldn't complain about them for being long.  I think an error about using a constant may still be thrown though.

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -14,8 +14,10 @@ class Arrowhead
     },
   }
 
-  ERRs = "Unknown shape value. Are you sure you know what you're talking about?"
-  ERRr = "Unknown region, please provide a valid region"
+  ERRORS = {
+    shape: "Unknown shape value. Are you sure you know what you're talking about?",
+    region: "Unknown region, please provide a valid region"
+  }
 
   def self.classify(region, shape)
     if check_status(region, shape)
@@ -28,9 +30,9 @@ class Arrowhead
     if valid_region?(region) && valid_shape?(region, shape)
       return true
     elsif valid_region?(region)
-      raise NameError, ERRs
+      raise NameError, ERRORS[:shape]
     else
-      raise NameError, ERRr
+      raise NameError, ERRORS[:region]
     end
   end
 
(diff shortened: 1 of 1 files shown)


arrowhead/arrowhead.rb
```diff
@@ -15,20 +14,33 @@ class Arrowhead
     },
   }
 
-  # FIXME: I don't have time to deal with this.
+  ERRs = "Unknown shape value. Are you sure you know what you're talking about?"
+  ERRr = "Unknown region, please provide a valid region"
```
Tornquist: I can't or I'll violate the line length for the first error message.  In this case, I feel like the 80 character rule should be ignored.  What's best practice?

kotp: You have a few options for this.

```
ERR_REGION = "Unknown region, " +
  + "please provide a valid region."
```

or 

```
ERR_REGION =
"Unknown region, please provide a valid region."
```

I think I might prefer a ERR constant as a hash, and store the keys that you want to use with the string values in that key.  Makes it easier to remember and easier to label.

```
ERRORS = {
r: "Unknown region, please provide a valid region",
s: "Unknown shape value. Are you sure you know what you're talking about?"
}

I would say the 80 character rule is more a guideline, than a rule.  There are times that it won't make sense to preserve it.

File: arrowhead/arrowhead.rb [hunk 2]
```diff
@@ -15,20 +14,35 @@ class Arrowhead
...
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
+    if check_status(region, shape)
+      arrowhead = CLASSIFICATIONS[region][shape]
+      puts "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
+    end
+  end
+
+  def self.check_status(region, shape)
+    if valid_region?(region) && valid_shape?(region, shape)
```
incazteca: This `if` block is a little confusing due to the last last `else` returning a region NameError. This `if` block can be removed and you could instead just have two lines that raise errors if the region is not included. The method itself can be renamed to `valid_classification?`as it's not really checking for a status message.

incazteca:
@Tornquist Fix up the `check_status` method. A resubmission is not necessary.

arrowhead/arrowhead.rb
Tornquist: How's this look? It simplified the classify method as well

PR #785, opened 2015-07-07
Dino Catalog (Rev 2) jnyman

jnyman:
This is my second attempt at the Dino Catalog example, based on initial reviewer comments.

I made a lot of little structural changes here that I think reflect more idiomatic Ruby code. I did put two reviewer notes in Converter.rb There were suggestions about my code there but in two cases I found it difficult to accommodate them but am certainly open to further suggestions.


dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,119 @@
...
+  def process_condition_array(condition, data)
+    condition.each do |term|
+      data = @entity.send(term, data) if term.class == Symbol
+      data = search_data_text(data, term) unless term.class == Symbol
+    end
+    data
+  end
+
+  def process_condition_hash(condition, data)
+    condition.reduce(data) do |entity_list, (key, value)|
```
jnyman: I don't know whose preference this is. Rubocop had no issues with the code. From what I've read online, [including from Uncle Bob](http://martinfowler.com/articles/collection-pipeline/reduce.html) and the [rspec team decision](https://github.com/rspec/rspec-core/issues/1110), `reduce` is often favored. Not sure if this requires a resubmit.

incazteca: We actually prefer not using `inject` since it can be difficult to comprehend.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,33 @@
+class Dinosaur
+  attr_accessor :name, :period, :continent, :diet, :weight, :movement,
+    :description
+
+  HEADERS = %w(Name Period Continent Diet Weight Movement Description)
+  MAX_SMALL_WEIGHT = 2000
+
+  def self.columns
+    HEADERS.map(&:downcase)
```
incazteca: Why not just downcase `HEADERS` when you create them?

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def self.big(data_set)
```
incazteca: The `self` is not necessary

incazteca: No wait... I see what you're trying to do here.

dino_catalog/dinosaur.rb
jnyman: You know, come to think of it, you do raise a really good point. In reality, I could do exactly that. That might even obviate the need for the `columns` method since I could just reference `HEADERS` directly.

dino_catalog/rolodex.rb
jnyman: Do you mean folks here prefer `reduce` over `inject` in terms of the name to use or that there is a preference to not use `reject`/`inject` at all?

dino_catalog/converter.rb
```diff
@@ -0,0 +1,58 @@
...
+  # NOTE TO REVIEWER:
+  # This method was brought up as being unnecessary and probably good to
+  # remove. This method exists to call other conversions should they be
+  # necessary. This seems like good design to me because in my setup_options()
+  # in dinodex.rb I don't want to have to pass in multiple method calls for
+  # different converters. Instead I always just call the one below. This keeps
+  # the interface consistent and less subject to change. So even though this
+  # is planning ahead a bit, it seems a useful bit of planning ahead.
+  # Thoughts?
+  def self.dino_data_convert
```
incazteca: I think this method is a case of premature optimization. At the moment you only need to do a conversion for carnivore and you might find that if something did change in the future it might lead to a solution that you did not foresee now leading to tech debt being created due to loose ends being left here and there.

jnyman: Yeah, it's a good point. I can honestly see both sides of this. The way I was looking at it is that the reality is this is a dinosaur catalog program. It's known that different data sources may be used and, in those cases, conversion must occur in order to allow queries.

So, given that I use converters to handle that, I can opt to treat the converters as _always_ calling `dino_data_convert`. That then never needs to change -- unless the program requirements change so drastically that everything would be up in the air anyway. This means `dino_data_convert` thus is responsible for calling however many utility methods are needed for specific conversions. It may only be one conversion (as in my case) but it will likely involve many more.

In fact, even with this example challenge, it could be argued that there already would be more conversions, if I had a customer to talk with. For example, if a continent is blank, do I convert the data to put in a continent if the file is specified as being Africa? What about blank weight values? Should those be converted? So the challenge is missing a key component that would help us determine if we are optimizing or gold plating too much: customer feedback on the problem domain. (To be critical of my own approach, though, one could argue that the lack of this means I should _not_ have done even the small optimizing that I did.)

So while some of this is outside the challenge proper, the challenge itself sets itself up to suggest that these are immediate concerns.

So, again, I do genuinely see both sides of this issue. I think, however, that there is a balance between thinking ahead when the customer (problem space) clearly has already indicated the direction things will go versus optimizing for conditions that have yet to manifest in any way. All this being said, it's certainly nothing I'm religious about and I can most definitely see how this could be seen as an unnecessary optimization.

dino_catalog/dinosaur.rb
incazteca: Ok, So I looked back at the conversation that led to this in the previous PR. `self.small` and `self.big` should not be in the Dinosaur model. I know you stated that you felt that dinodex was doing too much but it looks to be a better place for these methods since they're doing the search on dinosaur. It wouldn't really fit in rolodex though because rolodex is a bit more generalized in it's search functionality.

jnyman: Yeah, I was kind of torn on that. I know Josh felt DinoDex was doing too much and I can sort of see that. My original concern about moving `big` and `small` to `Dinosaur` was that technically they are not aspects of a particular dinosaur but rather aspects used to qualify a given set of dinosaurs.

On the other hand, though, having these on `Dinosaur` does allow me to question if a `Dinosaur` is big or small, based on what that condition means. Since each dinosaur will have a weight, I suppose it makes sense for the dinosaur itself to be able to answer if it's big or small -- if it's told what "big" or "small" means.

So I have two reviewers who felt differently about this (which is kinda cool, because it forces me to think more strongly about it). Given that, I wonder if this is a case where these methods don't quite fit in `DinoDex` or `Rolodex`. In which case, that leads to an interesting question: _where do they fit?_ :)

dino_catalog/rolodex.rb
incazteca: There's a preference not to use `reduce` or `inject`. However in this case it doesn't seem a good choice since anyways since it isn't doing a mathematical task.

incazteca:
@jnyman I feel that this dinodex is complete and a resubmission isn't needed.

dino_catalog/rolodex.rb
jnyman: Interesting. So one of the comments in the previous PR was that I _should_ consider `reduce`/`inject` here. My understanding is that if you have an array that you want to reduce to one result by applying a recurrent logic, then you would use `reduce`, even if that was not at a mathematical task. Is that not true?

jnyman:
@djkotowski , @incazteca : Just checking -- should I be the one that closes this, assuming no further submissions are needed? I didn't want do anything prematurely.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,151 @@
...
+    @options[:path] += '/' unless ends_with_slash?(options[:path])
+    @options[:path] += '*.csv' if no_extension?(options[:path])
+  end
+
+  def no_extension?(path)
+    File.extname(path).empty?
+  end
+
+  def ends_with_slash?(path)
+    path.split('').last == '/'
```
djkotowski: Could also be
```ruby
path.chars.last == '/'
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,151 @@
...
+  def present_results(conditions)
+    results = @rolodex.query(@data.data_set, conditions)
+    return puts 'No results found with your criteria.' if results.empty?
+    table(results.values, fields: Dinosaur.columns)
+  end
+
+  def create_file_path
+    return unless File.extname(options[:path]).empty?
+    @options[:path] += '/' unless ends_with_slash?(options[:path])
+    @options[:path] += '*.csv' if no_extension?(options[:path])
```
djkotowski: Use [`File.join`](http://ruby-doc.org/core-2.2.0/File.html#method-c-join) here.

dino_catalog/dinodex.rb
jnyman: Ah, interesting. Was not even aware of that one. That looks cleaner to me.

dino_catalog/dinosaur.rb
djkotowski: `big` and `small` queries should be in DinoDex, but it makes sense for Dinosaur to have the predicates `big?` and `small?`.

dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,119 @@
...
+    term_found = false
+    value.instance_variables.each do |header_variable|
+      column_cell = value.instance_variable_get(header_variable)
+      term_found = column_cell.to_s.include?(term) unless term_found
+    end
+    term_found
+  end
+
+  def create_header_methods
+    @headers.each do |header|
```
djkotowski: Interesting use of metaprogramming.  Be cautious, as extensive metaprogramming can make code difficult to change and debug.

dino_catalog/rolodex.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+  end
+
+  def search_by_header_column(value, term)
+    term_found = false
+    value.instance_variables.each do |header_variable|
+      column_cell = value.instance_variable_get(header_variable)
+      term_found = column_cell.to_s.include?(term) unless term_found
+    end
+    term_found
```
djkotowski: A better way to handle this loop is to use `#any?` instead of `#each`.  `#any?` will return true if the provided code block returns true for at least one element in the collection.

dino_catalog/rolodex.rb
jnyman: Amen to that. I was trying to stretch my Ruby muscles a bit here but I absolutely understand what you are saying because this was an area that was worrying me a bit in terms of being obtuse. I felt this code was straying from obvious expressiveness based on the fact that I kept feeling the need to comment it!

djkotowski:
Good work!  Take a look at the comments.  No need for a resubmission, though.

PR #786, opened 2015-07-08
Robot Name v2

Tornquist:


robot_name/robot_name.rb
```diff
@@ -12,21 +12,55 @@ def initialize(args = {})
     if @name_generator
```
johnmcconnell: Okay this is actually for line 9, but you should probably not want to initialize a class variable inside the instance scope. Might as well make line 6 @@registry = []

johnmcconnell:
Looks good.

robot_name/robot_name.rb
```diff
@@ -12,21 +12,55 @@ def initialize(args = {})
...
+    rand(10)
+  end
+
+  def validate_name(name)
+    check_valid_name(name)
+    check_new_name(name)
+  end
+
+  def check_valid_name(name)
+    if (name =~ /\A[[:alpha:]]{2}[[:digit:]]{3}\Z/).nil?
```
djkotowski: Prefer
```ruby
unless name =~ /\A[[:alpha:]]{2}[[:digit:]]{3}\Z/
```

robot_name/robot_name.rb
```diff
@@ -12,21 +12,55 @@ def initialize(args = {})
...
+      raise NameError, "Name is not valid"
+    end
+  end
+
+  def check_new_name(name)
+    if @@registry.include?(name)
+      raise NameCollisionError, 'There was a problem generating the robot name!'
+    end
```
djkotowski: This is good.  One thing some have done to further avoid arrowhead coding is to create "raise" methods for the different errors.

Example:
```ruby
def check_new_name(name)
  raise_name_collision if @@registry.include?(name)
end

def raise_name_collision
  raise NameCollisionError, "There was a problem generating the robot name!"
end
```

It's a matter of preference, though.

robot_name/robot_name.rb
```diff
@@ -12,21 +12,55 @@ def initialize(args = {})
...
     @@registry << @name
   end
+
+  private
+
+  def generate_name
+    "".tap do |name|
+      2.times { name << generate_char }
+      3.times { name << generate_num.to_s }
+    end
```
djkotowski: Nice using `#tap`.

djkotowski:

PR #791, opened 2015-07-09
Triangle

oharace:
Made style changes to ruby file.

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -1,55 +1,55 @@
 # Killer facts about triangles AWW YEAH
 class Triangle
-	attr_accessor :side1,:side2,:side3
+  attr_accessor :side1,:side2,:side3
 
-	def initialize(side1,side2,side3)
-		@side1 = side1
-		@side2 = side2
-		@side3 = side3
-	end
+  def initialize(side1,side2,side3)
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+  end
 
-	def equilateral
-		side1 == side2 && side2 == side3
-	end
+  def equilateral
+    side1 == side2 && side2 == side3
+  end
 
-	def isosceles
-		[side1,side2,side3].uniq.length == 2
-	end
+  def isosceles
+    [side1,side2,side3].uniq.length == 2
+  end
 
-	def scalene
-		equilateral || isosceles ? false : true
-	end
+  def scalene
+    equilateral || isosceles ? false : true
+  end
 
-	def recite_facts
-		puts 'This triangle is equilateral!' if equilateral
-		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+  def recite_facts
+    puts 'This triangle is equilateral!' if equilateral
+    puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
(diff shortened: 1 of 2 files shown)


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,55 @@
...
+  def equilateral
+    side1 == side2 && side2 == side3
+  end
+
+  def isosceles
+    [side1,side2,side3].uniq.length == 2
+  end
+
+  def scalene
+    equilateral || isosceles ? false : true
```
djkotowski: Just return the negated condition.

```ruby
!(equilateral? || isosceles?)
```

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,55 @@
...
-end
+  attr_accessor :side1,:side2,:side3
+
+  def initialize(side1,side2,side3)
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+  end
+
+  def equilateral
```
djkotowski: This and the two methods below it are predicate methods (returning a boolean value).  Predicate methods should be suffixed with a `?`

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,55 @@
...
+
+  def recite_facts
+    puts 'This triangle is equilateral!' if equilateral
+    puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
+    puts 'This triangle is scalene and mathematically boring.' if scalene 
 
+    angles = self.calculate_angles(side1,side2,side3)
+    puts 'The angles of this triangle are ' + angles.join(',')
+
+    puts 'This triangle is also a right triangle!' if angles.include? 90
```
djkotowski: Wrap args with parentheses.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,55 @@
...
+    puts ''
```
djkotowski: `puts` can be called without an argument.  No need for the empty string.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,55 @@
...
+    equilateral || isosceles ? false : true
+  end
+
+  def recite_facts
+    puts 'This triangle is equilateral!' if equilateral
+    puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
+    puts 'This triangle is scalene and mathematically boring.' if scalene 
 
+    angles = self.calculate_angles(side1,side2,side3)
+    puts 'The angles of this triangle are ' + angles.join(',')
```
djkotowski: Prefer `<<` over `+` when the leftmost string can be altered. `+` creates a new string from two others.  `<<` appends to the leftmost string.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,55 @@
...
-		angleC = radians_to_degrees(Math.acos((a**2 + b**2 - c**2) / (2.0 * a * b)))
-
-		return [angleA, angleB, angleC]
-	end
-
-	def radians_to_degrees(rads)
-		return (rads * 180 / Math::PI).round
-	end
-end
+  attr_accessor :side1,:side2,:side3
```
djkotowski: Spacing.

djkotowski:
Nice work.  There are a few comments throughout.  Keep them in mind for future exercises.

PR #793, opened 2015-07-09
Arrowhead

zsyed91:
Refactor arrowhead so that it better follows Sandi Metz rules. Also moved the error status closer to the error checking


djkotowski:

PR #794, opened 2015-07-09
Fixed arrowhead rule issues

oharace:
Fixed some of the rule-breaking within Arrowhead. Split the logic into separate methods.


djkotowski:
Avoid the "true or raise error" style of error checking.  Either validate with a predicate method _or_ raise an error if a condition is met.  The latter would let you structure your code like:
```ruby
check_region(region)
check_shape(region, shape)
# retrieve info
```

djkotowski:
Looks good.  Just one comment to keep in mind for the next exercise.  No need for a resubmission.

PR #796, opened 2015-07-09
robot_name

zsyed91:
Refactor robot_name to use methods and clean up the code

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -2,24 +2,51 @@ class NameCollisionError < RuntimeError; end
 
 class Robot
   attr_accessor :name
+  attr_accessor :name_generator
 
-  @@registry
+  @@registry = []
 
   def initialize(args = {})
-    @@registry ||= []
     @name_generator = args[:name_generator]
+    @name = generate_name
 
-    if @name_generator
-      @name = @name_generator.call
+    check_name_collision
+    @@registry << name
+  end
+
+  def generate_name
+    if name_generator
+      name_generator.call
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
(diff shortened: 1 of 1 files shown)


robot_name/robot_name.rb
```diff
@@ -2,24 +2,51 @@ class NameCollisionError < RuntimeError; end
...
+    @@registry << name
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def generate_name
+    if name_generator
+      name_generator.call
+    else
+      generate_character_sequence(2) + generate_number_sequence(3)
     end
```
djkotowski: Prefer guard clauses to decision blocks here.  Also, prefer `<<` to `+` (the former appends to the first string, the latter combines into a new string)
```ruby
return name_generator.call if name_generator
generate_character_sequence(2) << generate_number_sequence(3)
```

djkotowski:

robot_name/robot_name.rb
```diff
@@ -2,24 +2,51 @@ class NameCollisionError < RuntimeError; end
...
+  def generate_character
+    ('A'..'Z').to_a.sample
+  end
+
+  def generate_number
+    rand(10).to_s
+  end
+
+  def check_name_collision
+    return if name =~ /[[:alpha]]{2}[[:digit:]]{3}/
```
kotp: Does line 49 ever get reached if this is valid?  Does line 47?

djkotowski: Yep, the logic is a bit messed up here.

robot_name/robot_name.rb
```diff
@@ -2,24 +2,51 @@ class NameCollisionError < RuntimeError; end
 
 class Robot
   attr_accessor :name
+  attr_accessor :name_generator
```
djkotowski: Just combine the two accessors into one line.

PR #797, opened 2015-07-10
Dino catalog

rcaileff:
Initial LevelUp dinodex solution.


dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,159 @@
+require 'optparse'
+require 'ostruct'
+
+require_relative 'dinodex_catalog'
+
+class DinoOptParser
```
jmtai: Break up these files so you have one class per file. It's not a big deal for this exercise, but once you start working with Rails, the autoloader will search for files based on the class names. So, for example, Rails would expect `DinoOptParser` to be in `dino_opt_parser.rb`.

jmtai: However, custom error or exception classes can be included in the same file where they are being used (assuming they aren't used by multiple classes, in which case it would make sense to put them in a separate file).

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,159 @@
...
+  attr_accessor :parser, :parsed_options
+
+  def initialize
+    @parser = OptionParser.new
+    @parsed_options = OpenStruct.new
+    configure_parser
```
jmtai: Suggestion:  Place your methods in the file in the order in which they are called (if possible). This is not critical but I find that it helps reduce the amount of jumping around the file when examining unfamiliar code. If you put your `configure_parser` method right after the initialize, it's super easy to just read the code from top to bottom.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,159 @@
...
+  end
+
+  def configure_list_dinos_option
+    @parser.on(
```
jmtai: Since you've used `attr_accessor` for `parser` and `parsed_options`, you don't need to use the `@` here.  Not a big deal but will improve readability.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,159 @@
...
+    @parser.separator ""
+    @parser.separator "Abbreviations:"
+    explain_attr
+    explain_value
+  end
+
+  def explain_attr
+    text = "  [ATTR] represents the name of a dinosaur attribute."
+    valid_attrs = Dinosaur.new.searchable_attrs
+    text << "  Can be any of: #{valid_attrs.join(', ')})"
```
jmtai: Nice use of the shovel operator.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,159 @@
...
+    @parser.separator text
+  end
+
+  def explain_value
+    text = "  [VALUE] represents the desired value of a dinosaur attribute."
```
jmtai: For longer multi-line strings, you can also use Ruby HEREDOCs.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,159 @@
...
+    @parser.on(
+      "-w", "--write [FILENAME]",
+      "Write entire dinodex in JSON format to named file") do |filename|
+      @parsed_options.filename = filename
+      @parsed_options.command = :write
+    end
+  end
+
+  def verify_search_arguments_list_length(list)
+    list_length_error = "List must have an even number of elements"
```
jmtai: This should probably be defined as a class-level constant (i.e. define it outside of any method, usually at the top of the class) instead of a local variable, as it shouldn't be modified. In Ruby, constants aren't actually constant but it signals to the reader that it shouldn't be modified.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,159 @@
...
+  attr_accessor :dino_dex, :parsed_options
+
+  def initialize
+    @dino_dex = DinoCatalog.new(dino_files: default_dino_files)
+    @parsed_options = DinoOptParser.new.parse(ARGV)
+  end
+
+  def default_dino_files
+    [
+      DinoFile.new("dinodex.csv", :original),
```
jmtai: It's kind of a bummer that a user of the DinoFile class has to specify an explicit format for a CSV input file. Wouldn't it be nicer to just handle the differences automatically in some way?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,159 @@
...
+      DinoFile.new("african_dinosaur_export.csv", :p_b_africa),
+    ]
+  end
+
+  def list_dinos
+    puts @dino_dex.dinosaurs
+  end
+
+  def write_file
+    dino_dex.write_dinosaurs_as_json_file(parsed_options.filename)
```
jmtai: I like the descriptive method names, but I think you might be able to shorten them a bit while still keeping them descriptive. Something like `export_as_json` might still be clear while making the method name a bit more succinct. Naming, of course, can be pretty subjective but we definitely want to try to strike a balance between expressiveness and brevity.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,159 @@
...
+  end
+
+  def search
+    attr = nil
+    hash = {}
```
jmtai: `hash` is not a great variable name. In general, we try to avoid using the data type in variable names. Perhaps `search_params` or something like that might be more descriptive?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,159 @@
...
+    parsed_options.list.each_with_index do |element, index|
+      index.even? ? attr = element : hash[attr] = element
```
jmtai: Here we can see how naming can make it difficult to understand what a block of code is doing. While this is nice and compact, it might be clearer if you did something like this:
```ruby
params = parsed_options.list
while !params.empty?
  key = params.shift
  value = params.shift
  hash[key] = value
end
```

jmtai: I'd argue that it would be even better to change the command line interface to force the user to delineate key-value pairs with a different character. That way you can just split the pairs on that character and not have to use even/odd indexing logic.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,159 @@
...
+    end
+    puts dino_dex.search_by_hash(hash)
+  end
+
+  def execute_command
+    return list_dinos if @parsed_options.command == :list
+    return write_file if @parsed_options.command == :write
+    return search if parsed_options.command == :search
+  end
```
jmtai: What happens if no command is passed in? It'd be nice if the app printed the help text in that case since otherwise how are you to know what commands are available?

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+
+  def initialize(dinosaurs: [], dino_files: nil)
+    @dinosaurs = dinosaurs
+    @dino_files = dino_files
+    load_dino_files
+  end
+
+  def load_dino_files
+    return if @dino_files.nil?
+    @dino_files.each { |dino_file| @dinosaurs.concat(dino_file.read) }
```
jmtai: Since this is a one-liner, you don't really need to call the local variable `dino_file`, since we know from the instance variable name that we are operating on dino_files. It could just be `file` and you wouldn't lose anything. What you have is perfectly fine, but it might help tighten up your code in other cases.

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+  end
+
+  def search_by_attr(dino_attr, desired_value)
+    filter = DinoFilter.new(dino_attr, desired_value)
+    filter.search(@dinosaurs)
+  end
+
+  def search_by_hash(hash)
```
jmtai: `search_by_hash` isn't a great name. What you're really doing is searching on multiple attributes, so why not call it that? Maybe something like `search_by_multiple_attrs`?

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+    hash.inject(@dinosaurs) do |dinos, attr_value_pair|
```
jmtai: We generally try to avoid using `inject` in most cases because it tends to be difficult to understand what's happening at a glance. Unless you're doing something like calculating a sum, it almost always violates the 5-second rule.

jmtai: In this case, I'd say it's ok, although using `each_with_object` might be a little clearer.

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+      filter = DinoFilter.new(attr_value_pair[0], attr_value_pair[1])
+      filter.search(dinos)
+    end
+  end
+
+  def write_dinosaurs_as_json_file(path)
+    path << ".json" unless path.include?(".json")
+    File.open(path, mode: "w") do |file|
+      file.write(JSON.generate(@dinosaurs))
```
jmtai: Note that since your Dinosaur class does not have a `to_h` method, the JSON file generated here probably isn't quite what you want. Ideally, you want something like this:
```ruby
[
  {
    "name": "DinoName1"
    "weight": "2000"
    etc...
  }
  {
    "name": "DinoName2"
    "weight": "4000"
    etc...
  }
]
```
Instead, since you've only got a `to_s` method for your Dinosaur class, you're getting this:
[
  "name: DinoName1, weight: 2000, ...",
  "name: DinoName2, weight: 4000, ...",
]
```
which is just an array of strings. If you were to parse this with the JSON parser, you wouldn't end up with a nice array of hashes with easily accessible dino attributes, it'll just be an array of strings and would require further parsing to extract each of the individual attributes.

File: dino_catalog/dinodex_catalog.rb [hunk 16]
```diff
@@ -0,0 +1,192 @@
...
+    else
+      search_using_block(dinosaurs, ->(a, b) { a == b })
+    end
+  end
+
+  def permissive_search?
+    "period" == @dino_attr
+  end
+
+  def search_using_block(dinosaurs, block)
```
jmtai: Just as we want to avoid using data types in variable names, we should try to avoid using parameter types in method names. You can use the Ruby standard library as a guide - methods that accept block parameters don't explicitly say `with_block`.

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+    dinos = dinosaurs.map do |dino|
```
jmtai: Using `select` instead of map here will clean this up a bit.

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+    dinos.compact
+  end
+end
+
+class Dinosaur
+  ATTRS = [:name, :period, :continent, :diet, :weight, :walking, :description]
+  attr_accessor(*ATTRS)
+
+  def initialize(options = {})
+    @name = options[:name] || "template"
```
jmtai: What does "template" signify?

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+    @description = options[:description]
+  end
+
+  def to_s
+    strings = []
+    ATTRS.each { |attr| strings << attr_to_s(attr) }
+    strings.compact.join(', ')
+  end
+
+  def attr_to_s(attr)
```
jmtai: Since this method is named `attr_to_s`, it's somewhat surprising behavior that it can also return nil. The convention would be for any method named `to_s` to return a string.

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+    value = method(attr).call
+    "#{attr.capitalize}: #{value}" unless value.nil?
+  end
+
+  def size
+    return if @weight.nil?
+    two_tons = 4000
```
jmtai: This should be declared as a constant.

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+    @weight.to_f > two_tons ? "big" : "small"
+  end
+
+  def carnivore?
+    return if @diet.nil?
```
jmtai: The convention for predicate methods is that they should return only `true` or `false`. If you break this convention, a future user could run into problems if they explicitly check for something like `carnivore? == false`.

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+    diet = @diet.to_s.downcase
+    %w(carnivore insectivore piscivore omnivore).include? diet
```
jmtai: Make this list a constant.

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+  end
+
+  def carnivore
+    carnivore?.to_s
+  end
+
+  def searchable_attrs
+    attr_symbols = ATTRS + [:size, :carnivore]
+    attr_symbols.map(&:id2name)
```
jmtai: I was going to comment that id2name isn't a great method name, but then realized it's a built-in. So I'll complain that the Ruby implementors picked a bad name. :smile:

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+      attr_value = dino.method(@dino_attr).call
+      downcase_attr_value = attr_value.to_s.downcase
+      dino if block.call(downcase_attr_value, @desired_value)
+    end
+    dinos.compact
+  end
+end
+
+class Dinosaur
+  ATTRS = [:name, :period, :continent, :diet, :weight, :walking, :description]
```
jmtai: Since your code depends on these built-in attributes to have specific values, you should consider using `freeze` to prevent any modification of this array.

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+    @continent  = options[:continent]
+    @diet = options[:diet]
+    @weight = options[:weight]
+    @walking = options[:walking]
+    @description = options[:description]
+  end
+
+  def to_s
+    strings = []
+    ATTRS.each { |attr| strings << attr_to_s(attr) }
```
jmtai: If you use `each_with_object`, you could collapse these two lines into a single line.

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+      downcase_attr_value = attr_value.to_s.downcase
+      dino if block.call(downcase_attr_value, @desired_value)
+    end
+    dinos.compact
+  end
+end
+
+class Dinosaur
+  ATTRS = [:name, :period, :continent, :diet, :weight, :walking, :description]
+  attr_accessor(*ATTRS)
```
jmtai: Alternately, instead of doing this, you could explicitly declare the accessors, then dynamically generate the `ATTR` array using `get_instance_variables`.

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+  def initialize(path, format)
+    @path = path
+    @reader = format_specific_reader(format)
+  end
+
+  def format_specific_reader(format)
+    return nil unless [:original, :p_b_africa].include?(format)
+    lambda do
+      converter_method = method("#{format}_format_hash_to_dinosaur")
+      csv_file_to_hashes.map do |dino_hash|
```
jmtai: The naming convention would be `to_h` for conversion to a hash.

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+  def csv_file_to_hashes
+    CSV::Converters[:blank_to_nil] = blank_to_nil_converter
+    options = { converters: :blank_to_nil, headers: true, return_headers: false,
+                header_converters: [:downcase, :symbol] }
+    CSV.open(@path, options) do |csv|
+      csv.to_a.map(&:to_hash)
+    end
+  end
+
+  def original_format_hash_to_dinosaur(dino_hash)
```
jmtai: There are only a few headers that need to be remapped, so why not do this using the CSV header_converters?

dino_catalog/dinodex_catalog.rb
```diff
@@ -0,0 +1,192 @@
...
+    dino_hash[:weight] = dino_hash[:weight_in_lbs]
+    Dinosaur.new(dino_hash)
+  end
+
+  def p_b_africa_format_hash_to_dinosaur(dino_hash)
+    dino_hash[:name] = dino_hash[:genus]
+    dino_hash[:continent] = "Africa"
+    return Dinosaur.new(dino_hash) unless dino_hash.key?(:carnivore)
```
jmtai: Can you think of a better way to structure this logic? Having an early return in the middle of a method isn't necessarily bad but in this case there's almost certainly a cleaner way to implement this.

jmtai:
Nice first pass! I like the use of the OptionParser for the command line interface. Read over the comments, refactor a bit (definitely separate your classes into individual files), and resubmit when ready.

rcaileff:
Thanks!  And thank you for the review.  I'll see about beating my code into
better shape early next week.  :)

-Rachel

dino_catalog/dinodex.rb
rcaileff: Thanks for the pointer; I hadn't learned about HEREDOCs yet.  I tried rewriting explain_value using it.  Unfortunately, by the time I've gotten the whitespace parsed so that the text displays the way I want, the code looks less clear to me.  In this case I prefer the original version.

  def explain_value
    text = <<-TEXT.gsub(/^ {6}/, '').gsub(/\n/, ' ')
        [VALUE] represents the desired value of a dinosaur attribute.
      Value is not case sensitive, but requires an otherwise exact
      match for all attributes except Period.
    TEXT
    parser.separator text
  end

dino_catalog/dinodex_catalog.rb
rcaileff: I'd prefer to keep ATTR attached to the class rather than to an instance.

dino_catalog/dinodex_catalog.rb
rcaileff: A template is a blank form with the shape/format of an object but not the content.  For example a template of a complaint letter might start "Dear <name>, I am writing in regard to <incident> that happened on <date>."

dino_catalog/dinodex.rb
rcaileff: An early iteration of the code had separate flags for the user to specify the search key and the search value.  It was annoying to use, so I removed the requirement.  It was a design trade-off; I decided that under the current circumstances, it made my life easier to use the even/odd logic in the code instead of the additional flags on the command line.

dino_catalog/dinodex_catalog.rb
jmtai: Sorry, I should have been more clear. Why is "template" necessary here? I don't see it being used anywhere else in your code.

dino_catalog/dinodex.rb
jmtai: Sorry, maybe I wasn't being clear enough. I'm not suggesting that the user should have to specify different flags for the key and value, just that key value pairs are separated by a special character. Something like
`key1,value1; key2,value2`
where the semicolon delineates key,value pairs. That way you don't need even/odd logic, you can just split on the semicolon and you'll have a nice list of individual key,value pairs.

dino_catalog/dinodex_catalog.rb
rcaileff: Ah!  There used to be a spot in the code that had problems if name was empty.  When I fixed that, I missed that I could remove the default name.  Thanks for spotting it.  :)

dino_catalog/dinodex_catalog.rb
rcaileff: I prefer to keep all of the format-specific processing separate from the generic processing.  Everything currently in csv_file_to_hashes needs to be done to every file that is read, no matter what format it is.  The format-specific changes happen in format-specific functions.

I did look more closely at header_converters.  I can't find a way to use converters + header_converters to handle the carnivore:yes/no to diet:carnivore/herbivore conversion in a single step.
I could convert carnivore to diet with a header_converter in csv_file_to_hashes (which is getting renamed in my next commit) then convert diet:yes to diet:carnivore and diet:no to diet:herbivore later.  I strongly prefer to keep the data in some sort of consistent state, so I don't want to do the header conversion and the data conversion in separate functions.
Besides, I really don't like the idea of diet:yes for a dinosaur;  I'm pretty sure dieting is a human concept.  ;)

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,82 @@
...
+  TWO_TONS = 4000
+
+  ATTRS = [:name, :period, :continent, :diet, :weight, :walking, :description]
+  ATTRS.freeze
+  attr_accessor(*ATTRS)
+
+  def initialize(options = {})
+    @name = options[:name]
+    @period = options[:period]
+    @continent  = options[:continent]
```
rcaileff: Ready for another round of reviewing!

-Rachel

PR #799, opened 2015-07-10
Data science

yuvalroth:
changes based on review


data_science/lib/data_manager.rb
```diff
@@ -0,0 +1,55 @@
...
+class DataManager
+  @sample_size = 0
+  attr_accessor :reader
+
+  def initialize(reader = nil)
+    load(reader) if reader
+  end
+
+  def load(reader)
+    @reader = reader
```
djkotowski: This may be a bit too much abstraction.  The following works just fine:
```ruby
def initialize(reader = nil)
  @reader = reader if reader
end
```
Even better, since `reader` defaults to `nil`, your initialize method could just be:
```ruby
def initialize(reader = nil)
  @reader = reader
end
```

data_science/lib/data_manager.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def sample_size(cohort_name = nil)
```
djkotowski: This class has too much responsibility.  You have a lot of methods that rely solely on a single cohort.  Why not just make a Cohort class, and extract much of this logic out to that?

djkotowski: That will also give you the benefit of efficiency.  At the moment, you call `cohort_conversion_size` many times, and each time it has to have the reader parse the information and filter by cohort.  Do the reading process _once_, and save that processed data in memory.

data_science/specs/data_manager_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+    end
+
+    context 'when initialized with reader it should not blow up' do
+      it 'should not be nil' do
+        expect(manager).not_to be_nil
+      end
+    end
+  end
+
+  describe '.load' do
```
djkotowski: This and the methods below should be preceded with a `#`, not a `.`.
# = instance method
. = class method

data_science/specs/data_manager_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      it 'should have a total conversion rate for cohort B' do
+        expect(manager.conversion_size("A")).to eq(47)
+      end
+    end
+  end
+
+  describe '.conversion_rate' do
+    context 'before loading a reader' do
+      it 'should raise an error' do
+        expect { DataManager.new.conversion_rate }.to raise_error ArgumentError
```
djkotowski: Wrap `ArgumentError` in parentheses

data_science/specs/data_manager_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      it 'B should have a conversion rate of 5.12%' do
+        expect(manager.conversion_rate("B")).to eq(5.12)
+      end
+    end
+  end
+
+  describe '.calculate_chi_square' do
+    context 'before loading a reader' do
+      it 'should raise an error' do
+        expect { DataManager.new.calc_chi_square }.to raise_error ArgumentError
```
djkotowski: Ditto above.

djkotowski:
Good work!  I'd look at extracting out some of the logic from DataManager.  It's doing an awful lot here.

It's also making calculations very inefficiently by reading the JSON data each time.  Just read it once and save the necessary summary values (number of conversions &amp; number of views).

Please refactor and resubmit when ready.

PR #800, opened 2015-07-13
Fixed robot file, ready to check

oharace:


robot_name/robot_name.rb
```diff
@@ -3,23 +3,34 @@ class NameCollisionError < RuntimeError; end
...
-    @@registry ||= []
+    @registry ||= []
     @name_generator = args[:name_generator]
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+    @name_generator ? @name = @name_generator.call : generate_name
```
djkotowski: Should be
```ruby
@name = @name_generator ? @name_generator.call : generate_name
```

robot_name/robot_name.rb
```diff
@@ -3,23 +3,34 @@ class NameCollisionError < RuntimeError; end
...
+    @registry << @name if valid_name?
+  end
+
+  def generate_name
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
```
djkotowski: The two procs above could be made into methods for clarity.

robot_name/robot_name.rb
```diff
@@ -3,23 +3,34 @@ class NameCollisionError < RuntimeError; end
...
+    @name = "#{generate_char.call}#{generate_char.call}"
+    @name << "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
```
djkotowski: You may also use the `\` operator to append lines.  Still, this may not be the clearest way to represent it.

robot_name/robot_name.rb
```diff
@@ -3,23 +3,34 @@ class NameCollisionError < RuntimeError; end
...
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def valid_name?
+    if !name_is_correct_format || name_in_registry
+      raise NameCollisionError, 'There was a problem generating the robot name!'
     end
+    true
+  end
```
djkotowski: A predicate method should only return true or false, not true or raise error.  Pick one paradigm here.

robot_name/robot_name.rb
```diff
@@ -3,23 +3,34 @@ class NameCollisionError < RuntimeError; end
...
+
+  def name_is_correct_format?
+    @name =~ /[[:alpha:]]{2}[[:digit:]]{3}/
```
djkotowski: This is acceptable, though if you find yourself wanting to ensure `true` or `false` are returned (not 'truthy' or 'falsy' values), you can use a double-bang (`!!`) to convert truthy->true & falsy->false.

robot_name/robot_name.rb
```diff
@@ -3,23 +3,34 @@ class NameCollisionError < RuntimeError; end
 class Robot
   attr_accessor :name
 
-  @@registry
-
   def initialize(args = {})
-    @@registry ||= []
+    @registry ||= []
```
djkotowski: So now you have an instance variable `@registry` that doesn't carry over between instances of Robot.

djkotowski:
Good first iteration.  I don't think the registry works, though.  Take a look at the comments, and fix and refactor where necessary.  Resubmit when ready.

PR #801, opened 2015-07-13
Dinodex

zsyed91:
Dinodex catalog. Command line script that uses sql style syntax for the custom query ability. Does not support or statements but only and statements. For example:

weight > 1000 and continent eq "North America"

### DIFF
--- a/dino_catalog/dinodex.rb
+++ b/dino_catalog/dinodex.rb
@@ -1,6 +1,7 @@
 require 'json'
 require 'table_print'
 require_relative 'dinosaur_loader'
+require_relative 'dinosaur_query'
 
 def csv_file
   csv_file = ARGV[0]
@@ -32,14 +33,15 @@ dinosaurs = DinosaurLoader.load(csv_file)
 
 input = ''
 until input.eql?('q')
-  putsq
+  puts
   print menu
   input = gets
   input.chomp!.downcase!
 
   case input
     when '1'
-      next
+      query = DinosaurQuery.new(dinosaurs)
+      tp query.process
     when '2'
       tp dinosaurs.select { |dinosaur| dinosaur.filter('walking', 'Biped') }
     when '3'
--- a/dino_catalog/dinodex.rb
+++ b/dino_catalog/dinodex.rb
@@ -1,37 +1,57 @@
-require_relative 'parser'
 require 'json'
+require 'table_print'
+require_relative 'dinosaur_loader'
 
-def show_menu
-  print "Enter any of the options below\n" \
-    "1  - Query dinodex\n" \
-    "2  - Output JSON\n" \
-    "q  - Quit\n" \
-    '> '
+def csv_file
+  csv_file = ARGV[0]
+  puts "test" if csv_file
+  return csv_file if csv_file
+
+  ask_user_for_csv_file
 end
 
-p 'Welcome to Dinodex (Gotta catch em all)'
-
-csv_file = ARGV[0]
-unless csv_file
+def ask_user_for_csv_file
   print "Enter the csv filename for the dinosaurs (ext optional)\n> "
   csv_file = gets
   csv_file.chomp!
 end
 
-parser = Parser.new(csv_file)
-dinosaurs = parser.parse
+p 'Welcome to Dinodex (Gotta catch em all)'
+
+menu = "Enter any of the options below\n" \
+    "\t1  - Query dinodex\n" \
+    "\t2  - Bipeds only\n" \
+    "\t3  - Carnivores only\n" \
+    "\t4  - Specific periods only\n" \
+    "\t5  - Only big (2 ton +)]\n" \
+    "\t9  - Output JSON\n" \
+    "\tq  - Quit\n" \
+    '> '
+
+dinosaurs = DinosaurLoader.load(csv_file)
 
 input = ''
 until input.eql?('q')
-  show_menu
+  putsq
+  print menu
   input = gets
   input.chomp!.downcase!
 
   case input
     when '1'
-      p dinosaurs
+      next
     when '2'
-      p JSON.generate(dinosaurs)
+      tp dinosaurs.select { |dinosaur| dinosaur.filter('walking', 'Biped') }
+    when '3'
+      tp dinosaurs.select { |dinosaur| dinosaur.filter('diet', 'Carnivore') }
+    when '4'
+      print "Enter period:\n> "
+      value = gets.chomp!
+      tp dinosaurs.select { |dinosaur| dinosaur.contains('period', value) }
+    when '5'
+      tp dinosaurs.select { |dinosaur| dinosaur.greater_than?('weight', 2000) }
+    when '9'
+      p JSON.generate(dinosaurs.map(&:to_hash))
     when 'q'
       next
     else
--- a/dino_catalog/dinodex.rb
+++ b/dino_catalog/dinodex.rb
@@ -1,0 +1,41 @@
+require_relative 'parser'
+require 'json'
+
+def show_menu
+  print "Enter any of the options below\n" \
+    "1  - Query dinodex\n" \
+    "2  - Output JSON\n" \
+    "q  - Quit\n" \
+    '> '
+end
+
+p 'Welcome to Dinodex (Gotta catch em all)'
+
+csv_file = ARGV[0]
+unless csv_file
+  print "Enter the csv filename for the dinosaurs (ext optional)\n> "
+  csv_file = gets
+  csv_file.chomp!
+end
+
+parser = Parser.new(csv_file)
+dinosaurs = parser.parse
+
+input = ''
+until input.eql?('q')
+  show_menu
+  input = gets
+  input.chomp!.downcase!
+
+  case input
+    when '1'
+      p dinosaurs
+    when '2'
+      p JSON.generate(dinosaurs)
+    when 'q'
+      next
+    else
+      p 'Invalid input...'
+      next
+  end
+end
--- a/dino_catalog/dinosaur.rb
+++ b/dino_catalog/dinosaur.rb
@@ -17,9 +17,9 @@ class Dinosaur < Filterable
     end
   end
 
-  def diet=(value)
-    value = 'Carnivore' if value =~ /Insect|Pisci|Yes/
-    write_attribute(:diet, value)
+  def diet
+    return 'Carnivore' if @diet =~ /Insect|Pisci|Yes/
+    @diet
   end
 
   def to_hash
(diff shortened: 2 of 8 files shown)


dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,63 @@
+require 'json'
+require 'table_print'
+require_relative 'dinosaur_loader'
+require_relative 'dinosaur_query'
+
+def csv_file
+  csv_file = ARGV[0]
+  puts "test" if csv_file
```
zsyed1: Remove line 8 :smile:

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,63 @@
...
+  return csv_file if csv_file
```
zsyed1: You can combine line 9 into line 7:
```
return ARGV[0] if ARGV[0]
```
Though I like the fact that put in a descriptive name.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,63 @@
...
+
+menu = "Enter any of the options below\n" \
+    "\t1  - Query dinodex\n" \
+    "\t2  - Bipeds only\n" \
+    "\t3  - Carnivores only\n" \
+    "\t4  - Specific periods only\n" \
+    "\t5  - Only big (2 ton +)]\n" \
+    "\t9  - Output JSON\n" \
+    "\tq  - Quit\n" \
+    '> '
```
zsyed1: Probably want to put lines 22 - 30 into a method:
```
def menu
["\t1 - Query dinodex",
 "\t2 - Bipeds only",
 ...
 '> '].join("\n")
end
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,63 @@
...
+      tp dinosaurs.select { |dinosaur| dinosaur.greater_than?('weight', 2000) }
+    when '9'
+      p JSON.generate(dinosaurs.map(&:to_hash))
+    when 'q'
+      next
+    else
+      p 'Invalid input...'
+      next
+  end
+end
```
zsyed1: Using a command word is a nicer way of specifying menu options. Instead of `3` you can use `meat` or `carnivores` for example.

Also the Ruby community dislikes `case` statements (I don't agree with this btw). The _preferred_ way to do this is to set up a map from `command` to `action` where the latter is a reference to a method with performs the task. We can discuss this in crapchat if you'd like.

So take option 2 for example. You could move the code:
```
  tp dinosaurs.select { |dinosaur| dinosaur.filter('walking', 'Biped') } 
```
into a method on `DinosaurLoader` called `def walking`. And the you could bind that method to the option `2` in a map:
```
actions = {}
...
...
actions['2'] = &dinosaurs.walking
```

djkotowski:
Nice to see @zsyed1 reviewing @zsyed91

dino_catalog/parser.rb
```diff
@@ -0,0 +1,35 @@
...
+class Parser
+  attr_accessor :headers, :content, :file
+
+  def initialize(filepath)
+    @file = filepath
+    @file += '.csv' unless filepath =~ /\.csv$/
+    @content = []
+  end
+
+  def parse
```
zsyed1: Nice - This is set up to be called sequentially to parse multiple files

dino_catalog/dinosaur_query.rb
```diff
@@ -0,0 +1,81 @@
+class DinosaurQuery
```
zsyed1: I really like this SQL parsing component!

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,39 @@
...
+      instance_variable_set("@#{key}", value) unless value.nil?
+    end
+  end
+
+  def diet
+    return 'Carnivore' if @diet =~ /Insect|Pisci|Yes/
+    @diet
+  end
+
+  def to_hash
```
zsyed1: You can shorten this name to `to_h`
Also, if you wanted to practice the `reduce` method, you can do it here. Any time you create an empty container, populate it within an `each` block, then return the container, there is probably a shorter way to do it with `each_with_object` or `map`.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,39 @@
...
+    instance_variables.each do |variable|
+      attributes[variable.to_s.delete('@')] = instance_variable_get(variable)
+    end
+    attributes
+  end
+
+  private
+
+  def map_keys(key)
+    return KEY_MAPS[key.to_sym] if KEY_MAPS.member?(key.to_sym)
```
zsyed1: I think there is a method `key?` that is more commonly used than `member?`

dino_catalog/dinosaur_loader.rb
```diff
@@ -0,0 +1,17 @@
...
+require_relative 'parser'
+
+class DinosaurLoader
+  def self.load(file)
+    dinosaurs = []
+    parse_file(file).each do |dino|
+      dinosaurs << Dinosaur.new(dino)
+    end
+    dinosaurs
+  end
```
zsyed1: You can use `map` - see comment above about containers and `reduce`. One line implementation of lines 6 - 10:

```
parse_file(file).map { |dino| Dinosaur.new(dino) }
```

dino_catalog/dinosaur_loader.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def self.parse_file(file)
+    parser = Parser.new(file)
+    parser.parse
+  end
```
zsyed1: Lines 14, 15: You can shrink these two lines into one: `Parser.new(file).parse`

dino_catalog/dinosaur_query.rb
```diff
@@ -0,0 +1,81 @@
...
+      run_criteria if index % 4 == 3
+    end
+  end
+
+  def run_criteria
+    results = run_query_part
+    results.each do |dinosaur|
+      filtered << dinosaur unless filtered.include?(dinosaur)
+    end
+  end
```
zsyed1: Lines 51 - 53: You could also do:
```
filtered += run_query_part
filtered.uniq!
```

dino_catalog/dinosaur_query.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  def run_query_part
```
djkotowski: This definitely should be split into smaller methods.

dino_catalog/parser.rb
```diff
@@ -0,0 +1,35 @@
...
+    @headers = lines.shift.each(&:downcase!)
+    lines
+  end
+
+  def load_content(lines)
+    lines.each do |line|
+      object = {}
+      line.each_with_index do |value, index|
+        object[headers[index]] = value
+      end
```
zsyed1: Lines 28 - 31: I thought the CSV gem had a way of doing this for you?

dino_catalog/dinosaur_query.rb
```diff
@@ -0,0 +1,81 @@
...
+          d.greater_than?(criteria[:key], criteria[:value])
+        end
+      when '<'
+        dinosaurs.select { |d| d.less_than?(criteria[:key], criteria[:value]) }
+      when '='
+        dinosaurs.select { |d| d.filter(criteria[:key], criteria[:value]) }
+      when 'eq'
+        dinosaurs.select { |d| d.contains(criteria[:key], criteria[:value]) }
+      when '!='
+        dinosaurs.select { |d| !d.filter(criteria[:key], criteria[:value]) }
```
djkotowski: Instead of negating a condition in a `select`, make it a `reject` call.

dino_catalog/parser.rb
```diff
@@ -0,0 +1,35 @@
...
+    content
+  end
+
+  private
+
+  def load_lines_and_process_headers
+    lines = CSV.read(file)
+    @headers = lines.shift.each(&:downcase!)
+    lines
+  end
```
zsyed1: Line 22: Same here - I though CSV could do this for you. You've done it in one line of code so I guess it doesn't really matter,

dino_catalog/parser.rb
```diff
@@ -0,0 +1,35 @@
...
+  def load_content(lines)
+    lines.each do |line|
+      object = {}
+      line.each_with_index do |value, index|
+        object[headers[index]] = value
+      end
+      content << object
+    end
+  end
+end
```
zsyed1: Very nice work! - I don't think this needs a resubmission.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,39 @@
+require_relative('filterable')
+class Dinosaur < Filterable
+  attr_accessor :name, :period, :continent, :diet, :weight,
+                :walking, :description
```
djkotowski: I definitely prefer this syntax over Hound's two-space indent on accessor lists.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,63 @@
+require 'json'
+require 'table_print'
+require_relative 'dinosaur_loader'
+require_relative 'dinosaur_query'
+
+def csv_file
```
djkotowski: To clarify, perhaps this should be `csv_filename`

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,63 @@
...
+  csv_file = ARGV[0]
+  puts "test" if csv_file
+  return csv_file if csv_file
+
+  ask_user_for_csv_file
+end
+
+def ask_user_for_csv_file
+  print "Enter the csv filename for the dinosaurs (ext optional)\n> "
+  csv_file = gets
```
djkotowski: Why not just
```ruby
csv_file = gets.chomp!
```

dino_catalog/dinodex.rb
djkotowski: Or make it a constant.

dino_catalog/dinodex.rb
djkotowski: Yep, this file is doing too much.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,39 @@
+require_relative('filterable')
```
djkotowski: Add a blank line between require and class definition.

dino_catalog/filterable.rb
```diff
@@ -0,0 +1,21 @@
+class Filterable
```
djkotowski: This may serve better as a mixin than a class.

dino_catalog/parser.rb
```diff
@@ -0,0 +1,35 @@
+require "CSV"
+
+class Parser
+  attr_accessor :headers, :content, :file
+
+  def initialize(filepath)
+    @file = filepath
+    @file += '.csv' unless filepath =~ /\.csv$/
```
djkotowski: What if the file doesn't include the `.csv` extension?

djkotowski:
Nice work!  This could use a bit of refactoring, though.  Simplify the methods and follow the single responsibility principle.  Take a look at the comments, refactor and resubmit when ready.

PR #802, opened 2015-07-13
Dino catalog

jklemen:
Updated based on feedback.


dino_catalog/dinotest.rb
```diff
@@ -0,0 +1,102 @@
...
+  def test2
+    case_str = "Grab all the dinosaurs that were carnivores"
+    case_str << " (fish and insects count)"
+
+    query = dinodex.not_equal('diet', 'herbivore')
+    execute(case_str, query)
+  end
+
+  def test3
+    case_str =  "Grab dinosaurs for specific periods (no need to"
```
jklemen: oops >_<

dino_catalog/dino.rb
```diff
@@ -0,0 +1,57 @@
+class Dino
+  ATTRIBUTES = %w(name period continent diet weight walking description)
+
+  attr_accessor :name, :period, :continent
+  attr_accessor :diet, :weight, :walking, :description
```
johnmcconnell: You could just
```ruby
attr_accessor *ATTRIBUTES
```

johnmcconnell:
Looks good :+1:

dino_catalog/dino.rb
jklemen: No. Way. Never knew that.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  def less_than(field, value)
+    Dinodex.new(lt(field, value))
+  end
+
+  def greater_than(field, value)
+    slt = select_less_than(field, value)
+    seq = select_equal(field, value)
+    Dinodex.new(@dinos - slt - seq)
```
djkotowski: This is not a very efficient way of finding greater than.  You're filtering the collection twice, then removing those subsets from the whole.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,89 @@
...
+  def sort(field)
+    sorted = @dinos.sort do |a, b|
+      a.compare(field, b.send(field))
+    end
+    Dinodex.new(sorted)
+  end
+
+  def to_json
+    {
+      'dinos' => @dinos.map { |dino| Hash[dino.properties] }
```
djkotowski: Shouldn't Dino handle the conversion of its properties to a Hash?

dino_catalog/dino.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def properties
+    ATTRIBUTES.map { |attr| [attr, send(attr)] }
+  end
+
+  def dino_facts
+    properties.each do |attr, val|
+      fattr = "#{attr.capitalize}:".ljust(20)
+      fval = val.nil? ? '---' : val
+      puts "#{fattr}#{fval}"
```
djkotowski: Dino should not be reliant on the app using it being a console app.  Separate your concerns.  If you want to return a formatted string, that would be clearer.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,57 @@
...
+  def equal?(field, value)
+    compare(field, value) == 0
+  end
+
+  def like?(field, value)
+    myval = send(field)
+    if myval.nil? || value.nil?
+      myval.nil? && value.nil?
+    else
+      myval.downcase =~ /#{value.downcase}/
```
djkotowski: While not necessary, if you want to ensure this method returns true or false, rather than truthy or falsy values, use a double-bang (!!) to negate the expression twice.

dino_catalog/dinodex.rb
djkotowski: If you include methods `greater_than?`, `greater_than_or_equal?`, etc. in Dino, then these methods become much simpler.

dino_catalog/dinotest.rb
```diff
@@ -0,0 +1,102 @@
+require './dinodex.rb'
+require './dino.rb'
+require './dinoparse.rb'
+
+class Dinotest
```
djkotowski: Interesting testing solution.  You'll be looking at RSpec pretty heavily in the testing unit.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,57 @@
...
+  attr_accessor :diet, :weight, :walking, :description
+
+  def initialize(data)
+    data.keys.each do |attr|
+      send("#{attr}=", data[attr])
+    end
+  end
+
+  def properties
+    ATTRIBUTES.map { |attr| [attr, send(attr)] }
```
djkotowski: Does it make sense to return this as an array of tuple arrays?  For key/value combinations, a Hash is typically better.

dino_catalog/dinoparse.rb
```diff
@@ -0,0 +1,36 @@
...
+require './dino.rb'
+
+class Dinoparse
+  attr_reader :dinos
+
+  def initialize(dinofile)
+    dino_data = parse(dinofile)
+    @dinos = dino_data.map { |dino_dat| Dino.new(dino_dat) }
+  end
+
```
djkotowski: Everything below this point is only used locally.  They should be private methods.

djkotowski:
Good iteration.  Take a look at the comments and refactor.  Resubmit when ready.

Work on making your code more readable.  Also consider where responsibility belongs between your classes.  Let me know if you have any questions.

dino_catalog/dinodex.rb
jklemen: Well, I had a method that did a single pass before. So I guess I'll just change it back.

dino_catalog/dinodex.rb
jklemen: Yeah.

dino_catalog/dino.rb
jklemen: Makes sense.

dino_catalog/dinotest.rb
jklemen: This isn't a test, I just exhibited that what I wrote was capable of answering the example queries posed in the README.

dino_catalog/dino.rb
jklemen: I agree, an oversight on my part.

dino_catalog/dinoparse.rb
jklemen: Yes.

djkotowski:
I'd also like to see a CLI if you can develop it.  Some command interface for users to interact with your library.

PR #806, opened 2015-07-13
Dino catalog

lmenova:
run main.rb.


jmtai:
Make sure you create your exercise branches off of your `master` branch. This branch contains your changes from your arrowhead exercise as well.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,48 @@
...
+  attr_accessor :dino_data
+
+  def initialize(dinodata = [])
+    @dino_data = dinodata
+  end
+
+  def populate_data
+    dl = DinoLoader.new
+    data = dl.load_dino_data
+    data.each { |x| @dino_data << Dinosaur.new(x) }
```
jmtai: In other languages, it's common to use single letter variable names but in Ruby we try to make our code as expressive as possible. In this case, each element corresponds to a row of the loaded data, so why not call it that?

jmtai: What you're doing here is transforming, or **mapping** each row of dino data to an instance of the Dinosaur class:
```ruby
@dino_data = data.map { |row| Dinosaur.new(row) }
```

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def show_data
+    @dino_data.each { |dino| dino }
```
jmtai: What's the purpose of this operation? It's named `show_data` but it's not outputting anything to the console.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def add_to_catalog(data)
```
jmtai: Why is this method defined if it's never used?

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,48 @@
...
+    @dino_data << data
+  end
+
+  def search_by_trait(trait, criteria = nil)
+    trait = trait.downcase
+    criteria = criteria.downcase unless criteria.nil?
+    filter_data = @dino_data.select do |dino|
+      next if dino.public_send(trait).nil? || dino.public_send(trait).empty?
+      filter_by_criteria(dino.public_send(trait), criteria)
```
jmtai: Note that for a trait that exists, you are calling `dino.public_send(trait)` **three** times. For something simply like this, it's not a big deal, but if the method call were an expensive one (i.e. long-running or computationally intensive), this would be a waste of resources, since you only need to compute the result once.

In this case, although it would add an extra line of code, storing the result in a temporary variable would be much more efficient.
```ruby
result = dino.public_send(trait)
next if result.nil? || result.empty?
filter_by_criteria(result, criteria)
```

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def show_data
+    @dino_data.each { |dino| dino }
+  end
+
+  def add_to_catalog(data)
+    @dino_data << data
+  end
+
+  def search_by_trait(trait, criteria = nil)
```
jmtai: Does it make sense to search by trait if the criteria is nil? Should this case even be allowed?

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,48 @@
...
+    trait = trait.downcase
+    criteria = criteria.downcase unless criteria.nil?
+    filter_data = @dino_data.select do |dino|
+      next if dino.public_send(trait).nil? || dino.public_send(trait).empty?
+      filter_by_criteria(dino.public_send(trait), criteria)
+    end
+    DinoCatalog.new(filter_data)
+  end
+
+  def filter_by_criteria(dino, criteria)
```
jmtai: Naming is hard and often times subjective, but here your choice of parameter names is obscuring the actual function of this method. `dino` makes it seem that a `Dinosaur` instance is being passed into this method, but when we look at the code, we see `dino.downcase`, which wouldn't make any sense if dino was indeed a `Dinosaur` instance. Using your existing terminology, this is actually a `dino_trait`, so it should be named appropriately.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,48 @@
...
+    if criteria.nil?
+      dino.downcase
+    else
+      dino.downcase.include?(criteria)
+    end
+  end
+
+  def search_by_size(size)
+    filter_data = @dino_data.select { |dino| dino.public_send("#{size}?") }
```
jmtai: What if I pass in 'Big' or 'Small' for the `size`?

dino_catalog/dino_loader.rb
```diff
@@ -0,0 +1,45 @@
+require 'csv'
+
+class DinoLoader
+  attr_accessor :dino_data, :path, :header_mapper
+  attr_reader :catalog
```
jmtai: What is the purpose of these? Are they leftovers from a previous implementation?

dino_catalog/dino_loader.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def initialize
+    @csv_table = []
+  end
+
+  def load_dino_data
+    Dir.glob("data/*.csv").each do |file|
+      data = CSV.read(file, headers: true, header_converters: header_converters,
+        converters: [diet_field_converter, weight_field_converter]
+                     )
```
jmtai: The formatting here makes the code hard to read. Aligning the parameters of a method call helps improve readability:
```ruby
data = CSV.read(file, headers: true, header_converters: header_converters,
                converters: [diet_field_converter, weight_field_converter])
```

dino_catalog/dino_loader.rb
```diff
@@ -0,0 +1,45 @@
...
+class DinoLoader
+  attr_accessor :dino_data, :path, :header_mapper
+  attr_reader :catalog
+
+  def initialize
+    @csv_table = []
+  end
+
+  def load_dino_data
+    Dir.glob("data/*.csv").each do |file|
```
jmtai: Nice way to handle an arbitrary number of data files.

dino_catalog/dino_loader.rb
```diff
@@ -0,0 +1,45 @@
...
+    @csv_table
+  end
+
+  private
+
+  def header_converters
+    lambda do |header_value|
+      header_value.gsub!('genus', 'name')
+      header_value.gsub!('weight_in_lbs', 'weight')
+      header_value.gsub!('carnivore', 'diet')
```
jmtai: Note that you are evaluating three different regexes for **each** header_value. Regexes are slow, so this is not super efficient. For a small data set like this, it won't be noticeable but this is something to be careful of in the future.

dino_catalog/dino_loader.rb
```diff
@@ -0,0 +1,45 @@
...
+      header_value.downcase.to_sym
+    end
+  end
+
+  def diet_field_converter
+    lambda do |value, fields|
+      return value unless (fields.header == :diet) && (value =~ /^(Yes|No)$/)
```
jmtai: What if value is 'yes' or 'no'?

dino_catalog/dino_loader.rb
```diff
@@ -0,0 +1,45 @@
...
+      value == 'Yes' ? 'Carnivore' : 'Herbivore'
+    end
+  end
+
+  def weight_field_converter
+    lambda do |value, fields|
+      return value unless (fields.header == :weight)
+      value.to_i
```
jmtai: If the weight field is `nil`, `to_i` will return 0. Is this what you want? Should dinosaurs without any defined weight be given a weight of 0 pounds?

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,53 @@
+class Dinosaur
+  attr_accessor :name, :period, :weight, :continent, :diet, :walking,
+    :description, :trait, :dino_name
```
jmtai: Hound likes this formatting but the convention here is to align the symbol names:
```ruby
attr_accessor :name, :period, :weight, :continent, :diet, :walking,
              :description, :trait, :dino_name
```

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  CARNIVORE_DIET = %w(carnivore insectivore piscivore)
+  DINO_TRAITS = %w(name period continent diet weight walking description size)
```
jmtai: Nice use of constants.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+    @output = ''
+  end
+
+  def size
+    big? ? 'Big' : 'Small'
+  end
+
+  def big?
+    weight.to_i > 4000
```
jmtai: We want to avoid having magic numbers embedded in our code. This number should be defined as a constant and given a descriptive name (e.g. `TWO_TONS`)

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def continent
+    if @continent.nil? || @continent.empty?
```
jmtai: `blank?(@continent)`
You've defined a nice convenience method to perform this check, why not use it?

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,53 @@
...
+      @continent = 'Africa'
+    else
+      @continent
+    end
+  end
+
+  def diet
+    if CARNIVORE_DIET.include? @diet.downcase
```
jmtai: Wrap method parameters is parentheses.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+  end
+
+  def to_s
+    @output = "*" * 80
+    DINO_TRAITS.each { |x| retrieve_trait(x) }
+    @output
+  end
+
+  def retrieve_trait(dino_trait)
```
jmtai: `retrieve_trait` sounds like it is just returning the value of a particular trait, but it's actually returning a formatted string containing the name of the trait as well as its value. Perhaps there's a more suitable name for this method?

jmtai:
Nice first pass! Your code is clean and easy to understand. What we'd really like for this exercise is a simple command line interface that allows a user to interact with the app. Take a look at the comments, refactor a bit (and add an interactive interface - it doesn't have to be super fancy) and resubmit when ready.

PR #807, opened 2015-07-13
Dinodex V2

zsyed91:
Made the changes per peer reviewers.


dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,55 @@
+require 'table_print'
+require_relative 'dinosaur_loader'
+
+MENU =
```
jmtai: Since `MENU` just contains static text, you could also use Ruby's HEREDOC support:
```ruby
MENU = <<TEXT
query       - Query dinodex
bipeds      - Bipeds only
carnivores  - Carnivores only
periods     - Specific periods only
big         - Only big (2 tons+)
json        - Output JSON
quit        - Quit
>
TEXT
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,55 @@
...
+   'bipeds      - Bipeds only',
+   'carnivores  - Carnivores only',
+   'periods     - Specific periods only',
+   'big         - Only big (2 tons+)',
+   'json        - Output JSON',
+   'quit        - Quit',
+   '> ',
+  ].join("\n")
+
+def csv_filename
```
jmtai: This method is returning a filename provided as a command line parameter or prompting a user to interactively enter a filename. If you were reading someone else's code and saw a method named `csv_filename`, would you expect it to be reading input from the console or would you expect it to simply return a filename?

Naming is tricky but we definitely want to choose method names that accurately describe what the method is actually doing.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,55 @@
...
+  ask_user_for_csv_filename
+end
+
+def ask_user_for_csv_filename
+  print "Enter the csv filename for the dinosaurs (ext optional)\n> "
+  gets.chomp!
+end
+
+p 'Welcome to Dinodex (Gotta catch em all)'
+dinosaurs = DinosaurLoader.new(csv_filename)
```
jmtai: A big problem with this implementation is that you are only loading one CSV file at a time. That means that you aren't able to query all available dinosaurs.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,55 @@
...
+
+actions = {}
+actions['query'] = dinosaurs.method(:query)
+actions['bipeds'] = dinosaurs.method(:walking)
+actions['carnivores'] = dinosaurs.method(:carnivore)
+actions['periods'] = dinosaurs.method(:period)
+actions['big'] = dinosaurs.method(:big)
+actions['json'] = dinosaurs.method(:json)
```
jmtai: You can clean this up a bit by just defining the hash explicitly:
```ruby
actions = {
  'query' => dinosaurs.method(:query),
  'bipeds' => dinosaurs.method(:walking),
  'carnivores' => dinosaurs.method(:carnivore),
  'periods' => dinosaurs.method(:period),
  'big' => dinosaurs.method(:big),
  'json' => dinosaurs.method(:json)
}
```

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,39 @@
+require_relative('filterable')
```
jmtai: The parens aren't necessary here.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,39 @@
...
+  def initialize(args)
+    args.each do |key, value|
+      key = map_keys(key)
+      next unless self.respond_to?(key)
+      instance_variable_set("@#{key}", value) unless value.nil?
+    end
+  end
+
+  def diet
+    return 'Carnivore' if @diet =~ /Insect|Pisci|Yes/
```
jmtai: What happens if the file contains lowercase values?

dino_catalog/dinosaur_loader.rb
```diff
@@ -0,0 +1,49 @@
...
+
+class DinosaurLoader
+  attr_accessor :dinosaurs
+
+  def initialize(file)
+    @dinosaurs = load(file)
+  end
+
+  def walking
+    dinosaurs.select { |d| d.filter('walking', 'Biped') }
```
jmtai: Should your search methods be case-sensitive?

jmtai: The naming of this method seems a bit misleading. What you are returning is a list of dinosaurs that are **bipeds**. Perhaps it'd be more accurate to name this `bipeds`?

dino_catalog/dinosaur_loader.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def carnivore
```
jmtai: Again, here we are returning a list of (possibly) many dinosaurs. `carnivores` would be more descriptive.

dino_catalog/dinosaur_loader.rb
```diff
@@ -0,0 +1,49 @@
+require_relative 'dinosaur'
+require_relative 'parser'
+require_relative 'dinosaur_query'
+require 'json'
+
+class DinosaurLoader
```
jmtai: Is this class name accurate? This class loads, parses, and queries dinosaurs. `DinosaurLoader` seems to imply that it only does the first of that list of functions.

dino_catalog/dinosaur_loader.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def walking
+    dinosaurs.select { |d| d.filter('walking', 'Biped') }
+  end
+
+  def carnivore
+    dinosaurs.select { |d| d.filter('diet', 'Carnivore') }
+  end
+
+  def period
```
jmtai: This method is both reading user input and filtering the list of dinosaurs based on that user input. Ideally, we want our methods to be single responsibility, so this should probably be split into two separate methods (and named accordingly).

dino_catalog/dinosaur_loader.rb
```diff
@@ -0,0 +1,49 @@
...
+    print "Enter period:\n> "
+    value = gets.chomp!
+    dinosaurs.select { |d| d.contains('period', value) }
+  end
+
+  def big
+    dinosaurs.select { |d| d.greater_than?('weight', 2000) }
```
jmtai: Hmmm... a single ton is 2000 pounds. The definition of `big` in the requirements is `> 2 tons`.

Also, we definitely want to avoid having magic numbers embedded in our code. The number here should be defined as a constant and given a descriptive name. (e.g. ONE_TON_IN_POUNDS = 2000)

dino_catalog/dinosaur_loader.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def query
+    query = DinosaurQuery.new(dinosaurs)
+    query.process
+  end
+
+  private
+
+  def load(file)
+    parse_file(file).map { |dinosaur| Dinosaur.new(dinosaur) }
```
jmtai: The naming of your local variable here is a bit confusing. `|dinosaur|` implies that what is being returned from `parse_file` are `Dinosaur` objects, but the body of the map block is
```ruby
Dinosaur.new(dinosaur)
```

dino_catalog/dinosaur_query.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  private
+
+  def normalize_query
+    query.squeeze!(' ')
+  end
+
+  def filter_criteria
+    print "Enter your query (sql style)\n> "
+    @query = gets.chomp!
```
jmtai: Much like the `csv_filename` method in your `dinodex.rb` file, this method has multiple responsibilities.

Also, this prompt assumes the user knows what `sql style` querying means. An example would be much more useful.

dino_catalog/dinosaur_query.rb
```diff
@@ -0,0 +1,79 @@
...
+    normalize_query
+    process_query
+  end
+
+  def process_query
+    parts = query.split(/\s(?=(?:[^"]|"[^"]*")*$)/)
```
jmtai: Regexes are great for cramming a lot of functionality into a single line. Not so great for readability and maintainability. This is a case where a comment would actually be useful to explain what this regex is doing. This helps in two ways:
1. It helps someone unfamiliar with the code understand the logic without having to parse the regex.
2. If a bug is discovered later, helps whoever is debugging understand what the original author actually intended.

dino_catalog/dinosaur_query.rb
```diff
@@ -0,0 +1,79 @@
...
+    return if parts.length < 3
+    return run_single_query(parts) if parts.length == 3
+    run_multiple_queries(parts)
+  end
+
+  def run_single_query(parts)
+    criteria[:key] = parts[0]
+    criteria[:filter] = parts[1]
+    criteria[:value] = parts[2].gsub(/\"/, '')
```
jmtai: I prefer this form to avoid having to repeat the `criteria[]` references:
```ruby
self.criteria = {
  key: parts[0],
  filter: parts[1],
  value: parts[2].gsub(/\"/, '')
}

File: dino_catalog/filterable.rb [hunk 16]
```diff
@@ -0,0 +1,21 @@
+module Filterable
+  def filter(key, value)
+    instance_variable_get("@#{key}") == value
+  end
+
+  def greater_than?(key, value)
+    instance_variable_get("@#{key}").to_i >= value.to_i
+  end
+
+  def less_than(key, value)
```
jmtai: This is missing the trailing `?`

dino_catalog/filterable.rb
```diff
@@ -0,0 +1,21 @@
+module Filterable
+  def filter(key, value)
```
jmtai: `filter` is not a great name for this method. It's checking for equality so why is it called `filter`?

dino_catalog/dinosaur_query.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  def run_multiple_queries(parts)
+    parts.each_with_index do |part, index|
+      build_criteria(part, index)
+      run_criteria if index % 4 == 3
+    end
+  end
+
+  def run_criteria
+    @filtered += run_query_part
```
jmtai: This doesn't do quite what I'd expect. Say I want all dinosaurs that are from Europe **AND** weigh more than 3000 pounds. By simply adding the results from the two filtering operations, I'll get the union of all dinosaurs from Europe plus all the dinosaurs that weigh more than 3000 pounds. Instead, what I really want is the intersection of those two groups.

dino_catalog/parser.rb
```diff
@@ -0,0 +1,34 @@
...
+    lines = load_lines_and_process_headers
+    load_content(lines)
+    content
+  end
+
+  private
+
+  def load_lines_and_process_headers
+    lines = CSV.read(file)
+    @headers = lines.shift.each(&:downcase!)
```
jmtai: This can actually be done automatically by the CSV class. There is also built-in support for doing header and data conversions using user-supplied lambdas.

jmtai:
Nice job! Your code is clean and easy to read. Read over the comments, refactor a bit, and resubmit when ready. In particular, make sure your dinodex loads dinosaurs from all the CSV files so a user can query from the entire list of dinosaurs.

zsyed91:
Ready to review again

PR #808, opened 2015-07-14
Overlord v1

Tornquist:
Complete Overlord Project.  

By default server needs to run at localhost:4567.  Client can be run at any port.  Client is looking for server at location specific at the top of public/overlord.js


overlord/overlord.rb
```diff
@@ -1,14 +1,48 @@
...
+
+# Temporary Debug Method
+put '/bomb' do
+  bomb = Bomb.new
+  "new bomb"
+end
+
+def try_method(action, success)
+  action.()
+  success.()
```
Tornquist: I was unable to find an alternative to this.  I'd be interested in learning if there is one.

overlord/overlord.rb
```diff
@@ -1,14 +1,48 @@
...
+end
+
+post '/set_activation_key' do
+  try_method(-> { bomb.activation_key = params[:key] },
+             -> { bomb.activation_key })
+end
+
+post '/set_deactivation_key' do
+  try_method(-> { bomb.deactivation_key = params[:key] },
+             -> { bomb.deactivation_key })
```
Tornquist: They look pretty aligned to me.  Does anyone know what hound is looking for?

jmmastey:
Hey Nathan, I see nobody reviewed this, so I figured I would. Testing being so near to my heart.

overlord/client/features/custom.feature
```diff
@@ -0,0 +1,23 @@
+Feature:
+  The bomb will respond to custom settings and will ignore the defaults
+
+  Background:
+    Given the user is on the home page
+    And there is a new bomb
+    And the user submits activation code 5567
+    And the user submits deactivation code 0943
+    And the user presses boot bomb
```
jmmastey: This makes for fewer step definitions, but make the test hard to read. Better:

`Background: Given there is a booted bomb with codes 5567 and 0943`

overlord/client/features/custom.feature
```diff
@@ -0,0 +1,23 @@
...
+
+  Scenario:
+    Then submit code 1234 is invalid
+    And the bomb is inactive
```
jmmastey: This is odd without a When.

```
Scenario: Invalid boot code
  When the user submits code 1234
  Then the bomb is inactive
  And I see a message that the bomb failed to boot
```

overlord/client/features/default.feature
```diff
@@ -0,0 +1,56 @@
+Feature:
+  Default Bomb Interactions
```
jmmastey: Traditionally you should start features with a preamble:

```
As an evil overlord
I want a way to disable the bomb while foiling interlopers
So that only I can turn off the bomb.
```

overlord/client/features/default.feature
```diff
@@ -0,0 +1,56 @@
...
+    Then the deactivation code fails
+
+  Scenario:
+    When the user submits activation code 5432
+    Then the activation code fails
+
+  Scenario:
+    When the user submits code 1234
+    Then the bomb is active
+    And the time is 2:00
```
jmmastey: What does `And the time is 2:00` mean? Something more literal like `And the timer displays 2:00` would be better.

overlord/client/features/default.feature
```diff
@@ -0,0 +1,56 @@
...
+
+  Scenario:
+    When the user submits code 1234
+    And the user submits code 0000
+    Then the bomb is inactive
```
jmmastey:
```
Scenario: Submitting the correct deactivation code
  Given the bomb is active
  When I submit the code 0000
  Then the bomb is inactive
```

The scenario description is important too :)

overlord/client/features/default.feature
```diff
@@ -0,0 +1,56 @@
...
+    And the user submits code 0391
+    And the user submits code 1236
+    Then the bomb is active
+
+  Scenario: 2 misses and a correct code defuses it
+    When the user submits code 1234
+    And the user submits code 0391
+    And the user submits code 1236
+    And the user submits code 0000
+    Then the bomb is inactive
```
jmmastey: One option you don't necessarily have to take, but might want to in other cases where the setup is expensive, is to combine a few of these cases:

```
Scenario: Disarming the bomb after consecutive failed attmepts
  Given the bomb is active
  When I submit the code 4444
  And I submit the code 5555
  Then the bomb is still active
  When I submit the code 0000
  Then the bomb is inactive
```

This is not super common but it works well in some cases. Also, I figured out why the first step feels better as a `Given`. Given is for preconditions, When is for the action that you're testing. Since the activation of the bomb isn't really what's under test (you tested it elsewhere), it's not part of When.

overlord/client/features/new_bomb.feature
```diff
@@ -0,0 +1,30 @@
+Feature:
+  This client utility will allow a given user to monitor the status of a
+  remote bomb and communicate with it.
+
+  Background:
+    Given the user is on the home page
+    And there is a new bomb
+
+  Scenario:
+    Then the bomb is off
```
jmmastey: Still weird without a When. I suppose it sort-of works here, since it's a test for default behavior/state.

overlord/client/features/new_bomb.feature
```diff
@@ -0,0 +1,30 @@
...
+
+  Scenario:
+    Then the user should see the bomb state
+    And the bomb state is valid
```
jmmastey: May as well combine this with the above test.

overlord/client/features/step_definitions/client.rb
```diff
@@ -0,0 +1,91 @@
+require 'selenium-webdriver'
```
jmmastey: Nooooooooooooooooooooooooooo

(use poltergeist, it's 10x faster)

overlord/client/features/step_definitions/client.rb
```diff
@@ -0,0 +1,91 @@
...
+require 'pry'
+browser = Selenium::WebDriver.for :firefox
+
+bomb_states = ["off", "active", "inactive", "exploded"]
+booted_states = bomb_states.reject { |e| e == "off" }
```
jmmastey: `bomb_states - ["off"]`

overlord/client/features/step_definitions/client.rb
```diff
@@ -0,0 +1,91 @@
...
+
+Given /^there is a new bomb$/ do
+  reset_button = browser.find_element(:id => "bomb_reset_btn")
+  reset_button.click
+end
+
+Given /^the user is on the home page$/ do
+  visit "/"
+  browser.get("http://localhost:1234")
```
jmmastey: Not sure what the deal is with all these `browser.` prefixes. Usually with capybara you don't need them.

overlord/client/features/step_definitions/client.rb
```diff
@@ -0,0 +1,91 @@
...
+end
+
+When /^the user presses boot bomb$/ do
+  boot = browser.find_element(:id => "bomb_boot_btn")
+  boot.click
+  # Wait for view to update
+  sleep(0.5)
```
jmmastey: Fwiw, Capybara automatically waits for elements to appear that didn't exist before, so depending on what your next step is, this should be unnecessary (and is also slow)

overlord/client/features/step_definitions/client.rb
```diff
@@ -0,0 +1,91 @@
...
+end
+
+Given /^the user is on the home page$/ do
+  visit "/"
+  browser.get("http://localhost:1234")
+end
+
+When /^the user presses boot bomb$/ do
+  boot = browser.find_element(:id => "bomb_boot_btn")
+  boot.click
```
jmmastey: `click "Boot"`

overlord/client/features/step_definitions/client.rb
```diff
@@ -0,0 +1,91 @@
...
+    field_id = "submit_code"
+  else
+    raise "Invalid Submission Type"
+  end
+  region = browser.find_element(id: field_id)
+  textbox = region.find_element(class: "textbox")
+  textbox.clear
+  textbox.send_keys(code)
+  region.find_element(class: "btn").click
+end
```
jmmastey: This is quasi-efficient, at the cost of readability. Better:

```
def submit_code(field)
  # region stuff etc
end

When /^the user submits deactivation code (\d+)$/ do |code|
  submit_code("deactivation_code")
end
```

overlord/client/features/step_definitions/client.rb
```diff
@@ -0,0 +1,91 @@
...
+
+Then /^the user should see the bomb state$/ do
+  browser.find_elements(:id => "bomb_state").any?.should eq(true)
+end
```
jmmastey: `expect(page).to have_css("#bomb_state")`

Should is deprecated and going away.

overlord/client/features/step_definitions/client.rb
```diff
@@ -0,0 +1,91 @@
...
+Then /^the user should see the time$/ do
+  browser.find_elements(:id => "timer").any?.should eq(true)
+end
+
+Then /^the time is (.*)$/ do |time|
+  browser.find_element(:id => "timer").text.should have_content(time)
+end
+
+Then /^the bomb is (.*)$/ do |state|
+  sleep(0.5)
```
jmmastey: Try hard not to have these, they really do cripple the speed of the test suite.

overlord/client/features/step_definitions/client.rb
```diff
@@ -0,0 +1,91 @@
...
+  field_id = "submit_code"
+  region = browser.find_element(id: field_id)
+  textbox = region.find_element(class: "textbox")
+  textbox.clear
+  textbox.send_keys(code)
+  region.find_element(class: "btn").click
+  sleep(0.1)
+  result = browser.find_element(class: "status").text
+  result.should have_content("Incorrect Code")
+end
```
jmmastey: This is a `When` but contains an assertion. I think you meant `Then`.

overlord/client/overlord.rb
```diff
@@ -0,0 +1,9 @@
+# run `ruby overlord.rb` to run a webserver for this app
+
+require 'sinatra'
+require 'pry'
+require 'haml'
+
+get '/' do
+  haml :index
+end
```
jmmastey: Gee, don't make it complicated ;)

overlord/client/public/overlord.js
```diff
@@ -0,0 +1,122 @@
...
+      success: function(data) {
+        message = data['message'];
+        $('#timer').text(message)
+      },
+      error: function(data) {
+        alert('Error Getting Time');
+      }
+    });
+  }, 500);
+});
```
jmmastey: Since this is a single-page app, you could replace your sleeps by firing JS events and letting capybara listen to them, or keeping a spinner or some element that you can watch for the disappearance of.

overlord/client/public/reset.css
```diff
@@ -0,0 +1,4 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
jmmastey: O.O

overlord/client/public/overlord.js
```diff
@@ -0,0 +1,122 @@
+$(function() {
+  remote_url = "http://localhost:4567";
```
jmmastey: Use relative URLs. This can't be deployed even on alternative ports.

overlord/server/bomb.rb
```diff
@@ -0,0 +1,127 @@
...
+  end
+
+  def exploded?
+    return true if @state == :exploded
+    if @state == :active && (Time.now - @timer > 120)
+      @state = :exploded
+      return true
+    end
+    false
+  end
```
jmmastey:
```
def check_timer
  @state = :exploded if state == :active && (Time.now - @timer > 120)
end

def exploded?
  !!(@state == :exploded || check_timer)
end
```

Technically that not-not isn't necessary if you don't care that the return value is truthy versus being true.

overlord/server/bomb.rb
```diff
@@ -0,0 +1,127 @@
...
+
+  def inspect
+    "It's a bomb"
+  end
+
+  def to_s
+    "Bomb"
+  end
+
+  def time_left
```
jmmastey: I really have no idea what this means.

overlord/server/bomb.rb
```diff
@@ -0,0 +1,127 @@
...
+    return "0:00" unless state == :active
+    end_time = timer + 120
+    time_remaining = (end_time - Time.now).round(0)
+    return "0:00" if time_remaining < 0
+    seconds = time_remaining % 60
+    minutes = (time_remaining - seconds) / 60
+    minutes.to_s + ":" + "%02d" % seconds
+  end
```
jmmastey: This method is too hard to follow. Refactor.

overlord/server/bomb.rb
```diff
@@ -0,0 +1,127 @@
...
+
+  private
+
+  def attempt_activation(code)
+    raise "Incorrect Code" unless @activation_key == code
+    @timer = Time.now
+    @misses = 0
+    @state = :active
+  end
```
jmmastey: I'm not sure an incorrect code should raise. I'm generally a fan of raising, though, so I don't know.

overlord/server/overlord.rb
```diff
@@ -0,0 +1,71 @@
...
+
+before do
+   content_type :json
+   headers 'Access-Control-Allow-Origin' => '*',
+            'Access-Control-Allow-Methods' => 'OPTIONS, GET, POST, PUT'
+end
+
+set :protection, false
+
+bomb = Bomb.new
```
jmmastey: I'm surprised this works, honestly.

overlord/server/overlord.rb
```diff
@@ -0,0 +1,71 @@
...
+post '/set_activation_code' do
+  content_type :json
+  { :message => try_method(-> { bomb.activation_key = params[:code] },
+                           -> { "Alert: Activation Code set to: " +
+                                bomb.activation_key }) }.to_json
+end
+
+post '/set_deactivation_code' do
+  content_type :json
+  { :message => try_method(-> { bomb.deactivation_key = params[:code] },
```
jmmastey: This way of doing `:message` is probably too clever, though I see where it simplifies your JS. I'd expect the server to respond with something more generic like `{ timer: 87, status: 'active', deactivation_key: '1234' }`. I would probably implement `Bomb#to_h` to make that happen.

Also, on second thought, I would not let the bomb _tell_ people the deactivation code.

overlord/server/rspec/activate_deactivate_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    it "doesn't allow deactivate to be used when inactive" do
+      bomb.boot
+      expect { bomb.deactivate("0000") }.to raise_error(StandardError)
+    end
+
+    it "doesn't allow activate or deactivate when off" do
+      expect { bomb.activate("1234") }.to raise_error(StandardError)
+      expect { bomb.deactivate("0000") }.to raise_error(StandardError)
```
jmmastey: If you're catching StandardError, just skip it and let it stay as `expect { }.to raise_error`

overlord/server/rspec/basic_bomb_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+
+    it "explodes after three wrong attempts" do
+      bomb.boot
+      bomb.enter_code("1234")
+      expect(bomb.status).to eq(:active)
+      expect { bomb.enter_code("0001") }.to raise_error(StandardError)
+      expect { bomb.enter_code("0001") }.to raise_error(StandardError)
+      expect { bomb.enter_code("0001") }.to raise_error(StandardError)
+      expect(bomb.status).to eq(:exploded)
```
jmmastey: This probably argues against the raise as a way to deal with the "normal" condition of failed input codes. This is a lot of exception handling.

overlord/server/rspec/timer_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      bomb.timer = Time.now - 10
+      expect(bomb.time_left).to eq("1:50")
+    end
+
+    it "detenates at 0:00" do
+      bomb.boot
+      bomb.enter_code("1234")
+      bomb.timer = Time.now - 119
+      expect(bomb.time_left).to eq("0:01")
+      sleep(1)
```
jmmastey: No sleeping. Use dependency injection or fake the current time.

jmmastey:
Made some notes. Nothing serious. Keep moving, and good job.

overlord/server/rspec/basic_bomb_spec.rb
Tornquist: It's a bomb! Things should be dangerous.

overlord/server/rspec/activate_deactivate_spec.rb
Tornquist: But I expect a standard error.  If I catch all errors than should something else break that raises an unexpected error I won't know about it?

overlord/server/overlord.rb
Tornquist: It only tells people the activation code when you set it initially.  That's validation that the remote bomb received your instructions.  Otherwise should something fail how would you know? You don't want to turn it on and then not know the keys.

overlord/client/features/step_definitions/client.rb
Tornquist: The catch is that all my elements exist from the beginning.  The only thing that changes is the text of those elements.  I use JS to update the text all throughout the page.   With that in mind, how should this be handled? I don't like sleep, but it seems to accomplish the goals.

overlord/client/public/overlord.js
Tornquist: I wanted to use some yaml config files, but most of the super easy and clean gems are built for rails.  I can't use relative urls between distinct apps can I?

overlord/server/overlord.rb
Tornquist: ???

overlord/server/rspec/timer_spec.rb
Tornquist: Noted.  This is a consistent issue throughout.

PR #809, opened 2015-07-14
Shopping Cart v1

Tornquist:


shopping_cart/address.feature
```diff
@@ -0,0 +1,27 @@
+  Scenario:
```
djkotowski: Remove

shopping_cart/address.feature
```diff
@@ -0,0 +1,25 @@
+Feature:
+  As a user I want to be able to add my address to display shipping information
+
+  Background:
+    Given there is 1 "xBox One" and 1 "WiiU" in the cart
+    And the user is on the cart page
+    And no shipping estimates are displayed
```
djkotowski: This sounds more like a test (`Then`) than a `Given` preparation step.

shopping_cart/address.feature
```diff
@@ -0,0 +1,25 @@
...
+
+  Scenario: Anonymous users can enter an address
+    Given the user is not logged in
```
djkotowski: From an ordering standpoint, you may want to have the login step occur before the page visit step.  If you visit the cart page, then log in, you will no longer be on the cart page.

The only exception to this is if your app allows login from any page and reloads the current page on log-in.  If you do that, nice work!

shopping_cart/address.feature
```diff
@@ -0,0 +1,25 @@
...
+    When the user logs in
+    And the user has address information
+    Then the cart will display a shipping estimate
+
+  Scenario: Users can log and still need to provide an address
+    Given the user is not logged in
+    When the user logs in
+    And the user does not have address information
+    And the user enters address information
```
djkotowski: I'm not sure I understand what L23-24 mean.

shopping_cart/coupon.feature
```diff
@@ -0,0 +1,31 @@
+Feature:
+  As a user either logged in or anonmyous
+  Entering a coupon should either apply the coupon and update the totals
+  or alert me that the coupon was invalid
+
+  Note: Each item is $100 for testing
+        And value is a % off original price
```
djkotowski: Maybe a better term than `value` should be used here.

shopping_cart/coupon.feature
```diff
@@ -0,0 +1,31 @@
+Feature:
+  As a user either logged in or anonmyous
+  Entering a coupon should either apply the coupon and update the totals
+  or alert me that the coupon was invalid
```
djkotowski: I recommend sticking with the 
```
As a _______
I want to _________
In order to _________
```
scheme.  You'll find this used often on real projects, and it will help you focus your features and tests.

shopping_cart/item_quantity.feature
```diff
@@ -0,0 +1,71 @@
+Feature:
+  As a user either logged in or anonymous
+  I want to be able to add, remove, and change the quantity of items in the cart
+  So that I can ultimatly have control over what I order.
+
+  Background:
+    Given a user is on the home page
+    And the user is EITHER logged in OR anonymous
```
djkotowski: This sounds like two different branches for your tests, rather than a single step.

shopping_cart/item_quantity.feature
```diff
@@ -0,0 +1,71 @@
...
+    And the cart is empty
+
+  Scenario: Add item to the cart
+    Given the user opens the product page for <product>
+    When the user clicks "add to cart"
```
djkotowski: If the text changes (e.g. i18n) or the app goes through some implementation changes, this step may not work.  Write the test at a higher level.  A good example would be `When the user adds the item to the cart`.  It doesn't concern itself with the app implementation, rather the _features_ of the app.

shopping_cart/item_page.feature
```diff
@@ -0,0 +1,27 @@
+Feature:
+  As a user either logged in or anonmyous
+  I want to be able to access an item's page from the shopping cart
+
+  Background:
+    Given a user is on the cart page
+    And there is 1 "xBox One" and 3 "PS4" and 2 "SNES" in the cart
```
djkotowski: For reusability, you should split this into three different steps.

shopping_cart/coupon.feature
```diff
@@ -0,0 +1,31 @@
...
+
+  Note: Each item is $100 for testing
+        And value is a % off original price
+          100 = 100%
+          75 = 75%
+          60 = 60%
+
+  Background:
+    Given there is 2 "WiiU" and 1 "PS4" and 3 "SNES" in the cart
```
djkotowski: For reusability, you should split this into three different steps.

shopping_cart/item_quantity.feature
```diff
@@ -0,0 +1,71 @@
+Feature:
+  As a user either logged in or anonymous
+  I want to be able to add, remove, and change the quantity of items in the cart
+  So that I can ultimatly have control over what I order.
+
+  Background:
+    Given a user is on the home page
+    And the user is EITHER logged in OR anonymous
+    And the cart is empty
```
djkotowski: What would the step definition for this step look like?  If the step is empty, it's unnecessary.  You could write a test to ensure the cart is empty by default.

shopping_cart/item_page.feature
```diff
@@ -0,0 +1,27 @@
...
+
+  Scenario: Link to product page
+    When the user clicks "<product>" in the cart
+    Then the user will be shown the <product> page
+    And the product page will show "<quantity> <product>'s" in cart
```
djkotowski: You may run into some issues here if you end up using `#pluralize` from ActiveSupport to pluralize your products.  Just something to keep in mind.

shopping_cart/address.feature
djkotowski: Never mind.  I got it.

shopping_cart/address.feature
```diff
@@ -0,0 +1,25 @@
+Feature:
```
djkotowski: You should name your features.  This one, for example, could be "Shipping estimation".  Just something so anyone who looks at the feature can understand it from the title.

shopping_cart/coupon.feature
```diff
@@ -0,0 +1,31 @@
...
+    And the user is on the cart page
+    And the total cost is 600
+
+  Scenario: Check coupon process
```
djkotowski: Since you're using examples, this is a Scenario Outline, not a Scenario.

shopping_cart/login.feature
```diff
@@ -0,0 +1,32 @@
+Feature:
+  As a user my cart is remembered between sessions.
+  Additionally, logging has the ability to transfer items into my user's cart
+
+  Background:
+    Given there is 2 "WiiU" and 1 "PS4" and 1 "SNES" in the cart
```
djkotowski: Is this an anonymous cart or a user cart?  It's hard to tell, especially with your merge scenario.

djkotowski:
This is a good first iteration.  Take a look at the comments and refactor.  Let me know if you have any questions.

shopping_cart/coupon.feature
Tornquist: My thinking was that you could use regex to match groups and then refer to an infinite number of items in a single line.  I was imagining this scaling where <num> "<item>" (and) could be repeated.

djkotowski: That's making the step way too complex.  Keep them simple, and debugging will be simple.

shopping_cart/item_page.feature
Tornquist: That's just for readability in the feature.  Quantity and general product name is all that the step definition would actually search for.

PR #818, opened 2015-07-15
Triangle Facts

ssapra:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,63 @@
...
+  end
+
+  def triangle_type
+    if equilateral?
+      'This triangle is equilateral!'
+    elsif isosceles?
+      'This triangle is isosceles! Also, that word is hard to type.'
+    else
+      'This triangle is scalene and mathematically boring.'
+    end
```
djkotowski: You can skip these large if-blocks with guard clauses.

```ruby
return "This triangle is equilateral!" if equilateral?
return "This triangle is isosceles! Also that word is hard to type" if isosceles?
"This triangle is scalene and mathematically boring."
```

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,63 @@
...
+    puts 'The angles of this triangle are ' + angles.join(', ')
+    puts 'This triangle is also a right triangle!' if angles.include?(90)
+    puts
+  end
+
+  def calculate_angles(a, b, c)
+    angle_a = calculate_angle(a, b, c)
+    angle_b = calculate_angle(b, a, c)
+    angle_c = calculate_angle(c, a, b)
+    %w(angle_a angle_b angle_c)
```
djkotowski: This doesn't work.  Run the code and check the output.

djkotowski:
Nice iteration.  Be sure to test your code before you submit, though.

Take a look at the comments, fix/refactor where needed, and resubmit.  Let me know if you have questions.

triangle_facts/triangle_facts.rb
ssapra: Running ruby 2.2.2, I don't see any errors. I can just use the brackets to make it less fancy

PR #819, opened 2015-07-15
Arrowhead

ssapra:
Cleaned up method so that error conditions are near their conditions

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -1,5 +1,4 @@
 class Arrowhead
-  # This seriously belongs in a database.
   CLASSIFICATIONS = {
     far_west: {
       notched: "Archaic Side Notch",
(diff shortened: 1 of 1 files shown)


arrowhead/arrowhead.rb
```diff
@@ -15,19 +14,14 @@ class Arrowhead
...
-      end
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+    raise "Please provide valid region" unless CLASSIFICATIONS.include?(region)
+
+    shapes = CLASSIFICATIONS[region]
+    raise "Please provide valid shape" unless shapes.include?(shape)
+
+    arrowhead = shapes[shape]
```
djkotowski: I may split L18-20 and L21-23 into separate methods.

djkotowski:
Looks good!  You're done with this one.

PR #820, opened 2015-07-15
Refactored Robot class for better validations

ssapra:


robot_name/robot.rb
```diff
@@ -0,0 +1,50 @@
...
+    true
+  end
+
+  def generate_name
+    @name = ""
+    3.times { @name << CHAR_GENERATOR.call }
+    3.times { @name << NUM_GENERATOR.call.to_s }
+  end
+
+  def initialize(args = {})
```
djkotowski: `initialize` should be the first method defined.

robot_name/robot.rb
```diff
@@ -0,0 +1,50 @@
...
+    name_generator = args[:name_generator]
+
+    if name_generator
+      @name = name_generator.call
+    else
+      generate_name
+    end
```
djkotowski: Use guard clauses to simplify here.

robot_name/robot.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def generate_name
+    @name = ""
+    3.times { @name << CHAR_GENERATOR.call }
+    3.times { @name << NUM_GENERATOR.call.to_s }
+  end
+
+  def initialize(args = {})
+    name_generator = args[:name_generator]
```
djkotowski: Since the name generator is the only thing you're passing in, no need to make it an arg hash.

robot_name/robot.rb
```diff
@@ -0,0 +1,50 @@
...
+      raise InvalidName, 'Invalid robot name'
+    elsif @@robot_names.include?(name)
+      raise DuplicateName, 'There was a problem generating the robot name!'
+    end
+    true
+  end
+
+  def generate_name
+    @name = ""
+    3.times { @name << CHAR_GENERATOR.call }
```
djkotowski: This should be 2 times.

robot_name/robot.rb
```diff
@@ -0,0 +1,50 @@
...
+    @@robot_names
+  end
+
+  def valid?
+    if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/)
+      raise InvalidName, 'Invalid robot name'
+    elsif @@robot_names.include?(name)
+      raise DuplicateName, 'There was a problem generating the robot name!'
+    end
+    true
```
djkotowski: These should probably be split into separate methods.  Also, it's bad practice to have a predicate method that raises an error when it should return false.

robot_name/robot.rb
```diff
@@ -0,0 +1,50 @@
+class DuplicateName < RuntimeError; end
+class InvalidName < RuntimeError; end
+
+class Robot
+  attr_accessor :name
+
+  @@robot_names = []
```
djkotowski: Prefer class instance variables over class variables.

robot_name/robot.rb
```diff
@@ -0,0 +1,50 @@
...
+  def initialize(args = {})
+    name_generator = args[:name_generator]
+
+    if name_generator
+      @name = name_generator.call
+    else
+      generate_name
+    end
+
+    @@robot_names << @name if valid?
```
djkotowski: Since you have the class method `robot_names`, use that instead of the variable.

djkotowski:
Good first iteration.  Take a look at the comments, refactor and resubmit when ready.  Let me know if you have any questions.

robot_name/robot.rb
ssapra: I don't understand how to use a guard clause here. Both cases need to continue to use the last line of adding the name to the array. I've also added two more methods in my refactored code to perform the two validations, so now I have 3 methods that are being called after the if-else.

djkotowski: You could do a little restructuring.  This is an example:
```ruby
# in Robot#initialize
@name = generate_name(name_generator)
```
```ruby
def generate_name(generator)
  return generator.call if generator

  # custom generation logic here
end
```

An alternative would be to use a ternary (this being one of the few cases where a ternary is simple enough to use).
```ruby
@name = name_generator ? name_generator.call : generate_name
```

PR #821, opened 2015-07-15
TriangleFacts

ssapra:
Made changes to address feedback

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -1,58 +1,63 @@
 # Killer facts about triangles AWW YEAH
 class Triangle
-	attr_accessor :side1,:side2,:side3
+  attr_accessor :side1, :side2, :side3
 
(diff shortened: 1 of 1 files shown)


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,59 @@
...
+    return "This triangle is isosceles!" if isosceles?
+    "This triangle is scalene and mathematically boring."
+  end
+
+  def recite_facts
+    puts triangle_type
+
+    angles = calculate_angles(side1, side2, side3)
+
+    puts 'The angles of this triangle are ' + angles.map(&:to_s).join(', ')
```
djkotowski: Use `<<` instead of `+` to improve speed and save memory.
`<<` appends to the existing string, `+` creates a new one.

djkotowski:
Great job!  You're done with this one.

PR #822, opened 2015-07-16
Dinodex - Resubmission

oharace:
Made changes based on feedback. Re-ordered logic & added method chaining for filtering.


dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,22 @@
+require 'csv'
+require_relative 'dinosaur'
+
+class Parser
+  def self.read_csv_file(file_path)
+    csv_file_rows_array = CSV.read(file_path)
```
jmtai: In general, you should avoid including the data type in your variable names. In this case `csv_file_rows` is sufficiently descriptive - the pluralization implies that multiple items are being returned. Additionally, if the interface changes, you'll have to remember to go through and rename every instance of the variable or it becomes confusing for someone that has to maintain the code later because now the variable names don't accurately reflect what is happening in the code.

dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,22 @@
...
+    analyze_csv_array(csv_file_rows_array)
+  end
+
+  def self.analyze_csv_array(array)
+    headers_array = array.shift
```
jmtai: The CSV parser can do this for you automatically. Check out the `headers` option.

dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,22 @@
+require 'csv'
+require_relative 'dinosaur'
+
+class Parser
+  def self.read_csv_file(file_path)
+    csv_file_rows_array = CSV.read(file_path)
+    analyze_csv_array(csv_file_rows_array)
+  end
+
+  def self.analyze_csv_array(array)
```
jmtai: You don't want to restrict yourself to a particular data type with method names, either. Since this is a public method, it becomes part of your `Parser`'s public API. What if you wanted to change the internal implementation of your parser so it uses hashes instead of arrays? Now you either have to add a new method `analyze_csv_hash` and maintain both the old and new methods for some time so you don't break any existing code using the old API, or you break all existing code and force them to switch to the new method.

dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,22 @@
...
+    headers_array = array.shift
+    array.each_with_object([]) do |row, dino_array|
```
jmtai: When you are transforming each element of an array, what you really want is `map`. If you use `map`, you don't have to explicitly add each new element to a new array:
```ruby
dino_array = array.map { |row| Dinosaur.new(get_facts_from_each_row(headers_array, row) }
```

jmtai: And, again, something like `dinosaurs` would probably be a better variable name here.

dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,22 @@
...
+      dino_array << Dinosaur.new(get_facts_from_each_row(headers_array, row))
+    end
+  end
+
+  def self.get_facts_from_each_row(headers, row)
```
jmtai: Naming is hard and often subjective, but there are some general rules we can follow. One is to make sure your method name accurately reflect the operation being performed.
In this case, the method name is `get_facts_from_each_row`. Since `each` is included in the name, it would be reasonable for someone to assume that this method iterates over each row and extracts the facts from each row. However, what is actually happening is that this method is extracting data from a single row, so the name should reflect that. Maybe `get_facts_from_row` would be more appropriate? Or maybe even something more succinct like `parse_row`?

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,35 @@
...
+  def assign_name(facts)
+    @name = facts["name"] || facts["genus"]
+  end
+
+  def assign_weight(facts)
+    @weight = facts["weight"] || facts["weight_in_lbs"]
+  end
+
+  def assign_diet(facts)
+    @diet = "Carnivore" if facts["Carnivore"] == "Yes"
```
jmtai: Uh-oh. In your CSV reader you use `header.downcase` when populating the facts hash but here you're using a mixed-case key.

jmtai: Also, what if a row was added to the file with a lower-case "yes" or "no" for the `Carnivore` entry? Do you think this comparison should fail for those cases?

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,109 @@
+require_relative 'csv_parser'
+require_relative 'printer'
+
+class DinosaurCatalog
+  TWO_TONS = 4000
```
jmtai: Good use of a constant here.

dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,22 @@
+require 'csv'
+require_relative 'dinosaur'
+
+class Parser
```
jmtai: The name convention for classes is that the camel-case class name should map directly to a snake-case filename. So `class Parser` would be stored in a file called `parser.rb` and `class CsvParser` would be stored in a file called `csv_parser.rb`.

For these exercises, the names don't have to match, but once you start working with Rails, you'll want to follow the convention as the class autoloader uses that rule to determine which files to load.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  def pass_new_instance_for_filter(dinosaurs)
+    new_catalog = DinosaurCatalog.new
+    new_catalog.dinosaurs = dinosaurs
+    new_catalog
+  end
+
+  def all_dinosaurs
+    pass_new_instance_for_filter(@dinosaurs)
```
jmtai: Why create an entirely new instance for this method? All you need is the data stored in the `@dinosaurs` instance variable, so you easily just pass `self` instead of creating a new instance.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,109 @@
...
+    @dinosaurs = []
+    file_paths.each do |file_path|
+      @dinosaurs.concat(Parser.read_csv_file(file_path))
+    end
+  end
+
+  def pass_new_instance_for_filter(dinosaurs)
+    new_catalog = DinosaurCatalog.new
+    new_catalog.dinosaurs = dinosaurs
+    new_catalog
```
jmtai: Instead of incurring the overhead of creating a new instance each time you want to filter, can you think of a way to do this by just returning `self` from each filter call?

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  def all_dinosaurs
+    pass_new_instance_for_filter(@dinosaurs)
+  end
+
+  def filter_by_name(name)
+    dino = @dinosaurs.detect { |dinosaur| dinosaur.name == name }
```
jmtai: Should all these filters be case-sensitive?

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,109 @@
...
+    else
+      dinos = @dinosaurs.select { |dino| dino.diet == diet }
+    end
+    pass_new_instance_for_filter(dinos)
+  end
+
+  def big_dinos
+    dinos = []
+    @dinosaurs.each do |dino|
+      dinos << dino if !dino.weight.nil? && dino.weight.to_i > TWO_TONS
```
jmtai: You can use `select` here, as well.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,109 @@
...
+    end
+    pass_new_instance_for_filter(dinos)
+  end
+
+  def small_dinos
+    dinos = []
+    @dinosaurs.each do |dino|
+      dinos << dino if !dino.weight.nil? && dino.weight.to_i <= TWO_TONS
```
jmtai: Again, `select` works even for multiple conditions.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,109 @@
...
+    end
+    pass_new_instance_for_filter(dinos)
+  end
+
+  def filter_by_weight(weight)
+    if weight > TWO_TONS
+      big_dinos
+    else
+      small_dinos
+    end
```
jmtai: For a simple conditional such as this, the ternary operator can be used to compact the expression:
```ruby
weight > TWO_TONS ? big_dinos : small_dinos
```

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  def filter_by_continent(continent)
+    dinos = @dinosaurs.select { |dino| dino.continent == continent }
+    pass_new_instance_for_filter(dinos)
+  end
+
+  def print
+    Printer.print_just_names(@dinosaurs)
```
jmtai: I'd argue that the `just` modifier here is unnecessary. `print_names` is enough to describe what the method is doing.

dino_catalog/printer.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def self.print_description_string(description)
+    puts "Fun fact: #{description}" if description
+  end
+
+  def self.print_diet_string(dinosaur)
+    if dinosaur.diet
+      puts "This dinosaur was a #{dinosaur.diet.downcase}."
+    elsif dinosaur.carnivorous
```
jmtai: Make sure you test all code paths - this method call is missing the `?`

dino_catalog/printer.rb
```diff
@@ -0,0 +1,61 @@
...
+      puts "This dinosaur was a carnivore."
+    else
+      puts "This dinosaur was not a carnivore."
+    end
+  end
+
+  def self.print_with_facts(dinosaurs)
+    if dinosaurs.empty?
```
jmtai: Simple if-else constructs are good candidates for replacement with **guard clauses**:
```ruby
return print_no_dinosaurs_found if dinosaurs.empty?
dinosaurs.each { |dino| print_fact_string(dino) }
```

jmtai:
Nice job! Your code is nice and clean and easy to follow. Read over the comments and refactor a bit - in particular, see if you can come up with a way to allow filter chaining without having to create a new instance of DinoCatalog each time.

Also, although it's not explicitly stated in the exercise description, it'd be really nice to have a simple command-line interface so a user can interact with the catalog without having to modify the source code.

oharace:
Thanks! I will work on what you suggested. Lots of helpful things!

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def assign_other_facts(facts)
+    @period = facts["period"]
+    @walking = facts["walking"]
+    @continent = facts["continent"]
+    @description = facts["description"]
+  end
+
+  def carnivorous?
+    @diet == "Carnivore" || @diet == "Piscivore" || @diet == "Insectivore"
```
paulghaddad: Very nice.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,35 @@
...
+  attr_reader :period, :weight, :walking, :name, :diet, :continent, :description
+
+  def initialize(facts)
+    assign_name(facts)
+    assign_weight(facts)
+    assign_diet(facts)
+    assign_other_facts(facts)
+  end
+
+  def assign_name(facts)
```
paulghaddad: The `assign_*` methods are only called within this class, so consider making them private.

PR #823, opened 2015-07-16
Robot name

ssapra:
Addressing feedback with new class to handle array of names and duplicate name error.


robot_name/robot.rb
```diff
@@ -0,0 +1,53 @@
...
+  def self.add_name(name)
+    check_for_duplicate(name)
+    @robot_names << name
+  end
+end
+
+class Robot
+  attr_accessor :name
+
+  CHAR_GENERATOR = -> { ('A'..'Z').to_a.sample }
```
paulghaddad: Do you necessarily need to use lambdas here, or would a method work better? I'd personally use a private method here.

robot_name/robot.rb
```diff
@@ -0,0 +1,53 @@
...
+  NUM_GENERATOR = -> { rand(10) }
+
+  def initialize(generator = nil)
+    @name = generator ? generator.call : generate_name
```
paulghaddad: Consider extracting this as a method.

robot_name/robot.rb
```diff
@@ -0,0 +1,53 @@
...
+    Registry.add_name(@name)
+  end
+
+  def validate_name
+    is_invalid = !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/)
+    raise InvalidName, 'Invalid name' if is_invalid
+  end
+
+  def generate_name
+    @name = ""
```
paulghaddad: This pattern of starting with an initial object, building it up and then returning the resulting object leads itself to using `inject`. See if you can simplify this method by using `inject`.

paulghaddad: Since `name` has an accessor method, do you need the `@` sign?

robot_name/robot.rb
```diff
@@ -0,0 +1,53 @@
...
+
+class Robot
+  attr_accessor :name
+
+  CHAR_GENERATOR = -> { ('A'..'Z').to_a.sample }
+  NUM_GENERATOR = -> { rand(10) }
+
+  def initialize(generator = nil)
+    @name = generator ? generator.call : generate_name
+    validate_name
```
paulghaddad: Nice job validating the name as its own method.

robot_name/robot.rb
```diff
@@ -0,0 +1,53 @@
+class DuplicateName < RuntimeError; end
+class InvalidName < RuntimeError; end
+
+class Registry
+  @robot_names = []
```
paulghaddad: I like how you eliminated the class variable from the original implementation.

robot_name/robot.rb
```diff
@@ -0,0 +1,53 @@
+class DuplicateName < RuntimeError; end
```
paulghaddad: I'd personally name these as `DuplicateNameError` and `InvalidNameError` to follow the Ruby convention. But nice job creating two specific errors instead of one general error.

robot_name/robot.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def self.check_for_duplicate(name)
+    if @robot_names.include?(name)
+      raise DuplicateName, 'There was a problem generating the robot name!'
+    end
+  end
+
+  def self.add_name(name)
+    check_for_duplicate(name)
+    @robot_names << name
```
paulghaddad: Nice use of the append operator!

robot_name/robot.rb
```diff
@@ -0,0 +1,53 @@
+class DuplicateName < RuntimeError; end
+class InvalidName < RuntimeError; end
+
+class Registry
```
paulghaddad: Great job extracting the Registry as its own class and adhering to the Single Responsibility Principle!

robot_name/robot.rb
```diff
@@ -0,0 +1,53 @@
...
+  @robot_names = []
+
+  def self.check_for_duplicate(name)
```
paulghaddad: This can be a private method.

robot_name/robot.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+  end
+
+  def self.add_name(name)
+    check_for_duplicate(name)
+    @robot_names << name
+  end
+end
+
+class Robot
```
paulghaddad: Check for opportunities in this class to use private methods. Only expose what you need to .

paulghaddad:
Overall, this is a very nice implementation. I like how you extracted the `Registry` class, and had really easy to follow methods. Simplicity is a great thing. Your code was really clear and easy to understand.

Some things to keep in mind:

1. Make methods not called from outside the class as private.
2. Take a look at the `inject` and `each_as_object` methods when incrementally building objects.
3. See if you can change the lambdas to simple methods. Does it improve the code?

PR #824, opened 2015-07-16
Clean up triangle_facts to follow the ruby style guide.

abake-enova:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,68 @@
...
+      'This triangle is equilateral!'
+    elsif isosceles?
+      'This triangle is isosceles! Also, that word is hard to type.'
+    elsif scalene?
+      'This triangle is scalene and mathematically boring.'
+    end
+  end
 
+  def recite_facts
+    angles = calculate_angles(side1, side2, side3)
```
paulghaddad: Consider renaming `calculate_angles` to `angles`. This will allow you to interpolate the `angles` method in the line below and eliminate the need for this variable assignment.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,68 @@
...
+    puts 'The angles of this triangle are ' + angles.join(',')
+
+    puts triangle_type_message
+    puts 'This triangle is also a right triangle!' if right_triangle?
+    puts ''
```
paulghaddad: You should be able to just use `puts` without the `''`.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,68 @@
...
+  def right_triangle?
+    angles.include? 90
+  end
+
+  def triangle_type_message
+    if equilateral?
+      'This triangle is equilateral!'
+    elsif isosceles?
+      'This triangle is isosceles! Also, that word is hard to type.'
+    elsif scalene?
```
paulghaddad: What happens if none of these conditions is true? :)

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,68 @@
...
+  def equilateral?
+    side1 == side2 && side2 == side3
+  end
+
+  def isosceles?
+    [side1, side2, side3].uniq.length == 2
+  end
+
+  def scalene?
+    !(equilateral? || isosceles?)
```
paulghaddad: Instead of determining whether a triangle is scalene based on two method calls, I'd prefer to use the a similar implementation as line 15:

`[side1, side2, side3].uniq.length == 3`

paulghaddad:
You did a really nice job in cleaning this file up in regards to style. I didn't see anything to point out. Just a couple pointers for future exercises:

1. Methods not called outside of a class should be `private`.
2. See if you can use `inject` or `each_with_object` to simplify the `calculate_angles` method.

Other than that, nice work!

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,68 @@
...
+  end
+
+  def right_triangle?
+    angles.include? 90
```
djkotowski: Wrap method args in parentheses.

triangle_facts/triangle_facts.rb
djkotowski: Impossible.  You can use guard clauses to simplify this to 3 lines though.

djkotowski:
Looks good.  Take a look at the comments, but no need to resubmit.

PR #826, opened 2015-07-16
DinoCatalog

ssapra:
First pass.

### DIFF
--- a/dino_catalog/dino_query.rb
+++ b/dino_catalog/dino_query.rb
@@ -4,7 +4,54 @@ require_relative './dinosaur.rb'
 class DinoQuery
   attr_accessor :dinos
 
-  def initialize
-    @dinos = Parser.parse
+  def initialize(dinos = [])
+    @dinos = get_dinos(dinos)
+  end
+
+  def get_dinos(dinos)
+    return dinos unless dinos.empty?
+    parser = Parser.new
+    parser.parse
+    parser.dinos
+  end
+
+  def biped
+    results = dinos & dinos.select(&:biped?)
+    DinoQuery.new(results)
+  end
+
+  def carnivore
+    results = dinos & dinos.select(&:carnivore?)
+    DinoQuery.new(results)
+  end
+
+  def big
+    results = dinos & dinos.select(&:big?)
+    DinoQuery.new(results)
+  end
+
+  def small
+    results = dinos & dinos.select(&:small?)
+    DinoQuery.new(results)
+  end
+
+  def from_period(prd)
+    results = dinos & dinos.select { |d| d.from_period?(prd) }
+    DinoQuery.new(results)
+  end
+
+  def print
+    dinos.each(&:print_facts)
+    nil
+  end
+
+  def to_json
+    dinos.map(&:to_json)
   end
 end
+
+# Tests
+# dq = DinoQuery.new
+# dq.from_period("Jurassic").small.print
+# dq.biped.big.print
+# dq.carnivore.from_period("Cretaceous").print
--- a/dino_catalog/dino_query.rb
+++ b/dino_catalog/dino_query.rb
@@ -1,0 +1,10 @@
+require_relative './parse.rb'
+require_relative './dinosaur.rb'
+
+class DinoQuery
+  attr_accessor :dinos
+
+  def initialize
+    @dinos = Parser.parse
+  end
+end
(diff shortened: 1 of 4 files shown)


dino_catalog/dino_query.rb
```diff
@@ -0,0 +1,70 @@
...
+    @dinos = get_dinos(dinos)
+  end
+
+  def get_dinos(dinos)
+    return dinos unless dinos.nil?
+
+    parser = Parser.new
+    parser.parse
+    parser.dinos
+  end
```
jmmastey: This usage of the default dinos parameter is confusing (and I'd argue surprising to the user). Better to have them explicitly pass in the parser if they want it.

dino_catalog/dino_query.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def small
+    results = dinos & dinos.select(&:small?)
+    DinoQuery.new(results)
+  end
+
+  def from_period(prd)
+    results = dinos & dinos.select { |d| d.from_period?(prd) }
+    DinoQuery.new(results)
+  end
```
jmmastey: These methods all have a lot of duplication. Can you reduce them to calling a single method that accepts a block and returns relevant entries?

Also, what's the purpose for the `&` in these methods? It seems extraneous.

dino_catalog/dino_query.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def print
+    puts "No dinosaurs in this query" if dinos.empty?
+    dinos.each(&:print_facts)
+    nil
+  end
+
+  def to_json
+    dinos.map(&:to_json)
+  end
```
jmmastey: This will return an array of strings. Better to use `dinos.to_json`.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def carnivore?
+    return false if diet.nil?
+    %w(Carnivore Piscivore Insectivore).include?(diet)
+  end
+
+  def biped?
+    return false if walking.nil?
+    walking == "Biped"
+  end
```
jmmastey: For this, and the others, it might be better to go with `!walking.nil? && walking == "Biped"`. It's a more common pattern. In rails, we can use `.present?` instead (or `.try`).

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,57 @@
...
+    return false if period.nil?
+    period.include?(prd)
+  end
+
+  def print_facts
+    to_hash.each_pair do |key, value|
+      puts "#{key}: #{value}" unless value.nil?
+    end
+    puts
+  end
```
jmmastey: I'd rather have a `.to_s` than a print.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def to_hash
```
jmmastey: Okay, so this is super esoteric, but the method you want to define is `to_h`, which is what most methods will call.

The method `.to_hash` is defined only if an object has a "natural" representation as a hash, for use with the method `Kernel#Hash()`. However, _even that method_ ends up calling `to_h` in the end.

dino_catalog/parser.rb
```diff
@@ -0,0 +1,58 @@
...
+
+class Parser
+  FILES = %w(dinodex.csv african_dinosaur_export.csv)
+
+  attr_reader :dinos
+
+  def parse
+    @dinos = []
+    FILES.each do |file|
+      csv = CSV.open(file)
```
jmmastey: CSV has a `header: true` option that you should use.

dino_catalog/parser.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    @header[index] = "diet"
+    adjust_diet_values(index)
+  end
+
+  def fix_discrepancies
+    handle_weight
+    handle_name
+    handle_diet
+  end
```
jmmastey: This doesn't correctly assign "Africa" as the continent from the second CSV.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def biped?
+    return false if walking.nil?
+    walking == "Biped"
+  end
+
+  def big?
+    return false if weight.nil?
+    weight.to_i > TWO_TONS
```
jmmastey: Fwiw, if you do this as you originally create the dino, you won't have to keep casting it to an int.

jmmastey:
This is a really solid implementation. I left a few notes, but it's good as-is.

PR #827, opened 2015-07-16
Robot name

abake-enova:


avavita:
This looks good to me. Talk to Dan, but I don't see a reason to resubmit.

robot_name/robot_name.rb
```diff
@@ -1,25 +1,54 @@
-class NameCollisionError < RuntimeError; end
+InvalidNameError = Class.new(RuntimeError)
+NameCollisionError = Class.new(RuntimeError)
+
+class RobotNameRegistry
```
djkotowski: Separate your classes to separate files, and follow the snake case file naming scheme.

robot_name/robot_name.rb
```diff
@@ -1,25 +1,54 @@
...
+
+  def self.add_name(name)
+    @names << name if self.valid?(name) && self.unique?(name)
+  end
+
+  def self.valid?(name)
+    unless name =~ /[[:alpha:]]{2}[[:digit:]]{3}/
+      raise InvalidNameError, 'The given/generated name is invalid.'
+    end
+    true
```
djkotowski: Don't mix predicate methods with raising errors.  If it's a predicate method, it should return true or false.  This method either breaks execution, or returns true, making the if-statement it's used for useless.  Pick one scheme for the method and go with it.

robot_name/robot_name.rb
```diff
@@ -1,25 +1,54 @@
...
+  end
+end
 
 class Robot
   attr_accessor :name
 
-  @@registry
+  GEN_CHAR = -> { ('A'..'Z').to_a.sample }
+  GEN_NUM = -> { rand(10) }
```
djkotowski: Why not make these two methods?  They never change.

djkotowski:
Good first iteration.  Take a look at the comments, refactor and resubmit.  Let me know if you have questions.

PR #829, opened 2015-07-17
Data Science - First Iteration

oharace:
Ready for review.


data_science/data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
+require_relative 'json_reader'
+require 'abanalyzer'
+
+class DataAnalyzer
+  attr_accessor :data
```
jmtai: What's the purpose of this accessor? This doesn't appear to be used anywhere.

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def initialize(file_path)
+    @data_array = JSONReader.read_json_file(file_path)
```
jmtai: Do you ever expect your `JSONReader` class to read files that **aren't** JSON? If not, there's no need to name the read method `read_json_file`. `read_file` would suffice.

jmtai: Also, variable names shouldn't include data types (check your DinoDex review for some reasons why).

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def sample_size(cohort)
```
jmtai: Notice how a bunch of your methods have the same `cohort` input parameter? This can be an indication that these methods might be better off in a separate class. In this case, your `DataAnalyzer` class is responsible for performing per-cohort calculations as well as the overall A/B analysis. The single-responsibility principle states that our classes and methods should have only one responsibility.

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+    @data_array.count { |data| data["cohort"] == cohort }
```
jmtai: A side effect of not having a `Cohort` class to store cohort-related data is shown here, where you are processing the entire data set with `count` each time `sample_size` is being called. Since your methods `conversion_rate` and `standard_error` are using `sample_size` to perform their calculations, you are parsing the data multiple times when you really only need to do it once.

Although you won't notice it for this particular data set, which is small, imagine if your file had millions of entries. Would you want to be processing a huge data set like that multiple times?

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def num_conversions(cohort)
+    @data_array.count { |data| data["cohort"] == cohort && data["result"] == 1 }
```
jmtai: Just like `sample_size`, this number doesn't change, right? So there's no need to recalculate the value every time you want to get the number of conversions per cohort. Ideally, this value would be calculated once and simply stored when the file is initially parsed.

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def num_fails(cohort)
+    @data_array.count { |data| data["cohort"] == cohort && data["result"] == 0 }
+  end
+
+  def conversion_rate(cohort)
+    rate = num_conversions(cohort).to_f / sample_size(cohort)
```
jmtai: What if the sample_size is 0?

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+    round_value(rate)
```
jmtai: Why is a separate rounding function necessary here? The build-in `round` method for floating point values can round with arbitrary precision. i.e. round(2) would round the number with 2 decimal places of precision.

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def standard_error(cohort)
+    rate = conversion_rate(cohort)
+    se = Math.sqrt(rate * (1 - rate) / sample_size(cohort))
+    round_value(se)
+  end
+
+  def better_than_random?
+    tester = ABAnalyzer::ABTest.new(values_to_hash)
+    tester.different?
```
jmtai: This isn't *quite* what the requirements are asking for. The requirement states "**Confidence level** that the current leader is in fact better than random". So what we are looking for is the **confidence level** of the split test result.

The requirements mention that you should use the Chi-square test for this, so look that up and see if you can come up with the necessary code to calculate the confidence level using that test.

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def values_to_hash
+    values = {}
+    values[:a] = { conversion: num_conversions("A"), fails: num_fails("A") }
+    values[:b] = { conversion: num_conversions("B"), fails: num_fails("B") }
+    values
```
jmtai: Instead of repeating yourself with values, values, values, ... , you can simply define an explicit hash:
```ruby
{
  a: { conversion: num_conversions("A"), fails: num_fails("A") },
  b: { conversion: num_conversions("B"), fails: num_fails("B") }
}
```

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def round_value(value)
+    (value * 1000).round / 1000.0
+  end
+
+  def conversion_rate_range(cohort)
+    range = []
+    left_value = conversion_rate(cohort) - 1.96 * standard_error(cohort)
```
jmtai: `left_value` and `right_value` would probably be better if named something like `range_min` and `range_max`. While technically correct, `left` and `right` aren't super descriptive.

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+    right_value = conversion_rate(cohort) + 1.96 * standard_error(cohort)
+    left_value = round_value(left_value)
+    right_value = round_value(right_value)
+    range << (left_value < 0 ? 0 : left_value)
```
jmtai: Since `range` isn't being used in any intermediate operations, we can just return the constructed array directly without the need for the temporary variable:
```ruby
[ (left_value < 0 ? 0 : left_value),
  (right_value > 1 ? 1 : right_value) ]
```

jmtai: To make it even cleaner, you could simply create a clamping method and the return value could be this:
```ruby
[ clamp(left_value), clamp(right_value) ]
```

data_science/spec/data_analyzer_spec.rb
```diff
@@ -0,0 +1,45 @@
+require_relative 'spec_helper'
+
+describe DataAnalyzer do
+  before :each do
```
jmtai: Take a look at this: https://www.launchacademy.com/codecabulary/learn-test-driven-development/rspec/before-vs-let

Usually let will be a better choice than before for the reasons mentioned in that article.

data_science/spec/data_analyzer_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  it "returns the correct range for the conversion rate" do
+    a_cohort_range = @data_analyzer.conversion_rate_range("A")
+    expect(a_cohort_range[0]).to eq(0.022)
+    expect(a_cohort_range[1]).to eq(0.07)
+    b_cohort_range = @data_analyzer.conversion_rate_range("B")
+    expect(b_cohort_range[0]).to eq(0.039)
+    expect(b_cohort_range[1]).to eq(0.089)
+  end
+end
```
jmtai: Generally speaking, we don't want any magic numbers in our code (which includes our tests). In this case, I don't have a big problem with the hardcoded result numbers since none are being reused and it's clear what each number represents.

In other cases, you'll want to use `let` to define your constants instead of hardcoding numbers into the tests. e.g.
```ruby
let(:sample_size_a) { 284 }
```
You might also be tempted to use constants but if you do so, make sure to use self::SOME_CONSTANT to ensure that the constant doesn't accidentally end up globally scoped, which could break other unrelated specs: http://stackoverflow.com/questions/5230070/scope-constants-to-an-rspec-context

jmtai: Your tests are good for checking the happy path (with the exception of confidence level, which needs to be added), but they don't cover any sad or error path cases. Think about what could go wrong and add some tests to cover those paths, as well.

jmtai:
Nice first pass! Read over the comments, refactor, and resubmit when ready. Feel free to come talk to me if you have any questions.

data_science/spec/data_analyzer_spec.rb
djkotowski: Nice resource.

PR #830, opened 2015-07-17
DataScience v1

ssapra:
First pass through data science exercise. Focused on separating logic into concise methods while keeping it highly readable.


data_science/analyzer.rb
```diff
@@ -0,0 +1,79 @@
...
+  def initialize(file_name)
+    @file_name = file_name
+  end
+
+  def load_raw_data_from_file
+    JSON.parse(File.read(@file_name))
+  end
+
+  def load_cohort_data(data)
+    cohort_data = Hash.new { [] }
```
katieschilling: `each_with_object` will remove 2 lines from this method

data_science/analyzer.rb
```diff
@@ -0,0 +1,79 @@
...
+    data.each do |row|
+      cohort_name = row["cohort"]
+      cohort_data[cohort_name] <<= row["result"]
+    end
+    cohort_data
+  end
+
+  def build_cohort_array(cohort_data)
+    cohort_data.each_pair.map do |name, sample_data|
```
katieschilling: nice!

data_science/analyzer.rb
```diff
@@ -0,0 +1,79 @@
...
+  def convert_data_for_test
+    raw_data = load_raw_data_from_file
+    cohort_data = load_cohort_data(raw_data)
+    build_cohort_array(cohort_data)
+  end
+
+  def initialize_test
+    @cohorts = convert_data_for_test
+    values = {}
+    @cohorts.each { |chrt| values[chrt.name] = chrt.conversion_statistics }
```
katieschilling: another `each_with_object`

data_science/cohort.rb
```diff
@@ -0,0 +1,44 @@
...
+    @upper_ci = @conversion_rate + COND_LEVEL * @stnd_error
+  end
+
+  def calc_standard_error
+    p = @conversion_rate
+    n = @sample_size
+    @stnd_error = Math.sqrt(p * (1 - p) / n)
+  end
+
+  def conversion_statistics
```
katieschilling: I like how you separated this.

data_science/cohort.rb
```diff
@@ -0,0 +1,44 @@
...
+  def calc_conversion_rate
+    @conversion_rate = @conversions.to_f / @sample_size
+  end
+
+  def calc_confidence_intervals
+    @lower_ci = @conversion_rate - COND_LEVEL * @stnd_error
+    @upper_ci = @conversion_rate + COND_LEVEL * @stnd_error
+  end
+
+  def calc_standard_error
```
katieschilling: you can call it `calculate`. this isn't python or c. don't be scared of long names as long as they're intention-revealing.

data_science/cohort.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def calculate_statistics
+    calc_conversion_rate
+    calc_standard_error
+    calc_confidence_intervals
+  end
+
+  def calc_conversion_rate
+    @conversion_rate = @conversions.to_f / @sample_size
```
katieschilling: Most of the numbers you'll be dealing with are money. Never cast those values as floats. Always use `BigDecimal`. `Float` can give rounding errors. 

Just get in the habit of never using floats. You'll always want to specify the precision and scale, anyway.

data_science/cohort.rb
```diff
@@ -0,0 +1,44 @@
...
+  attr_accessor :stnd_error, :lower_ci, :upper_ci
+  COND_LEVEL = 1.96
+
+  def initialize(name, sample)
+    @name = name
+    @sample_size = sample.count
+    @conversions = sample.count(1)
+  end
+
+  def calculate_statistics
```
katieschilling: In small projects, this design pattern is fine. However, in a larger project you can quickly get a tangle of tiny method calls. Be careful not to have too many levels of method calls. It's like arrowhead code, but less obvious to the eye because of how neat it seems initially.

I can show you examples of this in our codebase. It makes refactoring challenging, and IMO it's one of the annoying things about Ruby.

I typically comment that code should be separated out into another method, so take it as a compliment that you are already doing that and should be thinking about the larger architecture of your code.

data_science/cohort.rb
```diff
@@ -0,0 +1,44 @@
+class Cohort
+  attr_accessor :name, :sample_size, :conversions, :conversion_rate
+  attr_accessor :stnd_error, :lower_ci, :upper_ci
+  COND_LEVEL = 1.96
```
katieschilling: name should be more explicit

data_science/cohort_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+require 'json'
+require_relative './cohort.rb'
+
+describe Cohort do
+  let(:test_file_name) { 'test_data.json' }
+  let(:data) { JSON.parse(File.read(test_file_name)) }
+  let(:cohort_name) { 'A' }
+  let(:cohort_data) do
+    data.select { |row| row["cohort"] == cohort_name }
+      .collect { |row| row["result"] }
```
katieschilling: hound is right :)

data_science/cohort_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+  let(:cohort) { Cohort.new(cohort_name, cohort_data) }
+
+  describe '#initialize' do
+    it 'sets Cohort instance with its name, 24 data points, and 10 convs' do
```
katieschilling: you could name this `sets cohort instance with proper name and data` so that you don't have to change the test title if you change the numbers

data_science/cohort_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      cohort.calc_confidence_intervals
+    end
+
+    context '#calc_conversion_rate' do
+      subject { cohort.conversion_rate }
+      it { is_expected.to be_within(0.01).of(0.416) }
+    end
+
+    context '#calc_standard_error' do
+      subject { cohort.stnd_error }
```
katieschilling: You should use subjects only when you have several tests related to the same subject. Using subjects for each test isn't necessary. Just use `expect(whatever).to be_within(whatever).of(whatever)`.

data_science/analyzer.rb
```diff
@@ -0,0 +1,79 @@
...
+      cohort_data[cohort_name] <<= row["result"]
+    end
+    cohort_data
+  end
+
+  def build_cohort_array(cohort_data)
+    cohort_data.each_pair.map do |name, sample_data|
+      cohort = Cohort.new(name, sample_data)
+      cohort.calculate_statistics
+      cohort
```
djkotowski: L28-30 could be shortened to
```ruby
Cohort.new(name, sample_data).tap(&:calculate_statistics)
```
if you choose.

data_science/analyzer.rb
```diff
@@ -0,0 +1,79 @@
...
+      puts "There is a significant difference between the cohorts"
+    else
+      puts "There is no significant difference between the cohorts"
+    end
+  end
+
+  def print_cohort_summary
+    puts "Cohort Summary:"
+    puts
+    @cohorts.each(&:print_summary)
```
djkotowski: When reading instance variables that have corresponding readers, prefer the readers over the instance vars.

data_science/analyzer.rb
```diff
@@ -0,0 +1,79 @@
...
+  def load_cohort_data(data)
+    cohort_data = Hash.new { [] }
+    data.each do |row|
+      cohort_name = row["cohort"]
+      cohort_data[cohort_name] <<= row["result"]
+    end
+    cohort_data
+  end
+
+  def build_cohort_array(cohort_data)
```
djkotowski: Just reading your interface, I don't understand the difference between `#cohorts` and `#build_cohort_array`.  You'll want to clarify this.

data_science/analyzer.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+  end
+
+  def convert_data_for_test
+    raw_data = load_raw_data_from_file
+    cohort_data = load_cohort_data(raw_data)
+    build_cohort_array(cohort_data)
+  end
+
+  def initialize_test
```
djkotowski: It looks like this is just used from `#perform_experiment`.  Shouldn't this be private?

data_science/analyzer.rb
```diff
@@ -0,0 +1,79 @@
+require 'json'
+require 'abanalyzer'
+require_relative './cohort.rb'
+
+class Analyzer
```
djkotowski: Go through your methods and determine which need to be public and which could be private.

data_science/analyzer_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      it 'groups data under the cohort names' do
+        expect(cohort_data.keys).to eq(%w(B A))
+      end
+    end
+
+    describe '#build_cohort_array' do
+      it 'creates an array of two cohort objects' do
+        expect(cohorts.count).to eq(2)
+        expect(cohorts).to be_a(Array)
```
djkotowski: Is it important that the result be an array, or could it just as easily be any Enumerable?

data_science/analyzer_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        expect(cohorts[1]).to be_a(Cohort)
+      end
+    end
+  end
+
+  describe '#initialize_test' do
+    let(:tester) { analyzer.initialize_test }
+
+    it 'creates an ABAnalyzer::ABTest instance' do
+      expect(tester).to be_a(ABAnalyzer::ABTest)
```
djkotowski: Do you need access to the inner `ABAnalyzer::ABTest`?  Does it make sense for the interface to provide this?

data_science/analyzer_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+    end
+
+    context '#load_cohort_data' do
+      it 'builds a hash to group the data' do
+        expect(cohort_data).to be_a(Hash)
+      end
+
+      it 'groups data under the cohort names' do
+        expect(cohort_data.keys).to eq(%w(B A))
```
djkotowski: Rather than ensuring that the keys equal the collection provided, check to see that they [contain exactly](https://relishapp.com/rspec/rspec-expectations/v/3-3/docs/built-in-matchers/contain-exactly-matcher) the elements provided.

data_science/cohort.rb
djkotowski: For a conversion rate where precision isn't as critical, it doesn't matter, though.

data_science/cohort.rb
```diff
@@ -0,0 +1,44 @@
...
+    n = @sample_size
+    @stnd_error = Math.sqrt(p * (1 - p) / n)
+  end
+
+  def conversion_statistics
+    { conversions: @conversions, non_conversions: @sample_size - @conversions }
+  end
+
+  def print_summary
+    puts "Cohort: #{@name}"
```
djkotowski: Cohort shouldn't be responsible for console output.  You could make this a `summary` method that returns a string, though, and not break the single-responsibility principle.

data_science/cohort_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+describe Cohort do
+  let(:test_file_name) { 'test_data.json' }
+  let(:data) { JSON.parse(File.read(test_file_name)) }
+  let(:cohort_name) { 'A' }
+  let(:cohort_data) do
+    data.select { |row| row["cohort"] == cohort_name }
+      .collect { |row| row["result"] }
+  end
+  let(:cohort) { Cohort.new(cohort_name, cohort_data) }
```
djkotowski: Minor nitpick:  Add blank lines between `do`/`end` blocks and the surrounding code.

data_science/cohort_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      expect(cohort.conversions).to eq(10)
+    end
+  end
+
+  describe 'calculating statistics' do
+    before do
+      cohort.calc_conversion_rate
+      cohort.calc_standard_error
+      cohort.calc_confidence_intervals
+    end
```
djkotowski: Isn't this what `cohort.calculate_statistics` is supposed to do?

data_science/cohort_spec.rb
djkotowski: To expand upon this, think about how the shorthand test will be outputted.  This will read as "it should be within 0.01 of 0.1006" rather than "it is the standard error value of the conversion rate".

As a side note, this also looks like a method that should not be available as part of the public interface.

data_science/cohort_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        it { is_expected.to be_within(0.01).of(0.613176) }
+      end
+    end
+  end
+
+  describe '#conversion_statistics' do
+    let(:stats) { cohort.conversion_statistics }
+
+    it 'creates a hash of conversions and non-conversions' do
+      expect(stats).to be_a(Hash)
```
djkotowski: This check is unnecessary.  As long as it has the `[]` method, does it really matter exactly what structure this is?

djkotowski:
Good first iteration.  Take a look at the comments, refactor and resubmit.  Think more about what classes should have which responsibilities.  Also focus heavily on the public interface.  What should be publicly accessible?  What shouldn't?

Let me or @katieschilling know if you have any questions.

PR #831, opened 2015-07-17
Triangle Facts Refactor

mcdaniel67:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+  def recite_facts
+    puts 'This triangle is equilateral!' if equilateral
+    puts 'This triangle is isosceles!' \
+         ' Also, that word is hard to type.' if isosceles
+    puts 'This triangle is scalene and mathematically boring.' if scalene
+
+    puts_angle_info
+  end
+
+  def equilateral
```
djkotowski: Predicate methods should be suffixed with a `?`

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+    if equilateral || isosceles
+      true
+    else
+      false
+    end
+  end
+
+  def puts_angle_info
+    angles = calculate_angles(side1, side2, side3)
+    puts 'The angles of this triangle are ' + angles.join(',')
```
djkotowski: Use `<<` instead of `+`.  `<<` appends to the original; `+` creates a new string

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+
+    puts 'This triangle is also a right triangle!' if angles.include? 90
```
djkotowski: Wrap method arguments with parentheses (`puts` is one of the exceptions to this rule)

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+  def equilateral
+    side1 == side2 && side2 == side3
+  end
+
+  def isosceles
+    [side1, side2, side3].uniq.length == 2
+  end
 
+  def scalene
+    if equilateral || isosceles
```
djkotowski: Don't use `if` to return true or false.  Just return the boolean expression.

djkotowski:
Nice first iteration.  Take a look at the comments, refactor and resubmit.  You're just about done with this one.

PR #833, opened 2015-07-17
Robot name

mcdaniel67:


mcdaniel67:
Retrospectively, I should have done each exercise independently instead of building on each exercise. I've switched to doing that starting with DinoDex.

arrowhead/arrowhead.rb
```diff
@@ -2,32 +2,34 @@ class Arrowhead
   # This seriously belongs in a database.
```
djkotowski: Again, clean out this branch.

robot_name/robot_name.rb
```diff
@@ -3,30 +3,49 @@ class NameCollisionError < RuntimeError; end
...
     @@registry ||= []
     @name_generator = args[:name_generator]
 
+    check_name_generator_arg
+
+    raise_name_error unless name_ok
+    @@registry << @name
+  end
+
+  def check_name_generator_arg
```
djkotowski: The name implies you're checking an argument, when in fact you're checking an instance variable.  You also generate a name in this step.  Make sure your methods do what they say they do, and only that.

robot_name/robot_name.rb
```diff
@@ -3,30 +3,49 @@ class NameCollisionError < RuntimeError; end
...
     if @name_generator
       @name = @name_generator.call
+      generate_name if already_registered(@name)
```
djkotowski: And what happens if it's already registered?

robot_name/robot_name.rb
```diff
@@ -3,30 +3,49 @@ class NameCollisionError < RuntimeError; end
...
+
+  def already_registered(name)
+    @@registry.include?(name)
+  end
+
+  def raise_name_error
+    raise NameCollisionError, 'There was a problem generating the robot name!'
+  end
+
+  def name_ok
```
djkotowski: Naming is a bit off here.  First, this is a predicate method, so it should be suffixed with a `?`.  Second, `ok` is a very vague term.  Maybe `name_valid?` would be a better name?

robot_name/robot_name.rb
```diff
@@ -3,30 +3,49 @@ class NameCollisionError < RuntimeError; end
...
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def generate_name
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
+    @name = "#{generate_char.call}#{generate_char.call}" \
+      "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  end
+
+  def already_registered(name)
```
djkotowski: This is a predicate method.

robot_name/robot_name.rb
```diff
@@ -3,30 +3,49 @@ class NameCollisionError < RuntimeError; end
...
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+      generate_name
     end
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def generate_name
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
```
djkotowski: These two lambdas never change.  Why not make them methods?

robot_name/robot_name.rb
```diff
@@ -3,30 +3,49 @@ class NameCollisionError < RuntimeError; end
...
+    @name = "#{generate_char.call}#{generate_char.call}" \
+      "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
```
djkotowski: Think about some clearer ways to write this.  As it stands, it is difficult to read.  I have to carefully look to ensure that there are two chars and three nums.

djkotowski:
Nice first iteration.  Take a look at the comments, refactor and resubmit.  You'll also want to clear out the repo.  Let me know if you have any questions.

PR #834, opened 2015-07-20
Dinodex V1

mcdaniel67:
My first draft of the dinodex. Rubocop complains about the initialize method in the Dinosaur class. I'm not entirely sure if that's something that should be broken down farther.

### DIFF
--- a/dino_catalog/csv_parser.rb
+++ b/dino_catalog/csv_parser.rb
@@ -1,0 +1,56 @@
+require_relative('dinosaur.rb')
+
+class DinodexParserCSV
+  def initialize(filepath)
+    parse(filepath)
+  end
+
+  def parse(filepath)
+    @file = open(filepath, 'r')
+    generate_categories
+    generate_data
+    @dinosaurs
+  end
+
+  def generate_categories
+    @categories = clean_row(@file.first)
+    convert_african
+  end
+
+  def generate_data
+    @dinosaurs = []
+    @file.each do |row|
+      row_data = Hash[@categories.zip(clean_row(row))]
+      @dinosaurs << Dinosaur.new(row_data)
+    end
+  end
+
+  def clean_row(row)
+    line = row.split(',')
+    line.each do |value|
+      value.strip!
+      value.capitalize!
+    end
+    line
+  end
+
+  def convert_african
+    convert('Genus', 'Name')
+    convert('Carnivore', 'Diet')
+    convert('Weight', 'Weight_in_lbs')
+  end
+
+  def convert(from, to)
+    index = @categories.index(from)
+    unless index.nil?
+      @categories[index] = to
+    end
+  end
+
+end
+
+dex = DinodexParserCSV.new('dinodex.csv')
+list = dex.parse('african_dinosaur_export.csv')
+list.each do |dino|
+  puts dino
+end
--- a/dino_catalog/csv_parser.rb
+++ b/dino_catalog/csv_parser.rb
@@ -31,7 +31,6 @@ class DinodexParserCSV
       value.strip!
       value.capitalize!
     end
-    line
   end
 
   def convert_african
@@ -42,15 +41,6 @@ class DinodexParserCSV
 
   def convert(from, to)
     index = @categories.index(from)
-    unless index.nil?
-      @categories[index] = to
-    end
+    @categories[index] = to unless index.nil?
   end
-
-end
-
-dex = DinodexParserCSV.new('dinodex.csv')
-list = dex.parse('african_dinosaur_export.csv')
-list.each do |dino|
-  puts dino
 end
--- a/dino_catalog/dinodex.rb
+++ b/dino_catalog/dinodex.rb
@@ -1,0 +1,62 @@
+# Our Dinodex is stored as an array. We perform various array ops to query.
+
+require_relative('csv_parser')
+require_relative('dinosaur.rb')
+require('json')
+
+class Array
+  def query_walking(type)
+    walkers_of_type = []
+    each do |dino|
+      walkers_of_type << dino if dino.walking_type?(type)
+    end
+    walkers_of_type
+  end
+
+  def find_all_carnivores
+    carnivores = []
+    each do |dino|
+      carnivores << dino if dino.carnivore?
+    end
+    carnivores
+  end
+
+  def find_all_in_period(period)
+    in_period = []
+    each do |dino|
+      in_period << dino if dino.in_period?(period)
+    end
+    in_period
+  end
+
+  def find_all_above_weight(weight)
+    above_weight = []
+    each do |dino|
+      above_weight << dino if dino.above_weight?(weight)
+    end
+    above_weight
+  end
+
+  def find_all_below_weight(weight)
+    below_weight = []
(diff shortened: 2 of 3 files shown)


dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,46 @@
...
+    convert_african
+  end
+
+  def generate_data
+    @dinosaurs = []
+    @file.each do |row|
+      row_data = Hash[@categories.zip(clean_row(row))]
+      @dinosaurs << Dinosaur.new(row_data)
+    end
+  end
```
jmmastey: If you turn this method into `@file.map` and just leave a dino on line 24, you'll return an array of dinos from this method. You can even make it `@dinosaurs = @file.map` if you like. Anyway, that lets you delete the weird extraneous `@dinosaurs` on line 12 and everything gets a little cleaner.

dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def clean_row(row)
+    line = row.split(',')
+    line.each do |value|
+      value.strip!
+      value.capitalize!
+    end
+  end
```
jmmastey: Fwiw, the CSV library would be much more appropriate here, though this works.

dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def convert_african
+    convert('Genus', 'Name')
+    convert('Carnivore', 'Diet')
+    convert('Weight', 'Weight_in_lbs')
+  end
```
jmmastey: What about the continent? :)

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,62 @@
+# Our Dinodex is stored as an array. We perform various array ops to query.
+
+require_relative('csv_parser')
+require_relative('dinosaur.rb')
+require('json')
+
+class Array
```
jmmastey: Generally, monkey-patching Array is generally considered to be a bad idea. It works the first few times, but before long you end up having namespace issues. Much better just to have a new class ot use as a container.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,62 @@
...
+  def query_walking(type)
+    walkers_of_type = []
+    each do |dino|
+      walkers_of_type << dino if dino.walking_type?(type)
+    end
+    walkers_of_type
+  end
```
jmmastey:
```
def query_walking(type)
  select { |dino| dino.walking_type?(type) }
end
```

Any time you see the pattern:

```
def my_method
  some_arg
  another_arg.iterate_and_populate
  some_arg
end
```

That pattern can almost always be refactored to use `map`, `inject`, `select`, `reject` or `each_with_object`. Those methods are your friend in ruby.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,62 @@
...
+    above_weight
+  end
+
+  def find_all_below_weight(weight)
+    below_weight = []
+    each do |dino|
+      below_weight << dino if dino.below_weight?(weight)
+    end
+    below_weight
+  end
```
jmmastey: Same w/ all these as the first: `select`.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,62 @@
...
+    in_period
+  end
+
+  def find_all_above_weight(weight)
+    above_weight = []
+    each do |dino|
+      above_weight << dino if dino.above_weight?(weight)
+    end
+    above_weight
+  end
```
jmmastey: Generally, the name `find_all_` is not a great prefix. For methods that don't modify data (known as Queries, as contrasted to those that do modify, called Commands`, we tend to use noun names. What would that look like here? Maybe it's just `above_weight`, maybe not. Either way, see if you can find a more concise name.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,62 @@
...
+puts 'All Carnivores:'
+puts list.find_all_carnivores
+puts 'All dinosaurs in cretaceous period:'
+puts list.find_all_in_period('cretaceous')
+puts 'All large dinosaurs:'
+puts list.find_all_above_weight(2000)
+puts 'All small Dinosaurs (includes dinosaurs without weight):'
+puts list.find_all_below_weight(1000)
+puts 'Chaining example'
+puts list.query_walking('Biped').find_all_carnivores
```
jmmastey: This belongs in its own file. We can't use this library as part of a larger program right now, as it would `puts` a bunch of spam into the logs.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,103 @@
...
+  def initialize(info)
+    @name = info['Name']
+    @period = info['Period']
+    @continent = info['Continent'] || ''
+    setup_diet(info['Diet'])
+    @weight = info['Weight_in_lbs']
+    @walking = info['Walking']
+    @description = info['Description'] || ''
+    capitalize_relevant_members
+  end
```
jmmastey: So to your question on rubocop: this is not bad per se. Some folks use `info.each_pair`, but it's not necessary.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  def setup_diet(diet)
+    @diet = diet unless diet == 'Yes' || diet == 'No'
+    @diet = 'Carnivore' if diet == 'Yes'
+    @diet = 'Herbivore' if diet == 'No'
+  end
```
jmmastey: This is too repetitive, but more importantly this logic doesn't belong in the Dinosaur class. Dinosaurs don't know the vagaries of our data format, and they shouldn't. If I added a second CSV with "Herbivore: Yes", this code would break, and it really shouldn't. Move this into the parser.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  def capitalize_relevant_members
+    @period.split.map(&:capitalize).join(' ')
+    @continent.split.map(&:capitalize).join(' ')
+  end
+
+  def to_s
+    name_to_s + period_to_s + continent_to_s + diet_to_s + \
+      weight_to_s + walking_to_s + description_to_s
+  end
```
jmmastey: This is... valid? But distracting. I'd probably rather have more lines in this method and do something cleverer than to have 40 lines of stringifying code for a model that isn't primarily about stringifying.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,103 @@
...
+    @period.downcase.include?(period.downcase)
+  end
+
+  def above_weight?(weight)
+    @weight.to_i > weight.to_i
+  end
+
+  def below_weight?(weight)
+    @weight.to_i < weight.to_i
+  end
```
jmmastey: There's an off-by-one error here for dinos at exactly the "big" weight.

jmmastey:
Hey Kyle! Good work on this. There are a few changes suggested in the notes, take a look and see if you can make the changes.

Keep it up!

PR #835, opened 2015-07-20
Dino catalog

abake-enova:


dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,216 @@
+require 'json'
+require_relative 'dinoparser'
+
+class Dinosaur
```
jmtai: Classes should be separated - one class/module per file. The naming convention is that class/module names in camel-case map directly to the filename in snake-case. So `module DinodexPrinter` would be stored in a file named 'dinodex_printer.rb'

For these exercises it's not critical (although it makes reviewing a little easier) but when you start working with Rails, the autoloader uses that rule to determine which files to load.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,216 @@
...
+  attr_accessor :data
+
+  def initialize(data = {})
+    @data = data
+  end
+
+  def has?(key)
+    @data.include?(key)
```
jmtai: Since you've got a `data` accessor declared, you don't need to use the `@` when accessing the `data` variable after it has been initialized.

jmtai: For a hash, I'd probably prefer `has_key?` over 'include?` for clarity but either way is fine.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,216 @@
...
+  end
+
+  def name_is?(name)
+    @data['name'] == name if has?('name')
```
jmtai: The `has?` check is unnecessary here. If the key `name` is not found in the data hash, it will just return `nil`, so all you need is the equivalence check.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,216 @@
...
+  end
+
+  def from_period?(period)
+    @data['period'].include?(period) if has?('period')
+  end
+
+  def diet_is?(diet)
+    return carnivore? if diet == 'carnivore'
+    @data['diet'] == diet if has?('diet')
```
jmtai: Again, the `has?` is unnecessary here.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,216 @@
...
+  end
+
+  def carnivore?
+    @data['diet'] != 'herbivore' && !@data['diet'].empty? if has?('diet')
```
jmtai: You can eliminate the `has?` here if instead of using `empty?` you directly check for an empty string:
```ruby
@data['diet'] != 'herbivore' && @data['diet'] != ""
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,216 @@
...
+  end
+
+  def size_is?(size)
+    return false if !has?('weight') || has?('weight') && @data['weight'].empty?
```
jmtai: This is a lot of checking that could be avoided by simply moving the exact weight check to the bottom of the method. Then you'd have this:
```ruby
return Float(@data['weight']) >= 2000 if size == 'big'
return Float(@data['weight']) < 2000 if size == 'small'
@data['weight'] == weight
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,216 @@
...
+    return Float(@data['weight']) >= 2000 if size == 'big'
```
jmtai: Two tons is 4000 pounds, not 2000. Also, we want to avoid having magic numbers embedded in our code. 4000 should be made a constant and declared at the top of the class. e.g.
```ruby
TWO_TONS = 4000
```

jmtai: What if I call `size_is?("SMALL")` or `size_is?("Small") ?
Do you think this check should fail if the case of the input parameter doesn't match?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,216 @@
...
+    return Float(@data['weight']) < 2000 if size == 'small'
+  end
+
+  def walking_is?(walking)
+    @data['walking'] == walking if has?('walking')
```
jmtai: Again, you can eliminate the `has?` here.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,216 @@
...
+  end
+
+  def to_s
+    max_len = @data.keys.max_by(&:length).length
+    @data.each do |k, v|
+      spaces = ' ' * (max_len - k.length) + ' '
+      puts k + ':' + spaces + v if v.length > 0
```
jmtai: In general, prefer using string interpolation instead of `+`, as concatenation creates a new string for each `+` operation.
```ruby
puts "#{k} : #{spaces}#{v}"
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,216 @@
...
+    end
+  end
+end
+
+class Dinodex
+  SEARCH_NAME = lambda do |res, name|
```
jmtai: Ruby's constant syntax should be reserved for constant **values**, not objects (a lambda is a Proc object).

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,216 @@
...
+
+  SEARCH_SIZE = lambda do |res, size|
+    res.select { |dino| dino.size_is?(size) }
+  end
+
+  SEARCH_WALKING = lambda do |res, walking|
+    res.select { |dino| dino.walking_is?(walking) }
+  end
+
+  SEARCH_DISPATCHER = {
```
jmtai: The constant syntax here is appropriate.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,21 @@
...
+dinodex.print_search(diet: 'carnivore')
+
+puts "\nAll the dinosaurs from the cretaceous period:"
+dinodex.print_search(period: 'cretaceous')
+
+puts "\nAll the big dinosaurs:"
+dinodex.print_search(big: true)
+
+puts "\nAll the small dinosaurs:"
+dinodex.print_search(small: true)
```
jmtai: Nice job moving this testing code into a separate file. Now the DinoDex is usable as as standalone library.

Although it's not explicitly stated in the requirements, what we really want is for this app to be interactive. A user should be able to use a command line interface to query the DinoDex without having to write any Ruby code.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,85 @@
...
+    small: search_small,
+    weight: search_weight,
+    walking: search_walking,
+  }
+
+  def initialize
+    @registry = []
+  end
+
+  def load_data(data)
```
jmtai: Naming is tricky and often subjective but we definitely want our method and variables to be as accurate and descriptive as possible. Here your method is named `load_data` but it takes `data` as a parameter. What is this method really doing? The data was already loaded from files elsewhere. If we look at the actually code, the data is being converted into `Dinosaur`s and added to an internal `registry`. Perhaps something like `build_registry` is more appropriate?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,85 @@
...
+    data.each do |row|
+      @registry << Dinosaur.new(row)
+    end
```
jmtai: Any time you are transforming an array of elements to a new array of elements, you're really doing a `map`. Using `map` can reduce this block to a single line of code:
```ruby
@registry = data.map { |row| Dinosaur.new(row) }
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  def load_csv(filename)
+    return unless File.file?(filename)
```
jmtai: Good use of a guard clause here.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,85 @@
...
+    data = DinoTranslator.translate(DinoParser.parse_csv(filename))
```
jmtai: I'd argue that `Translator` and `translate` might not be the best names here since translation carries a strong connotation of converting from one language to another.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,85 @@
...
+    raise InvalidDataError unless DinoValidator.valid_data?(data)
+    load_data(data)
+  end
+
+  def search(args = {}, results = @registry)
+    args.each do |k, v|
```
jmtai: In general, try to avoid using single letter variable names. The goal is to be succinct but expressive. In this case I can see that **k** means **key** and **v** means **value** but why not be a little more descriptive?
```ruby
args.each do |dino_attr, value|
  results = SEARCH_DISPATCHER[dino_attr].call(results, value)
end
```
Do you think someone looking at this code for the first time would have an easier time understanding that version?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,85 @@
...
+      results = SEARCH_DISPATCHER[k].call(results, v)
+    end
+    results
+  end
+
+  def print_search(args = {})
+    res = search(args)
```
jmtai: No need to truncate your variable name here - you're using `results` consistently everywhere else so why not just use `result` here?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,85 @@
...
+    DinodexPrinter.print(res)
+  end
+
+  def export_json(results = @registry)
+    { "dinosaurs" => results.map(&:data) }.to_json
```
jmtai: Nice.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  def load_csv(filename)
+    return unless File.file?(filename)
+    data = DinoTranslator.translate(DinoParser.parse_csv(filename))
+    raise InvalidDataError unless DinoValidator.valid_data?(data)
+    load_data(data)
+  end
+
+  def search(args = {}, results = @registry)
```
jmtai: Note that since you've used `attr_accessor` for your registry, the `@` prefix is unnecessary to read the variable.

dino_catalog/dinodex_printer.rb
```diff
@@ -0,0 +1,75 @@
...
+module DinodexPrinter
+  COLUMN_MAX_LENGTHS = {
+    'name' => 0,
+    'period' => 0,
+    'continent' => 0,
+    'diet' => 0,
+    'weight' => 0,
+    'walking' => 0,
+    'description' => 0,
+  }
```
jmtai: While this is ok, your `DinodexPrinter` is now tightly coupled with your Dinosaur class since these keys correspond directly with the internal keys in the `Dinosaur` class. If `Dinosaur` is modified, this class must also be modified which makes the codebase harder to maintain. Perhaps there's a way to pull these keys directly from a Dinosaur instance?

dino_catalog/dinodex_printer.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def self.build_table_header(columns)
+    header = "|"
+    columns.each do |key, max_len|
+      spaces = " " * (max_len - key.length) + " |"
```
jmtai: It's kind of odd that variable named `spaces` includes a character that isn't a space, no?

dino_catalog/dinodex_printer.rb
```diff
@@ -0,0 +1,75 @@
...
+      header += " " + key + spaces
```
jmtai: Prefer using the shovel operator `<<` and string interpolation when dynamically building strings, as it modifies the original object instead of creating new string objects like the `+` operator.
```ruby
header << " #{key}#{spaces}"
```

dino_catalog/dinodex_printer.rb
```diff
@@ -0,0 +1,75 @@
...
+    end
+    header
+  end
+
+  def self.build_cell(dino, key, max_len)
```
jmtai: Since all your other methods use the `table` descriptor, you should use it here as well for consistency: `build_table_cell`

dino_catalog/dinodex_printer.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def self.build_table_body(subset, columns)
+    body = ""
+    subset.each { |dino| body += build_table_row(dino, columns) }
+    body
+  end
+
+  def self.calc_max_column_lengths(subset)
+    columns_max_len = COLUMN_MAX_LENGTHS
```
jmtai: Nitpick: You're using three different naming schemes here: `max_column_lengths`, `columns_max_len`, and `COLUMN_MAX_LENGTHS`. I'd argue that the second doesn't really read well. Regardless, pick one naming scheme and use it consistently.

dino_catalog/dinodex_printer.rb
```diff
@@ -0,0 +1,75 @@
...
+    " #{data}#{'.' * (max_len - data_len)} |"
+  end
+
+  def self.build_table_row(dino, columns)
+    row = "|"
+    columns.each { |key, max_len| row += build_cell(dino, key, max_len) }
+    row += "\n"
+  end
+
+  def self.build_table_body(subset, columns)
```
jmtai: `subset` isn't a great parameter name. Looking at this for the first time, I'm thinking "subset of what?"  If I then look at the code below, I see that subset is an array of dinos, so it would be much clearer to just name this parameter `dinos`.

dino_catalog/dinodex_printer.rb
```diff
@@ -0,0 +1,75 @@
...
+    body = ""
+    subset.each { |dino| body += build_table_row(dino, columns) }
+    body
```
jmtai: Whenever you see the pattern
variable
some_array.each {}
variable

You can express it more succinctly using `each_with_object`.

dino_catalog/dinodex_printer.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def self.calc_column_lengths(subset)
+    columns_max_len = calc_max_column_lengths(subset)
+    remove_blank_columns(columns_max_len)
+  end
+
+  def self.remove_blank_columns(columns_max_len)
+    columns_max_len.keys.each do |key|
+      columns_max_len.delete key if columns_max_len[key] == 0
```
jmtai: Wrap method parameters in parens. Without them, it's a little hard to parse this line at first glance.

dino_catalog/dinodex_printer.rb
```diff
@@ -0,0 +1,75 @@
...
+      next if columns_max_len[key] == 0
+      v = columns_max_len[key]
+      columns_max_len[key] = v > key.length ? v : key.length
```
jmtai: Here again judicious use of parens would make it easier to parse this line:
```ruby
columns_max_len[key] = (v > key.length) ? v : key.length
```

jmtai: A bigger question is why is a method named `remove_blank_columns` also adjusting the max column lengths? Ideally, we are trying to follow the principle of single responsibility, so our methods should be doing just one thing.

dino_catalog/dinoparser.rb
```diff
@@ -0,0 +1,23 @@
+module DinoParser
+  def self.split_strip_downcase(line)
+    line.split(',').map { |d| d.strip.downcase }
```
jmtai: One problem with this method is you've now lost the original formatting from the source files and can never get it back. Since the source contains mixed case strings, it would be somewhat surprising to me as a user that I have no way of printing out the data with the original formatting intact.

dino_catalog/dinoparser.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def self.grab_header
+    split_strip_downcase(@f.first)
```
jmtai: Again, avoid single letter variable names. If I'm reading this file from top the bottom, the first time I encounter this variable `@f` I have no idea what it is. I have to jump to the bottom of the file to figure out what's going on.

dino_catalog/dinoparser.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def self.grab_data
```
jmtai: `grab_data` doesn't really tell me what this method is doing. Can you think of a better name?

dino_catalog/dinoparser.rb
```diff
@@ -0,0 +1,23 @@
...
+    dino_data = []
+    @f.each do |line|
+      dino_data << Hash[@header.zip(split_strip_downcase(line))]
+    end
+    dino_data
```
jmtai: This is the pattern mentioned earlier that can be simplified by using `each_with_object`.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,57 @@
+class Dinosaur
+  TWO_TONS = 4000
+  SMALL = 2000
```
jmtai: Hmmm... why is `SMALL` defined as 2000? If big is greater than TWO_TONS and small is less than 2000, what happens in between?

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,57 @@
...
+    return carnivore? if diet == 'carnivore'
+    data['diet'] == diet
+  end
+
+  def carnivore?
+    data['diet'] != 'herbivore' && data['diet'] != ""
+  end
+
+  def big?(b)
+    return unless data['weight']
```
jmtai: Predicate methods should return `true` or `false`. Here the method is returning `nil`.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,57 @@
...
+    b == (Float(data['weight']) > TWO_TONS)
```
jmtai: Why not just this?
```ruby
Float(data['weight']) > TWO_TONS
```
What's the purpose of the boolean input parameter? A dinosaur is either big or small, so calling `big?(false)` would just duplicate the functionality of the `small?` method.

dino_catalog/dinotranslator.rb
```diff
@@ -0,0 +1,44 @@
+module DinoTranslator
+  HEADER_MAP = {
+    "genus" => "name",
+    "carnivore" => "diet",
+    "weight_in_lbs" => "weight",
+  }
+
+  def self.translate_key(hash, key)
+    hash[HEADER_MAP[key]] = hash.delete key if HEADER_MAP.include?(key)
```
jmtai: This is not super readable. Sometimes it's ok to use temporary variables if it makes the code easier to understand.
Do you think this is clearer?
```ruby
if HEADER_MAP.include?(key)
  translated_key = HEADER_MAP[key]
  value = hash.delete(key)
  hash[translated_key] = value
end
```

dino_catalog/dinotranslator.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def self.translate_carnivore_key(h)
+    carnivore_map = {
```
jmtai: This can be declared as a constant outside of this method since it's never modified.

dino_catalog/dinotranslator.rb
```diff
@@ -0,0 +1,44 @@
+module DinoTranslator
+  HEADER_MAP = {
+    "genus" => "name",
+    "carnivore" => "diet",
+    "weight_in_lbs" => "weight",
+  }
+
+  def self.translate_key(hash, key)
```
jmtai: In general, try to avoid using data types as or in variable names. Since this hash represents a dino object, just call it that.

dino_catalog/dinotranslator.rb
```diff
@@ -0,0 +1,44 @@
...
+    hash[HEADER_MAP[key]] = hash.delete key if HEADER_MAP.include?(key)
+  end
+
+  def self.translate_carnivore_key(h)
+    carnivore_map = {
+      "yes" => "carnivore",
+      "no" => "",
+    }
+    h["carnivore"] = carnivore_map[h["carnivore"]] if h.include?("carnivore")
```
jmtai: This could also be cleaned up a bit.

dino_catalog/dinotranslator.rb
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  def self.translate_key(hash, key)
+    hash[HEADER_MAP[key]] = hash.delete key if HEADER_MAP.include?(key)
+  end
+
+  def self.translate_carnivore_key(h)
+    carnivore_map = {
+      "yes" => "carnivore",
+      "no" => "",
```
jmtai: If a dinosaur is not a carnivore, should it be left blank? It might be safe to assume that if it doesn't eat meat, it eats plants (i.e. it's an herbivore).

dino_catalog/dinotranslator.rb
```diff
@@ -0,0 +1,44 @@
...
+    "genus" => "name",
+    "carnivore" => "diet",
+    "weight_in_lbs" => "weight",
+  }
+
+  def self.translate_key(hash, key)
+    hash[HEADER_MAP[key]] = hash.delete key if HEADER_MAP.include?(key)
+  end
+
+  def self.translate_carnivore_key(h)
```
jmtai: Is this method translating the carnivore key itself or the data?

dino_catalog/dinotranslator.rb
```diff
@@ -0,0 +1,44 @@
...
+    }
+    h["carnivore"] = carnivore_map[h["carnivore"]] if h.include?("carnivore")
+  end
+
+  def self.translate_empty_weight(h)
+    if h.include?("weight")
+      h.delete("weight") if h["weight"].length == 0
+    elsif h.include?("weight_in_lbs")
+      h.delete("weight_in_lbs") if h["weight_in_lbs"].length == 0
+    end
```
jmtai: Couldn't you avoid the need to check for multiple header values here if you performed the header translation first?

dino_catalog/dinotranslator.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def self.translate_header(data)
+    data.each { |h| h.keys.each { |k| translate_key(h, k) } }
```
jmtai: Seeing nested iterators should definitely be a red flag. For a small set of keys it won't matter but for larger data sets this will be very, very slow.

jmtai: Perhaps there's a better place to perform the header translation?

dino_catalog/dinovalidator.rb
```diff
@@ -0,0 +1,42 @@
...
+  DISPATCHER = {
+    "name" => valid_name,
+    "period" => valid_period,
+    "continent" => valid_continent,
+    "diet" => valid_diet,
+    "weight" => valid_weight,
+    "walking" => valid_walking,
+  }
+
+  def self.validate(k, v)
```
jmtai: This is returning `true` or `false` so it should probably be a predicate method.

dino_catalog/dinovalidator.rb
```diff
@@ -0,0 +1,42 @@
...
+    return true unless DISPATCHER.include?(k)
+    DISPATCHER[k].call(v)
+  end
+
+  def self.valid_row?(row)
+    row.inject(0) { |valid, (k, v)| valid && validate(k, v) }
```
jmtai: Using inject here makes it much harder to understand what this method is doing. The `all?` method from the `Enumerable` module can help clean this up:
```ruby
row.all? { |key, value| validate(key, value) }
```
If you rename the validate method, it becomes even clearer:
```ruby
row.all? { |key, value| valid?(key, value) }
```

dino_catalog/dinovalidator.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def self.valid_data?(data)
+    data.inject(0) { |valid, row| valid && valid_row?(row) }
```
jmtai: Ditto for the inject here.

jmtai:
Looking good! Read over the comments, refactor a bit (and add an interactive command line interface), and resubmit when ready.

PR #836, opened 2015-07-20
Triangle refactor V2

mcdaniel67:
I have fixed what was noted on the last PR.

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -8,6 +8,15 @@ class Triangle
     @side3 = side3
   end
 
+  def recite_facts
+    puts 'This triangle is equilateral!' if equilateral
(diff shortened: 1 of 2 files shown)


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,60 @@
...
+  def scalene?
+    equilateral? || isosceles?
+  end
+
+  def puts_angle_info
+    angles = calculate_angles(side1, side2, side3)
+    puts 'The angles of this triangle are ' << angles.join(',')
+
+    puts 'This triangle is also a right triangle!' if angles.include?(90)
+    puts ''
```
djkotowski: Forgot to mention this.  No need for an empty string on `puts`.  It accepts 0 args if you just want a newline.

djkotowski:
Looks good.  You're done here!

PR #838, opened 2015-07-20
Robot name V2

mcdaniel67:


robot_name/robot_name.rb
```diff
@@ -3,30 +3,62 @@ class NameCollisionError < RuntimeError; end
...
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+      generate_name
     end
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def generate_name
+    @name = "#{generate_char}" \
```
djkotowski: I recommend trying something else to DRY this out.  What about something like?

```ruby
def generate_name
  name = ""
  2.times { name << generate_char }
  3.times { name << generate_num.to_s }
  name
end
```

robot_name/robot_name.rb
```diff
@@ -3,30 +3,62 @@ class NameCollisionError < RuntimeError; end
...
+            "#{generate_char}" \
+            "#{generate_num}" \
+            "#{generate_num}"\
+            "#{generate_num}"
+  end
+
+  def generate_new_name
+    generate_name
```
djkotowski: Why this abstraction?

robot_name/robot_name.rb
```diff
@@ -3,30 +3,62 @@ class NameCollisionError < RuntimeError; end
 class Robot
   attr_accessor :name
 
-  @@registry
-
   def initialize(args = {})
     @@registry ||= []
     @name_generator = args[:name_generator]
 
+    check_name_instance_and_generate_name
```
djkotowski: If your method has an `and` in it, it's probably doing too much.

robot_name/robot_name.rb
```diff
@@ -3,30 +3,62 @@ class NameCollisionError < RuntimeError; end
...
+
+    raise_name_error unless name_valid?
+    @@registry << @name
+  end
+
+  def check_name_instance_and_generate_name
     if @name_generator
       @name = @name_generator.call
+      generate_new_name if already_registered?(@name)
```
djkotowski: Why not call the provided generator again?  This seems a bit counter-intuitive.

robot_name/robot_name.rb
```diff
@@ -3,30 +3,62 @@ class NameCollisionError < RuntimeError; end
...
+  def generate_char
+    ('A'..'Z').to_a.sample
+  end
+
+  def generate_num
+    rand(10)
+  end
+
+  def already_registered?(name)
+    @@registry.include?(name)
```
djkotowski: Nice abstraction.

robot_name/robot_name.rb
```diff
@@ -3,30 +3,62 @@ class NameCollisionError < RuntimeError; end
...
-  @@registry
-
   def initialize(args = {})
     @@registry ||= []
     @name_generator = args[:name_generator]
 
+    check_name_instance_and_generate_name
+
+    raise_name_error unless name_valid?
+    @@registry << @name
```
djkotowski: Generally prefer class instance vars to class vars.

Class instance vars = available to this class and all its instances
Class vars = available to this class, its subclasses, and all their instances

djkotowski:
Nice job.  No need to resubmit.  I would take a look at the comments, though, and figure out how you would refactor and simplify this.

PR #840, opened 2015-07-20
Dino catalog

jklemen:


dino_catalog/dino.rb
```diff
@@ -0,0 +1,72 @@
+class Dino
+  ATTRIBUTES = %w(name period continent diet weight walking description)
+
+  attr_accessor *ATTRIBUTES
+
+  def initialize(data)
+    data.keys.each do |attr|
+      send("#{attr}=", data[attr])
+    end
+  end
```
jmmastey: You can grab `key` and `value` both with `Hash#each`. No need to iterate keys specifically. It'll be slightly shorter and a more common pattern people will understand.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def properties
+    Hash[ATTRIBUTES.map { |attr| [attr, send(attr)] }]
+  end
```
jmmastey: I squint my eyes at this one a little bit for being both clever and dense. That Hash initializer is always a little weird. What about:

```
ATTRIBUTES.each_with_object({}) { |attr, hash| hash[attr] = send(attr) }
```

dino_catalog/dino.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def to_s
+    properties.each.inject("") do |str, tuple|
+      fattr = "#{tuple[0].capitalize}:".ljust(20)
+      fval = tuple[1].nil? ? '---' : tuple[1]
+
+      str << "#{fattr}#{fval}\n"
+    end
+  end
```
jmmastey: The `.each` here should be superfluous. On 19, what if it's an empty string? (I haven't read all the way through, maybe you filter it elsewhere.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,72 @@
...
+  def compare(field, with)
+    myval = send(field)
+    if !(myval && with)
+      myval.__id__ <=> with.__id__
+    elsif field == 'weight'
+      myval.to_i <=> with.to_i
+    else
+      myval.downcase <=> with.downcase
+    end
+  end
```
jmmastey: So... this is more complex than I would hope. The case on 27 is confusing, and `weight` being a special case is not awesome. I see what you were doing with the name `with`, but outside the method sig, I'm not sure it's super understandable.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,72 @@
...
+    !less_than?(field, value)
+  end
+
+  def not_equal?(field, value)
+    !equal?(field, value)
+  end
+
+  def not_like?(field, value)
+    !like?(field, value)
+  end
```
jmmastey: Overall, these methods don't have a lot to do with Dinosaurs, but with generic search. If they're going to exist, it should probably be elsewhere (like a module that's mixed in, or a specific search class).

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,102 @@
...
+    Dinodex.new(match)
+  end
+
+  def not_like(field, value)
+    match = @dinos.select do |dino|
+      dino.not_like?(field, value)
+    end
+
+    Dinodex.new(match)
+  end
```
jmmastey: Lots of duplication here. This is probably the most ripe for refactoring so far. Find a way not to write this same method eight times.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,102 @@
...
+      a.compare(field, b.send(field))
+    end
+    Dinodex.new(sorted)
+  end
+
+  def to_json
+    {
+      'dinos' => @dinos.map(&:properties)
+    }.to_json
+  end
```
jmmastey: Nice

dino_catalog/dinoparse.rb
```diff
@@ -0,0 +1,38 @@
...
+
+class Dinoparse
+  attr_reader :dinos
+
+  def initialize(dinofile)
+    dino_data = parse(dinofile)
+    @dinos = dino_data.map { |dino_dat| Dino.new(dino_dat) }
+  end
+
+  private
```
jmmastey: I like that most of this class is private.

dino_catalog/dinoparse.rb
```diff
@@ -0,0 +1,38 @@
...
+  
+  def filter(d_attrs)
```
jmmastey: Filter is not a great name for this. What else could you call it that's more descriptive (even if more verbose)?

jmmastey:
Methods are short and sweet, well done. Two notes that don't sit anywhere in the code:

1. Right now African dinos have no continent from the looks of it. You're going to have to be cleverer about parsing for that one.
2. Your repo has commits from the first exercise in it. You should delete your local `master` and set up a new one tracking upstream.

Feel free to move on if Dan doesn't disagree, this is solid. Let me know if notes are confusing.

jklemen:
Great comments, I'll submit a final revision and find a way to DRY out those methods and update my master.

PR #841, opened 2015-07-21
Data science

abake-enova:


data_science/analyzer.rb
```diff
@@ -0,0 +1,27 @@
+require 'abanalyzer'
+
+class Analyzer
+  attr_accessor :cohorts, :abanalyzer, :p_value
+
+  def initialize(cohorts, p_value = 0.05)
```
johnmcconnell: If you are using ruby 2 or higher you can support keyword args and default values. Example:
```ruby
def add(_x = 0, _y = 0, x: _x, y: y)
   x + y
end

add(1, 2) # => 3
add(x: 3, y: 1) # => 4
```

data_science/analyzer.rb
```diff
@@ -0,0 +1,27 @@
...
+    @cohorts = cohorts
+    @p_value = p_value
+    initialize_abanalyzer
```
johnmcconnell: Why are we forcing the program to initialize here? Don't force data to be stored until it is necessary. Remove initialize_abanalyzer and instead have
```ruby
def abanalyzer
   @abanalyzer ||= Hash.new do |hash, key|
      hash[key] = values[key].conversion_stats
    end
end
```

data_science/analyzer.rb
```diff
@@ -0,0 +1,27 @@
...
+    @cohorts.each { |k, v| values[k] = v.conversion_stats }
+    @abanalyzer = ABAnalyzer::ABTest.new(values)
+  end
+
+  def better_than_random?
+    @abanalyzer.chisquare_p < @p_value
+  end
+
+  def print_results
+    @cohorts.values.each(&:print_info)
```
johnmcconnell: Methods with side effects are using better split up into methods with no side effect and methods which just perform the side effect.
Instead of:
```ruby
@cohorts.values.each(&:print_info)
```
```ruby
puts @cohorts.values.map(&:info)
```

data_science/cohort.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def calculate_standard_error
+    p = calculate_conversion_rate
+    n = @sample_size.to_f
+    Math.sqrt(p * (1 - p) / n)
+  end
+
+  def calculate_confidence_interval
+    lower_limit = calculate_conversion_rate - calculate_standard_error
+    upper_limit = calculate_conversion_rate + calculate_standard_error
```
johnmcconnell: Okay none of these values change after initialization. Cache the value in an instance variable and use `||=` (just be careful of false values. You will notice that most of your method named `calculate_(some_concept_here)` will just become the concept itself after doing this.
For example,
```ruby
def standard_error
   @standard_error ||= begin
     p = conversion_rate
     n = @sample_size.to_f
     Math.sqrt(p * (1 - p) / n)
   end
end

File: data_science/dataloader.rb [hunk 5]
```diff
@@ -0,0 +1,23 @@
...
+module DataLoader
+  def self.parse_json(filename)
+    return {} unless File.file?(filename)
+    data = JSON.parse(File.read(filename))
+    cohort_data = data.group_by { |r| r["cohort"] }
+    cohort_data.each do |k, v|
+      cohort_data[k] = v.map { |r| r["result"] }
+    end
+    cohort_data
+  end
```
johnmcconnell: You did a nice job trying to hide it but this method really looks like.
```ruby
 def self.parse_json(filename)
    return {} unless File.file?(filename)
    data = JSON.parse(File.read(filename))
    cohort_data = data.group_by do |r|
      r["cohort"]
    end
    cohort_data.each do |k, v|
      cohort_data[k] = v.map do |r|
        r["result"]
      end
    end
    cohort_data
  end
```

data_science/dataloader.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def self.load_json(filename)
+    cohorts = {}
+    cohort_data = parse_json(filename)
+    cohort_data.each do |k, v|
+      cohorts[k] = Cohort.new(k, v)
+    end
+    cohorts
+  end
+end
```
johnmcconnell: Okay, instead of:
```ruby
class A
   def self.hello
      'hello'
   end
end
A.hello # => 'hello'
```
Use:
```ruby
class B
   class << self
     def hello
       'hello'
     end
   end
end
```

data_science/spec/analyzer_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'spec_helper'
+
+describe Analyzer do
+  let(:filename) { "test_data.json" }
+  let(:cohorts) { DataLoader.load_json(filename) }
+  let(:analyzer) { Analyzer.new(cohorts) }
+
+  describe "#initialize" do
+    it "takes one argument and returns an Analyzer object", happy: true do
+      expect(analyzer).to be_a(Analyzer)
```
johnmcconnell: glad you are practicing being real careful in your test, but you just initialized the guy 4 lines above. If
```ruby
Analyzer.new(cohorts)
```
does not return an Analyzer, I think we have bigger problems.

data_science/spec/analyzer_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      expect(analyzer.cohorts.size).to eq(2)
+    end
+  end
+
+  describe "#initialize_abanalyzer" do
+    it "takes zero arguments and initializes the ABAnalyzer", happy: true do
+      expect(analyzer.abanalyzer).to be_a(ABAnalyzer::ABTest)
```
johnmcconnell: Same as above. Tests like:
```ruby
x.y = 2
expect(x.y).to be 2
```
do not return nearly as much value as something that tests business logic. These tests can even be destructive. What if we don't like the naming convention and we want to change it. Now, our class names are hard coded in our test suite. Wow, what a nightmare.

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  let(:cohort_name) { "A" }
+  let(:cohort_data) do
+    data.select { |d| d["cohort"] == cohort_name }.map { |d| d["result"] }
+  end
+  let(:cohort) { Cohort.new(cohort_name, cohort_data) }
+  let(:conversion_stats) { cohort.conversion_stats }
+
+  describe "#initialize" do
+    it "takes two parameters and returns a Cohort object", happy: true do
+      expect(cohort).to be_a(Cohort)
```
johnmcconnell: Stop please....

johnmcconnell: def self.parse_json(filename)
    return {} unless File.file?(filename)
    data = JSON.parse(File.read(filename))
    cohort_data = data.group_by { |r| r["cohort"] }
    cohort_data.each do |k, v|
      cohort_data[k] = v.map { |r| r["result"] }
    end
    cohort_data
  end

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      expect(cohort.name).to eq("A")
+      expect(cohort.sample_size).to eq(30)
+    end
+  end
+
+  describe "#calculate_conversion_rate" do
+    it "should have a conversion rate of about 23.34%", happy: true do
+      expect(cohort.calculate_conversion_rate).to be_within(0.01).of(0.234)
```
johnmcconnell: Good stuff

data_science/analyzer.rb
paulghaddad: Excellent suggestion.

data_science/analyzer.rb
paulghaddad: +1

data_science/cohort.rb
```diff
@@ -0,0 +1,38 @@
...
+    { converts: @conversions, nonconverts: @sample_size - @conversions }
+  end
+
+  def print_info
+    puts "Cohort: #{@name}"
+    puts "Sample Size: #{@sample_size}"
+    puts "Conversions: #{@conversions}"
+    puts "Conversion Rate #{calculate_conversion_rate}"
+    puts "Confidence Interval #{calculate_confidence_interval}"
+    puts
```
paulghaddad: You can just put `\n` at the end of line 35.

data_science/dataloader.rb
```diff
@@ -0,0 +1,23 @@
...
+    cohort_data.each do |k, v|
+      cohort_data[k] = v.map { |r| r["result"] }
+    end
+    cohort_data
+  end
+
+  def self.load_json(filename)
+    cohorts = {}
+    cohort_data = parse_json(filename)
+    cohort_data.each do |k, v|
```
paulghaddad: Consider the naming of your block variables.

data_science/spec/analyzer_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+  end
+
+  describe "#better_than_random?" do
+    it "returns true if the p-value is below 0.05", happy: true do
+      expect(analyzer.better_than_random?).to eq(true)
```
paulghaddad: Look at this matcher:

`expect(actual).to be true`

Check out: https://github.com/rspec/rspec-expectations

data_science/spec/cohort_spec.rb
paulghaddad: +1 Great matcher here.

paulghaddad:
Take a look at the comments here and refactor. You should be good to go after that.

Nice work overall!

PR #844, opened 2015-07-21
Overlord v1

ssapra:
Initial pass at bomb. Relied on AJAX and bootstrap for clean experience


overlord/bomb.rb
```diff
@@ -0,0 +1,66 @@
+class Bomb
+  OFF = 1
```
schepedw: Think about what this does to the external interface to this class. A fellow programmer might be confused to find that `Bomb::OFF` = 1. You also expose state via attr_accessor, meaning that `Bomb.new.state` would give 1 or 2 - also a bit confusing.

Consider using a BombState class, such that `Bomb.new.state` would yield a BombState instance that would have one of the 5 states enumerated here

overlord/bomb.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def initialize(activation_code = '1234', deactivation_code = '0000')
+    @activation_code = activation_code
+    @deactivation_code = deactivation_code
+    @state = valid_codes? ? BOOTED : OFF
+    @attempts = DEFAULT_NUM_ATTEMPTS
+  end
+
+  def valid_codes?
+    @activation_code =~ /^[0-9]+$/ && @deactivation_code =~ /^[0-9]+$/
```
schepedw: For error messaging, might be nice to know which was invalid - activation code or deactivation code

overlord/features/deactivate.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Deactivating bombs
+  As a visitor to the site
+  I want to deactivate the bomb
+  Because I need to save the world
+
+  Scenario: Bomb is deactivated
```
schepedw: You're in `deactive.feature`, 'Bomb is deactivated' isn't particularly descriptive. Describe what you're doing a little more - 'Deactivating with default config', perhaps?

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,86 @@
...
+end
+
+When(/^I do nothing$/) do
+end
+
+When(/^I boot the bomb$/) do
+  click_button('Boot')
+end
+
+Then(/^the bomb should be booted$/) do
```
schepedw: I like that you've put the bomb status in the same place each time, it simplifies steps.

You could boil it down even further. Use a capture expression at the end of 'the bomb should be booted', then map the bomb state to the words you expect to see.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,125 @@
+require_relative '../bomb.rb'
+
+describe Bomb do
+  describe '#initialize' do
+    context 'with default codes' do
+      let(:bomb) { Bomb.new }
+      it 'is booted correctly' do
+        expect(bomb).to be_booted
```
schepedw: Nice use of rspec predicate method magic :+1:

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+      it 'is not deactivated' do
+        expect(bomb).to_not be_deactivated
+      end
+
+      it 'has 2 attempts left' do
+        expect(bomb.attempts).to eq(2)
+      end
+    end
+
+    context 'with the 3 failed attempts' do
```
schepedw: Instead of 3 one line `before` blocks, just use 1 with `do` .. `end` over 3 lines

overlord/bomb.rb
djkotowski: Also, prefer symbols over numeric values for these enumerations.

overlord/bomb.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def update_codes(params)
```
djkotowski: `codes` may make more sense than `params` here.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,86 @@
+Given(/^I visit the home page$/) do
+  visit('/')
+end
+
+When(/^I do nothing$/) do
```
djkotowski: Remove empty steps.

overlord/features/boot.feature
```diff
@@ -0,0 +1,58 @@
+Feature: Dropping bombs
+  As a visitor to the site
+  I want to boot and activate the bomb
+  Because I need to be a supervillain
+
+  Scenario: Bomb is not booted
+    Given I visit the home page
+    When I do nothing
```
djkotowski: Remove unnecessary steps.  This should be just
```ruby
When I visit the home page
Then the bomb is off
```

overlord/features/boot.feature
```diff
@@ -0,0 +1,58 @@
...
+    Then the bomb should be off
```
djkotowski: Avoid using the word "should".  It isn't confident.

overlord/features/deactivate.feature
```diff
@@ -0,0 +1,45 @@
...
+    Given the bomb is booted with default config
+    And the bomb is activated with default config
+    When I submit the deactivation code "3333"
+    Then I see a deactivation error message
+
+  Scenario: Wrong deactivation code first
+    Given the bomb is booted with default config
+    And the bomb is activated with default config
+    When I submit the deactivation code "3333"
+    When I submit the deactivation code "0000"
```
djkotowski: **Nitpick:**  `And` instead of `When`.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,86 @@
...
+  click_button('Boot')
+end
+
+Then(/^the bomb should be booted$/) do
+  using_wait_time 10 do
+    expect(find('.bomb_status').text).to eq("Booted up!")
+  end
+end
+
+When(/^I configure the activation code "([^"]*)"$/) do |code|
```
djkotowski: Try sorting your step definitions by Given/When/Then for easy searching.

overlord/views/index.haml
```diff
@@ -0,0 +1,41 @@
+!!!
+%html
+  %head
+    %title Overlord
+    %script{src: '//code.jquery.com/jquery-1.11.3.min.js'}
```
djkotowski: It's good practice to put scripts at the end of the HTML file.  This speeds up page loading for slower clients, allowing them to see the UI from the HTML tags before the browser starts loading the JS.

overlord/features/boot.feature
```diff
@@ -0,0 +1,58 @@
+Feature: Dropping bombs
+  As a visitor to the site
+  I want to boot and activate the bomb
+  Because I need to be a supervillain
```
djkotowski: From this perspective, the visitor _is_ the supervillain.  No need to be generic in the role section ("As a _____")

djkotowski:
Nice first iteration.  Take a look at the comments, refactor and resubmit.  Let me know if you have any questions.

PR #845, opened 2015-07-21
Dino catalog

jklemen:


dino_catalog/dino_test.rb
```diff
@@ -0,0 +1,102 @@
+require './dinodex.rb'
+require './dino.rb'
+require './dino_parse.rb'
+
+class DinoTest
+  attr_accessor :dinodex
+
+  def initialize
+    file1 = 'dinodex.csv'
```
katieschilling: I should be able to pass in a file name and have it parsed.

jklemen: That's what DinoParse does. DinoTest shows that the queries in the README are all possible. Sorry for any confusion.

jmmastey:
Looks great, move on! :D :dancer:

PR #849, opened 2015-07-21
Data science first iteration 

lmenova:
run main.rb


data_science/lib/ab_tester.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def initialize(all_visits)
+    @all_visits = all_visits
+  end
+
+  def conversion_rate(visits)
+    conversions(visits) / visits.size.to_f
+  end
+
+  def cohort_by_name(cohort_name)
```
zsyed1: Line 14: A better name might be:

```
def visits_by(cohort)
  all_visits.select ...
end
```

data_science/lib/ab_tester.rb
```diff
@@ -0,0 +1,36 @@
...
+    visits.count { |visit| visit["result"] == 1 }
+  end
+
+  def chi_square(cohort_a, cohort_b)
+    values = {}
+    values[:cohort_a] = create_values_for_chi_square(cohort_a)
+    values[:cohort_b] = create_values_for_chi_square(cohort_b)
+    tester = ABAnalyzer::ABTest.new values
+    (1 - tester.chisquare_p) * 100
+  end
```
zsyed1: Lines 23 - 27: These seem to be doing a lot of things. It might be better to split them into (at least) two methods. Also, is there really need to pass in the two cohorts?

```
def chi_square
  tester = ABAnalyzes::ABTest.new(values)
  (1 - tester.chisquare_p) * 100
end
```
and move values into its own method:
```
def values
  {
    cohort_a: create_values_for_chi_square(visits_for('A'),
    cohort_a: create_values_for_chi_square(visits_for('B'),
  }
end
```

data_science/lib/cohort_data.rb
```diff
@@ -0,0 +1,17 @@
...
+
+class CohortData
+  attr_accessor :cohort_visits, :file_name
+
+  def initialize(file_name)
+    @file_name = file_name
+  end
+
+  def load_json_data_into_array
+    puts "File #{file_name} not found" unless File.exist?(file_name)
```
zsyed1: Line 11: At this point you want to raise an error `raise "FIle #{filename} not found" unless File.exist?(file_name)"`

data_science/lib/cohort_data.rb
```diff
@@ -0,0 +1,17 @@
...
+    file = File.read(file_name)
+    @cohort_visits = JSON.parse(file)
+  rescue
```
zsyed1: Instead of rescuing can you let the `JSON` package raise the error and just erase lines 14 -16?

data_science/lib/ab_tester.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def create_values_for_chi_square(visits)
```
zsyed1: You can shorten the name to `values_for_chi_square`.

data_science/lib/cohort_data.rb
```diff
@@ -0,0 +1,17 @@
+require 'json'
+
+class CohortData
+  attr_accessor :cohort_visits, :file_name
+
+  def initialize(file_name)
```
zsyed1: Why not call the load method directly from initialize and make the former private. Is there ever a time when the user of this class would initialize and _not_ call `load_json_data_into_array`?

data_science/lib/cohort_data.rb
```diff
@@ -0,0 +1,17 @@
...
+    @file_name = file_name
+  end
+
+  def load_json_data_into_array
```
zsyed1: You can shorten this method name to `load_json`. The `...data...` part is redundant. The `...into_array` exposes the internal design which is better to hide. After all, what if you decide to change the internal data structure to something other than an array?

data_science/lib/cohort_data.rb
```diff
@@ -0,0 +1,17 @@
+require 'json'
+
+class CohortData
+  attr_accessor :cohort_visits, :file_name
```
zsyed1: The class name is `CohortData` so the name `cohort_visits` can be reduced to `visits`. The `cohort` is redundant.

data_science/spec/ab_tester_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+      { 'date' => '2014-03-18', 'cohort' => 'B', 'result' => 1 },
+      { 'date' => '2014-03-18', 'cohort' => 'B', 'result' => 0 },
+      { 'date' => '2014-03-18', 'cohort' => 'B', 'result' => 1 },
+      { 'date' => '2014-03-19', 'cohort' => 'B', 'result' => 0 },
+      { 'date' => '2014-03-19', 'cohort' => 'B', 'result' => 1 },
+      { 'date' => '2014-03-20', 'cohort' => 'B', 'result' => 1 },
+      { 'date' => '2014-03-20', 'cohort' => 'B', 'result' => 1 },
+      { 'date' => '2014-03-20', 'cohort' => 'B', 'result' => 1 },
+    ]
+  end
```
zsyed1: You shouldn't have to explicitly create all three arrays here. The first is just the concatenation of the the other two:

```
  let(:cohort_visits) { cohort_a + cohort_b }
```

zsyed1: Btw I like how you are creating your own data-sets to test the functionality!

data_science/spec/ab_tester_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+  let(:ab_tester) { ABTester.new(cohort_visits) }
+  describe '#initialize' do
+    it 'accepts and assigns arg of data array to variable' do
+      expect(ab_tester.all_visits).to eq cohort_visits
+    end
+  end
+
+  describe '#conversion_rate' do
+    it 'computes conversion rate for cohort A' do
+      expect(ab_tester.conversion_rate(cohort_a)).to eq 3 / 20.to_f
```
zsyed1: Line 111: I like the fraction 3 / 20 here. You can also use a floating point value (in this case 0.15) but have to be careful of errors the tail end of the decimal. For that you can use RSpec's `witihin`:
```
expect (ab_tester....).to be_within(1e-8).of(0.15)
```
Here `1e-8 = 0.00000001`

data_science/spec/ab_tester_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+    end
+    it 'computes conversion rate for cohort B' do
+      expect(ab_tester.conversion_rate(cohort_b)).to eq 10 / 22.to_f
+    end
+  end
+
+  describe '#cohort_by_name' do
```
zsyed1: See comment above about making this method private.

data_science/spec/ab_tester_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+    it 'creates a subset of cohorts based on name' do
+      expect(ab_tester.cohort_by_name('A')).to match_array(cohort_a)
+    end
+  end
+
+  describe '#conversions' do
+    it 'sums of cohort A where result = 1' do
+      expect(ab_tester.conversions(cohort_a)).to eq 3
```
zsyed1: Instead of passing in the data hashes, they should really stay inside the implementation of `ABTester`. Notice you are instantiating the object with all the data. Then you are extracting the data for each cohort. Then you are resubmitting the extracted data. That data should all live inside the class. Hit me up on hipchat @zamir if you want to discuss.

data_science/spec/ab_tester_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+    end
+    it 'sums of cohort B where result = 1' do
+      expect(ab_tester.conversions(cohort_b)).to eq 10
+    end
+  end
+
+  describe '#chi_square' do
+    it 'Confidence level the current leader is in fact better than random.' do
```
zsyed1: Better description `it 'computes the correct chi_square significance' do`

data_science/spec/cohort_data_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  let(:cohort_data) { CohortData.new(file) }
+  describe '#intialize' do
+    it 'loads cohort data' do
+      cohort_data.load_json_data_into_array
+      expect(cohort_data.cohort_visits.count).not_to eq 0
+    end
+  end
+
+  describe '#load_json_data_into_array' do
+    it 'returns an array from a json file' do
```
zsyed1: See comments above about hiding internals. The `Array` is only needed by the software internally and probably shouldn't be exposed publicly.

data_science/spec/spec_helper.rb
```diff
@@ -0,0 +1,99 @@
...
+  #     --seed 1234
+  config.order = :random
+
+  # Seed global randomization in this process using the `--seed` CLI option.
+  # Setting this allows you to use `--seed` to deterministically reproduce
+  # test failures related to randomization by passing the same `--seed` value
+  # as the one that triggered the failure.
+  Kernel.srand config.seed
+=end
+end
```
zsyed1: Nice work implementing the data-science exercise!
I would recommend another round of submission since this is early on in the program.

BTW My name is Zamir and my hipchat name is @zamir. Unfortunately there is a collision between my public github username and the username of a recent hire so that has caused confusion! :smile: 

Zamir

data_science/lib/ab_tester.rb
paulghaddad: Agree with Zamir. Only provide public access if needed.

paulghaddad:
@lmenova Take a look at Zamir's comments and resubmit. A second iteration should be all that's needed.

PR #851, opened 2015-07-21
Dinodex V2

mcdaniel67:
A corrected version of my Dinodex

### DIFF
--- a/dino_catalog/csv_parser.rb
+++ b/dino_catalog/csv_parser.rb
@@ -1,0 +1,56 @@
+require_relative('dinosaur.rb')
+
+class DinodexParserCSV
+  def initialize(filepath)
+    parse(filepath)
+  end
+
+  def parse(filepath)
+    @file = open(filepath, 'r')
+    generate_categories
+    generate_data
+    @dinosaurs
+  end
+
+  def generate_categories
+    @categories = clean_row(@file.first)
+    convert_african
+  end
+
+  def generate_data
+    @dinosaurs = []
+    @file.each do |row|
+      row_data = Hash[@categories.zip(clean_row(row))]
+      @dinosaurs << Dinosaur.new(row_data)
+    end
+  end
+
+  def clean_row(row)
+    line = row.split(',')
+    line.each do |value|
+      value.strip!
+      value.capitalize!
+    end
+    line
+  end
+
+  def convert_african
+    convert('Genus', 'Name')
+    convert('Carnivore', 'Diet')
+    convert('Weight', 'Weight_in_lbs')
+  end
+
+  def convert(from, to)
+    index = @categories.index(from)
+    unless index.nil?
+      @categories[index] = to
+    end
+  end
+
+end
+
+dex = DinodexParserCSV.new('dinodex.csv')
+list = dex.parse('african_dinosaur_export.csv')
+list.each do |dino|
+  puts dino
+end
--- a/dino_catalog/csv_parser.rb
+++ b/dino_catalog/csv_parser.rb
@@ -1,28 +1,21 @@
 require_relative('dinosaur.rb')
+require_relative('dinodex.rb')
 
 class DinodexParserCSV
-  def initialize(filepath)
-    parse(filepath)
-  end
-
   def parse(filepath)
-    @file = open(filepath, 'r')
+    @path = filepath
+    @file = open(@path, 'r')
     generate_categories
     generate_data
-    @dinosaurs
   end
 
   def generate_categories
     @categories = clean_row(@file.first)
-    convert_african
+    convert_african if african_csv?
   end
 
   def generate_data
-    @dinosaurs = []
-    @file.each do |row|
-      row_data = Hash[@categories.zip(clean_row(row))]
-      @dinosaurs << Dinosaur.new(row_data)
-    end
+    @dinosaurs = @file.map { |row| generate_dinosaur(row) }
   end
 
   def clean_row(row)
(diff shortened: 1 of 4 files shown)


dino_catalog/dinodex_queries.rb
```diff
@@ -0,0 +1,20 @@
+require_relative('dinodex.rb')
+
+dex = DinodexParserCSV.new
+list = dex.parse('african_dinosaur_export.csv')
+dex = DinoDex.new(list)
```
avavita: Don't reuse variable names.

dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def generate_dinosaur(row)
+    dino_hash = Hash[@categories.zip(clean_row(row))]
+    convert_african_hash(dino_hash) if african_csv?
+    Dinosaur.new(dino_hash)
+  end
+
+  def african_csv?
+    @path == 'african_dinosaur_export.csv'
```
avavita: It's worth mentioning that you've tied the format of the input to your code in several places within this class and others. I imagine this was simply to contain the scope of the assignment, but these classes and methods are quite fragile as a result. So long as you understood the trade-off, I'm fine with it.

dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,57 @@
...
+    convert_african if african_csv?
+  end
+
+  def generate_data
+    @dinosaurs = @file.map { |row| generate_dinosaur(row) }
+  end
+
+  def clean_row(row)
+    line = row.split(',')
+    line.each do |value|
```
avavita: While "line" makes sense as a collection of values, reading ````line.each```` isn't totally intuitive. I prefer to name collections as plurals. For instance: ````attributes```` rather than ````row````.

But honestly I'm just nitpicking.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,43 @@
...
+    puts @dinosaurs
+  end
+
+  def at_index(index)
+    @dinosaurs[index]
+  end
+
+  def walking(type)
+    new_collection = @dinosaurs.select { |dino| dino.walking_type?(type) }
+    DinoDex.new(new_collection)
```
avavita: Having each of your ````Dinodex```` methods return a new instance of ````Dinodex```` threw me off just because when I saw the output, I assumed that the ````Dinodex```` instance was returning ````self````, and was therefore modified by the query methods. 

That said, this design does very elegantly provide chainable queries. I'm just wondering if there's a more intuitive way to write it.

avavita:
Good changes! I'll leave it up to Paul as to whether this is a final version.

dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,57 @@
+require_relative('dinosaur.rb')
```
paulghaddad: Use `require_relative 'dinosaur'`

dino_catalog/csv_parser.rb
paulghaddad: I'd prefer a different design on having the african export file work with your program. You're doing a lot of checks in your method. There's probably a better way of going about this.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def carnivores
+    new_collection = @dinosaurs.select(&:carnivore?)
```
paulghaddad: Nice job using this idiom: `&:carnivore?`

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,55 @@
+# Dinosaur Class. We store these dinosaurs in an array in
+# order to query the array.
+
+class Dinosaur
+  def initialize(info)
```
paulghaddad: Consider setting your instance variables dynamically. This will clean up this method.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,55 @@
...
+    @name = info['Name']
+    @period = info['Period']
+    @continent = info['Continent']
+    @diet = info['Diet']
+    @weight = info['Weight_in_lbs']
+    @walking = info['Walking']
+    @description = info['Description'] || ''
```
paulghaddad: Nice job using a default.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,55 @@
...
+    capitalize_relevant_members
+  end
+
+  def setup_diet(diet)
+    @diet = diet unless diet == 'Yes' || diet == 'No'
+    @diet = 'Carnivore' if diet == 'Yes'
```
paulghaddad: You can extract the `diet == x` logic into another method.

dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,57 @@
...
+    if map['Carnivore'] == 'Yes'
+      map['Diet'] = 'Carnivore'
+    else
+      map['Diet'] = 'Not Specified'
+    end
+  end
+
+  def convert(from, to)
+    index = @categories.index(from)
+    @categories[index] = to unless index.nil?
```
paulghaddad: Could you alternatively use `if index`?

paulghaddad:
@mcdaniel67 Nice work on this iteration. You made a lot of good changes from your first iteration. You have satisfied the requirements of this exercise. Keep up the good work!

PR #852, opened 2015-07-22
Data science

rcaileff:


data_science/lib/cohort.rb
```diff
@@ -0,0 +1,71 @@
...
+
+class Cohort
+  attr_accessor :name, :size, :conversion_count
+  attr_reader :conversion_rate_interval
+
+  def initialize(name: nil, size: 0, conversion_count: 0)
+    @name = name
+    @size = size
+    @conversion_count = conversion_count
+    @conversion_rate_interval = calculate_conversion_rate_interval
```
djkotowski: Why cache this?  The second you change `size` or `conversion_count` via the accessors above, this value becomes inaccurate.  Just rename `calculate_conversion_rate_interval` to `conversion_rate_interval`.

data_science/lib/cohort.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  def calculate_conversion_rate_interval
+    return 0 if size == 0
+    interval_confidence = 0.95
```
djkotowski: This should be a constant.

data_science/lib/cohort.rb
```diff
@@ -0,0 +1,71 @@
...
+    conversion_rate_interval[1]
+  end
+
+  def significance_of_difference(other)
+    cohort = { yes: conversion_count, no: size - conversion_count }
+    other_cohort = { yes: other.conversion_count,
+                     no: other.size - other.conversion_count }
+    test = ABAnalyzer::ABTest.new(name => cohort, other.name => other_cohort)
+    p = test.chisquare_p
+    p_to_percentage(p)
```
djkotowski: Is this where we should be making this conversion?

data_science/lib/cohort.rb
```diff
@@ -0,0 +1,71 @@
...
+    interval_confidence = 0.95
+    ABAnalyzer.confidence_interval(conversion_count, size, interval_confidence)
+  end
+
+  def better_than?(other)
+    interval_midpoint > other.interval_midpoint
+  end
+
+  def interval_midpoint
+    (rate_min + rate_max) / 2
```
djkotowski: The naming here gets confusing.  This is the conversion rate.

data_science/lib/cohort.rb
```diff
@@ -0,0 +1,71 @@
...
+    (1 - p) * 100
+  end
+
+  def to_s
+    "#{readable_basic_data}  #{readable_conversion_rate}"
+  end
+
+  def readable_basic_data
+    string = "Cohort #{name} contains #{size} samples with #{conversion_count} "
+    string << "conversions."
```
djkotowski: You can split the string into separate lines with a `\`.

```ruby
def readable_basic_data
  "Cohort #{name} contains #{size} samples with #{conversion_count} " \
    "conversions."
end
```

djkotowski: Should `Cohort` be responsible for outputting this data?

data_science/lib/data_reader.rb
```diff
@@ -0,0 +1,24 @@
...
+  def read_json
+    data_s = File.read(file_name)
+    return [] if data_s.empty?
+    JSON.parse(data_s, symbolize_names: true)
+  rescue JSON::ParserError
+    raise BadDataFile, "Unable to parse #{file_name}"
+  end
+end
+
+class BadDataFile < StandardError
```
djkotowski: Put errors at the beginning of the file.

data_science/lib/data_point.rb
```diff
@@ -0,0 +1,23 @@
+class DataPoint
```
djkotowski: This seems to be too much abstraction for the little time you're going to use the individual data points.  It also means you're going to be running through the data in the JSON more times than you need to.  Why not calculate conversions and views the first time you read the data?

data_science/lib/cohort.rb
```diff
@@ -0,0 +1,71 @@
+require 'abanalyzer'
+
+class Cohort
+  attr_accessor :name, :size, :conversion_count
+  attr_reader :conversion_rate_interval
+
+  def initialize(name: nil, size: 0, conversion_count: 0)
+    @name = name
+    @size = size
```
djkotowski: Size may not be the best term here.  Maybe `views`?

data_science/lib/cohort.rb
```diff
@@ -0,0 +1,71 @@
+require 'abanalyzer'
+
+class Cohort
+  attr_accessor :name, :size, :conversion_count
+  attr_reader :conversion_rate_interval
+
+  def initialize(name: nil, size: 0, conversion_count: 0)
```
djkotowski: Should these parameters be optional?

data_science/lib/data_point.rb
```diff
@@ -0,0 +1,23 @@
...
+  def initialize(hash)
+    @contents = hash
+  end
+
+  def value(measure)
+    contents.fetch(measure, nil)
+  end
+
+  def match?(requirements)
+    requirements.each do |key, value|
```
djkotowski: `Enumerable#all?` can help shrink this.

data_science/lib/data_reader.rb
```diff
@@ -0,0 +1,24 @@
...
+require_relative 'data_point'
+
+class DataReader
+  attr_reader :file_name, :data
+
+  def initialize(file_name)
+    @file_name = file_name
+    @data = []
+    data_hashes = read_json
+    data_hashes.each { |hash| @data << DataPoint.new(hash) }
```
djkotowski: `#map` would be simpler here, though as mentioned earlier, you shouldn't have to store the individual data points when reading the JSON.

data_science/lib/robo_researcher.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def conclude_significant(significance)
+    best = cohorts[0].better_than?(cohorts[1]) ? cohorts[0] : cohorts[1]
+    sig = shorten_significance(significance)
+    string = "Cohort #{best.name} is better with 95% confidence.  The "
+    string << "difference is significant with #{sig}% confidence."
+  end
+
+  def shorten_significance(significance)
+    return 99.99 if significance >= 99.99
```
djkotowski: Why is this here?

data_science/lib/robo_researcher.rb
```diff
@@ -0,0 +1,66 @@
...
+    return 99.9 if significance >= 99.9
+    return 99 if significance >= 99
+    significance.round
+  end
+
+  def details
+    cohorts.each_with_object([]) do |cohort, info|
+      info << cohort.to_s
+    end
```
djkotowski: This could be shortened to
```ruby
cohorts.map(&:to_s)
```

data_science/spec/lib/cohort_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'spec_helper'
+require 'cohort'
+
+describe Cohort do
+  context "when it calculates its .conversion_rate_interval" do
```
djkotowski: Contexts are used to define different circumstances that alter the expected actions or output of your class.  What you mean to do here is to `describe` the methods.

For example, below you should have two tests:

```ruby
describe "#rate_min"
# and
describe "#rate_max"
```
or just
```ruby
describe "#conversion_rate_interval`
```

data_science/spec/lib/cohort_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    let(:cohort) { Cohort.new(name: "A", size: 20, conversion_count: 10) }
+    it "gets the right range" do
+      expect(cohort.rate_min).to be_within(0.01).of(0.28)
+      expect(cohort.rate_max).to be_within(0.01).of(0.72)
+    end
+  end
+
+  context "when A is better than B" do
+    let(:cohort_b) { Cohort.new(name: "B", size: 20, conversion_count: 10) }
+    context "when the difference is significant" do
```
djkotowski: These are good contexts.

data_science/spec/lib/cohort_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      let(:a_sig_more) { Cohort.new(name: "A", size: 20, conversion_count: 18) }
+      it "is .better_than? B" do
```
djkotowski: No need to include the `.` or `?` here.  If you want to include the method name, put it in a `describe`

data_science/spec/lib/cohort_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        expect(a_sig_more.better_than?(cohort_b)).to be true
+      end
+      it "gets the right .significance_of_difference percentage" do
+        significance = a_sig_more.significance_of_difference(cohort_b)
+        expect(significance).to be_within(0.1).of(99.5)
+      end
+      it "correctly converts itself .to_s" do
+        description = "Cohort A contains 20 samples with 18 conversions.  The "
+        description << "conversion rate is 77% - 100% with a 95% confidence."
+        expect(a_sig_more.to_s).to eq(description)
```
djkotowski: This is a bit of a brittle test.  Any minor change to the formatting and you have to rewrite this.  Why not check for the key values?

Example:
```ruby
expect(a_sig_more.to_s).to include("Cohort A")
expect(a_sig_more.to_s).to include("20 samples")
# etc
```

data_science/spec/lib/cohort_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+    end
+    context "when the difference is not significant" do
+      let(:insig_more) { Cohort.new(name: "A", size: 20, conversion_count: 15) }
+      it "is .better_than?" do
+        expect(insig_more.better_than?(cohort_b)).to be true
```
djkotowski: This could be replaced with
```ruby
expect(insig_more).to be_better_than(cohort_b)
```

Also, avoid abbreviations like `insig_more` and `a_sig_more`.

data_science/spec/lib/data_point_spec.rb
```diff
@@ -0,0 +1,70 @@
+require 'spec_helper'
+require 'data_point'
+
+describe DataPoint do
+  let(:data_point) { DataPoint.new(foo: 1, bar: "bar", baz: nil) }
+
+  context "when there is a value associated with a measure" do
+    it "can return the value" do
+      expect(data_point.value(:foo)).to be(1)
```
djkotowski: Use `eq` for comparing numeric values.

data_science/spec/lib/cohort_spec.rb
djkotowski: Look at other places where predicate matchers can replace true/false comparisons.

data_science/spec/lib/data_point_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      expect(data_point.match?({})).to be true
+    end
+  end
+
+  context "when there is only one requirement" do
+    context "when it matches" do
+      it ".match? returns true" do
+        expect(data_point.match?(foo: 1)).to be true
+      end
+      context "and nil is a required value" do
```
djkotowski: I don't understand this context.
```
DataPoint when there is only one requirement when it matches and nil is a required value .match? returns true
```

If you're describing `.match?`, put it in a `describe "#match?"` block.  Also clarify your contexts and spec descriptions to make it more readable.

data_science/spec/lib/data_point_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+    end
+    context "when it only matches one requirement" do
+      it ".match? returns false" do
+        expect(data_point.match?(foo: 1, bar: "fred")).to be false
+      end
+    end
+  end
+
+  context "when key/value pairs are the same" do
```
djkotowski: I'll use this to clarify what I mentioned above.  This part of your spec could be rewritten as:
```ruby
describe DataPoint do
  # other tests

  describe "#eql?" do
    context "when key/value pairs are the same" do
      # your lets here

      it "is true" do
        expect(data_point_1).to eql(data_point_2)
      end
    end

    # other contexts/tests for #eql? here
end

Comment (djkotowski) [inline data_science/spec/lib/data_point_spec.rb]: I highly recommend checking out [Better Specs](http://betterspecs.org/) for tips and suggestions on writing readable, solid specs.

File: data_science/spec/lib/data_reader_spec.rb [hunk 22]
```diff
@@ -0,0 +1,37 @@
...
+    let(:data) { [DataPointFactory.create, DataPointFactory.create(day: "21")] }
+    it "returns the DataPoints" do
+      expect(data_reader.data).to eql(data)
+    end
+  end
+
+  context "reading an empty file" do
+    let(:data_reader) { ReaderFactory.create("empty") }
+    it "stores an empty list" do
+      expect(data_reader.data).to eq([])
```
djkotowski: Wouldn't `expect(data_reader.data).to be_empty` make more sense?

data_science/spec/spec_helper.rb
```diff
@@ -0,0 +1,96 @@
+# This file was generated by the `rspec --init` command. Conventionally, all
```
djkotowski: Delete generated comments.

data_science/spec/lib/robo_researcher_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+end
+
+module RoboResearcherFactory
+  def self.create(fake_data)
+    researcher = RoboResearcher.new
+    researcher.cohorts = [
+      Cohort.new(name: "A", size: fake_data[0], conversion_count: fake_data[1]),
+      Cohort.new(name: "B", size: fake_data[2], conversion_count: fake_data[3]),
+    ]
+    researcher
```
djkotowski: This could be made more readable using a `#tap` block.

data_science/lib/robo_researcher.rb
```diff
@@ -0,0 +1,66 @@
...
+    Cohort.new(name: name, size: size, conversion_count: conversion_count)
+  end
+
+  def conclude
+    significance = cohorts[0].significance_of_difference(cohorts[1])
+    if significance < 95.0
+      return "There is no significant difference between cohorts."
+    else
+      conclude_significant(significance)
+    end
```
djkotowski: You could simplify this using guard clauses.

```ruby
return conclude_significant(significance) unless significance < 95
"There is no significant difference between cohorts."
```

djkotowski:
Nice iteration.  I'd separate the console output concern from classes like Cohort.  There's no need for those classes to concern themselves with how itheir properties are outputted to a console.  If you want to change the formatting, you shouldn't have to change the classes responsible for calculating the values.

Take a look at the comments, refactor and resubmit when ready.  Let me know if you have any questions.

PR #853, opened 2015-07-22
Data science V1

mcdaniel67:
This is my first draft of data_science


data_science/data_science.rb
```diff
@@ -0,0 +1,43 @@
+require_relative('json_parser')
+require('abanalyzer')
+
+class DataScience
+  def load(filepath)
+    @parser = JSONParser.new
```
zsyed1: Line 6: The variable `@parser` does not need to be a member variable. So you can just do:
```
parse = JSONParser.new
```
But then you might as well combine lines 6 and 7 into one line:
```
@data = JSONParser.new.parse(filepath)
```
There is probably a single class-level method on `JSONParser` that is equivalent to this, but I don't remember off hand what it is.

data_science/data_science.rb
```diff
@@ -0,0 +1,43 @@
...
+    @data = @parser.parse(filepath)
+  end
+
+  def data_reader
```
zsyed1: I don't see the need for this method `def data_reader`. If you want to expose the member variable `@data`, simply declare it as
```
attr_accessor: data
```
at the top of your class definition. This will generate a (hidden) getter method for outside access.

data_science/data_science.rb
```diff
@@ -0,0 +1,43 @@
...
+    @data.count { |entry| entry['cohort'] == cohort }
+  end
+
+  def number_of_conversions(cohort)
+    conversions = 0
+    @data.each do |entry|
+      conversions += 1 if entry['result'] == 1 && entry['cohort'] == cohort
+    end
+    conversions
+  end
```
zsyed1: Lines 23 - 27 can be compressed into a single call to `count`:
```
@data.count { |entry| entry['result'] == 1 && entry['cohort'] == cohort }
```

data_science/data_science.rb
```diff
@@ -0,0 +1,43 @@
...
+  def confidence_interval(cohort, confidence)
+    successes = number_of_conversions(cohort)
+    trials = total_size(cohort)
+    ABAnalyzer.confidence_interval(successes, trials, confidence)
+  end
+
+  def chi_square_confidence
+    values = {}
+    values[:a] = { success: number_of_conversions('A'), total: total_size('A') }
+    values[:b] = { success: number_of_conversions('B'), total: total_size('B') }
```
zsyed1: Lines 37 - 39: You can declare the values directly on creation:
```
values = {
  a: { success: number_of_conversions('A'), total: total_size('A') }
  b: { success: number_of_conversions('B'), total: total_size('B') }
}
```

data_science/data_science_example.rb
```diff
@@ -0,0 +1,24 @@
+require_relative('data_science.rb')
+
+calculator = DataScience.new
```
zsyed1: If the class is called `DataScience` why call the variable `calculator`? The two names should have some agreement, perhaps you can change `DataScience` to `Calculater`?

data_science/data_science_spec.rb
```diff
@@ -0,0 +1,44 @@
+require_relative('data_science.rb')
+require('abanalyzer')
+
+describe DataScience do
+  before do
+    @calculator = DataScience.new
+    @calculator.load('json_test_files/small_test_data.json')
+  end
```
zsyed1: Lines 5 - 8: The preferred way to do this is to use `let` statements:

```
let(:calculator) do
  calculator = DataScience.new
  calculator.load('...')
  calculator
end
```
Then you can use `calculator` in your tests (no need for `@`).

data_science/data_science_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  it 'should correctly load a small json dataset' do
+    expect(@calculator.data_reader).to eq([
+      { 'date' => '2014-03-20', 'cohort' => 'A', 'result' => 1 },
+      { 'date' => '2014-03-20', 'cohort' => 'A', 'result' => 1 },
+      { 'date' => '2014-03-20', 'cohort' => 'A', 'result' => 0 },
+      { 'date' => '2014-03-20', 'cohort' => 'A', 'result' => 0 },
+    ])
+  end
```
zsyed1: I really like what you did here in creating a tiny data set for which you can test every value. There may be a better way to write this particular test (though I am fine with the way you wrote it here!) Suggestions @djkotowski ?

data_science/data_science_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  it 'should calculate the conversion rate of visitors' do
+    expect(@calculator.conversion_rate('A')).to eq(0.50)
```
zsyed1: Your `conversion_rate` method does some floating-point arithmetic so you may get `0.499999999998` :smile: 
However, there is an rspec expectation that goes like `expect(x).to_be within(1e-8).of(0.5)` thought I may have the underscores wrong.

data_science/data_science_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    expect(@calculator.number_of_conversions('A')).to eq(2)
+  end
+
+  it 'should calculate conversion rate with 95% confidence' do
+    @calculator.load('json_test_files/statistics_test.json')
+    conversion_rate = @calculator.conversion_rate('A')
+    interval = @calculator.confidence_interval('A', 0.95)
+    expect(conversion_rate).to be >= interval[0]
+    expect(conversion_rate).to be <= interval[1]
+  end
```
zsyed1: Lines 32 - 36, too much going on in this test. Use let statements to encapsulate these quantities. You are using a new calculator in line 32, you can move that into a `let` and then change the name of the other calculator to `tiny_calculator` or something. Also, since you may be replicating code to build those calculators, this might be a good opportunity to write an rspec-helper method say `def build_calculator(filename)` that returns a calculator given a stats file. Then your `let`s would look like:

```
let(:tiny_calculator) { build_calculator('json_test_files/small_test_data.json') } 
let(:full_calculator) { build_calculator('json_test_files/statistics_test_data.json') } 
```

data_science/json_parser.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def parse(filepath)
+    raise_file_error unless File.exist?(filepath)
+    file = File.read(filepath)
+    if file.empty?
+      {}
+    else
+      JSON.parse(file)
+    end
```
zsyed1: Lines 9 - 13:
```
return {} if file.empty?
JSON.parse(file)
```
Though I'm curious how `JSON.parse` handles an empty file. Does it raise an exception or simply return an empty json?

data_science/json_parser_spec.rb
```diff
@@ -0,0 +1,28 @@
+require_relative('json_parser.rb')
+
+describe JSONParser do
+  before do
+    @parser = JSONParser.new
+  end
```
zsyed1: Use `let`, see comment above about `let`s

data_science/json_test_files/statistics_test.json
```diff
@@ -0,0 +1 @@
...:1},{"date":"2014-03-20","cohort":"A","result":0},{"date":"2014-03-20","cohort":"A","result":0},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1},{"date":"2014-03-20","cohort":"B","result":1}]
```
zsyed1: Nice work implementing all the features! I would recommend another round of submission since this is early on in your training and it's nice to ingrain these conventions sooner than later.

data_science/json_parser.rb
mcdaniel67: I just tried it out, it raises a ParseError saying "A JSON test must at least contain two octets!"

PR #854, opened 2015-07-22
ShoppingCart v1

ssapra:


shopping_cart/carts.feature
```diff
@@ -0,0 +1,31 @@
+Feature: Shopping carts
```
magicalbanana: This would be better if your directory structure would look like this:

<pre>
|- features
|   |- Shopping Cart
|     |- Cart Management
|     |- Cart Reconciliation    
</pre>

You already have a shopping cart feature. But in this case, you are testing our the feature for shopping cart reconcilation when a user is anonymous and when a user is authenticated.

shopping_cart/coupon.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Shopping coupons
+  As an online shopper
+  I want to use coupons
+  Because I can save money
+
+Scenario: Using valid coupon
+  Given I am viewing the cart
+  And the cart total is "100" dollars
```
magicalbanana: 100 dollars of what? It would be best if you wrote this step to be more expressive. Because when you start writing the steps or the page object class that will handle this step I am certain you will be dealing with items.

So it would be better to write this one out to:

```
And the cart has "100" dollars worth of "items"
```

shopping_cart/coupon.feature
```diff
@@ -0,0 +1,32 @@
...
+  When I apply the coupon code "x3ifwf"
+  Then the coupon is marked valid
+  And the cart total is "86" dollars
+
+Scenario: Using expired coupon
+  Given I am viewing the cart
+  And the cart total is "100" dollars
+  When I apply the coupon code "goldtier"
+  Then the coupon is marked expired
+  And the cart total is "100" dollars
```
magicalbanana: Using affirmation statements when validating an action. In this case, it would be:

```
And the expected cart total should still be "100" dollars
```

shopping_cart/coupon.feature
```diff
@@ -0,0 +1,32 @@
...
+  Given I am viewing the cart
+  And the cart total is "100" dollars
+  When I apply the coupon code "boguscoupon"
+  Then the coupon is marked invalid
+  And the cart total is "100" dollars
+
+Scenario: Leaving coupon field blank
+  Given I am viewing the cart
+  And the cart total is "100" dollars
+  When I type "" for the coupon code
```
magicalbanana: Best to leave a space between the quotes, or better yet, pass in a value that can be translated to a blank or nil.

shopping_cart/shipping_estimates.feature
```diff
@@ -0,0 +1,39 @@
+Feature: Shopping cart estimates
```
magicalbanana: You're feature for this one is not estimating the shopping cart value but rather the shipping cost. Rewrite this title.

shopping_cart/shipping_estimates.feature
```diff
@@ -0,0 +1,39 @@
...
+  Because I want to manage how much I spend
+
+  Scenario: Load shipping address form
+    Given I am logged in
+    And I have a cart of 4 "computers"
+    And I visit the cart page
+    When I click on "See shipping estimate"
+    Then I see the shipping address form
+
+  Scenario: Load shipping address form
```
magicalbanana: You can not have the same scenario. Your scenario should explain or summarize what you are going to be testing.

shopping_cart/shipping_estimates.feature
```diff
@@ -0,0 +1,39 @@
...
+    Given I am not logged in
+    And I have a cart of 4 "computers"
```
magicalbanana: You can move this to a background step. Since you have a reconicilation feature, it would be ok to move this to a background step and have the login be the first Given statement of your scenario.

magicalbanana:
This is off to the right track but would like to see this resubmitted.

paulghaddad:
@ssapra In your resubmission, see if you can write these scenarios at a higher level of abstraction. See this as an example:

And this article explains this concept in more detail:

Please refactor and resubmit a new PR.

PR #855, opened 2015-07-22
shopping cart features

yuvalroth:
cucumber specs

### DIFF
--- a/shopping_cart/Features/add_item_from_product_page.feature
+++ b/shopping_cart/Features/add_item_from_product_page.feature
@@ -1,0 +1,48 @@
+Feature: Add an Item from the product page
+  In order to add an item from the product page
+  As a Customer
+  I want to select an item and add to the cart
+
+  Scenario: Customer is logged in and wants to add a first item to the cart
+    Given There are no items in the cart
+    And I am on an item page
+    And I am logged in
+    When I operate the add the item control
+    Then I should see that the item was added to the cart
+    And I should stay on the item page
+
+  Scenario: Customer is logged in and wants to add another item to the cart
+    Given There are items in the cart
+    And I am on an item page
+    And I am logged in
+    When I operate the add the item control
+    Then I should see that the item was added to the cart
+    And I should stay on the item page
+
+  Scenario: Customer is logged in and wants to add another item with the same sku to the cart
+    Given There are items in the cart
+    And I am on an item page
+    And I am logged in
+    When I operate the add the item control with an existing sku
+    Then I should see that the item was added to the cart
+    And I should stay on the item page
+
+
+  Scenario: Customer is not logged in and wants to add a first item to the cart
+    Given There are no items in the cart
+    And I am on an item page
+    And I am not logged in
+    When I operate the add item control
+    Then I should be prompted to login
+    And I should see that the item was added to the cart
+    And I should stay on the item page
+
+  Scenario: Customer is not logged in and wants to add a first item to the cart but does not want to log in
+    Given There are no items in the cart
+    And I am on an item page
+    And I am not logged in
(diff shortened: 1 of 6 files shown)


shopping_cart/Features/reconcile_shopping_cart_content.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Reconcile Shopping cart content when customer is logging in
+  In order to allow reconciling shopping cart content
```
magicalbanana: Your title is expected to give a brief summary or a macro level for what you are trying to do. In this case, reconciling shopping carts.

So the feature title would be besser if it was:

Feature: Shopping cart reconcilation from an anonymous visitor to a authenticated user.

It's more concise to use "authenticated" when stating a user is logged in, don't you think you? :dancer: :bee:

shopping_cart/Features/manage_shopping_cart_items.feature
```diff
@@ -0,0 +1,62 @@
+Feature: Manage Quantities Of Shopping Cart Items
+  In order to manage items I am about to order
+  As a customer
+  I want to be a be able to make changes to item selection in the cart
+
+  Scenario: Customer wants to remove all items from the cart should see a warning
+    Given Shopping Cart has items
```
magicalbanana: If the given step is not complicated and is used in your other scenarios and does not require a definite sequence in running, es ist besser if this was moved to a background group.

shopping_cart/Features/reconcile_shopping_cart_content.feature
```diff
@@ -0,0 +1,25 @@
...
+  As a Customer
+  I want to keep content of my shopping cart from previous sessions and make sure I do not lose any items while logging in
+
+  Scenario: Customer with empty anon shopping cart has a saved authenticated shopping cart with items and is logging in
```
magicalbanana: Keep scenario short as possible and be sure it's easily understood.

To be honest I could not understand thoroughly what you are trying to do here without reading the subsequent steps.
You can write the scenario as:
```
Scenario: A unauthenticated visitor has items in the shopping cart.
```
The reason there is no need to elaborate is because you're feature is self explanatory. This feature explores or tests the behavior of reconciling shopping carts from an unauthenticated user to a authenticated user. Hence your scenarios will explore the different cases that are involved with your intended feature which in this case is reconciling shopping carts for an unauthenticated user to a authenticated user.

:dancer: :bee:

magicalbanana:
So far so good.

But would like this resubmitted. Take notes of the comments. Try to split your features a bit more. Right now I like them because they're precise to what you are trying to do and that's good.

But be more detailed. The feature tests or UA tests should try and explore a latitude of scenarios.

magicalbanana:
Look at this for reference:

https://github.com/jmmastey/level_up_exercises/pull/860

shopping_cart/Features/add_item_from_product_page.feature
```diff
@@ -0,0 +1,48 @@
...
+    And I should stay on the item page
+
+  Scenario: Customer is logged in and wants to add another item with the same sku to the cart
+    Given There are items in the cart
+    And I am on an item page
+    And I am logged in
+    When I operate the add the item control with an existing sku
+    Then I should see that the item was added to the cart
+    And I should stay on the item page
+
```
paulghaddad: Spacing

paulghaddad:
@yuvalroth Take a look at Kareem's comments and resubmit. Also, consider keeping these features at a slightly higher level of abstraction so you don't have too many steps per scenario. It is much easier to read. 

Check out this file as an example:

And this article for staying at a consistent level of abstraction:

PR #857, opened 2015-07-22
RobotName v2

ssapra:
Addressed Paul's comments in second iteration.


robot_name/robot.rb
```diff
@@ -0,0 +1,63 @@
...
+    @robot_names << name
+  end
+
+  def self.check_for_duplicate(name)
+    if @robot_names.include?(name)
+      raise DuplicateNameError, 'Robot name already exists!'
+    end
+  end
+
+  private_class_method :check_for_duplicate
```
paulghaddad: Actually, I wouldn't make this class method private, just instance methods.

robot_name/robot.rb
```diff
@@ -0,0 +1,63 @@
...
+class Robot
+  attr_accessor :name
+
+  def initialize(generator = nil)
+    @name = generate_name(generator)
+    validate_name
+    Registry.add_name(@name)
+  end
+
+  private
```
paulghaddad: Nice work making these instance methods private.

robot_name/robot.rb
```diff
@@ -0,0 +1,63 @@
...
+    digits = 3.times.map { generate_digit }
+    (chars + digits).inject("") { |name, char| name << char }
+  end
+
+  def validate_name
+    is_invalid = !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/)
+    raise InvalidNameError, "#{name} is invalid" if is_invalid
+  end
+
+  def generate_char
```
paulghaddad: Excellent refactorings to make these methods. Consider naming this `generate_character` for clarity though.

robot_name/robot.rb
```diff
@@ -0,0 +1,63 @@
...
+    validate_name
+    Registry.add_name(@name)
+  end
+
+  private
+
+  def generate_name(generator)
+    return generator.call if generator
+
+    chars = 2.times.map { generate_char }
```
paulghaddad: To eliminate the need for lines 35 and 36, you can refactor `generate_char` and `generate_digit` to accept an argument for how many characters/digits should be generated, such as `generate_char(2)`.

robot_name/robot.rb
```diff
@@ -0,0 +1,63 @@
...
+    digits = 3.times.map { generate_digit }
+    (chars + digits).inject("") { |name, char| name << char }
+  end
+
+  def validate_name
+    is_invalid = !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/)
```
paulghaddad: Extract this as a separate method, and since it returns a boolean, change its name to `is_invalid?`. Or you could name it `valid_name?`, and change the error check to an `unless` in line 42. This will eliminate the need for the `!` operator.

paulghaddad:
@djkotowski Great work on the refactor! I think your code is more well structured. Overall, you write very clear and understandable code, so keep that up! :)

You can try some of the comments I suggest above, but this is complete for LevelUp.

robot_name/robot.rb
ssapra: Would it make sense to take it further and have a single generate method that took a range and integer and sampled x number of values from that range? Something akin to `generate(('A'..'Z'), 2) << generate((0..9), 3)`

paulghaddad: @ssapra I personally wouldn't but that's not to say it's not correct.

robot_name/robot.rb
djkotowski: This is one abbreviation where I think it's clear due to precedence (e.g. `String#chars`).

robot_name/robot.rb
```diff
@@ -0,0 +1,63 @@
+class DuplicateNameError < RuntimeError; end
+class InvalidNameError < RuntimeError; end
+
+class Registry
+  @robot_names = []
```
djkotowski: Using a class instance variable instead of a class variable.  Nice!

djkotowski:
Nice work!  There are some comments throughout to look at, but you're done with this exercise.

PR #862, opened 2015-07-23
Shopping cart

zsyed91:


shopping_cart/product_add.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Add Product
+  As a user
+  I want to be able to add products to my cart
+
+  Scenario Outline: Add product to cart
+    Given I view <product>
```
djkotowski: If this app were to expand, this step may have to be altered to identify what you're viewing.  Also, the tense of "view" sounds off.

Example:
```cucumber
Given I am viewing product <product>
```

shopping_cart/product_add.feature
```diff
@@ -0,0 +1,25 @@
...
+    When I enter in <quantity>
```
djkotowski: As above, `When I enter in <quantity>` sounds a bit too vague on the step definition side.  `When I enter in <quantity> quantity` will read better.

djkotowski: To elaborate, let's say you're using an example
```cucumber
| product | quantity |
| Xbox    | 1        |
```

This will read as
```cucumber
Given I view Xbox
When I enter in 1
And click buy
Then I expect to have 1 Xbox in my cart
```

Everything reads pretty clearly except for the quantity.

shopping_cart/product_add.feature
```diff
@@ -0,0 +1,25 @@
...
+    And click buy
```
djkotowski: This is too implementation-specific.  What if the button is changed to `Next >` or `Add to Cart`?

shopping_cart/product_add.feature
```diff
@@ -0,0 +1,25 @@
...
+  Scenario Outline: Invalid quantity
+    Given I view <product>
+    When I enter in an invalid <quantity>
+    And click buy
+    Then I expect to have a quantity error
+
+  Scenario Outline: Invalid product page
+    Given <product> is an old listing or invalid listing
+    When I try to view the <product> page
+    Then I expect to have a listing error
```
djkotowski: Where are your examples for these scenarios?

shopping_cart/product_add.feature
djkotowski: As an alternative, prefer something like
```cucumber
And I add the item to the cart
```

shopping_cart/shipping_estimates.feature
```diff
@@ -0,0 +1,31 @@
...
+    When I click on the shipping estimate button
+    Then I should see an estimate
+
+  Scenario: Authenticated user without a shipping address
+    Given I do not have a default shipping address
+    When I click on the shipping estimate button
+    And enter in my shipping information
+    Then I should see an estimate
+
+  Scenario Anonymous user without a shipping address
```
djkotowski: Missing `:`

shopping_cart/user_signup.feature
```diff
@@ -0,0 +1,26 @@
+Feature: New User
+  As a new user
+  I want to be able to create an account
+
+  Background:
+    Given I don't have an account
```
djkotowski: Is this a necessary step?

shopping_cart/user_account.feature
```diff
@@ -0,0 +1,41 @@
+Feature: Existing User
+  As an existing user
+  I want to be able to interact with my personal account
+
+  Background:
+    Given I have an account
+    And I securely visit the login page
```
djkotowski: The "securely" part looks like it should be a separate test.

shopping_cart/product_listing.feature
```diff
@@ -0,0 +1,17 @@
+Feature: Product listing
+  As a user
+  I want to be able to view the product page from the shopping cart
+
+  Background:
+    Given I view my cart
+    And I have products in my cart
```
djkotowski: From an implementation perspective, this should be reversed.  If you view your cart, and you _then_ have items in your cart, your cart should look empty.

djkotowski:
Looks good.  A few comments throughout, but no need for a resubmission.

PR #864, opened 2015-07-24
Arrowhead v1

mccormjt:
Finished the arrowhead exercise

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -1,4 +1,8 @@
 class Arrowhead
+  UNKNOWN_REGION_MSG = 'Unknown region, please provide a valid region.'
+  UNKNOWN_SHAPE_MSG  = "Unknown shape value."\
+                       "Are you sure you know what you're talking about?"
+
   # This seriously belongs in a database.
   CLASSIFICATIONS = {
     far_west: {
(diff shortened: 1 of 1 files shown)


arrowhead/arrowhead.rb
```diff
@@ -1,4 +1,8 @@
 class Arrowhead
+  UNKNOWN_REGION = 'Unknown region, please provide a valid region.'
+  UNKNOWN_SHAPE = "Unknown shape value."\
```
djkotowski: Check the spacing here.

arrowhead/arrowhead.rb
```diff
@@ -15,19 +20,14 @@ class Arrowhead
...
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+    region_shapes = CLASSIFICATIONS[region]
+    arwhead = region_shapes.nil? ? raise(UNKNOWN_REGION) : region_shapes[shape]
```
djkotowski: Avoid ternary operators, especially when determining if you need to raise an error.  Ternary operators should only be used for the _most basic_ of expressions.

djkotowski:
Looks good.  Take a look at the comments, but no need for a resubmission.

PR #865, opened 2015-07-24
Triangle facts v1

mccormjt:
Finished first attempt at triangle facts exercise

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -1,58 +1,72 @@
 # Killer facts about triangles AWW YEAH
 class Triangle
-	attr_accessor :side1,:side2,:side3
+  attr_accessor :side1, :side2, :side3
 
-	def initialize(side1,side2,side3)
-		@side1,@side2,@side3 = side1,side2,side3
-	end
+  def initialize(side1, side2, side3)
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+  end
 
-	def equilateral()
-		return side1 == side2 && side2 == side3
-	end
+  def equilateral
+    @side1 == @side2 && @side2 == @side3
+  end
 
-	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
-	end
+  def isosceles
+    [@side1, @side2, @side3].uniq.length == 2
+  end
 
-	def scalene()
-		unless ! (equilateral || isosceles)
-	    false
-	  else
-	    true
-	  end
-	end
+  def scalene
+    !(equilateral || isosceles)
+  end
 
-	def recite_facts
-		puts 'This triangle is equilateral!' if equilateral
-		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+  def recite_facts
+    puts shape_type_msg
+    puts angles_type_msg
+  end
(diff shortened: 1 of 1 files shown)


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
+  end
+
+  def shape_type_msg
+    if equilateral?
+      'This triangle is equilateral'
+    elsif isosceles?
+      'This triangle is isosceles! Also, that word is hard to type.'
+    else
+      'This triangle is scalene and mathematically boring.'
+    end
```
djkotowski: You can avoid big if-blocks like this with guard clauses.  Also, avoid abbreviations where possible.
```ruby
def shape_type_message
  return "This triangle is equilateral" if equilateral?
  return "This triangle is isosceles! Also, that word is hard to type." if isosceles?
  "This triangle is scalene and mathematically boring."
end
```

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
+  end
 
+  def angles_type_msg
+    angles = Triangle.calculate_angles(@side1, @side2, @side3)
+    msg = "The angles of this triangle are #{angles.join(',')}\n"
+    msg << "This triangle is also a right triangle!\n" if angles.include? 90
```
djkotowski: Wrap method arguments in parentheses.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
+      'This triangle is equilateral'
+    elsif isosceles?
+      'This triangle is isosceles! Also, that word is hard to type.'
+    else
+      'This triangle is scalene and mathematically boring.'
+    end
+  end
 
+  def angles_type_msg
+    angles = Triangle.calculate_angles(@side1, @side2, @side3)
```
djkotowski: 1.  Prefer `self.class.method` over `ClassName.method` in the same class.
2.  Why not make this an instance method?

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
+      Triangle.calculate_angle(a, b, c),
+      Triangle.calculate_angle(b, a, c),
+      Triangle.calculate_angle(c, a, b),
+    ]
+  end
+
+  def self.calculate_angle(main_angle, a, b)
+    numerator   = a**2 + b**2 - main_angle**2
+    denominator = 2.0 * a * b
+    radians_to_degrees Math.acos(numerator / denominator)
```
djkotowski: Wrap method arguments in parentheses.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
 
+  def angles_type_msg
+    angles = Triangle.calculate_angles(@side1, @side2, @side3)
+    msg = "The angles of this triangle are #{angles.join(',')}\n"
+    msg << "This triangle is also a right triangle!\n" if angles.include? 90
+  end
+
+  def self.calculate_angles(a, b, c)
+    [
+      Triangle.calculate_angle(a, b, c),
```
djkotowski: No need for the class name here.

djkotowski:
Nice first iteration.  Take a look at the comments, refactor and resubmit.  You're almost done with this one.

PR #867, opened 2015-07-24
Data Science v1

mccormjt:
Finished first attempt at data_science


data_science/data_science/cohort.rb
```diff
@@ -0,0 +1,22 @@
+class Cohort
+  HIT_VAL = 1
+  MISSED_VAL = 0
+
+  attr_reader :dates, :results, :result_val
+
+  def initialize
+    @dates = []
```
zsyed1: You never use the `@dates` field in the calculations. I don't see the need for this field.

data_science/data_science/cohort.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def num_entries
+    @results.size
+  end
+
+  def insert(entry)
+    @dates << entry[:date]
+    @results << entry[:result]
+    @result_val += entry[:result]
```
zsyed1: It looks like `@result_val` is just the sum of the `@results`. So you can remove this member variable (`@result_val`) and write a method:
```
def conversions
  @results.sum
end
```
But then the question becomes "Do you need to store the individual results?" - Probably not. You only need to store the number of successes and the number of failures.
```

File: data_science/data_science/data_science.rb [hunk 3]
```diff
@@ -0,0 +1,67 @@
...
+    cohort_entries.each(&method(:add_cohort_entry))
+  end
+
+  def add_cohort_entry(entry)
+    if entry[:cohort] == @cohort_a_name
+      cohort_a.insert entry
+    elsif entry[:cohort] == @cohort_b_name
+      cohort_b.insert entry
+    end
+  end
```
zsyed1: Lines 35 - 39: You can create a method called: `def cohort(entry)` that returns the Cohort. Then this method becomes one line `cohort(entry).insert(entry)`. Then the method `def cohort(entry)` can raise an error if the cohort name is unrecognized.

Side note - put parentheses around method arguments: `cohort_a.insert(entry)`.

data_science/data_science/data_science.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def reset_cohort_data
+    @cohort_a = Cohort.new
+    @cohort_b = Cohort.new
+  end
+
+  def load_cohort_json(path = DEFUALT_COHORT_JSON_PATH)
+    json_file = File.read path
+    cohort_entries = JSON.parse(json_file, symbolize_names: true)
+    cohort_entries.each(&method(:add_cohort_entry))
```
zsyed1: Interesting :smile:

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  def missed_result_entry
+    { date: '2014-03-20', result: Cohort::MISSED_VAL }
+  end
+
+  def hit_result_entry
+    { date: '2011-05-22', result: Cohort::HIT_VAL }
+  end
+
+  it 'initializes with empty data fields' do
+    field_vals = @cohort.dates.size + @cohort.results.size + @cohort.result_val
```
zsyed1: Instead of adding these sizes, you can make a method on `Cohort` called `def empty?` than checks all three are empty. e.g. `dates.empty?`. But see the comments on `Cohort` about reducing the size of the class.

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,59 @@
+require_relative '../data_science/cohort.rb'
+
+RSpec.describe Cohort do
+  before(:example) do
```
zsyed1: Use `let` instead of `before`:
```
let(:empty) { Cohort.new }
```

data_science/spec/cohort_spec.rb
zsyed1: BTW When you use `let` you will remove the `@` symbol.

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    { date: '2011-05-22', result: Cohort::HIT_VAL }
+  end
+
+  it 'initializes with empty data fields' do
+    field_vals = @cohort.dates.size + @cohort.results.size + @cohort.result_val
+    expect(field_vals).to eq(0)
+  end
+
+  context 'has hit result entry' do
+    before(:example) do
```
zsyed1: Use `let` instead of `before`:

```
let(:cohort) { empty.insert(hit_result_entry) }
```

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      @cohort.insert hit_result_entry
+    end
+
+    it 'stores a result entrys date' do
```
zsyed1: Change these to double quotes so you can use the apostrophe: `...entry's date...`

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      expect(@cohort.results.first).to eq(hit_result_entry[:result])
+    end
+
+    it 'does increment result val on hit result entry' do
+      expect(@cohort.result_val).to eq(Cohort::HIT_VAL)
+    end
+  end
+
+  it 'does not increment result val on missed result entry' do
+    @cohort.insert missed_result_entry
```
zsyed1: Put arguments in parentheses: `@cohort.insert(missed...`

zsyed1: You can put these into a `let`:
```
let(:missed) { cohort.insert(missed_result_entry) }
```
Then the test becomes
```
expect(missed.result_val).to eq ...
```

zsyed1: Okay, scratch that comment. It looks like you are adding data points one by one and testing the results incrementally. I actually like this method though @djkotowski may disagree!

For these tests it may be better to pre-define certain test cases. But there are definitely times (e.g. event-based systems) where incremental testing is the way to go.

data_science/spec/data_science_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+
+RSpec.describe DataScience do
+  TEST_JSON_PATH = 'split_data/test_data.json'
+  NUM_TEST_ENTRIES_TOTAL = 40
+  NUM_TEST_ENTRIES_A = 20
+  NUM_TEST_ENTRIES_B = NUM_TEST_ENTRIES_TOTAL - NUM_TEST_ENTRIES_A
+  COHORT_NAME_A = 'A'
+  COHORT_NAME_B = 'B'
+
+  before(:example) do
```
zsyed1: Use `let`

data_science/spec/data_science_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    @data_science = DataScience.new(COHORT_NAME_A, COHORT_NAME_B)
+  end
+
+  def cohort_entry(cohort_name)
```
zsyed1: This method should go in a `spec_helper.rb` file

data_science/data_science/data_science.rb
```diff
@@ -0,0 +1,67 @@
...
+      cohort_b.insert entry
+    end
+  end
+
+  def sample_size
+    cohort_a.num_entries + cohort_b.num_entries
+  end
+
+  def cohort_a_conversion_rate
+    conversion_rate_for cohort_a
```
zsyed1: Put parentheses around method arguments

data_science/spec/data_science_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  it 'can load an entry into cohort B' do
+    @data_science.add_cohort_entry cohort_entry(COHORT_NAME_B)
+    expect(@data_science.cohort_b.num_entries).to eq(1)
+    expect(@data_science.cohort_a.num_entries).to eq(0)
+  end
+
+  context 'has loaded test json' do
+    before(:example) do
```
zsyed1: Use `let`

data_science/spec/data_science_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      expect(@data_science.cohort_a.num_entries).to eq(NUM_TEST_ENTRIES_A)
+    end
+
+    it 'laods correct num of entries from json file into B cohort' do
+      expect(@data_science.cohort_b.num_entries).to eq(NUM_TEST_ENTRIES_B)
+    end
+
+    it 'can calculate cohort A conversion rate' do
+      rate = @data_science.cohort_a_conversion_rate
+      min = rate[0].round(3)
```
zsyed1: Instead of rounding, you can use `expect(x).to be_within(0.001).of(0.025)`

data_science/spec/data_science_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      expect(min).to eq(0.141)
+      expect(max).to eq(0.559)
+    end
+  end
+
+  it 'can calculate cohort leader confidence' do
+    @data_science.load_cohort_json
+    expect(@data_science.cohort_leader_confidence).to eq(1.0)
+  end
+end
```
zsyed1: Nice job implementing the requirements. Please look at the comments, make revisions and resubmit.

BTW My name is Zamir and my hipchat username is @zamir. Feel free to hit me up there with questions/comments.

Zamir

PR #868, opened 2015-07-24
Overlord v1

abake-enova:


overlord/bomb.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def activate(code)
+    return unless on?
+    return unless code == @activation_code
+    @state = ACTIVATED
+    @start_time = Time.now
+  end
+
+  def deactivate(code)
+    return unless activated?
```
magicalbanana: return what? If activated is false, this will return "nil:.

overlord/bomb.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def set_codes(a_code, d_code)
+    raise InvalidCodeError unless valid_code?(a_code) && valid_code?(d_code)
+    @activation_code = a_code
+    @deactivation_code = d_code
+  end
+
+  def boot_up(activation_code = "1234", deactivation_code = "0000")
+    set_codes(activation_code, deactivation_code)
+    @state = ON
```
magicalbanana: I would avoid constants that are assigned with booleans just to set the states. I would rather see this to be either static or more dynamic.

overlord/features/bomb_activate_deactivate.feature
```diff
@@ -0,0 +1,41 @@
+Feature: Activating and Deactivating Bomb
```
magicalbanana: I would split this feature, one for activing and another deactivating.

From there, you can explore a multitude of scenarios that you can try and test out.

Like,

```
Feature: Bomb Activitation

  Scenario: Successful bomb activation
  Scenario: Failed bomb activation
```

overlord/public/js/countdowntimer.js
```diff
@@ -0,0 +1,77 @@
...
+        var blink = "blink";
+        if (parseInt($(self.s).html()) < 10)
+            blink += "-red";
+        if ($(self.t).hasClass(blink)) {
+            $(self.t).removeClass(blink);
+        } else {
+            $(self.t).addClass(blink);
+        }
+    }
+}
```
magicalbanana: Can you provide unit tests for these? You can use Jasmine if you want, but preference is yours.

overlord/public/js/overlord.js
```diff
@@ -0,0 +1,172 @@
...
+}
+
+function boom() {
+  $("#boom").html('<iframe allowfullscreen="" frameborder="0" height="360" '+
+    'src="https://www.youtube.com/embed/3KJnnMBHOVQ?rel=0&amp;controls=0&amp;'+
+    'showinfo=0&amp;autoplay=1&amp;start=5&amp;&cc_load_policy=1" '+
+    'width="640"></iframe>');
+}
+
+on_load();
```
magicalbanana: Same thing, I would like to see unit tests for these.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,252 @@
...
+
+  describe "#initialize" do
+    it "should return a Bomb object with the state set to OFF", happy: true do
+      expect(@bomb).to be_a(Bomb)
+      expect(@bomb.state).to eq(0)
+    end
+  end
+
+  describe "#boot_up" do
+    context "no arguments" do
```
magicalbanana: Verify that there are no arguments. And be more expressive of your intent with your context. RSpec gives you the ability to define or give a summary for each tests that you will be performing.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,252 @@
...
+      before(:each) do
+        @bomb.boot_up
+      end
+
+      it "should boot up with default codes", happy: true do
```
magicalbanana: No need to define a custom defined metadata for happy, sad and bad. You're context summary or it block summary is expected to be able to express the intent of your test.

overlord/features/bomb_activate_deactivate.feature
```diff
@@ -0,0 +1,41 @@
...
+
+  Scenario: Deactivating the bomb with custom code
+    Given I visit the home page
+    When I configure the deactivation code to be 1111
+    And I boot up the bomb
+    And I activate the bomb with code 1234
+    And I deactivate the bomb with code 1111
+    Then the bomb should be deactivated
+
+  Scenario: Activating and deactivating with custom codes
```
paulghaddad: Here is an overall comment. Your scenarios should be at a higher level of abstraction.

Something like this:

```
Given I have a bomb with custom activation and deactivation codes
When I activate and then deactivate the bomb
Then the bomb should be deactivated
```

This is much more readable and encapsulates more logic in each step definition.

Take a look at all of your features and strive for step definitions at a higher level of abstraction. See this article for reference: 

http://robots.thoughtbot.com/acceptance-tests-at-a-single-level-of-abstraction

paulghaddad:
@abake-enova Nice first iteration. Please take a look at the comments and submit a second iteration. The main thing to focus on is to refactoring your feature tests. Please see my comment there and apply it to all of your features.

PR #869, opened 2015-07-24
Dino catalog v1

mccormjt:
Finished first attempt at dino catalog


dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def catalog_dino(dino)
+    DinoNormalizer.normalize(dino)
+    dino_addr = @dinos.size
+    @dinos << dino
+    dino.each { |field, value| index_dino_field(dino_addr, field, value) }
+  end
+
+  def all_dinos
+    @dinos
```
djkotowski: Should this just be an attr_reader?

dino_catalog/dino_normalizer.rb
```diff
@@ -0,0 +1,44 @@
+class DinoNormalizer
+  CARNIVORE_DIETS = %w(Piscivore Insectivore Carnivore)
+  MAX_SMALL_DINO_WEIGHT = 4000
+
+  def self.normalize(fields = {})
```
liaden: `DinoNormalizer` gains nothing by being a class with these class methods since the class is only being used as a namespace. 

Push all these instances down onto the instance and initialize the instance with `fields` since every method already requires fields.

dino_catalog/dino_normalizer.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    normalize_period fields
+    normalize_carnivore fields
+    normalize_size fields
+  end
+
+  def self.normalize_field_aliases(fields, field_name, aliases)
+    aliases.each do |als|
+      alias_val = fields.delete(als)
+      unless alias_val.nil?
```
liaden: Simplify logic here: `if alias_val`

liaden:
In general, I feel you would be better supported by having a `Dino` class to wrap up some of your primitives. A `DinoPresenter` could be extracted as well to handle at least console output, if not being useful for json and serialization. `DinoCatalog` seems to be a custom collection data object, and you may want to check out ruby tapas 321 (http://www.rubytapas.com/episodes/321-Smart-Collection). Talk to Dan if you don't have access.

Message me on hipchat (jorel) if you have any questions.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,44 @@
+require_relative "dino_normalizer.rb"
+
+class DinoCatalog
+  def initialize
+    reset_data
+  end
+
+  def reset_data
+    @dinos = []
+    @dino_field_index = {}
```
jmtai: Naming can be hard but one general rule we should follow is that variable names should be as descriptive and self-explanatory as possible. `@dinos` here makes sense - it's clearly an array of dinosaur objects. `@dino_field_index` is a bit confusing at first glance, however. As someone coming in with no knowledge of the code, the singular name suggests to me that maybe this is simply storing a single index value. But then I see that it's actually a hash, so maybe each dino field has a particular index value, and it's being stored here?

I have to read through the read of the code to figure out what's actually being stored here, which is a hash of hashes that maps dino field **values** to an array of `@dinos` array object indexes. Just reading the english description of that data structure might imply that perhaps the implementation could be simplified a bit?

Regardless, a better name might be something like `field_value_to_dino_index_map` or something along those lines. It's on the edge of maybe being too wordy but I'd argue that clarity trumps brevity in most cases.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def catalog_dino(dino)
+    DinoNormalizer.normalize(dino)
+    dino_addr = @dinos.size
```
jmtai: Again, here the variable naming is a bit confusing. When I see something like `dino_addr`, I assume it's storing an address of some kind. But I can see it's being set to the current size of the `@dinos` array, so it must actually be an array index value. Since it's an index, why not just name it `dino_index` or `dino_array_index` and avoid the confusion?

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,44 @@
...
+    @dinos << dino
+    dino.each { |field, value| index_dino_field(dino_addr, field, value) }
+  end
+
+  def all_dinos
+    @dinos
+  end
+
+  def dinos_where(filters = {})
```
jmtai: Any particular reason you named this method `dinos_where` but in `Dinodex` you used `search`? Since `Dinodex.search` just calls this method with the same argument, why not keep the naming consistent and call this one `search` as well?

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,44 @@
...
+    dino_addrs = filters.map { |flt, val| dino_addrs_with(flt, val) }
```
jmtai: Again, I'd use `indexes` instead of `addrs` as anyone experienced with a lower-level language like C would expect `addr` to refer to a memory address.

liaden: `indices`

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,44 @@
...
+    dinos_at dino_addrs.inject(&:&)
```
jmtai: The recommended style is to wrap all method arguments in parentheses.

Also, `dino_addrs.inject(&:&)` is clever, but way **too** clever. Clever code is fun to write but not so fun to maintain. All this is doing is flattening the array, so it would be much clearer to simply use `dino_addrs.flatten`.

dino_catalog/dino_catalog.rb
jmtai: http://www.merriam-webster.com/dictionary/index
:stuck_out_tongue_winking_eye:

dino_catalog/dino_catalog.rb
liaden: `&:&` is no more clever than injecting `&:+` It is uncommon, but there already exists `flatten` so that is definitely better.

dino_catalog/dino_catalog.rb
liaden: Just because enough people have been wrong...

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  private
+
+  def index_dino_field(dino_addr, field, value)
+    @dino_field_index[field] ||= {}
+    @dino_field_index[field][value] ||= []
+    @dino_field_index[field][value] << dino_addr
```
jmtai: This is an interesting implementation (and quite clever). However, you're basically storing all of the dino data twice to make it work. It's not a big deal here but something to think about.
The main problem here is that the implementation is a bit hard to understand at first glance. Instead of operating directly on hash data, what if you had a `Dinosaur` class that you operated on instead? Do you think that would lead to an easier to understand implemenation?

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,95 @@
...
+class DinoDex
+  DINO_DATA_DIR = 'dino_data'
+  EMPTY_FIELD = '(null)'
+
+  def initialize
+    @catalog = DinoCatalog.new
+  end
+
+  def search(filters = {})
+    @catalog.dinos_where filters
```
jmtai: Again, wrap your method arguments in parentheses.

dino_catalog/dino_catalog.rb
liaden: :+1: https://github.com/jmmastey/level_up_exercises/pull/869#issuecomment-131830804

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def catalog_all_dino_data
+    @catalog.reset_data
+    dino_data_dir_path = "#{DINO_DATA_DIR}/*.csv"
+    Dir.glob(dino_data_dir_path).each(&method(:catalog_dino_csv))
+  end
+
+  def catalog_dino_csv(csv_file)
+    dino_data = CSV.read(csv_file, encoding: "UTF-8", headers: true,
+                                  header_converters: :symbol, converters: :all)
```
jmtai: Align the parameters when splitting across multiple lines:
```ruby
dino_data = CSV.read(csv_file, encoding: "UTF-8", headers: true,
                    header_converters: :symbol, converters: :all)
```

dino_catalog/dino_normalizer.rb
```diff
@@ -0,0 +1,44 @@
+class DinoNormalizer
```
jmtai: Have you looked at the CSV class `converters` and `header_converters`? If not, take a look. They allow you to specify custom conversion routines for the headers and data being read. It's possible to use those to normalize the data instead of this separate class.

jmtai:
Nice first pass! This is a unique implementation that I haven't seen before. For your next pass, try implementing a `Dino` or `Dinosaur` class instead of operating directly on the hash data. Joe's suggestion to move output to a separate `DinoPresenter` class is a good one as well, as it helps enforce the single responsibility principle for your classes.

Also, it's not explicitly stated in the exercise description, but one of the goals of this exercise is to create an interactive command line app. Try adding a simple command line interface to the next iteration as well.

Read over the comments, refactor a bit, and resubmit when ready. Feel free to hunt me down or ping me on HipChat if you have any questions.

paulghaddad:
@mccormjt Please take a look at Joel's and Josh's excellent comments. Refactor a second iteration, making sure to create a command line interface. Nice work for this first iteration.

PR #870, opened 2015-07-24
Overlord Iteration #2

jklemen:


overlord/classes/bomb.rb
```diff
@@ -0,0 +1,81 @@
+require './classes/panel.rb'
```
jmmastey: I'm going to stay light on non-test commentary, but point out a few things worth thinking about.

overlord/classes/bomb.rb
```diff
@@ -0,0 +1,81 @@
...
+  STATES = { 0 => 'armed', 1 => 'disarmed', 2 => 'detonated', 3 => 'hacked' }
+
+  attr_accessor :state, :start_time, :panels
+  attr_accessor :hack_count, :attempts_remain
+
+  def initialize
+    @panels = initialize_panels
+    @hack_count = 0
+    @attempts_remain = 3
+    @state = 1
```
jmmastey: Generally IDs like this are opaque. Better to use symbols and just set `@state = :disarmed`.

overlord/classes/bomb.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def attempt_hack(binary, panel_num)
+    match = @panels[panel_num].validate(binary)
+    if match
+      @hack_count += 1
+      @state = 3 if @hack_count == 6
+    end
+    @panels[panel_num].hacked = match
+  end
```
jmmastey: I love the hack thing :D

overlord/classes/bomb.rb
```diff
@@ -0,0 +1,81 @@
...
+    if code != stored_code
+      @attempts_remain -= 1
+      @state = 2 if @attempts_remain == 0
+    else
+      @state = state_if_same
+    end
+  end
+
+  def arm(code)
+    return false if @state > 1
```
jmmastey: If someone adds another state later (frozen?), this check will become problematic. Better to use `[0, 1].include? state`

overlord/classes/bomb.rb
```diff
@@ -0,0 +1,81 @@
...
+  def disarm(code)
+    return false if @state > 1
+
+    check_code(code, @secret_disarm_code, 1)
+    code == @secret_disarm_code
+  end
+
+  def state_to_s
+    STATES[@state]
+  end
```
jmmastey: Ehhh, shaky name. The consumer of this object shouldn't care (and shouldn't have to) about how states are implemented internally, which is what happens here.

overlord/classes/panel.rb
```diff
@@ -0,0 +1,37 @@
+class Panel
+  attr_accessor :hacked, :code, :sequence
+
+  def initialize(code)
+    @code = code
+    @sequence = conv_to_binary
```
jmmastey: `conv_to_binary` is not a great name. better to use something like `code_to_binary`

overlord/features/bomb_bad_codes.feature
```diff
@@ -0,0 +1,20 @@
+#encoding: utf-8
+
+@bad-codes
+Feature: Entering incorrect bomb codes
+  In order to detonate the bomb
+  As an unfortunate soul
+  I should enter an incorrect code 3 times
```
jmmastey: Hah, nice.

jmmastey: Seriously, though, these are solid for understanding the thought process behind the tests.

overlord/features/bomb_bad_codes.feature
```diff
@@ -0,0 +1,20 @@
...
+
+  Background:
+    Given I'm at the bomb page
+    And I have already submitted my codes
```
jmmastey: It's unclear that this means that the bomb is armed. I think it only explodes when armed?

overlord/features/bomb_bad_codes.feature
```diff
@@ -0,0 +1,20 @@
...
+
+  @happy
+  Scenario: Entering the incorrect code 3 times
+    When I enter the incorrect disarm code
+    And I enter the incorrect disarm code again
+    And I enter the incorrect disarm code again
+    Then the bomb should explode
+    And I should be unable to arm the bomb
+    And I should be unable to disarm the bomb
```
jmmastey: Remove "should", it's out of fashion. `And I am unable to arm the bomb`

overlord/features/bomb_basic.feature
```diff
@@ -0,0 +1,12 @@
...
+@bad-codes
+Feature: Basic Bomb
+  In order to do anything to the bomb
+  As an eveil overlord
+  I should see correct behavior
+
+  @happy
+  Scenario: Launching the bomb page
+    Given I'm at the bomb page
+    Then the bomb should be disarmed
```
jmmastey: This might be a tad too specific to have its own file. Too much segmentation into different files makes for a tough read.

overlord/features/bomb_basic.feature
```diff
@@ -0,0 +1,12 @@
+#encoding: utf-8
+
+@bad-codes
+Feature: Basic Bomb
+  In order to do anything to the bomb
+  As an eveil overlord
+  I should see correct behavior
```
jmmastey: Ehh, this last sentence is weak. I think it's pointing towards this test being a little odd by itself. With another grouping of tests the feature's reason for being may be more obvious.

overlord/features/bomb_code_submit.feature
```diff
@@ -0,0 +1,73 @@
+#encoding: utf-8
+
+@code-submit
+Feature: Submitting The Arm / Disarm Codes
+  In order to submit codes
+  As the evil overlord
+  I should enter valid arm and disarm codes
+  And press submit
```
jmmastey: Too specific. Submit is an implementation detail.

overlord/features/bomb_code_submit.feature
```diff
@@ -0,0 +1,73 @@
...
+
+  Background:
+    Given I'm at the bomb page
+
+  @happy
+  Scenario: I enter nothing and arm
+    And I enter nothing
```
jmmastey: Don't use `And` as the first step. Change to `Given`.

jmmastey: I actually like this "null step" though, since it calls out that you're not actually doing anything.

overlord/features/bomb_code_submit.feature
```diff
@@ -0,0 +1,73 @@
...
+    When I press Submit Codes
+    And I enter the correct arm code
+    Then the bomb should be armed
+    And I should be able to disarm the bomb
```
jmmastey: Is this really part of the test itself? Ability to disarm is probably not a part of what you want to assert, though the other `Then` is good here.

overlord/features/bomb_code_submit.feature
```diff
@@ -0,0 +1,73 @@
...
+    When I enter the arm code only
+    And I press Submit Codes
+    And I enter the correct arm code
+    Then the bomb should be armed
+    And I should be able to disarm the bomb
+
+  @happy
+  Scenario: I enter only the disarm code
+    When I enter the disarm code only
+    And I press Submit Codes
```
jmmastey: The more I see this, the more I want it to change. I know it's easy for step definitions, but it feels awkward.

overlord/features/bomb_code_submit.feature
```diff
@@ -0,0 +1,73 @@
...
+    And the bomb is armed
```
jmmastey: Ahh, I think I see something here. Is this an assertion step? I think it is. In which case, change it to `Then`. It's okay to interleave When / Then / When / Then, as long as you do so judiciously.

If it's not an assertion, and just setup, put it into a single `Given`: `Given I arm the bomb with only the disarm code`.

overlord/features/bomb_code_submit.feature
```diff
@@ -0,0 +1,73 @@
...
+    And I enter the correct disarm code
+    Then the bomb should be disarmed
+    And I should be able to arm the bomb
+
+  @bad
+  Scenario: I enter either code incorrectly
+    When I enter incorrect codes
+    And I press Submit Codes
+    Then a flash message should appear
+    And the dialog should not disappear
```
jmmastey: What kind of flash message? This is a little odd.

overlord/features/bomb_code_submit.feature
```diff
@@ -0,0 +1,73 @@
...
+
+
+
+
```
jmmastey: What if I submit the disarm code and the bomb is already disarmed? What if I enter the arm code and the bomb is already armed?

If I put the arm code into the disarm box for a disarmed bomb?

overlord/features/bomb_explode.feature
```diff
@@ -0,0 +1,20 @@
+#encoding: utf-8
+
+@explode
+Feature: Bomb timer expires
+  In order to explode
+  As the bomb
+  I should hit zero without being disarmed
```
jmmastey: Heh, interesting choice of description. I could buy this.

overlord/features/bomb_panel_hack.feature
```diff
@@ -0,0 +1,45 @@
+#encoding: utf-8
+
+@panel
+Feature: Hacking into a panel
+  In order to hack into a panel
+  As an evil overlord
+  I should be able to toggle buttons
+  And submit hacks
```
jmmastey: Why would the overlord hack the panel? Presumably he knows the codes.

overlord/features/bomb_panel_hack.feature
```diff
@@ -0,0 +1,45 @@
...
+    When I press submit on that panel
+    Then the panel buttons should reset
+
+  @happy
+  Scenario: Hacking all panels
+    And I enter correct sequences for all panels
+    Then the timer should stop
+    And I should be unable to arm the bomb
+    And I should be unable to disarm the bomb
+    And the bomb should be hacked
```
jmmastey: No negative cases here?

overlord/features/bomb_panel_hack.feature
```diff
@@ -0,0 +1,45 @@
...
+    And the depressed buttons form a correct sequence
+    When I press submit on that panel
+    Then the panel should turn green
+
+  @happy
+  Scenario: Submitting incorrect panel hack
+    And I choose a panel
+    And the depressed buttons form an incorrect sequence
+    When I press submit on that panel
+    Then the panel buttons should reset
```
jmmastey: Ahh, I see. Before my below comment I breezed through this test, which is... sort of sad path?

overlord/features/step_definitions/bomb_code_submit.rb
```diff
@@ -0,0 +1,42 @@
+# encoding: utf-8
+
+When(/^I press Submit Codes$/) do
+  find('#submit-codes').click
+end
+
+When(/^I enter nothing$/) do
+  @arm_actual = @default_arm
+  @disarm_actual = @default_disarm
+end
```
jmmastey: This was surprising. I was expecting this step not to do anything.

overlord/features/step_definitions/bomb_code_submit.rb
```diff
@@ -0,0 +1,42 @@
...
+  if disarm
+    @disarm_actual = @my_disarm_code
+    @arm_actual = @default_arm
+    find('#disarm-code-input').set(@disarm_actual)
+  else
+    @arm_actual = @my_arm_code
+    @disarm_actual = @default_disarm
+    find('#arm-code-input').set(@arm_actual)
+  end
+end
```
jmmastey: If it's an `if` anyway, just make it two step definitions.

overlord/features/step_definitions/bomb_code_submit.rb
jmmastey: Also, I think this is really a `Given` step.

Given: Setup before the action
When: Action happens
Then: Assertion about the results

overlord/features/step_definitions/bomb_code_submit.rb
```diff
@@ -0,0 +1,42 @@
...
+When(/^I enter both the arm and disarm code$/) do
+  @disarm_actual = @my_disarm_code
+  @arm_actual = @my_arm_code
+  find('#disarm-code-input').set(@disarm_actual)
+  find('#arm-code-input').set(@arm_actual)
+end
+
+When(/^I enter incorrect codes$/) do
+  find('#disarm-code-input').set('THIS WONT WORK')
+  find('#arm-code-input').set('AND NEITHER WILL THIS')
```
jmmastey: More idiomatic: `fill_in 'Arm Code', with: 'THIS WONT WORK'`

Had to type it three times to skip the apostrophe. :D

overlord/features/step_definitions/bomb_code_submit.rb
```diff
@@ -0,0 +1,42 @@
...
+end
+
+Then(/^a flash message should appear$/) do
+  expect(find('.flash').text).not_to eq('')
+end
```
jmmastey: `expect(page).to have_css('.flash')`

Maybe...

`expect(find('.flash').text).not_to be_empty`

overlord/features/step_definitions/bomb_code_submit.rb
```diff
@@ -0,0 +1,42 @@
...
+
+Then(/^the dialog should not disappear$/) do
+  find('.code-input')
+end
```
jmmastey: Ehhhh, specify that it's visible. There's no real assertion here.

overlord/features/step_definitions/bomb_dis_arming.rb
```diff
@@ -0,0 +1,48 @@
+# encoding: utf-8
+require './classes/bomb.rb'
+
+Given(/^the bomb is armed$/) do
+  selector = '.bomb'
+  find('#password').set(@arm_actual)
+  find('#confirm').click
+
+  selector << ":not(.disarmed)"
+  find(selector)
```
jmmastey: Confused. Why the push into selector? It only ever has one value.

overlord/features/step_definitions/bomb_dis_arming.rb
```diff
@@ -0,0 +1,48 @@
...
+
+Then(/^the bomb should (not )?be (dis)?armed$/) do |should_not, disarm|
+  selector = '.bomb'
+  if disarm
+    selector << (should_not ? ':not(.disarmed)' : '.disarmed')
+  else
+    selector << (should_not ? '.disarmed' : ':not(.disarmed)')
+  end
+  find(selector)
+end
```
jmmastey: Too many things happening in here. Just write the `selector` all at once, and split based on arm/disarm, probably. Brevity is nice, unless it introduces complexity.

overlord/features/step_definitions/bomb_dis_arming.rb
```diff
@@ -0,0 +1,48 @@
...
+
+Then(/^the timer should (not )?(start|stop)$/) do |should_not, action|
+  sleep 1
```
jmmastey: Sleep is anathema to a well maintained test suite. Find some other way to do this. Basically, you'll introduce enough sleeps over time that the entire suite is unacceptably slow.

overlord/features/step_definitions/bomb_dis_arming.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  time = find('.minutes').text + find('.seconds').text
+  shouldnt_start = should_not && action == 'start'
+  should_stop = !should_not && action == 'stop'
```
jmmastey: Again, too much going on here.

overlord/features/step_definitions/bomb_explode.rb
```diff
@@ -0,0 +1,32 @@
...
+end
+
+When(/^the timer hits zero$/) do
+  submit_button = '.columns > '
+  submit_button << '.column:first-child > '
+  submit_button << 'input[type="submit"]'
+
+  12.times do
+    find(submit_button).click
+  end
```
jmmastey: ?? I'm unsure how this code relates to the timer having hit zero.

overlord/features/step_definitions/bomb_explode.rb
```diff
@@ -0,0 +1,32 @@
...
+
+Then(/^I should be (un)?able to (?:dis)?arm the bomb$/) do |unable|
+  if unable
+    find('#password[disabled]')
+    find('#confirm[disabled]')
+  else
+    find('#password:not([disabled])')
+    find('#confirm:not([disabled])')
+  end
+end
```
jmmastey: Two steps.

overlord/features/step_definitions/bomb_panel_hack.rb
```diff
@@ -0,0 +1,116 @@
...
+  @default_arm = '1234'
+  @default_disarm = '0000'
+
+  @my_arm_code = '8008'
+  @my_disarm_code = '0880'
+
+  @arm_actual = @my_arm_code
+  @disarm_actual = @my_disarm_code
+  @default_incorrect = '9999'
+end
```
jmmastey: This is a lot of setup to be in a method named "I'm at the bomb page".

overlord/features/step_definitions/bomb_panel_hack.rb
```diff
@@ -0,0 +1,116 @@
...
+      selector << ".connector:nth-child(#{button_number + 1})"
+      find(selector).click
+    end
+
+    submit_button = ".columns > "
+    submit_button << ".column:nth-child(#{panel}) > "
+    submit_button << "input[type='submit']"
+
+    find(submit_button).click
+  end
```
jmmastey: Overall, there's too much happening in these step definitions. Move some of it out to helper methods, refactor to eliminate branch conditions, etc.

overlord/overlord.rb
```diff
@@ -1,14 +1,74 @@
...
+
+  session[:bomb].disarm(params[:code]) if session[:bomb].armed?
+  session[:bomb].arm(params[:code]) if session[:bomb].disarmed?
+end
+
+def codes_valid?(params)
+  num_expr = /^[0-9]{4}$/
+
+  valid_arm_code = params[:arm] =~ num_expr
+  valid_disarm_code = params[:disarm] =~ num_expr
```
jmmastey: It's okay to keep the num_expr inlined in this case. It's short and comprehensible already, so the indirection is moving it to a variable is probably not warranted.

overlord/overlord.rb
```diff
@@ -1,14 +1,74 @@
...
+
+  response
+end
+
+def entered_code_matched?(params)
+  return false if session[:bomb].attempts_remain == 0
+
+  session[:bomb].disarm(params[:code]) if session[:bomb].armed?
+  session[:bomb].arm(params[:code]) if session[:bomb].disarmed?
+end
```
jmmastey: This method returns nil if you disarm correctly.

jmmastey: If you need a return value, switch to a single `if/else` statement, which will return either branch. This postfix form returns true/false if arming, nil otherwise.

overlord/public/js/index.js
```diff
@@ -0,0 +1,153 @@
...
+      if(e.keyCode == 13)
+      {
+          $(this).trigger("enterKey");
+      }
+  });
+
+  $('#password').bind('enterKey', enter_code);
+  $('#arm-code-input').bind('enterKey', initial_code_submit).focus();
+  $('#disarm-code-input').bind('enterKey', initial_code_submit);
+  $('#submit-codes').bind('enterKey', initial_code_submit);
```
jmmastey: Good use of methods. Fwiw, you can put these into one CSS selector if you feel like it. `$("#arm-code-input, #disarm-code-input")` works as expected.

overlord/public/js/index.js
```diff
@@ -0,0 +1,153 @@
...
+  $el.html(1 - parseInt($el.html()));
+}
+
+function detonate() {
+  clearInterval(ticker);
+  ticker = false;
+  $('.bomb').addClass('detonated');
+  $('#password, #confirm').attr('disabled', 'true');
+
+  var url = 'https://www.youtube.com/embed/qDMUekfOR-E';
```
jmmastey: I see what you did there.

overlord/public/js/index.js
```diff
@@ -0,0 +1,153 @@
...
+  var params = '?start=6&autoplay=1&controls=0&loop=1&rel=0';
+  var iframe = $('<iframe>')
+    .attr('src', url + params)
+    .attr('height', '390')
+    .attr('width', '640')
+    .addClass('explosion-overlay');
```
jmmastey: Better to use CSS For these properties.

overlord/public/js/index.js
```diff
@@ -0,0 +1,153 @@
...
+  var binary = "";
+  for(var idx = 0; idx < $seq.length; idx++) {
+    var dp = $seq.eq(idx).hasClass('depressed');
+    binary += dp ? 1 : 0;
+  }
+
+  attempt_hack($col, $seq, binary, panel);
+}
+
+function attempt_hack(col, seq, binary, panel) {
```
jmmastey: What an odd piece of feedback. Anyway, yeah, JS should typically be camelCase (really technically pascal case), not snake_case.

overlord/views/index.haml
```diff
@@ -0,0 +1,51 @@
...
+    %link{:rel => :stylesheet, :type => :"text/css", :href => "css/index.css"}
+    %script{:src => "https://code.jquery.com/jquery-1.11.3.min.js"}
+    %script{:src => "js/index.js"}
+  %body
+    .code-input
+      .message.box
+        %h4 Welcome back our one, true, dearest overlord.
+        %h1 Please enter your arm / disarm codes.
+        %input.enterable{:id => 'arm-code-input', :type => 'text', :placeholder => 'Arm Code', :autofocus => true}
+        %input.enterable{:id => 'disarm-code-input', :type => 'text', :placeholder => 'Disarm Code'}
```
jmmastey: Generally don't put the type of the thing into its name. `arm-code`, not `arm-code-input`.

overlord/views/index.haml
```diff
@@ -0,0 +1,51 @@
...
+        %h2 To whatever pathetic bomb defuser this may concern.
+        %p SECURITY WARNING: This ocean-annihilating bomb is currently inactive. Once armed with the secret code, there are three extremely tedious and difficult ways to disarm it again (so you might as well not try, especially since you will be underwater):
+        %ol
+          %li Hack into each of the 6 panels on the bomb by specifying the correct binary sequences.
+          %li Enter the super secret disarm code that only the overlord knows.
+          %li 
+            Simply 
+            %b [CLASSIFIED] 
+            until 
+            %b [REDACTED].
```
jmmastey: Heh.

jmmastey:
0. Fix your master branch. 
1. The non-test notes are superficial, don't sweat them much.
2. I'd mostly like to see a refactor on the step definitions, and maybe another swing at the right level of concision vs. reuse of the steps themselves. If you want to move ahead, that's cool too, just take it into account in the other stuff you do.

jmmastey:
Btw, good work, it's solid testing code.

jklemen:
Great input. Really helpful. I'm going to try to pull this one again just to make sure I'm solid on test writing.

PR #871, opened 2015-07-24
Overlord V1

mcdaniel67:
My first draft of overlord. I didn't commit enough along the way.


mcdaniel67:
Should I go through and correct the cucumber-rails generated files? HoundCI is complaining about some of them, but I didn't write most of those.

overlord/Guardfile
```diff
@@ -0,0 +1,25 @@
...
+## watching the project directory ('.'), then you will want to move
+## the Guardfile to a watched dir and symlink it back, e.g.
+#
+#  $ mkdir config
+#  $ mv Guardfile config/
+#  $ ln -s config/Guardfile .
+#
+# and, you'll have to watch "config/Guardfile" instead of "Guardfile"
+
+guard "cucumber" do
```
johnmcconnell: Do you actually feel guard lowers your development time?
How do you use it? Have it run in background and check tests as you go?
Legitimately curious

overlord/bomb.rb
```diff
@@ -0,0 +1,54 @@
...
+      false
+    end
+    @deactivation_attempts += 1
+  end
+
+  def status_reader
+    @status
+  end
+
+  def self.valid_code?(code)
```
johnmcconnell: This seems like a weird way to implement `valid_code?` but after you think about it, it's actually clever. I think I'd still like:
```ruby
def valid_code?(code)
  !!(/^[[:digit]+$/ =~ code)
end
```

overlord/bomb.rb
```diff
@@ -0,0 +1,54 @@
+class Bomb
+  def initialize(activation = '', deactivation = '')
+    setup_activation_code(activation)
+    setup_deactivation_code(deactivation)
```
johnmcconnell: Instead of forcing the strings to be of a certain format, during initialization, let's be lazy. Use a reader method that has the default value prepared. Also why are you defaulting to `''`? Clearly, it is not an accepted code.

```ruby
def initialize(activation = '1234', deactivation = '0000')
  ...
end
```
or
```ruby
def initialize(activation = nil, deactivation = nil)
  @activation_code = activation || '1234'
  @deactivation_code = deactivation || '0000'
end
```

overlord/bomb.rb
```diff
@@ -0,0 +1,54 @@
...
+      @status = 'Deactivated'
+      true
+    elsif @deactivation_attempts >= 2
+      @status = 'Blown Up'
+      false
+    end
+    @deactivation_attempts += 1
+  end
+
+  def status_reader
```
johnmcconnell: We can just name this `status`, naming things `[variable]_reader` or `get_[variable]` is pretty uncommon from what I have seen.

overlord/features/homepage.feature
```diff
@@ -0,0 +1,44 @@
+Feature: Homepage
```
johnmcconnell: I'm digging these tests

overlord/features/step_definitions/steps.rb
```diff
@@ -0,0 +1,50 @@
...
+  if page.respond_to? :should
+    page.should have_content(text)
+  else
+    assert page.has_content?(text)
+  end
+end
+
+Then(/^I should be on (.+)$/) do |name|
+  current_path = URI.parse(current_url).path
+  if current_path.respond_to? :should
```
johnmcconnell: What? Why? Are you running different versions of rspec?

overlord/overlord.rb
```diff
@@ -1,14 +1,96 @@
...
+    if session[:bomb].status_reader == 'Active'
+      redirect to('/deactivate/')
+    elsif session[:bomb].status_reader == 'Blown Up'
+      redirect to('/boom/')
+    else
+      redirect to('/hero/')
+    end
+  end
+
+  def validate_codes(activation, deactivation)
```
johnmcconnell: Put this logic into a model over a controller. It's fine now, but as each feature gets added or changed, the complexity of this class will grow

johnmcconnell:
:+1: good stuff

overlord/bomb.rb
paulghaddad: +1

overlord/bomb.rb
```diff
@@ -0,0 +1,54 @@
+class Bomb
+  def initialize(activation = '', deactivation = '')
+    setup_activation_code(activation)
+    setup_deactivation_code(deactivation)
+    @deactivation_attempts = 0
+    @status = 'Inactive'
+  end
+
+  def setup_activation_code(activation)
```
paulghaddad: John's comments in your initialize method will clean this method and the next up.

overlord/bomb.rb
```diff
@@ -0,0 +1,54 @@
...
+  def setup_deactivation_code(deactivation)
+    if deactivation == ''
+      @deactivation_code = '0000'
+    else
+      @deactivation_code = deactivation
+    end
+  end
+
+  def activate(code)
+    if @activation_code == code && @status != 'Active'
```
paulghaddad: Would it make sense to extract this into a method?

overlord/bomb_spec.rb
```diff
@@ -0,0 +1,60 @@
+require_relative('bomb.rb')
```
paulghaddad: Use `require_relative 'bomb'`

overlord/bomb_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+describe Bomb do
```
paulghaddad: For the future, using Factory Girl here would help out greatly. You can use different traits for the different types of bombs you need. Also, take a look at using `let`.

overlord/features/activation.feature
```diff
@@ -0,0 +1,48 @@
...
+  As an evil overlord
+  I should be able to input my specified activation code
+
+  Scenario: Attempt to jump directly to the activation page
+    Given I am on "the activation page" of the bomb
+    Then I should be on the starting page
+
+  Scenario: Correctly displays bomb status
+    Given I am on "the activation page" of the bomb
+    And I click the "Submit" button
```
paulghaddad: This step definition is too low level. Operate at a single, higher level of abstraction. A good step would read:

`When I create a new bomb`

See this article for reference: https://robots.thoughtbot.com/acceptance-tests-at-a-single-level-of-abstraction

This should also be a `When` statement.

overlord/features/activation.feature
```diff
@@ -0,0 +1,48 @@
...
+
+  Scenario: Accepts a specified activation code
+    Given I specified my "activation" code as "1111"
+    And I fill in "activation" with "1111"
+    And I click the "Activate" button
+    Then I should be on the deactivation page
+
+  Scenario: Double activation should have no effect
+    Given I am on "the starting page" of the bomb
+    And I click the "Submit" button
```
paulghaddad: Per my comment above, these low level step definitions can be encapsulated into higher level step definitions. See this file as an example:

https://github.com/jmmastey/level_up/blob/master/features/users/sign_in.feature

overlord/features/activation.feature
```diff
@@ -0,0 +1,48 @@
...
+    And I fill in "activation" with "1234"
+    And I click the "Activate" button
+    And I try to load "the activation page" of the bomb
+    And I fill in "activation" with "1234"
+    And I click the "Activate" button
+    Then I should be on the deactivation page
+
+  Scenario: Incorrect activation code
+    Given I specified my "activation" code as ""
+    And I fill in "activation" with "1111"
```
paulghaddad: A better way of conveying the purpose of a scenario like this to someone reading it would be:

```
Given I am on "the starting page" of the bomb
When I use an incorrect activation code
Then I should be on the activation page
```

overlord/overlord.rb
paulghaddad: +1 
This concept will carry to CYOA where you'll want to keep your controllers lean.

paulghaddad:
@mcdaniel67 Take a look at these comments and submit a second iteration. This is a good first iteration. Really focus on writing your feature tests at a higher level that read well to someone taking a look at your code for the first time.

overlord/Guardfile
mcdaniel67: I think that one of the steps I followed for setting up cucumber/rspec setup guard on it's own. I didn't use guard intentionally.

PR #872, opened 2015-07-24
Data science

zsyed91:
Refactored data science exercise


data_science/lib/cohort.rb
```diff
@@ -0,0 +1,46 @@
+class Cohort
+  attr_accessor :name, :visits
+
+  def initialize(name)
+    @name = name
+    @visits = []
+  end
+
+  def add_visit(visit)
+    @visits << visit
```
zsyed1: You don't really need to store the individual data points. All you need to track is the number of conversions/non-conversions. Imagine the data set is large.

data_science/lib/cohort.rb
```diff
@@ -0,0 +1,46 @@
...
+    size = sample_size
+    Math.sqrt((rate * (1 - rate)) / size)
+  end
+
+  def conversion_rate_range
+    [conversion_rate - standard_error,
+     conversion_rate + standard_error]
+  end
+
+  def generate_pass_fail_info
```
zsyed1: You can shorten the name to `pass_fail_info`.

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+  end
+
+  describe '#conversions' do
+    it 'should return the number of passing conversions' do
+      cohort = Cohort.new(name)
+      5.times { cohort.add_visit(passing_visit) }
+      3.times { cohort.add_visit(failing_visit) }
+      expect(cohort.conversions).to eq(5)
+    end
```
zsyed1: I really like the way you did these tests. Some may prefer you move lines 19 and 20 into a single `let`.

zsyed1: SInce you're doing this setup a lot, it's probably a good idea to create a helper method:

```
def create_cohort(name, passes, fails)
  cohort = Cohort.new(name)
  passes.times { cohort.add_visit(passing_visit) }
  fails.times { cohort.add_visit(failing_visit) }
end
```
And then make `passing_visit` and `failing_visit` methods as well in the same helper file.

data_science/spec/visitor_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  describe '#conversion?' do
+    it 'should return true if result is 1' do
+      expect(passing_visit.conversion?).to be(true)
+    end
+
+    it 'should return false if the result is 0' do
+      expect(failing_visit.conversion?).to be(false)
+    end
+  end
```
zsyed1: This is good work, some of my comments are real nitpicking. I think this is closed.

Zamir

data_science/spec/cohort_spec.rb
paulghaddad: Zamir is definitely correct on both of these points.

data_science/lib/cohort.rb
```diff
@@ -0,0 +1,46 @@
...
+    @name = name
+    @visits = []
+  end
+
+  def add_visit(visit)
+    @visits << visit
+  end
+
+  def conversions
+    visits.count(&:conversion?)
```
paulghaddad: Nice use of the & idiom!

data_science/lib/cohort.rb
```diff
@@ -0,0 +1,46 @@
+class Cohort
```
paulghaddad: Really nice short methods in this class!

data_science/spec/cohort_loader_spec.rb
```diff
@@ -0,0 +1,86 @@
+require_relative '../lib/cohort_loader'
+
+RSpec.describe CohortLoader do
+  let(:filename) { 'spec/data_export.json' }
+  let(:loader) { CohortLoader.new(filename).tap(&:load) }
+  let(:standard_error) { Math.sqrt(0.125) }
+  let(:total_standard_error) { 0.25 }
```
paulghaddad: Nice use of lets here.

data_science/spec/cohort_loader_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  describe '.new' do
+    it 'should take in a filename' do
```
paulghaddad: I'd personally have two separate `it` blocks here as the expectations are testing for different things, an error in one case.

data_science/spec/cohort_loader_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      expect { CohortLoader.new }.to raise_error(ArgumentError)
+      expect(CohortLoader.new(filename)).to be_a(CohortLoader)
+    end
+  end
+
+  describe '#load' do
+    it 'should successfully parse the json file' do
```
paulghaddad: I'm not really sure if this is testing that it successfully parsed the json file. It's only testing that an error wasn't raised.

data_science/spec/cohort_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+require_relative '../lib/visit'
+
+RSpec.describe Cohort  do
+  let(:name) { "A" }
+  let(:passing_visit) { Visit.new(1, '10-10-15') }
+  let(:failing_visit) { Visit.new(0, '10-10-15') }
+  let(:standard_error) { Math.sqrt(0.25 * 0.75 / 40) }
+
+  describe '.new' do
+    it 'should be a Cohort class' do
```
paulghaddad: I don't think this is a super valuable test.

paulghaddad:
@zsyed91 Very nice job overall! I like the organization of your tests, specifically. For something like this in the future, where you are testing numerous objects, check out Factory Girl.

You are good to go on this exercise.

PR #874, opened 2015-07-27
Dinodex - Resubmission

oharace:
I kept one detail the same from the last submission, mainly how I filter the calls. It was suggested that I do so without passing back a new instance of DinosaurCatalog every time and instead find a way to just pass back 'self'. However, this feature was how I was instructed to do it after conversing with Dan, so I didn't change it. If I should change this before this exercise will be accepted, I will spend more time trying to fix it.


dino_catalog/csv_parser.rb
```diff
@@ -0,0 +1,22 @@
+require 'csv'
+require_relative 'dinosaur'
+
+class CsvParser
```
paulghaddad: Fantastic class. I have nothing to comment on here. You have written a very clear class with methods that do one thing. Nice work!

dino_catalog/dino_experimentation.rb
```diff
@@ -0,0 +1,30 @@
+require_relative 'dinosaur_catalog'
+
+catalog = DinosaurCatalog.new("dino_catalog/dinodex.csv")
+
+puts "Carnivore, Weight 4001"
+catalog.filter_by_diet("Carnivore").filter_by_weight(4001).print
```
paulghaddad: To eliminate the need for the `print` call at the end of each filter operation here, consider redefining the `to_s` method on `Catalog`. That way you can do this:

`catalog.filter_by_diet("Carnivore").filter_by_weight(4001).print`

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,37 @@
+class Dinosaur
+  attr_reader :period, :weight, :walking, :name, :diet, :continent, :description
+
+  def initialize(facts)
+    facts.each_value { |val| val.downcase! unless val.nil? }
```
paulghaddad: Perhaps extract this line into a separate method and name it well so the reader understands its purpose.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,37 @@
...
+    assign_name(facts)
+    assign_weight(facts)
+    assign_diet(facts)
+    assign_other_facts(facts)
+  end
+
+  def assign_name(facts)
+    @name = facts["name"] || facts["genus"]
```
paulghaddad: Great way to handle the two conditions.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def assign_weight(facts)
+    @weight = facts["weight"].to_i unless facts["weight"].nil?
```
paulghaddad: You can also use: `unless facts["weight"]`

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def initialize(facts)
+    facts.each_value { |val| val.downcase! unless val.nil? }
+    assign_name(facts)
+    assign_weight(facts)
+    assign_diet(facts)
+    assign_other_facts(facts)
+  end
+
+  def assign_name(facts)
```
paulghaddad: The assign_* methods are only called within this class, so consider making them private.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,84 @@
+require_relative 'csv_parser'
+require_relative 'printer'
+
+class DinosaurCatalog
+  TWO_TONS = 4000
+  attr_accessor :dinosaurs
+
+  def initialize(*file_paths)
```
paulghaddad: Nice use of the splat operator.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,84 @@
...
+    @dinosaurs = []
+    file_paths.each do |file_path|
```
paulghaddad: You may be able to use `#each_with_object` to eliminate line 9.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,84 @@
...
+    new_catalog
+  end
+
+  def all_dinosaurs
+    self
+  end
+
+  def filter_by_name(name)
+    dino = @dinosaurs.detect { |dinosaur| dinosaur.name == name.downcase }
+    dinos = dino.nil? ? [] : [dino]
```
paulghaddad: You should also be able to do:

`dinos = dino? ? [dino] : []`

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,84 @@
...
+  attr_accessor :dinosaurs
+
+  def initialize(*file_paths)
+    @dinosaurs = []
+    file_paths.each do |file_path|
+      @dinosaurs.concat(CsvParser.read_csv_file(file_path))
+    end
+  end
+
+  def pass_new_instance_for_filter(dinosaurs)
```
paulghaddad: I really like how this method handles the method chaining. I'm thinking there is a more clear name, but can't think of one.

dino_catalog/printer.rb
```diff
@@ -0,0 +1,58 @@
...
+  def self.print_with_facts(dinosaurs)
+    return print_no_dinosaurs_found if dinosaurs.empty?
+    dinosaurs.each { |dino| print_fact_string(dino) }
+  end
+
+  def self.print_no_dinosaurs_found
+    puts "No dinosaurs were found. Please modify your query and try again.\n"
+  end
+
+  def self.print_names(dinosaurs)
```
paulghaddad: I'd personally flip the if-else to make the logic easier to comprehend:

```
if dinosaurs.empty?
  print_no_dinosaurs_found
else
  dinosaurs.each { |dino| puts "#{dino.name.capitalize}" }
end
```

paulghaddad:
@oharace This is one of the best DInodexes I've seen. Awesome work on it. Take a look at the comments for future work, but you're definitely good on this exercise. Your code is extremely clear and easy to understand.

PR #875, opened 2015-07-27
Data science iteration #3

jklemen:


data_science/cohort.rb
```diff
@@ -0,0 +1,73 @@
...
+    { low: low.round(3), high: high.round(3) }
+  end
+
+  def to_s
+    str = "======== Cohort #{@name} ========\n"
+    str << "#{report_conversions}\n"
+    str << "#{report_sample_size}\n"
+    str << "#{report_conversion_rate}\n"
+    str << "#{report_standard_error}\n"
+    str << "#{report_confidence}\n"
```
djkotowski: Just make this a Heredoc.

data_science/cohort.rb
```diff
@@ -0,0 +1,73 @@
+require 'abanalyzer'
+
+class Cohort
+  attr_accessor :conversions, :sample_size, :name
```
djkotowski: Why make these accessors (especially conversions)?  Readers seem like they'd make more sense here.

data_science/cohort.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def initialize(name, samples)
+    @name = name
+    @sample_size = samples.count
+    @conversions = calc_conversions(samples)
+  end
+
+  def conversion_hash
```
djkotowski: Given this is only here to provide a formatted hash for ABAnalyzer, and given that you're not using ABAnalyzer in this class, does this belong here?

data_science/cohort.rb
```diff
@@ -0,0 +1,73 @@
+require 'abanalyzer'
```
djkotowski: Unused require.

data_science/data_science.rb
```diff
@@ -0,0 +1,68 @@
+require 'json'
+require 'abanalyzer'
+require './cohort.rb'
+
+class DataScience
+  attr_accessor :cohorts, :sample_size
```
djkotowski: Given this class is initialized via a JSON file, these should be readers, not accessors.

data_science/data_science.rb
```diff
@@ -0,0 +1,68 @@
...
+    @sample_size = json.count
+    @cohorts = make_cohorts(json)
+  end
+
+  def winner
+    return nil unless chi_square_test.different?
+    @winner = @cohorts.max_by(&:conversion_rate)
+  end
+
+  def report_all
```
djkotowski: In Cohort, methods beginning with `report_` _return_ strings, whereas here they output them to console.  You'll want to clarify this difference with your naming scheme.

data_science/data_science.rb
```diff
@@ -0,0 +1,68 @@
...
+      sample['cohort'].downcase == 'a'
+    end
+
+    raw_cohorts.map do |info|
+      Cohort.new(info[0]['cohort'], info)
+    end
+  end
+
+  def chi_square_test
+    values = @cohorts.inject({}) do |vals, cohort|
```
djkotowski: Since you're returning the original hash each time, use `each_with_object` instead.

data_science/rspec/cohort.rb
```diff
@@ -0,0 +1,42 @@
+require '../cohort.rb'
+require '../data_science.rb'
+
+describe Cohort do
+  let(:data_science) do
+    data_file = '../data_export_2014_06_20_15_59_02.json'
+    DataScience.new(data_file)
+  end
+  let(:cohort_a) { data_science.cohorts[0] }
```
djkotowski: Minor nitpick: add a blank line before and after do/end blocks.

data_science/rspec/cohort.rb
```diff
@@ -0,0 +1,42 @@
...
+    expect(cohort_b.sample_size).to eq(1543)
+  end
+
+  it "should calculate the correct cohort conversions" do
+    expect(cohort_a.conversions).to eq(47)
+    expect(cohort_b.conversions).to eq(79)
+  end
+
+  it "should calculate the correct conversion rate" do
+    expect(cohort_a.conversion_rate).to be_within(0.0001).of(0.0348)
```
djkotowski: Nice use of `be_within` to avoid unnecessary precision.

data_science/rspec/cohort.rb
```diff
@@ -0,0 +1,42 @@
+require '../cohort.rb'
+require '../data_science.rb'
+
+describe Cohort do
```
djkotowski: This spec is good with happy path.  What about other paths?  What happens if you have no visits?

data_science/rspec/cohort.rb
```diff
@@ -0,0 +1,42 @@
...
+    data_file = '../data_export_2014_06_20_15_59_02.json'
+    DataScience.new(data_file)
+  end
+  let(:cohort_a) { data_science.cohorts[0] }
+  let(:cohort_b) { data_science.cohorts[1] }
+
+  let(:cohort_a_ci) { cohort_a.confidence_interval }
+  let(:cohort_b_ci) { cohort_b.confidence_interval }
+
+  it "should calculate the correct sample size" do
```
djkotowski: 1.  Avoid the word "should" in your specs.  It isn't confident.
2.  Since you're describing different methods for each `it` here, use `describe` to identify the methods.  See [Better Specs](http://betterspecs.org/) for examples.

djkotowski:
Nice work!  Take a look at the comments for things to consider in future projects.  No need for a resubmission.

PR #876, opened 2015-07-27
Robot name v2

abake-enova:


avavita:
Erp, I commented on the commit instead of the "files changed" tab. Just wanna make sure you can see which lines I was referencing. Sorry I didn't send this sooner, I wasn't aware it would come up that way. 

https://github.com/abake-enova/level_up_exercises/commit/ec3865663ddbb16af67da87041d9c444e92f4dbd

robot_name/robot.rb
```diff
@@ -0,0 +1,34 @@
...
+class Robot
+  attr_accessor :name
+
+  def initialize(args = {})
+    @name_generator = args[:name_generator]
+    give_name
+    add_name_to_registry
+  end
+
+  def generate_char
```
paulghaddad: Nice use of not using the proc in this and the next method.

robot_name/robot.rb
```diff
@@ -0,0 +1,34 @@
...
+    ('A'..'Z').to_a.sample
+  end
+
+  def generate_num
+    rand(10)
+  end
+
+  def generate_name
+    name = ""
```
paulghaddad: See if you can use `inject` here to build up a name.

robot_name/robot_name_registry.rb
```diff
@@ -0,0 +1,22 @@
+InvalidNameError = Class.new(RuntimeError)
+NameCollisionError = Class.new(RuntimeError)
+
+class RobotNameRegistry
```
paulghaddad: I'd personally create an instance of a registry when working with this class.

paulghaddad:
Good job, particularly in your robot class. As I said in the comments, I'd prefer the Registry to create an instance. This will remove the need for class methods and I think it better OOP.

You're good to go on this!

PR #882, opened 2015-07-29
Triangle facts

loringalta:


robot_name/robot_name.rb
```diff
@@ -1,32 +1,54 @@
...
-  @@registry
+  @@registry ||= []
+  class RobotName < Robot
+    attr_accessor :name
 
-  def initialize(args = {})
-    @@registry ||= []
-    @name_generator = args[:name_generator]
+    def initialize(args = {})
+      @name_generator = args[:name_generator]
```
djkotowski: An optional parameter may make more sense than an args hash.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,54 @@
...
+      @name = @name_generator.call unless @name_generator.nil?
+      create_name(2, 3) unless @name
```
djkotowski: You can avoid checking two conditions here by moving the name creation to a separate method.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,54 @@
...
+      check_name
+    end
+
+    def random_ascii(string = "", num, char)
+      num.times do
+        string += ('A'..'Z').to_a.sample if char
```
djkotowski: Prefer `<<` to `+=`.

`+=` returns a new string when it's called; `<<` appends to the leftmost string.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,54 @@
...
+        string += rand(10).to_s unless char
```
djkotowski: Try to avoid this pattern:
```ruby
do_something if condition
do_something_else unless condition
```

Guard clauses help here.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,54 @@
...
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+    def create_name(num_char, num_dig)
+      char = random_ascii(num_char, true)
+      digit = random_ascii(num_dig, false)
+      @name = char + digit
```
djkotowski: Ditto my above comment on `<<` vs `+`

robot_name/robot_name.rb
```diff
@@ -1,32 +1,54 @@
...
+    end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+    def name_structure?
```
djkotowski: This isn't a clear method name. Something like `valid_name?` would be clearer.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,54 @@
...
+      @name =~ /[[:alpha:]]{2}[[:digit:]]{3}/
     end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+    def check_name
+      if !(name_structure?) || @@registry.include?(@name)
+        raise NameCollisionError, 'There was a problem generating the name'
+      end
+      @@registry << @name
```
djkotowski: I wouldn't expect a method `check_name` to add the name to the registry if it's valid.  Try to enforce a single responsibility per method.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,54 @@
-class NameCollisionError < RuntimeError; end
+require_relative 'exceptions'
 
 class Robot
-  attr_accessor :name
+  attr_accessor :registry
 
-  @@registry
+  @@registry ||= []
+  class RobotName < Robot
```
djkotowski: Make sure in the future, you split your different exercises to different branches.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,92 @@
...
+  def check_isosceles
+    puts "This triangle is isosceles!" if isosceles?
+  end
+
+  def check_scalene
+    puts "This triangle is scalene!\n" if scalene?
+  end
+
+  def check_right(angles)
+    puts "This triangle is also a right triangle! \n" if angles.include? 90
```
djkotowski: Wrap arguments for `#include?` in parentheses.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,92 @@
...
+  end
 
+  def print_angle(angles)
+    puts "The angles of this triangle are " + angles.join(', ') + "\n"
```
djkotowski: Use `<<` instead of `+`.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,92 @@
...
+  def cosine_angles
+    angle_a = law_of_cosine(side2, side3, side1)
+    angle_b = law_of_cosine(side1, side3, side2)
+    angle_c = law_of_cosine(side1, side2, side3)
+    [angle_a, angle_b, angle_c]
+  end
+
+  def calculate_angles
+    converted_angles = []
+    triangle_angles = cosine_angles
```
djkotowski: This is an unnecessary step.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,92 @@
...
+    triangle_angles.each do |ang|
+      converted_angles << radians_to_degrees(ang)
```
djkotowski: Take a look at `Enumerable#map` to simplify this method.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,92 @@
...
+  def equilateral?
+    side1 == side2 && side2 == side3
+  end
+
+  def isosceles?
+    [side1, side2, side3].uniq.length == 2
+  end
+
+  def scalene?
+    equilateral? || isosceles? ? false : true
```
djkotowski: Never use an if-statement or ternary to return true or false.  Just return the result of the condition.
```ruby
!(equilateral? || isosceles?)
```

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,92 @@
...
 
+  def print_angle(angles)
+    puts "The angles of this triangle are " + angles.join(', ') + "\n"
+  end
+
+  def radians_to_degrees(rads)
+    (rads * 180 / Math::PI).round
+  end
+
+  def law_of_cosine(s1, s2, s3)
```
djkotowski: Since you're returning an angle in this method, why not just call it that?  `calculate_angle` or just `angle` would make sense here.

djkotowski:
Nice first iteration on both.  Take a look at the comments, refactor and resubmit.  You're almost done here.

PR #883, opened 2015-07-29
done with arrowhead

loringalta:


arrowhead/arrowhead.rb
```diff
@@ -16,19 +16,19 @@ class Arrowhead
...
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+  def self.classify_shape(shapes, shape)
+    raise ArgumentError, "Unknown shape value" unless shapes.include? shape
```
djkotowski: Wrap args for `include?` in parentheses.

arrowhead/arrowhead.rb
```diff
@@ -16,19 +16,19 @@ class Arrowhead
...
+    arrowhead = shapes[shape]
+    "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
+  end
+
+  def self.classify_region(region, shape)
+    regions = CLASSIFICATIONS.keys
+    raise ArgumentError, "Unknown region entered" unless regions.include? region
```
djkotowski: Ditto above.

arrowhead/arrowhead.rb
```diff
@@ -16,19 +16,19 @@ class Arrowhead
...
+    shapes = CLASSIFICATIONS[region]
+    classify_shape(shapes, shape)
```
djkotowski: Given you extracted this logic out to a method, I might do the same for finding the region above.

djkotowski:
Nice work!  Take a look at the comments for future exercises.  You're done with this one.

PR #885, opened 2015-07-30
Overlord v1

mccormjt:
finished first iteration of overlord


overlord/app/css/tank.scss
```diff
@@ -0,0 +1,82 @@
...
+
+.tank {
+  position: relative;
+  top: -50px;
+  margin: 10px;
+  width: 350px;
+  height: 300px;
+  z-index: 25;
+  -webkit-transform-style: preserve-3d; 
+  -webkit-transform-origin: 50% 50% 0;
```
mccormjt: need for 3D animations

overlord/bomb.rb
```diff
@@ -0,0 +1,13 @@
+require_relative './code_sequence.rb'
```
sshao: `require_relative` [tries to load the library named string relative to the requiring file’s path](http://ruby-doc.org/core-2.0.0/Kernel.html#method-i-require_relative), so this can just be `require_relative 'code_sequence'`

overlord/code_sequence.rb
```diff
@@ -0,0 +1,47 @@
...
+  def initialize(default_code)
+    raise('Invalid Default Code') unless valid_code?(default_code)
+    @default_code = default_code
+    @check_attempts = 0
+  end
+
+  def set(code)
+    code ||= @default_code
+    @is_valid = valid_code?(code)
+    @is_valid && (@code = code)
```
sshao: I'm personally not a big fan of using assignment in a boolean condition -- it's not very readable (too easy to mistake it for an `==` operator) -- but it's probably fine here.

Although the sequence of `@is_valid && (@code = code); @is_valid` is a little weird. Why not return just line 16? It seems that if either of those operations fail (code isn't valid, can't set `@code` to the desired code), then the `set` operation should fail.

overlord/code_sequence.rb
```diff
@@ -0,0 +1,47 @@
...
+    @is_valid
+  end
+
+  def check(compare_code)
```
sshao: `check` -> `check?`. Or maybe `same_as?` "check" as a method name isn't as clear as it could be -- what are you checking for?

Also, this is ok but if you're defining your own CodeSequence class, might as well define a `#==` operator on it. Then your `bomb.activation_code.check(code)` can become `bomb.activation_code == CodeSequence.new(code)` calls instead.

overlord/code_sequence.rb
```diff
@@ -0,0 +1,47 @@
...
+    @is_valid = @code == compare_code
```
sshao: This I'd definitely refactor, or at least throw in some parens to clarify what's actually being evaluated without having to know order of operations off-hand. Is this `(@is_valid = @code) == compare_code`? Or is it `@is_valid = (@code == compare_code)`?

Also, the reuse of `@is_valid` here (compared to its assignment on line 15) is confusing. Why is it being reassigned? The name of the instance variable suggests that it represents the validity of a CodeSequence, so it seems strange that the validity of a CodeSequence could potentially change outside of being initialized/set.

overlord/features/bomb_state_transitions.feature
```diff
@@ -0,0 +1,26 @@
...
+  Scenario: Initial bomb state is activation
+    Then the bomb state should be activation
+
+  Scenario: Submitting input at activation state switches bomb to deactivation state
+    When I submit a valid code input
+    Then the bomb state should be deactivation
+
+  Scenario: Submitting input at deactivation state switches bomb to standby state
+    When I submit a valid code input
+    And I submit a valid code input
```
sshao: Because the steps here are the same, the purpose and progression of this test becomes unclear. Differentiate between "valid deactivation code" and "valid activation code".

overlord/spec/code_sequence_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  DEFAULT_CODE = '1234'
+  VALID_CODE = '9264'
+  INVALID_NUMERIC_CODE = '92et'
+  INVALID_LENGTH_CODE = '973'
+
+  before :example do
+    @code = CodeSequence.new(DEFAULT_CODE)
+  end
+
+  it 'raises an error when initialized with invalid default code' do
```
sshao: Might want to test this for `INVALID_NUMERIC_CODE` as well.

overlord/code_sequence.rb
```diff
@@ -0,0 +1,47 @@
...
+  attr_reader :code, :default_code, :check_attempts
+
+  CODE_LENGTH = 4
+  MAX_CHECK_ATTEMPTS = 2
+
+  def initialize(default_code)
+    raise('Invalid Default Code') unless valid_code?(default_code)
+    @default_code = default_code
+    @check_attempts = 0
+  end
```
sshao: Seems weird that CodeSequence.new doesn't initialize `@code` with the default code, and that an explicit `#set` call is required to *really* initialize a CodeSequence.

overlord/spec/code_sequence_spec.rb
sshao: Also, a potential bad path that isn't covered:

```ruby
irb(main):001:0> require "./code_sequence"
=> true
irb(main):002:0> CodeSequence.new("123")
RuntimeError: Invalid Default Code 
irb(main):003:0> CodeSequence.new("123 ")
=> #<CodeSequence:0x007faf08b50f68 @default_code="123 ", @check_attempts=0> 
irb(main):004:0> CodeSequence.new(" 123")
=> #<CodeSequence:0x007faf08b68eb0 @default_code=" 123", @check_attempts=0>
```

sshao:
Take a look at comments and resubmit :+1: Tests look good (save for an uncaught bad path -- might want to consider those more), but individual steps + code just need some clean up.

arrowhead/arrowhead.rb
```diff
@@ -21,7 +21,7 @@ def self.classify(region, shape)
       shapes = CLASSIFICATIONS[region]
```
paulghaddad: Make sure your pull requests don't have other exercises in the future.

overlord/features/game_over_screen.feature
```diff
@@ -0,0 +1,16 @@
...
+  Background:
+    Given I am on the bomb page
+    And the bomb has exploded
+
+  Scenario: the game over screen appears after the bomb has exploded
+    Then I should see the game over screen
+
+  Scenario: Being able to resart the game from the game over screen
+    When I click the try again button
+    Then the bomb state should be activation
```
paulghaddad: Check your language in your step definitions for clarity.

overlord/features/disarming_bomb.feature
```diff
@@ -0,0 +1,18 @@
...
+    And the bomb is armed
+
+  Scenario: Disarming the bomb 
+    When I submit a valid code input
+    Then the bomb should be disarmed
+
+  Scenario: Entering in the wrong code three times cuases bomb to explode
+    And I submit an invalid code input
+    And I submit an invalid code input
+    And I submit an invalid code input
```
paulghaddad: Here is an example of three step definitions that can be condensed into one. Strive to write step definitions at a higher level of abstraction. A business owner would only want to read:

`When I enter an invalid code three times`

Check all of your features to ensure they are at a higher level of abstraction. See this article:

overlord/features/code_panel_input_switches.feature
```diff
@@ -0,0 +1,31 @@
+Feature: code panel input swithces
```
paulghaddad: I don't think these are at the level of feature tests. They are more of a view spec, which can be tested with Jasmine or RSpec.

paulghaddad:
@mccormjt Please take a look at our comments and submit a new version. Really focus on writing your feature tests at a higher level of abstraction. Here is a good example: 

PR #887, opened 2015-07-31
Dino catalog

loringalta:


dino_catalog/dino_csv_cleaner.rb
```diff
@@ -0,0 +1,66 @@
...
+require 'pp'
+class DinoCsvCleaner
+  attr_accessor :csv_headers
+  def initialize(csv_files)
+    @csv_data = []
+    @csv_headers = []
+    @csv_files = csv_files
+  end
+
+  def open_file(csv_file)
```
zsyed1: These methods should all be private.
When you initialize this class, you may as well call the parsing methods.

dino_catalog/dino_csv_cleaner.rb
```diff
@@ -0,0 +1,66 @@
...
+        row["weight"] = row.delete("weight_in_lbs")
+      end
+    end
+  end
+
+  def standardize_diet
+    @csv_data.each do |row|
+      if row.keys.include? "carnivore"
+        carn_status = row.delete("carnivore")
+        row["diet"] = "Carnivore" if carn_status == "Yes"
```
zsyed1: Do you want to set `row["diet"]` to something if `carn_status` is not `"Yes"`?

dino_catalog/dino_csv_cleaner.rb
```diff
@@ -0,0 +1,66 @@
...
+    @csv_headers.delete("carnivore")
+  end
+
+  def standardize_keys
+    standardize_headers
+    standardize_weight
+    standardize_diet
+  end
+
+  def create_merged_files
```
zsyed1: Interesting approach, you are creating a new file that contains a cleaned version of the supplied files - hence the name `DinoCsvCleaner`. I like this!

Every other dino-dex I've seen parses the supplied files and normalizes them internally. There is a lot to say about the way you did it!

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class Dinosaur
+  def initialize
+  end
+
+  def create_attribute(header)
+    header.each do |h|
+      instance_variable_set("@#{h}", nil)
+      self.class.send(:attr_accessor, "#{h}")
+    end
```
zsyed1: @johnmcconnell Check this out! Lines 8-12

dino_catalog/dinosaur_search.rb
```diff
@@ -0,0 +1,100 @@
...
+  def initialize(dinos)
+    @searched_dino = dinos
+    init_arrays
+  end
+
+  def init_arrays
+    file_array = []
+    ARGV.map { |a| file_array << a }
+    @dino_hash = DinoCsvCleaner.new(file_array)
+    @dino_hash.create_dinosaur_hash
```
zsyed1: Lines 15 + 16: These two lines will likely be called together - so see my comment above, you may as well call `create_dinosaur_hash` inside the initializer (of `DinoCsvCleaner`)

dino_catalog/dinosaur_search.rb
```diff
@@ -0,0 +1,100 @@
...
+    @new_csv = read_new_csv
```
zsyed1: I would split this into two apps: `dino_preprocessor` and `dinodex`

dino_catalog/dinosaur.rb
johnmcconnell: Well alright alright alright. You don't need `instance_variable_set`. It can be shortened to
```ruby
def create_attributes(*headers)
  self.class.class_exec do
    attr_accessor *headers
  end
end
```

johnmcconnell: FYI, this also has some unexpected behavior and you may want to move the method to the class.
For example,
```ruby
d = Dinosaur.new
e = Dinosaur.new
d.test_1
# = > NameError: undefined local variable or method `test_1'
e.test_1
# = > NameError: undefined local variable or method `test_1'
d.create_attributes('test_1', 'test_2')
d.test_1 # => nil
e.test_1 #=> nil
```
So they both get the attribute methods created even though you send the object to just a single instance

johnmcconnell: Also this sort of hash initialization problem is solved by this gem. https://github.com/johnmcconnell/attr_init

dino_catalog/dinosaur_search.rb
```diff
@@ -0,0 +1,100 @@
...
+require 'json'
+
+class DinosaurSearch
+  def initialize(dinos)
+    @searched_dino = dinos
+    init_arrays
+  end
+
+  def init_arrays
+    file_array = []
```
zsyed1: Better name for this variable is `files`. No need to include the type `array` in it - that's a general rule btw, unless you have a real need to include the type.

dino_catalog/dinosaur_search.rb
```diff
@@ -0,0 +1,100 @@
...
+    ARGV.map { |a| file_array << a }
```
zsyed1: I think you just want to do a copy here. Any reason why you can't pass `ARGV` in line 15?

dino_catalog/dinosaur_search.rb
```diff
@@ -0,0 +1,100 @@
...
+    @dino_hash.create_dinosaur_hash
+    @new_csv = read_new_csv
+    @all_dinos = create_dino_object
+  end
+
+  def read_new_csv
+    keys = open("new.csv")
+    csv = CSV.new(keys, headers: true, header_converters: :symbol, converters: :all)
+    csv = csv.to_a.map(&:to_hash)
+    csv
```
zsyed1: Lines 24-25: Replace with just one line:
```
csv.to_a.map(&:to_hash)
```

dino_catalog/dinosaur_search.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  def create_dino_object(all_dinos = [])
+    @new_csv.each do |d|
+      dino = Dinosaur.new
+      dino.create_attribute(@dino_hash.csv_headers)
+      d.each { |k, _| dino.send("#{k}=", d[k]) }
+      all_dinos << dino
+    end
+    all_dinos
+  end
```
zsyed1: Lines 29 - 35 should be done using `map`:
```
@new_csv.map do |d|
  dino = Dinosaur.new
  dino.create_attributes...
  d.each...
  dino
```

dino_catalog/dinosaur_search.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  def biped
+    dinos = []
+    @all_dinos.each do |dino|
+      dinos << (dino.name || dino.genus) if dino.walking == "Biped"
+    end
```
zsyed1: LIne 39-42: Use `map`. Any time you are converting an array into another array of the same size, you can probably use `map`

zsyed1: Actually I just realized that you are selecting dinosaurs from the array. So instead, use `select` :smile: 

```
dinos = @all_dinos.select { |d| d.walking == "Biped" }
dinos = dinos.map { |d| d.name || d.genus }
```
You can improve your design by moving the responsibility of `biped` into the dinosaur class:

```
def biped?
  walking == "Biped"
end
```
Then the select statements will look better. In fact, you should do this for all the methods below. The method `biped?` on class `Dinosaur` is called a _predicate_ method. Later you'll see RSpec plays nicely with predicate methods and in general they make for more readable code (e.g. the question mark!)

dino_catalog/dinosaur_search.rb
```diff
@@ -0,0 +1,100 @@
...
+  def big
+    dinos = []
+    @all_dinos.each do |dino|
+      dinos << (dino.name || dino.genus) if dino.weight && dino.weight > 2000
+    end
+    @searched_dino << dinos
+    DinosaurSearch.new(@searched_dino)
+  end
+
+  def print_dino
```
zsyed1: I don't think this method is used: `print_dino` you can remove it.

dino_catalog/dinosaur_search.rb
```diff
@@ -0,0 +1,100 @@
...
+    dino.instance_variables.each do |v|
+      dino_var = dino.instance_variable_get("#{v}")
+      str += "#{v}: " + "#{dino_var}\n" if dino_var
+    end
+    str
+  end
+
+  def print_dino_facts(name)
+    str = ""
+    @all_dinos.next do |d|
```
zsyed1: Haven't seen `next` before - What does it do?
Can you perform this method using `map` and `join`?

dino_catalog/new.csv
```diff
@@ -0,0 +1,18 @@
...
+Quetzalcoatlus,Late Cretaceous,North America,Carnivore,Quadruped,Largest known flying animal of all time.,,440
+Yangchuanosaurus,Oxfordian,Asia,Carnivore,Biped,,,7200
+Dracopelta,Early Cretaceous or Late Jurassic,South America,Herbivore,Quadruped,One of the most primitive known Ankylosauria.,,
+,Jurassic,,,Biped,,Abrictosaurus,100
+,Jurassic,,Carnivore,Biped,,Afrovenator,
+,Albian,,Carnivore,Biped,,Carcharodontosaurus,3000
+,Jurassic,,,Quadruped,,Giraffatitan,6600
+,Cretaceous,,,Quadruped,,Paralititan,120000
+,Cretaceous,,Carnivore,Biped,,Suchomimus,10400
+,Triassic,,,Quadruped,,Melanorosaurus,2400
```
zsyed1: Nice job so far! I would recommend moving the search method logic down into the `Dinosaur` class. Also try to use the `Enumerable` class's `map` and `select` methods instead of `each`. If you would like to discuss anything or need clarification, please feel free to hit me up on hipchat - my username is @zamir

Zamir

dino_catalog/dinosaur_search.rb
paulghaddad: +1

dino_catalog/dinosaur_search.rb
paulghaddad: This is a great suggestion. I had the same comments.

dino_catalog/dinosaur_search.rb
```diff
@@ -0,0 +1,100 @@
...
+      dinos << (dino.name || dino.genus) if dino.walking == "Biped"
+    end
+    @searched_dino << dinos
+    DinosaurSearch.new(@searched_dino)
+  end
+
+  def small
+    dinos = []
+    @all_dinos.each do |dino|
+      dinos << (dino.name || dino.genus) if dino.weight && dino.weight < 2000
```
paulghaddad: A similar thing can be done here as suggested in the `#biped` method above.

dino_catalog/dinosaur_search.rb
```diff
@@ -0,0 +1,100 @@
...
+    end
+    @searched_dino << dinos
+    DinosaurSearch.new(@searched_dino)
+  end
+
+  def big
+    dinos = []
+    @all_dinos.each do |dino|
+      dinos << (dino.name || dino.genus) if dino.weight && dino.weight > 2000
```
paulghaddad: And here, too.

dino_catalog/dino_csv_cleaner.rb
```diff
@@ -0,0 +1,66 @@
+require 'csv'
+require 'pp'
+class DinoCsvCleaner
```
paulghaddad: Space needed between the `require` statements and the `class`.

dino_catalog/dino_csv_cleaner.rb
```diff
@@ -0,0 +1,66 @@
...
+  attr_accessor :csv_headers
+  def initialize(csv_files)
```
paulghaddad: Space needed here, too. Make sure you adhere to the style guides.

dino_catalog/new.csv
paulghaddad: @loringalta Also, I would like to see a command line interface for running this program. Other than that, take a look at the comments and submit a second iteration. Nice work on this exercise!

PR #860, opened 2015-07-23
Shopping cart jnyman

jnyman:
I provided a `NOTES.md` page to highlight some of my thinking as I went through this example.

I took the "bare minimum" route for Cucumber here since there are numerous decision points that come up that are usually predicated upon how teams like to write feature files. For example, the generic (asterisk) step in Cucumber can be helpful to concisely state business rules. Scenario outlines and tables (or both together) can be good but it depends on how people prefer them: data condition based or test condition based. (Wars have been fought over less!)

PR #889, opened 2015-07-31
Shopping Cart v2

Tornquist:


shopping_cart/address.feature
```diff
@@ -0,0 +1,37 @@
...
+  Scenario: Users can log in and still need to provide an address for estimates
+    Given there is a user account without shipping information
+    And the user is not logged in
+    When the user logs in
+    And the user enters address information
+    Then the cart will display a shipping estimate
+
+  Scenario: Users can log in and not have an address or estimate
+    Given there is a user account without shipping information
+    And the user is not logged in
```
magicalbanana: This seems to be repeated, move it to background steps.

shopping_cart/address.feature
```diff
@@ -0,0 +1,37 @@
+Feature:
```
magicalbanana: Try to have a main phrase for your Feature.

e.g,

Feature: Add user/customer address
  As a user
  I expect to be able to add provide my address
  In order to give my shipping details

shopping_cart/address.feature
```diff
@@ -0,0 +1,37 @@
...
+  As a user
+  I want to be able to add my address to display shipping information
+  In order to get shipping estimates
```
magicalbanana: Es ist besser if this was moved to a different feature.

Feature: Calculating Shipping Estimates.
  As a user
  I expect to be able to get my shipping estimates
  With the address I have or will be adding.

shopping_cart/coupon.feature
```diff
@@ -0,0 +1,33 @@
+Feature:
```
magicalbanana: Again, explain directly what es ist you are trying todo. Because when you run this, the line where the "Feature:" prepend is what gets outputted to the console.

Not a big deal really, but as a stake holder or someone in the business side, I want to be able to see what the feature entails directly and not having to read the entire scenario/explanation.

shopping_cart/coupon.feature
```diff
@@ -0,0 +1,33 @@
...
+  As a user
+  I want to submit a potential coupon
+  In order to see if I can change the order price
+
+  Note: Each item is $100 for testing
```
magicalbanana: Incorporate this to the test instead of leaving a note.

magicalbanana:
So far so good.

But I would like to see this resubmitted. Split the features out to the macro level and make scenarios that will handle cases in a specific manner.

magicalbanana:
Look at this for reference:

https://github.com/jmmastey/level_up_exercises/pull/860

shopping_cart/address.feature
Tornquist: The feature is that addresses cause shipping estimates.

I could break it out and test the ability to save an address, but looking at that on it's own doesn't test anything other than the ability to save a value to a user account.  I can rename address.feature to shipping_estimates.feature, but the actual content of the feature wouldn't change.  

Additionally, I would argue that creating an address.feature isn't terribly useful.  This current feature set tests the interaction with the forms, and the backend ability to save the data would be exercised through unit tests.  Any errors that would be caught by a feature set dedicated to testing if an address field can be saved are already covered by the shipping estimate features.

magicalbanana: Fair enough.

But ideally you would test the actual entry of the address UNLESS the only place you can add an address is through estimating shipping.

Feature tests are user facing. The assumption that the unit tests will handle user cases can lead you to missed edge cases that can cause bugs that are not seen tested on the unit side.

But if your unit tests are extremely thorough that it's a mix of integration testing and unit testing then the validations, persistency etc is fine.

But like I said, if this was the only place to add an address then this feature would be sufficient. But if there's another place to do it, then I would say this feature suite is lacking.

Tornquist: That's a good point.  It wouldn't make sense to be the only place to add the address information.  I've added a profile.feature page. It's pretty simple, but should address that discrepancy.

magicalbanana: Good.

Ideally when you write out the feature tests this is where you would build out the UI/UX and with that can start to write good code based on the intended features.

Your initial feature tests does not have to be final. Ideally, you write out the general or macro requirements (like in this case) and think for more scenarios or missing features required (e.g, the need to have a profile).

But awesome that you got it.

Just keep in mind that it is common practice to have a title for your feature and a summary underneath it.

Because when you run the console, the first line is often the argument for std::cout. But if you use a formatter like Fivemat where it is replaced by dots, then it's really not necessary.

But great anaylsis based on my feedback.

:dancer: :bee:

shopping_cart/profile.feature
```diff
@@ -0,0 +1,41 @@
...
+  Scenario:
+    When the address information is blank
+    Then the user will be shown an add address information message
+
+  Scenario:
+    Given the address information is blank
+    When the user edits their address
+    Then address information will be on the profile page
+
+  Scenario:
```
paulghaddad: Great job overall in this feature file of staying at a high level of abstraction and not using low-level step definitions!

paulghaddad:
@Tornquist  Good job on this! You are good to go. 

Keep up the nice work!

PR #897, opened 2015-08-04
Let`s Be Rational

zsyed1:
@jmmastey @djkotowski Here is an alternative to learning TDD. It's a lot simpler than `data_science` and easier to code closer to TDD. Instead of just deleting `data_science` I'm suggesting replacing it with this as an alternative to bridge `DinoDex` and `Overlord`.

Zamir


djkotowski:
This doesn't look very TDD.  You're writing the implementation first based on an existing API spec, then the tests.  We want to teach that by using TDD and writing "the code we want to see", we can write clear, understandable code with an understandable API.

zsyed1:
@djkotowski Let's discuss this in person first (within the context of the firm)

jmmastey:
Apologies for not seeing this sooner, my notifications are turned off for pulls here because of the number of training reviews. Overall, I think it's a move in the right direction. Some good and bad things.

Good: First, this is not `data_science`, which biases me towards it immediately. Writing tests for already implemented code is a common activity, both at Enova and in the field. I'd be really interested to see a case where adding coverage to the existing code allowed the developer to refactor a shitty implementation. This is one of the major "aha" moments of TDD. Also, this could point to some smells like testing private methods while reinforcing operator overloading. Coders are likely familiar enough with fractions to need very little research to complete this.

Bad: This is not a lot less dry than `data_science`, so I'd probably grow not to like `rational` any better. The HTTP stuff is a bit contrived for my liking, and I'd prefer to give them even a half-assed rspec than a test script that a very junior person might mimic because they don't know better.

_Overall_: The suggestion of a "pick up this half implemented thing, test it, and finish it with TDD" could be a solid, fast, and educational exercise. Let me know what you two talk about today @djkotowski @zsyed1.

jmmastey:
@djkotowski @zsyed1 did you end up discussing this?

zsyed1:
@jmmastey I've been making noise, but no decision has been made. I'm hoping @djkotowski just makes some alterations and we can push it through. I'm tired of pretending learners are actually doing TDD on this exercise. :last_quarter_moon_with_face:

djkotowski:
I've been a bit busy with other projects at the moment.  Once I finish my production work, I can turn my focus back to Level UP.

paulghaddad:
@zsyed1 Let's run with our thoughts from this morning on tweaking this update. I'm going to close this for now so we can get our exercise count down. :)

PR #901, opened 2015-08-07
Second Drafts of Triangle Facts, Arrowhead, and Robot Name

DejaAugustine:
I've run them all through Rubocop, and the only issue that I did not correct was this one, as I'm a little stumped on how to accomplish it.

```
triangle_facts/triangle_facts.rb:81:3: C: Assignment Branch Condition size for calculate_angles is too high. [15.3/15]
  def calculate_angles
  ^^^
```

https://github.com/DejaAugustine/level_up_exercises/blob/deja/stylish_ruby/triangle_facts/triangle_facts.rb#L81

```
  def calculate_angles
    angle_a = radians_to_degrees(calculate_angle(side_b, side_c, side_a))
    angle_b = radians_to_degrees(calculate_angle(side_a, side_c, side_b))
    angle_c = radians_to_degrees(calculate_angle(side_a, side_b, side_c))

    [angle_a, angle_b, angle_c]
  end
```

NOTE:
This was resolved by moving the `radians_to_degrees` call into the `calculate_angle` method

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -1,34 +1,68 @@
+# Public: Classifies an arrowhead based on criteria such as region and shape
+#
+# Examples
+#
+#   Arrowhead.classify(:far_west, :notched)
+#   # => "You have an 'Archaic Side Notch' arrowhead. Probably priceless."
+#
+#   Arrowhead.classify(:northern_plains, :lanceolate)
+#   # => "You have a 'Humbolt Constricted Base' arrowhead. Probably priceless."
 class Arrowhead
-  # This seriously belongs in a database.
+  # Public: Static classification table. Consider migrating to database.
   CLASSIFICATIONS = {
-    far_west: {
-      notched: "Archaic Side Notch",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Agate Basin",
-      bifurcated: "Cody",
+    :far_west => {
+      :notched    => "Archaic Side Notch",
+      :stemmed    => "Archaic Stemmed",
+      :lanceolate => "Agate Basin",
+      :bifurcated => "Cody",
     },
-    northern_plains: {
-      notched: "Besant",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Humboldt Constricted Base",
-      bifurcated: "Oxbow",
+    :northern_plains => {
+      :notched    => "Besant",
+      :stemmed    => "Archaic Stemmed",
+      :lanceolate => "Humboldt Constricted Base",
+      :bifurcated => "Oxbow",
     },
   }
 
-  # FIXME: I don't have time to deal with this.
+  # Public: Locates an arrowhead classification based on the provided criteria
+  #
+  # region - The general region in which the arrowhead was found, as a Symbol
+  # shape  - The shape of the arrowhead as a Symbol
+  #
+  # Examples
+  #
+  #   classify(:northern_plains, :notched)
+  #   # => "You have a 'Besant' arrowhead. Probably priceless."
+  #
+  # Returns a String describing the arrowhead
+  # Raises an Exception if the region is not in the classification table
+  # Raises an Exception if the shape is not in the classification table
   def self.classify(region, shape)
-    if CLASSIFICATIONS.include? region
-      shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+    raise "Unknown region, please provide a valid region." unless CLASSIFICATIONS.include? region
+
+    shapes = CLASSIFICATIONS[region]
+    raise "Unknown shape value. Are you sure you know what you're talking about?" unless shapes.include? shape
+
+    arrowhead = shapes[shape]
+    a_an = %(a e i o u).include?(arrowhead[0].downcase) ? "an" : "a"
+
+    "You have #{a_an} '#{arrowhead}' arrowhead. Probably priceless."
   end
 end
 
+# Demonstrates the use of the class
+puts ""
+puts "****************************"
+puts "* Arrowhead Classification *"
+puts "****************************"
+puts ""
+
+puts "Region: Northern Plains"
+puts "Shape:  Bifurcated"
 puts Arrowhead.classify(:northern_plains, :bifurcated)
+puts ""
+
+puts "Region: Far West"
+puts "Shape:  Bifurcated"
+puts Arrowhead.classify(:far_west, :bifurcated)
+puts ""
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -10,17 +10,17 @@
 class Arrowhead
   # Public: Static classification table. Consider migrating to database.
   CLASSIFICATIONS = {
-    :far_west => {
-      :notched    => "Archaic Side Notch",
-      :stemmed    => "Archaic Stemmed",
-      :lanceolate => "Agate Basin",
-      :bifurcated => "Cody",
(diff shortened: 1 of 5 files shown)


arrowhead/arrowhead.rb
```diff
@@ -15,20 +24,45 @@ class Arrowhead
...
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+    raise "Invalid region." unless CLASSIFICATIONS.include? region
```
djkotowski: Wrap method args in parentheses for clarity.  This becomes more important the more complex your method calls get.

arrowhead/arrowhead.rb
```diff
@@ -15,20 +24,45 @@ class Arrowhead
...
+
+    shapes = CLASSIFICATIONS[region]
+    raise "Invalid shape." unless shapes.include? shape
+
+    arrowhead = shapes[shape]
+    a_an = %(a e i o u).include?(arrowhead[0].downcase) ? "an" : "a"
```
djkotowski: I might extract this out to a method for further clarity.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,156 @@
...
+#   robot = Robot.new(registry: RobotRegistry.new,
+#                     name_validator: -> (name) { !name.empty? })
 class Robot
-  attr_accessor :name
+  # Public: Establishes getters for the name and registry attributes
+  attr_reader :name, :registry
 
-  @@registry
+  # Internal: Establishes readers for the generator and validator
+  attr_reader :name_generator, :name_validator
```
djkotowski: These are public as well.  If you don't want them publically accessible, remove it.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,156 @@
-class NameCollisionError < RuntimeError; end
+require 'set'
 
+# Public: General Robot Exception
```
djkotowski: Your code should be self-documenting.  No need for all these comments.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,156 @@
...
+# Public: Exception denoting an invalid name
+class NameValidationError < RobotError; end
+
+# Public: A registry of unique robot names based on the Set class
+#
+# Examples
+#
+#   registry = RobotRegistry.new
+#   registry << "B.E.N.D.E.R."
+class RobotRegistry < Set
```
djkotowski: Move each class into its own file.  Exception classes with no custom code can be grouped together in the same file.

djkotowski: Also, having RobotRegistry inherit from Set could have unintended consequences.  It definitely results in a lack of control that you could otherwise have by wrapping a set with the registry class instead.

robot_name/robot_name.rb
DejaAugustine: Had merely been adhering to the "Documentation" standard in the Ruby Styleguide.

djkotowski: We mostly use these style guides:
https://github.com/bbatsov/ruby-style-guide
https://github.com/thoughtbot/guides/tree/master/style/ruby

robot_name/robot_name.rb
```diff
@@ -1,32 +1,92 @@
...
+  end
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  def generate_name
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
```
djkotowski: These two procs never change.  Why not make them methods?

robot_name/robot_name.rb
```diff
@@ -1,32 +1,92 @@
...
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
+    "#{generate_char.call}#{generate_char.call}" \
+      "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
```
djkotowski: Is there a clearer way to write these?

robot_name/robot_name.rb
```diff
@@ -1,32 +1,92 @@
...
+  puts "Robot '#{robot.name}' Registered! (custom generator, default validator)"
+rescue RobotError => e
+  puts "Error Generating Robot: #{e.message}"
+end
 
-# Errors!
-# generator = -> { 'AA111' }
-# Robot.new(name_generator: generator)
-# Robot.new(name_generator: generator)
+puts ""
```
djkotowski: Redundant empty string.  Just call `puts` without any args.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,77 @@
...
+    !(equilateral? || isosceles?)
+  end
+
+  def print_facts
+    puts "Facts about a triangle described by side lengths of " \
+         "#{side_a}, #{side_b}, and #{side_c} units"
+
+    print_type
+    print_angles
+    puts ""
```
djkotowski: See earlier comment about `puts`.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,77 @@
...
+    @side_b = side_b
+    @side_c = side_c
+  end
+
+  def equilateral?
+    side_a == side_b && side_b == side_c
+  end
+
+  def isosceles?
+    Set.new([side_a, side_b, side_c]).length == 2
```
djkotowski: Why not just use `#uniq`?

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,77 @@
...
+
+  def print_type
+    puts "This triangle is equilateral!" if equilateral?
+    puts "This triangle is isosceles!" if isosceles?
+    puts "This triangle is scalene!" if scalene?
+  end
+
+  def print_angles
+    angles = calculate_angles
+    puts "The angles of this triangle are #{angles.join("\xC2\xB0, ")}\xC2\xB0"
```
djkotowski: I would make the hexadecimal values constants to clarify their purpose.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,77 @@
...
+    angle_a = calculate_angle(side_b, side_c, side_a)
+    angle_b = calculate_angle(side_a, side_c, side_b)
+    angle_c = calculate_angle(side_a, side_b, side_c)
+
+    [angle_a, angle_b, angle_c]
+  end
+
+  def calculate_angle(left, right, trans)
+    radians_to_degrees(
+      Math.acos((left**2 + right**2 - trans**2) / (2.0 * left * right)))
```
djkotowski: It may be clearer to write this
```ruby
radians = Math.acos((left**2 + right**2 - trans**2) / (2.0 * left * right))
radians_to_degrees(radians)
```

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,77 @@
...
-}
+# Demonstrates the use of the class
+[
+  [5, 5, 5],    # Example Equilateral Triangle
+  [5, 12, 13],  # Example Scalene Triangle
+  [5, 10, 10],  # Example Isosceles Triangle
+].each do |sides|
+  tri = Triangle.new(*sides)
+  tri.print_facts
+end
```
djkotowski: This statement is doing too much.  With the original example, it was clear that you were defining triangles and then iterating through them.

djkotowski:
Nice work.  A few things to change throughout.  Take a look at the comments, refactor and resubmit when ready.

triangle_facts/triangle_facts.rb
DejaAugustine: Again, was conforming to the GitHub standard: "Use Set instead of Array when dealing with unique elements."

PR #903, opened 2015-08-10
deja/triangle_facts

DejaAugustine:
Incorporated latest feedback and split into separate feature branch.

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -1,58 +1,81 @@
-# Killer facts about triangles AWW YEAH
+require 'set'
+
 class Triangle
-	attr_accessor :side1,:side2,:side3
+  attr_accessor :side_a, :side_b, :side_c
 
-	def initialize(side1,side2,side3)
-		@side1,@side2,@side3 = side1,side2,side3
-	end
+  DEGREE_SYMBOL = "\u00B0"
 
-	def equilateral()
-		return side1 == side2 && side2 == side3
-	end
+  def initialize(side_a, side_b, side_c)
+    @side_a = side_a
+    @side_b = side_b
+    @side_c = side_c
+  end
 
-	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
-	end
+  def equilateral?
+    side_a == side_b && side_b == side_c
+  end
 
-	def scalene()
-		unless ! (equilateral || isosceles)
-	    false
-	  else
-	    true
-	  end
-	end
+  def isosceles?
+    [side_a, side_b, side_c].uniq.length == 2
+  end
 
-	def recite_facts
-		puts 'This triangle is equilateral!' if equilateral
-		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+  def scalene?
+    !(equilateral? || isosceles?)
+  end
 
-		angles = self.calculate_angles(side1,side2,side3)
-		puts 'The angles of this triangle are ' + angles.join(',')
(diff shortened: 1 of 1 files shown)


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,81 @@
...
-	end
-
-	def radians_to_degrees(rads)
-		return (rads * 180 / Math::PI).round
-	end
+  attr_accessor :side_a, :side_b, :side_c
+
+  DEGREE_SYMBOL = "\u00B0"
+
+  def initialize(side_a, side_b, side_c)
```
paulghaddad: This could also be an opportunity to use Ruby's Keyword Arguments. But perhaps not since the position of the argument is irrelevant since it's a triangle.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,81 @@
...
+
+  def calculate_angles
+    angle_a = calculate_angle(side_b, side_c, side_a)
+    angle_b = calculate_angle(side_a, side_c, side_b)
+    angle_c = calculate_angle(side_a, side_b, side_c)
+
+    [angle_a, angle_b, angle_c]
+  end
+
+  def calculate_angle(left, right, trans)
```
paulghaddad: Really nice methods here, particularly `calculate_angles`, `print_angles` and `calculate_angle`. They to one thing well.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,81 @@
...
+  end
+
+  def print_type
+    puts "This triangle is equilateral!" if equilateral?
+    puts "This triangle is isosceles!" if isosceles?
+    puts "This triangle is scalene!" if scalene?
+  end
+
+  def print_angles
+    angles = calculate_angles
```
paulghaddad: I like your explaining variable here. I'd normally say to just put `calculate_angles` in line 43, but this helps me understand better.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,81 @@
...
+    puts "The angles of this triangle are " \
+         "#{angles.join("#{DEGREE_SYMBOL}, ")}#{DEGREE_SYMBOL}"
+    puts "This triangle is also a right triangle!" if angles.include?(90)
```
paulghaddad: What do you think about a method that returns a boolean for `angles.include?(90)`? This may improve the clarity of the code and simplify this method.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,81 @@
...
+  def print_type
+    puts "This triangle is equilateral!" if equilateral?
+    puts "This triangle is isosceles!" if isosceles?
+    puts "This triangle is scalene!" if scalene?
+  end
+
+  def print_angles
+    angles = calculate_angles
+    puts "The angles of this triangle are " \
+         "#{angles.join("#{DEGREE_SYMBOL}, ")}#{DEGREE_SYMBOL}"
```
paulghaddad: I love your use of the degree symbol. That's great! I never thought of that.

triangle_facts/triangle_facts.rb
paulghaddad: Some of the methods in this class can be made private.

paulghaddad: I do think there are better parameter names for left, right and trans. I had to think for a second. What do you think about `side_1`, `side_2` and `side_3`?

paulghaddad:
As in Robot Name, excellent job on this exercise. Your code is super easy to understand. Let me know your thoughts about my comments. You are good on this exercise. :+1:

PR #904, opened 2015-08-10
Supportive v1

ssapra:


supportive/supportive.rb
```diff
@@ -9,94 +9,53 @@ class BlagPost
...
-        DISALLOWED_CATEGORIES.include? category
-      end
-    else
-      @categories = []
-    end
-
-    @comments = args[:comments] || []
+    @categories = handle_categories(args)
+    @comments = Array.wrap(args[:comments])
     @body = args[:body].gsub(/\s{2,}|\n/, ' ').gsub(/^\s+/, '')
```
zsyed1: Here you can use the support method
```
@body = args[:body].squish
```

supportive/supportive.rb
```diff
@@ -9,94 +9,53 @@ class BlagPost
...
-
-    if categories.length > 1
-      last_category = categories.pop
-      suffix = " and #{as_title(last_category)}"
-    else
-      suffix = ""
-    end
-
-    label + ": " + categories.map { |cat| as_title(cat) }.join(", ") + suffix
+  def byline
```
zsyed1: Snake-Case:  `by_line` for the method name.

supportive/supportive.rb
```diff
@@ -9,94 +9,53 @@ class BlagPost
...
+    return '' unless author
```
zsyed1: What if the fields `name` and `url` don't exist?

supportive/supportive.rb
```diff
@@ -9,94 +9,53 @@ class BlagPost
...
+    @categories = handle_categories(args)
+    @comments = Array.wrap(args[:comments])
     @body = args[:body].gsub(/\s{2,}|\n/, ' ').gsub(/^\s+/, '')
-    @publish_date = (args[:publish_date] && Date.parse(args[:publish_date])) || Date.today
+    @publish_date = args[:publish_date].try(&:to_date) || Date.today
   end
 
   def to_s
-    [ category_list, byline, abstract, commenters ].join("\n")
+    [category_list, byline, abstract, commenters].join("\n")
```
zsyed1: Reject blank values:
```
[category_list, by_line, abstract, commenters].reject(&:blank?).join("/n")
```
Then those individual methods an return `nil` and you don't have to explicitly return `''`.

supportive/supportive.rb
```diff
@@ -9,94 +9,53 @@ class BlagPost
...
-      body
-    else
-      body[0..200] + "..."
-    end
+    body.truncate(200)
   end
-
 end
 
 blag = BlagPost.new("author"        => "Foo Bar",
```
zsyed1: Instead of returning blanks you can just `return` then squash the `nil` values in the `to_s` method.

Nice job fixing up this code. I'm not sure if these needs another PR - will leave that up to @paulghaddad !

Zamir

supportive/supportive.rb
ssapra: I believe [byline](https://www.google.com/search?q=byline) is actually a noun meaning the name of the author of an article/blog.

supportive/supportive.rb
ssapra: Currently, an author is only created in the initialize method if both parameters are present in the args. So this line would raise ```undefined method `name' for nil:NilClass```.

I could change it so that an author is always created when a blag post is initialized and then in the byline method, return nil unless both parameters exist. 

Not sure which of these ideas is better, but perhaps moving the check for the parameters from the initialize method to the byline method where the author object is actually being used makes more sense.

zsyed1: Okay

supportive/supportive.rb
```diff
@@ -9,94 +9,53 @@ class BlagPost
...
-
-    ordinal = case comments.length % 10
-      when 1 then "st"
-      when 2 then "nd"
-      when 3 then "rd"
-      else "th"
-    end
-    "You will be the #{comments.length}#{ordinal} commenter"
+    return '' unless comments_allowed? && comments.present?
+    "You will be the #{comments.length.ordinalize} commenter"
```
paulghaddad: I personally think an `if-else` may make lines 48-49 more readable. It's doing the same thing.

supportive/supportive.rb
paulghaddad: +1

paulghaddad:
@ssapra Take a look at the comments provided for your future work with ActiveSupport. Another submission is not necessary. Nice work.

PR #905, opened 2015-08-10
deja/arrowhead

DejaAugustine:
Incorporated latest feedback and split into separate feature branch.

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -1,5 +1,4 @@
 class Arrowhead
-  # This seriously belongs in a database.
   CLASSIFICATIONS = {
     far_west: {
       notched: "Archaic Side Notch",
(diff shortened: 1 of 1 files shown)


arrowhead/arrowhead.rb
```diff
@@ -15,20 +14,34 @@ class Arrowhead
...
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+    raise "Invalid region." unless CLASSIFICATIONS.include?(region)
```
paulghaddad: These two `raise` statements look good. I'd consider extracting `CLASSIFICATIONS.include?(region)` and `shapes.include?(shape)` into their own methods but it is not necessary here as it is pretty clear already.

arrowhead/arrowhead.rb
```diff
@@ -15,20 +14,34 @@ class Arrowhead
...
+
+    shapes = CLASSIFICATIONS[region]
+    raise "Invalid shape." unless shapes.include?(shape)
+
+    arrowhead = shapes[shape]
+
+    "You have #{a_an(arrowhead)} '#{arrowhead}' arrowhead. Probably priceless."
+  end
+
+  def self.a_an(word)
```
paulghaddad: Great thinking on this edge case.

arrowhead/arrowhead.rb
```diff
@@ -15,20 +14,34 @@ class Arrowhead
     },
   }
 
-  # FIXME: I don't have time to deal with this.
   def self.classify(region, shape)
```
paulghaddad: Nice work on this method. It is very simple and easy to understand.

paulghaddad:
There isn't much to say on this one. You did a fine job and made it very simply, eliminating the arrowhead look. You're good to go on this exercise.

PR #907, opened 2015-08-10
Overlord submission

lmenova:
- use 'rackup' at project root to run application...just 'rackup'...nothing else
- localhost:9292/configure_bomb to configure/arm/shutdown bomb
- localhost:9292 will show an images when bomb is not armed. If bomb is armed, you'll be redirected to 'disarm page'


overlord/lib/bomb.rb
```diff
@@ -0,0 +1,86 @@
+require 'sinatra/base'
+require 'singleton'
+
+class Bomb
+  include Singleton
```
enovajuan: Why Singleton?

overlord/lib/bomb.rb
```diff
@@ -0,0 +1,86 @@
...
+  def disarmed?
+    status == 'Disarmed'
+  end
+
+  def detonated?
+    status == 'Detonated'
+  end
+
+  def boot_bomb
+    @status = 'Booted'
```
enovajuan: looks like you marked these as accessors `status`, but you're using the `@` syntax here. Since you've marked them as accessors, you should only set these values once (if at all) using the `@syntax`, otherwise calling `status = something` is the equivalent of `self.status = something` the difference being, if you use the implicit self syntax (i.e. you can remove `self.` altogether), you create a single point of failure (attr_accessor's dynamically created method) for setting or reading the value of this instance variable, which down the line can save you from weird bugs that crop up if you're not really sure which method is setting the value where, and keeps your code clean.

enovajuan: Good rule of thumb. Only ever use the `@syntax = x` inside of a class' initialize method. use implicit self. syntax everywhere else.
These 2 are essentially the same as the last thing:
1) `self.variable = x`
2) `variable = x`
in the background both become:
`Class.instance.variable = x`

however NOT the same as:
`@variable = x`

overlord/lib/bomb.rb
```diff
@@ -0,0 +1,86 @@
...
+    @disarming_count += 1
+    @status = "Detonated" if disarming_count == MAX_ATTEMPTS
+  end
+
+  def configure_arming_code(code = nil)
+    @arming_code = code unless blank?(code)
+  end
+
+  def configure_disarming_code(code = nil)
+    @disarming_code = code unless blank?(code)
```
enovajuan: take a look at activeSupport classes. Including them will let you do:

`code.blank?` or `code.present?`

http://guides.rubyonrails.org/active_support_core_extensions.html#blank-questionmark-and-present-questionmark

overlord/lib/bomb.rb
```diff
@@ -0,0 +1,86 @@
...
+  def disarm(code)
+    if code == disarming_code
+      @status = "Disarmed"
+      @disarming_count = 0
+    else
+      track_count
+    end
+  end
+
+  def blank?(str)
```
enovajuan: take a look at activeSupport classes. Including them will let you do:

`code.blank?` or `code.present?`

http://guides.rubyonrails.org/active_support_core_extensions.html#blank-questionmark-and-present-questionmark

overlord/overlord.rb
```diff
@@ -1,14 +1,81 @@
...
+require 'sinatra/base'
+require_relative 'lib/bomb'
 
-require 'sinatra'
+class Overlord < Sinatra::Base
+  enable :sessions
 
-enable :sessions
+  get '/' do
+    bomb = Bomb.instance
```
enovajuan: I guess this is a side-effect of making the class a Singleton.... 

But what would happen if two users (evil genius users) log into same server and take a shot at the bomb, are they all sharing the same in memory instance of it then?

overlord/overlord.rb
```diff
@@ -1,14 +1,81 @@
...
-  "Time to build an app around here. Start time: " + start_time
-end
+  get '/disarm_bomb' do
+    load_session
+    erb :disarm_bomb
+  end
+
+  post '/disarm_bomb' do
+    bomb = Bomb.instance
+    session[:bomb] = bomb
```
enovajuan: since the class is a singleton, would you really ever use this?

overlord/views/configure_bomb.erb
```diff
@@ -0,0 +1,45 @@
...
+  <link type="text/css" rel="stylesheet" href="styles/stylesheet.css"/>
+  <title>Configure Bomb</title>
+</head>
+<body>
+<div class="centerdiv">
+<h2>Bomb Status: <%= (session[:bomb]).status %> </h2>
+<h3>Arming code: <%= (session[:bomb]).arming_code %> </h3>
+<h3>Disarming code: <%= (session[:bomb]).disarming_code %> </h3>
+<p>Press the button below to boot up your bomb</p>
+<% if session[:bomb].not_booted? || session[:bomb].disarmed? %>
```
enovajuan: So it looks like you have controller logic leaking into your view here. All the if/else in these files, should be handled on the controller level. You would greatly simplify the view logic in these files, if you split each form into a partial(s), then determine which partial to render on the controller level then either pass the template name and have this file render it, or have a yield in the template which can yield the contents of the view.

enovajuan: The issue with the if/else pattern in view templates, is what happens if you'd like to add a new bomb state/view? Or if i need to add 5 or 10 new state views? 

also DRY's up the code nicely.

overlord/views/disarm_bomb.erb
```diff
@@ -0,0 +1,22 @@
+<html>
+<head>
+  <title>Disarm Bomb</title>
+  <link type="text/css" rel="stylesheet" href="styles/stylesheet.css"/>
+</head>
+   <body>
+     <div class="centerdiv">
+        <h2>Bomb Status: <%= (session[:bomb]).status %> </h2>
+        <% if (session[:bomb]).armed? %>
```
enovajuan: See comment on `configure.erb` file about using if/else in view.

overlord/views/outcome.erb
```diff
@@ -0,0 +1,37 @@
+<html>
+<head>
+  <link type="text/css" rel="styles/stylesheet" href="styles/stylesheet.css"/>
+  <% if (session[:bomb]).disarmed? %>
```
enovajuan: See comment on `configure.erb` file about using if/else in view.

enovajuan:
Overall good job on cukes/tests! Some comments on code, that need some cleanup. Main thing would be to separate the controller logic from view templates (if/else blocks in views should be DRY'ed up and moved into partials and controller logic)

overlord/lib/bomb.rb
```diff
@@ -0,0 +1,86 @@
...
+    @disarming_code = DEFAULT_DISARMING_CODE
+  end
+
+  def attempts_remaining
+    MAX_ATTEMPTS - disarming_count
+  end
+
+  def track_count
+    @disarming_count += 1
+    @status = "Detonated" if disarming_count == MAX_ATTEMPTS
```
enovajuan: try and stick with Single Responsibility (the S in SOLID)  principle here l:59 here could likely be it's own method maybe `detonate` or `detonate_bomb`? 

Since this method is doing 3 things (2 more than simply tracking count as the name implies)
1) Adds to disarming count
2) Sets status on bom
3) check if count is equal to max attempts 

[SOLID reference](https://en.wikipedia.org/wiki/SOLID_(object-oriented_design))

overlord/lib/bomb.rb
```diff
@@ -0,0 +1,86 @@
...
+  def configure_disarming_code(code = nil)
+    @disarming_code = code unless blank?(code)
+  end
+
+  def arm(code)
+    @status = "Armed"  if code == arming_code
+  end
+
+  def disarm(code)
+    if code == disarming_code
```
enovajuan: SOLID

overlord/features/reset_bomb.feature
```diff
@@ -0,0 +1,26 @@
+Feature: Villain resets arming and disarming codes
+
+Scenario: Villain resets the arming code
+  Given I have a booted bomb with default values
+  And I am on the "configure_bomb" page
+  When I enter '7777' for "arming_code_reset"
+  And I press "Reset Bomb Codes" within "configure_bomb"
```
paulghaddad: An overall comment for your resubmission is to create step definitions at a higher level of abstraction. Think of what a business owner would want to read. Something like this:

```
Given I have a booted bomb with a default configuration
When I reset the bomb code
Then I should see the updated arming code
```

Take a look at all your features and try to condense in this fashion, using a single level of abstraction. Your step definitions can encapsulate more capybara steps. This will make your features more readable.

See this article as a reference: https://robots.thoughtbot.com/acceptance-tests-at-a-single-level-of-abstraction

paulghaddad:
@lmenova Nice job on this PR. Take a look at the comments and resubmit. In particular, strive in all of your feature tests to create step definitions at a higher level of abstraction. Stop by in you would like to go over this.

PR #909, opened 2015-08-11
deja/dino_catalog

DejaAugustine:
I look forward to the feedback on this one.

### DIFF
--- a/dino_catalog/dino_catalog.rb
+++ b/dino_catalog/dino_catalog.rb
@@ -1,0 +1,54 @@
+require 'json'
+require_relative 'dinodex'
+
+puts
+puts "***********"
+puts "* Dinodex *"
+puts "***********"
+puts
+
+dex = Dinodex.new
+
+dex.import("dinodex.csv")
+dex.import("african_dinosaur_export.csv")
+
+puts "---------------"
+puts "  Query: All carnivores larger than 2,000 lbs"
+puts
+dex.filter(weight: 2000, op: ">")
+  .filter(carnivore: true)
+  .each do |dino|
+  puts dino.to_s
+  puts
+end
+
+puts "---------------"
+puts "  Query: All Cretaceous bipeds"
+puts
+dex.filter(walking: "Biped")
+  .filter(period: "Cretaceous", op: "include?")
+  .each do |dino|
+  puts dino.to_s
+  puts
+end
+
+puts "---------------"
+puts "  Query: All non-American dinosaurs ('they terk er jerbs!')"
+puts
+dex.filter(continent: "America", op: "include?", negate: true)
+  .each do |dino|
+  puts dino.to_s
+  puts
+end
+
+puts "--------------"
+puts "  Query: All \"Albert's\" as JSON"
+puts
+puts dex.filter(name: "Albert", op: "start_with?").to_json
+puts
+
+puts "--------------"
+puts "  Query: All \"Albert's\" as JSON (pretty)"
+puts
+puts JSON.pretty_generate(dex.filter(name: "Albert", op: "start_with?").to_hash)
+puts
--- a/dino_catalog/dinodex.rb
+++ b/dino_catalog/dinodex.rb
@@ -1,21 +1,12 @@
 require 'csv'
+require 'json'
 require_relative 'dinosaur'
 
 class Dinodex
   attr_reader :db
 
-  def initialize
-    @db = []
-  end
-
-  def import(filename)
-    db.concat(parse_csv(filename))
-  end
-
-  def parse_csv(filename)
-    csv = CSV.new(IO.read(filename), headers: true, converters: :all)
-
-    csv.to_a.map { |row| Dinosaur.new(row.to_hash) }
+  def each(&block)
+    to_a.each(&block)
   end
 
   def filters
@@ -24,37 +15,68 @@ class Dinodex
 
   def filter(args)
     args[:op] ||= "=="
+    args[:negate] ||= false
     filters.push(args)
     self
   end
 
-  def each(&block)
-    db.reject do |dino|
-      !match_dino(dino)
-    end.each(&block)
+  def import(filename)
+    db.concat(parse_csv(filename))
+  end
+
+  def initialize
+    @db = []
   end
 
   def match_dino(dino)
     match = 1
 
     filters.each do |filter|
-      op = filter[:op]
-      key,target = filter.first
+      key = filter.first.first
       value = dino.method(key).call
 
-      match = match & 0 unless value && value.method(op).(target)
+      match = test_match(match, value, filter)
     end
 
     match == 1
   end
-end
 
-dex = Dinodex.new
+  def parse_csv(filename)
+    csv = CSV.new(IO.read(filename), headers: true, converters: :all)
 
-dex.import("dinodex.csv")
-dex.import("african_dinosaur_export.csv")
+    csv.to_a.map { |row| Dinosaur.new(row.to_hash) }
+  end
 
-dex.filter({ weight: 2000, op: ">" }).filter({ carnivore: true }).each do |dino|
-  puts dino.to_s
-  puts
+  def reset_filters
+    @filter = []
+  end
+
+  def test_match(match, value, filter)
+    op = filter[:op]
+    negate = filter[:negate]
+    target = filter.first.last
+
+    match &= 0 unless value && (
+                 (!negate && value.method(op).call(target)) ||
+                 (negate && !value.method(op).call(target)))
+    match
+  end
+
+  def to_a
+    arr = db.reject do |dino|
+      !match_dino(dino)
+    end
+
+    reset_filters
+
+    arr
+  end
+
+  def to_hash
+    to_a.map(&:to_hash)
+  end
+
+  def to_json
+    JSON.generate(to_hash)
+  end
 end
--- a/dino_catalog/dinodex.rb
+++ b/dino_catalog/dinodex.rb
@@ -1,0 +1,60 @@
+require 'csv'
+require_relative 'dinosaur'
+
+class Dinodex
+  attr_reader :db
+
+  def initialize
+    @db = []
+  end
+
+  def import(filename)
+    db.concat(parse_csv(filename))
+  end
+
+  def parse_csv(filename)
+    csv = CSV.new(IO.read(filename), headers: true, converters: :all)
+
+    csv.to_a.map { |row| Dinosaur.new(row.to_hash) }
+  end
+
+  def filters
+    @filter ||= []
+  end
+
+  def filter(args)
+    args[:op] ||= "=="
+    filters.push(args)
+    self
+  end
+
+  def each(&block)
+    db.reject do |dino|
+      !match_dino(dino)
+    end.each(&block)
+  end
+
+  def match_dino(dino)
+    match = 1
+
+    filters.each do |filter|
+      op = filter[:op]
+      key,target = filter.first
+      value = dino.method(key).call
+
+      match = match & 0 unless value && value.method(op).(target)
+    end
+
+    match == 1
+  end
+end
+
+dex = Dinodex.new
+
+dex.import("dinodex.csv")
+dex.import("african_dinosaur_export.csv")
+
+dex.filter({ weight: 2000, op: ">" }).filter({ carnivore: true }).each do |dino|
+  puts dino.to_s
+  puts
+end
--- a/dino_catalog/dinosaur.rb
+++ b/dino_catalog/dinosaur.rb
@@ -5,12 +5,6 @@ class Dinosaur
   alias_method :genus=, :name=
   alias_method :weight_in_lbs=, :weight=
 
-  def initialize(args = {})
-    args.each do |key, value|
-      send("#{key.downcase}=", value)
-    end
-  end
-
   def carnivore
     @diet.downcase != "herbivore"
   end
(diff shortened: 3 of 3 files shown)


dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,54 @@
...
+puts "* Dinodex *"
+puts "***********"
+puts
+
+dex = Dinodex.new
+
+dex.import("dinodex.csv")
+dex.import("african_dinosaur_export.csv")
+
+puts "---------------"
```
paulghaddad: Awesome job on this file showing me how your program works!

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,46 @@
+class Dinosaur
+  attr_accessor :name, :period, :continent, :diet
+  attr_accessor :weight, :walking, :description
+
+  alias_method :genus=, :name=
```
paulghaddad: Excellent use case for `alias_method`.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,46 @@
...
+  alias_method :weight_in_lbs=, :weight=
+
+  def carnivore
```
paulghaddad: Consider renaming to `carnivore?` since this returns a boolean.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,46 @@
...
+    @diet.downcase != "herbivore"
+  end
+
+  def carnivore=(value)
+    @diet = "Carnivore" if value.downcase == "yes"
+    @diet = "Herbivore" if value.downcase == "no"
+  end
+
+  def initialize(args = {})
```
paulghaddad: Excellent way to generate the instance variables dynamically.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,82 @@
...
+require_relative 'dinosaur'
+
+class Dinodex
+  attr_reader :db
+
+  def each(&block)
+    to_a.each(&block)
+  end
+
+  def filters
```
paulghaddad: +1

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,82 @@
+require 'csv'
+require 'json'
+require_relative 'dinosaur'
+
+class Dinodex
+  attr_reader :db
```
paulghaddad: Consider renaming this variable to be more descriptive.

dino_catalog/dinosaur.rb
DejaAugustine: Good call. I was treating it as an attr_accessor override and completely missed that it has a boolean return

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,54 @@
...
+puts
+puts "***********"
+puts "* Dinodex *"
+puts "***********"
+puts
+
+dex = Dinodex.new
+
+dex.import("dinodex.csv")
+dex.import("african_dinosaur_export.csv")
```
paulghaddad: I really like this implementation where you are pulling in both csv files. Now I see why you initialized your `Dinodex` class with an empty array. I originally thought using an idiom like `dex = Dinodex.new("dinodex.csv")` would allow you to eliminate the `import` method. But your implementation is better as you pull in both. Nice work!

dino_catalog/dinodex.rb
paulghaddad: Nice work on building an array of your filters before actually filtering.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,54 @@
...
+  puts
+end
+
+puts "---------------"
+puts "  Query: All Cretaceous bipeds"
+puts
+dex.filter(walking: "Biped")
+  .filter(period: "Cretaceous", op: "include?")
+  .each do |dino|
+  puts dino.to_s
```
paulghaddad: I'd personally indent the lines inside your blocks (31 and 32).

DejaAugustine: I vaguely recall unindenting them to satisfy rubocop's tyrannical decree

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,54 @@
...
+dex.import("dinodex.csv")
+dex.import("african_dinosaur_export.csv")
+
+puts "---------------"
+puts "  Query: All carnivores larger than 2,000 lbs"
+puts
+dex.filter(weight: 2000, op: ">")
+  .filter(carnivore: true)
+  .each do |dino|
+  puts dino.to_s
```
jmtai: `puts` will automatically call the `to_s` method of whatever object your're printing, so it's not necessary to explicitly call the `to_s` method here.

jmtai: Chaining methods is fun but be careful that readability doesn't suffer. Even for something small like this, it might be nicer to create an intermediate result to improve readability.
```ruby
dinos = dex.filter(weight: 2000, op: ">").filter(carnivore: true)
dinos.each do |dino|
  puts dino.to_s
end
```
We can tighten this up even more by dropping the `to_s` and using curly brace notation since the code block is so small:
```ruby
dinos = dex.filter(weight: 2000, op: ">").filter(carnivore: true)
dinos.each { |dino| puts dino }
```

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,54 @@
...
+puts "--------------"
+puts "  Query: All \"Albert's\" as JSON"
+puts
+puts dex.filter(name: "Albert", op: "start_with?").to_json
+puts
+
+puts "--------------"
+puts "  Query: All \"Albert's\" as JSON (pretty)"
+puts
+puts JSON.pretty_generate(dex.filter(name: "Albert", op: "start_with?").to_hash)
```
jmtai: Conciseness is definitely a good thing, but you'll want to balance brevity with readability. I'd argue that using an intermediate result here is preferable since it will improve the readability of the code even though it adds an additional line:
```ruby
dinos = dex.filter(name: "Albert", op: "start_with?").to_hash
JSON.pretty_generate(dinos)
```
Ruby allows us to be extremely expressive, so anything that allows the code to read more fluidly and more like natural language is generally a good thing.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,82 @@
...
+  def import(filename)
+    db.concat(parse_csv(filename))
+  end
+
+  def initialize
+    @db = []
+  end
+
+  def match_dino(dino)
+    match = 1
```
jmtai: Avoid using 1 and 0 to signal true/false conditions. Ruby has native booleans so just use `true` and `false`.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    filters.each do |filter|
+      key = filter.first.first
```
jmtai: Hmm... `first.first` is a bit too mysterious. Without any context the reader is left to wonder what exactly is going on here. Can you think of a way to make this clearer?

dino_catalog/dinodex.rb
DejaAugustine: Unfortunately, in Ruby, you cannot apply bitwise operations to true/false. 

```
match = 1
# => 1
match &= 0
# => 0
```

versus

```
match = true
# => true
match &= 0
# => true
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,82 @@
...
+      value = dino.method(key).call
+
+      match = test_match(match, value, filter)
```
jmtai: Here the code is a bit confusing because at first glance it looks like match is simply being overwritten each time during the loop. A reader has to go look at the source of `test_match` to see that the current value of `match` is being used to calculate the new value of `match`.

Additionally, Ruby convention is to name methods that return true/false with a trailing question mark, so `test_match` might simply be named `match?`. These are called **predicate** methods.

However, you would then want to consider the naming of your variables because this:
```ruby
match = match?(match, value, filter)
```
isn't super readable.

dino_catalog/dinodex.rb
DejaAugustine: How about `filter.keys.first`?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,82 @@
...
+    end
+
+    match == 1
```
jmtai: If you make match a boolean, this would simply be `match`, as you could drop the comparison operation.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def parse_csv(filename)
+    csv = CSV.new(IO.read(filename), headers: true, converters: :all)
```
jmtai: No need to pull in `IO` here, you can just use CSV.read:
```ruby
csv = CSV.read(filename, headers: true, converters: :all)
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def reset_filters
+    @filter = []
+  end
+
+  def test_match(match, value, filter)
+    op = filter[:op]
+    negate = filter[:negate]
+    target = filter.first.last
```
jmtai: Again, `first.last` is too mysterious at first glance. Consider using intermediate results to make it more obvious what's going on here.

dino_catalog/dinodex.rb
DejaAugustine: To me, `match = test_match(match, ...)` seems a perfectly valid paradigm, though I have been struggling with a better name than `test_match`. Breaking it out into a separate method to satisfy rubocop/Hound made this code much less clear than it was originally written.

That said, `test_match` returns an Integer, not a Boolean, so styling it as a predicate method would be inappropriate.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    match &= 0 unless value && (
+                 (!negate && value.method(op).call(target)) ||
+                 (negate && !value.method(op).call(target)))
```
jmtai: The combination of using 0/1 for true/false, the formatting, and the negation logic make this line pretty difficult to parse. Try to see if you can clean this up a bit.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,82 @@
...
+    match
+  end
+
+  def to_a
+    arr = db.reject do |dino|
+      !match_dino(dino)
+    end
```
jmtai: For small, single line code blocks like this, it's generally preferred to use the curly brace format:
```ruby
arr = db.reject { |dino| !match_dino(dino) }
```

dino_catalog/dinodex.rb
DejaAugustine: Again, match is necessarily an Integer, so the comparison is required.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,82 @@
...
+
```
jmtai: The extra blank lines between statements generally isn't necessary. You'll find that most Ruby code in general uses blank lines sparingly since we aim to keep our methods as short as possible. When you've only got 5 lines of code in a method, it isn't necessary to separate them all with blank lines.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,46 @@
+class Dinosaur
+  attr_accessor :name, :period, :continent, :diet
+  attr_accessor :weight, :walking, :description
+
+  alias_method :genus=, :name=
+  alias_method :weight_in_lbs=, :weight=
+
+  def carnivore
+    @diet.downcase != "herbivore"
```
jmtai: Since you've got `diet` declared with an accessor, there's no need to use the `@` notation when reading the value of the variable.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def to_s
+    out = instance_variables.map do |ivar|
+      vname = ivar.to_s
+      vname[0] = ""
+
+      value = instance_variable_get(ivar)
+      "#{format('%12s', vname.capitalize)}:\t#{value}" if value
+    end
```
jmtai: Notice how your `to_hash` and `to_s` methods look very similar? Perhaps there's a way to reuse the common parts.

dino_catalog/dinodex.rb
jmtai: If this were C or C++, I'd be totally down with using bitwise operations. In Ruby that's going to be a bit of an anti-pattern. Since your bitwise operations are operating on a grand total of one single bit, there's not much to be gained by using them. Logical operators work just as well for this case and allow you to better express the intent of the code:
```
match = true
# => true
match &&= false
# => false
```

dino_catalog/dinodex.rb
jmtai: :+1:

dino_catalog/dinodex.rb
jmtai: Check the previous comment regarding the use of logical vs. bitwise operations. You can definitely turn `test_match` into a predicate method by using logical operators instead of bitwise operators. This will have the advantage of making the code more readable as well as making it more idiomatic Ruby.

jmtai:
Nice first pass! Although it's not stated explicitly in the problem description, one goal of the exercise is to get some experience creating a user-drive command line interface for the application. Read over the comments, refactor and bit, and try adding an interactive user interface to the app. Feel free to hunt me down if you have any questions.

dino_catalog/dinodex.rb
djkotowski: Shouldn't it be `select` instead of `reject` with a negation?

```ruby
arr = db.select do |dino|
  match_dino(dino)
end
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,82 @@
...
+    reset_filters
+
+    arr
```
djkotowski: To avoid assigning `arr` just to return it after resetting the filters, try:
```ruby
db.select { |dino| match_dino(dino) }.tap do
  reset_filters
end
```

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def to_hash
```
djkotowski: Why not use the traditional `#to_h`?

paulghaddad:
@DejaAugustine Take a look at some of the comments for a quick second iteration. This looks great!

PR #913, opened 2015-08-12
deja/data_science

DejaAugustine:
First draft of Data Science exercise


data_science/data_loader.rb
```diff
@@ -0,0 +1,18 @@
...
+    key, target = args.first
+    op = args[:op] || "=="
+
+    data.select do |i|
+      i[key].method(op).call(target)
+    end
+  end
+
+  def initialize(source)
+    @data = JSON.parse(IO.read(source), symbolize_names: true)
```
paulghaddad: I'd personally use `File.read`, but `IO.read` works just as well.

data_science/data_science.rb
```diff
@@ -0,0 +1,32 @@
+require_relative "data_analyzer"
```
paulghaddad: I really like these files. They do a great job of showing the usage of your program.

data_science/data_science.rb
```diff
@@ -0,0 +1,32 @@
...
+puts "****************"
+puts "* Data Science *"
+puts "****************"
+puts
+
+loader = DataLoader.new(
+  File.expand_path("../data/data_export_2014_06_20_15_59_02.json", __FILE__))
+
+analyzer = DataAnalyzer.new(loader)
+analyzer.load(agroup: { cohort: "A" }, bgroup: { cohort: "B" })
```
paulghaddad: I'd use `a_group` and `b_group` as variable names. Nice use of Keyword Arguments.

data_science/data_loader.rb
```diff
@@ -0,0 +1,18 @@
+require 'json'
+
+class DataLoader
+  attr_reader :data
+
+  def fetch(args)
```
paulghaddad: I really like this method and how it returns a subset of the data at this point. This really simplifies your `DataAnalyzer` class.

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,63 @@
...
+    { low: low, high: high }
+  end
+
+  def p_value
+    tester = ABAnalyzer::ABTest.new(compose_values)
+    tester.chisquare_p
+  end
+
+  def significant?
+    p_value < 0.05
```
paulghaddad: Could `0.05` be placed in a constant as its a magic number? Just a thought...maybe `PERCENTILE_95`?

data_science/data_science.rb
```diff
@@ -0,0 +1,32 @@
...
+
+puts "Total Sample Size: #{analyzer.sample_size}"
+puts "Significant? #{analyzer.significant? ? 'Yes' : 'No'}"
+puts "P-Value: #{analyzer.p_value.round(3)}"
```
paulghaddad: Would it be more clean to have `p_value` return a rounded result?

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def load(args)
+    @groups = {}
+    args.each do |group, criteria|
+      groups[group] = @loader.fetch(criteria)
+    end
+  end
+
+  def sample_size(group = nil)
```
paulghaddad: Some of the methods in this class can be made private, such as this method.

data_science/data_science.rb
DejaAugustine: While the spec only describes using two cohorts, the analyzer itself supports more than two.

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,63 @@
...
+  def compose_values
+    values = {}
+    groups.keys.each do |group|
+      successes = conversion_count(group)
+      failures = sample_size(group) - successes
+      values[group] = { successes: successes, failures: failures }
+    end
+    values
+  end
+end
```
paulghaddad: I really like this class. You have some nice methods that are very clear on their purpose.

data_science/data_science.rb
DejaAugustine: In this instance, rounding the value is part of the presentation layer to make it look better in the CLI. Consumers of the analyzer object should have access to the unadulterated results.

data_science/spec/data_analyzer_spec.rb
```diff
@@ -0,0 +1,72 @@
+require_relative "../data_analyzer"
+
+describe DataAnalyzer do
+  let(:filepath) { File.expand_path("../data/sample_data.json", __FILE__) }
+  let(:loader) { DataLoader.new(filepath) }
+
+  subject(:analyzer) { DataAnalyzer.new(loader) }
```
paulghaddad: I'm thinking there is a good way to stub the return of the data set from the JSON file instead of loading an actual file. What are your thoughts?

data_science/data_analyzer.rb
DejaAugustine: I disagree in this instance. The consuming procedure should have access to basic information about the data that's been loaded (see data_science.rb:16, which would not be possible if this method were private)

data_science/spec/data_analyzer_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  context "when data is loaded" do
+    before(:each) do
+      analyzer.load(agroup: { cohort: "A" }, bgroup: { cohort: "B" })
+    end
+
+    it "contains two distinct groups" do
+      expect(analyzer.groups.size).to eq(2)
+    end
+
+    it "reports the total sample size if no group is specified" do
```
paulghaddad: I'd personally have the tests for each method, such as `#sample_size` inside their own `describe` blocks.

data_science/spec/data_analyzer_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      expect(analyzer.conversion_rate(:bgroup).round(3)).to eq(0.083)
+    end
+
+    context "when reporting the 95% confidence interval" do
+      let(:interval) { analyzer.confidence_interval(:bgroup, 0.95) }
+      let(:upper_bound) { interval[:high] }
+      let(:lower_bound) { interval[:low] }
+
+      it "provides an upper boundary" do
+        expect(upper_bound.round(3)).to eq(0.135)
```
paulghaddad: Do you think the `be_within` matcher would be better here since the values are so precise?

data_science/spec/data_loader_spec.rb
```diff
@@ -0,0 +1,51 @@
+require_relative "../data_loader"
+
+describe DataLoader do
+  let(:filepath) { File.expand_path("../data/sample_data.json", __FILE__) }
+  let(:loader) { DataLoader.new(filepath) }
+
+  context "#initialize" do
```
paulghaddad: I think this should be a `describe` block; the same with `fetch`, below.

data_science/spec/data_analyzer_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      it "calculates a p-value" do
+        expect(analyzer.p_value).to be_kind_of(Float)
+      end
+
+      it "tests for statistical significance" do
+        expect(analyzer.significant?).to be false
+      end
+    end
+  end
+end
```
paulghaddad: Overall, nice clear specs!

data_science/spec/data_loader_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    it "creates a valid data array" do
+      expect(loader.data).to be_kind_of(Array)
+    end
+
+    it "loads more than 0 records" do
+      expect(loader.data.size).to be > 0
+    end
+
+    it "raises an exception if file doesn't exist" do
```
paulghaddad: Good test case.

data_science/data_science.rb
paulghaddad: Good point.

data_science/data_analyzer.rb
DejaAugustine: I went back and forth on this one. For most cases, a constant would be appropriate, but in this particular instance "p < .05" is such a ubiquitous and instantly recognizable fixture in statistics that `p_value < SIGNIFICANCE_THRESHOLD` actually looked less clear.

That said, you are right. For someone reading this without any knowledge in statistics, the latter would be better code.

data_science/data_analyzer.rb
paulghaddad: That's a good argument!

data_science/data_loader.rb
DejaAugustine: Noted

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+  end
+
+  def sample_size(group = nil)
+    if group
+      groups[group].size
+    else
+      groups.values.flatten.size
+    end
+  end
```
zsyed1: Lines 20-24: Here's a more rubyesque way of doing it:
```
return groups[group].size if group
groups.values.flatten.size
```

data_science/data_analyzer.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def conversion_count(group)
+    groups[group].count do |i|
+      i[:result] != 0
+    end
```
zsyed1: Lines 28-30 can go on one line

data_science/spec/data_analyzer_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  subject(:analyzer) { DataAnalyzer.new(loader) }
+
+  context "when an instance is created" do
+    it "is a valid instance" do
+      is_expected.to be_instance_of(DataAnalyzer)
+    end
+  end
+
+  context "when data is loaded" do
+    before(:each) do
```
zsyed1: Use `let` instead of `before` - firmwide convention! :smile:

data_science/spec/data_loader_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    it "produces a valid instance" do
+      expect(loader).to be_instance_of(DataLoader)
+    end
+
+    it "creates a valid data array" do
+      expect(loader.data).to be_kind_of(Array)
+    end
+
+    it "loads more than 0 records" do
+      expect(loader.data.size).to be > 0
```
zsyed1: Can you do:
```
expect(loader.data).not_to be_empty
```
?

data_science/spec/data_loader_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    it "returns an array" do
+      expect(filtered_loader).to be_kind_of(Array)
+    end
+
+    it "returns more than 0 records" do
+      expect(filtered_loader.size).to be > 0
+    end
+
+    it "does not contain rejected items" do
+      b_size = filtered_loader.count { |i| i[:cohort] == "B" }
```
zsyed1: Line 42: move into `let`
Honestly, I prefer the way you did it.

data_science/spec/data_analyzer_spec.rb
DejaAugustine: It's possible, however given the fact that the data is used across multiple specs and the test dataset should be a certain minimum size to reliably test the functionality, this is the DRYer and cleaner option.

A better option would be to use a factory to procedurally generate sample data.

data_science/spec/data_analyzer_spec.rb
DejaAugustine: Good to know. If that was mentioned in the lessons / docs, I missed it.

data_science/spec/data_analyzer_spec.rb
DejaAugustine: Makes sense. I'll refactor with that in mind.

data_science/spec/data_analyzer_spec.rb
DejaAugustine: It would indeed.

data_science/spec/data_loader_spec.rb
DejaAugustine: That looks like a more robust option. Thanks!

data_science/spec/data_loader_spec.rb
DejaAugustine: @zsyed1: What is the benefit of moving this into a `let` in this instance?

It is only used by the one test in which it appears, so putting it into a `let` won't dry out the code, and doing so will actually strip it of its context which, in this instance I would argue would actually reduce the simplicity and clarity of the block.

data_science/data_analyzer.rb
DejaAugustine: Much more concise, thank you.

data_science/spec/data_analyzer_spec.rb
paulghaddad: A Factory is definitely the best option for this data set in my opinion.

paulghaddad:
@DejaAugustine Really nice work on this exercise! You're good to go!

data_science/spec/data_loader_spec.rb
zsyed1: Yeah, that's why I prefer the way you did it :smile:
Here it is preferred to move all business logic into the preface of a test.

PR #922, opened 2015-08-18
Arrowhead Excercise

Manbearpixel:
Was a bit unsure about what to do for this exercise. From what I gathered, it seems like I was to make the if/then rabbit hole simple by checking and returning errors first and then going about the rest of the program. I also thought the addition of commas at the end of the array was an error so I omitted those too.

Cheers.

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -18,7 +18,7 @@ class Arrowhead
   # FIXME: I don't have time to deal with this. -- fixed
   def self.classify(region, shape)
 
-    if !CLASSIFICATIONS,include? region
+    if !CLASSIFICATIONS.include? region
       raise "Unknown region, please provide a valid region."
     end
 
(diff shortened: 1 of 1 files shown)


arrowhead/arrowhead.rb
```diff
@@ -5,30 +5,33 @@ class Arrowhead
...
-      shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
+
+    if !CLASSIFICATIONS.include? region
```
djkotowski: Prefer `unless condition` over `if !condition`.

djkotowski: Also, wrap arguments in parentheses.

arrowhead/arrowhead.rb
```diff
@@ -5,30 +5,33 @@ class Arrowhead
...
       raise "Unknown region, please provide a valid region."
     end
+
+    shapes = CLASSIFICATIONS[region]
+    if !shapes.include? shape
```
djkotowski: See comments above.

djkotowski:
Nice first iteration.  Take a look at the comments above, refactor and resubmit.  Let me know if you have any questions.

PR #923, opened 2015-08-18
Fixing dependency assumptions with Overlord exercise

DejaAugustine:
The WireBundle class is explicitly coupled with the Wire class, so the dependency should be explicitly declared.

### DIFF
--- a/overlord/lib/wire_bundle.rb
+++ b/overlord/lib/wire_bundle.rb
@@ -1,3 +1,5 @@
+require_relative "wire"
+
 class WireBundle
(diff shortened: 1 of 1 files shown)


jmmastey:
I'm game for this. Seems a pretty safe change. @djkotowski ?

paulghaddad:
I agree that this is a good change!

djkotowski:
Agreed.

paulghaddad:
@jmmastey @djkotowski Great...I'll merge.

PR #927, opened 2015-08-18
Overlord V2

mcdaniel67:
Second iteration of the Overlord exercise. I corrected the parts that I received feedback on.


overlord/bomb.rb
```diff
@@ -0,0 +1,44 @@
...
+  def initialize(activation = nil, deactivation = nil)
+    @activation_code = activation || '1234'
+    @deactivation_code = deactivation || '0000'
+    @deactivation_attempts = 0
+    @status = 'Inactive'
+  end
+
+  def activate(code)
+    @status = 'Active' if check_activation_code(code)
+    @status == 'Active'
```
enovajuan: Not sure why you're checking the status as a return?  Also if check_activation_code returns true/false it should be a predicate method, probably using a method name like `valid_activation_code?(code)` or `is_code_valid?(code)` also `@status == 'active'` could be it's own Bomb method of `Bomb.is_active?`

overlord/bomb.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def deactivate(code)
```
enovajuan: This method seems to have too many responsibilities (the S - in [SOLID](https://en.wikipedia.org/wiki/SOLID_(object-oriented_design))). 
Each one of these should be it's own method:
1. Checking deactivation code
2. Setting the bomb status
3. Setting deactivation attempts
4. Uncertain why it returns true or false, since it ALWAYS returns the count of `@deactivation_attempts`

overlord/features/deactivation.feature
```diff
@@ -0,0 +1,41 @@
...
+  Scenario: Should not be able to directly jump to the deactivation page
+    Given I am on "the deactivation page" of the bomb
+    Then I should be on the starting page
+
+  Scenario: Deactivation page status check
+    Given I activated the bomb with the deactivation code ""
+    Then The page should say "Bomb Status: Active"
+
+  Scenario: Bomb should deactivate with default code
+    Given I activated the bomb with the deactivation code ""
```
enovajuan: This step seems misleading (simlarly on Lines L#21, L#27, L#11, L#39), at first reading it seems like this line says `I activated the bomb WITH the Deactivation code`, I'm guessing this should read something like `I activated the bomb with the <default activation code> and deactivation code <deactivaiton_code>`

overlord/overlord.rb
```diff
@@ -1,14 +1,100 @@
...
+  def activate_bomb(key)
+    activated_this_turn = session[:bomb].activate(key)
+    already_activated = session[:bomb].status == 'Active'
+    redirect to('/activate/') unless activated_this_turn || already_activated
+  end
+
+  def root_error_codes(index)
+    case index
+      when 1
+        'Invalid activation code. Activation code must be only numbers.'
```
enovajuan: These statuses seem like a responsibility of the bomb, not the controller. Instead of having the bomb return a code of 1/2/3 it should just return these status strings, that way this controller can have no knowledge of the bomb. It just passes back what the bomb passes... This would allow for the controller to be able to handle anything else with the same status interface (i.e. instead of `Bomb` you could implement `AtomBomb`, `PipeBomb`, `KittenBomb`, `IceCreamBomb` etc, without having to change anything about the controller.

overlord/overlord.rb
```diff
@@ -1,14 +1,100 @@
...
+        'Invalid deactivation code. Deactivation code must be only numbers'
+      when 3
+        'Nice try hero, back to square one for you.'
+      else
+        ''
+    end
+  end
+
+  def replace_empty_params(activation, deactivation)
+    activation.replace('1234') if activation == ''
```
enovajuan: This is also logic which belongs in the Bomb.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+describe Bomb do
+  let(:bomb) { Bomb.new }
+
+  it 'should be inactive to start' do
+    expect(bomb.status).to eq('Inactive')
+  end
+
+  it 'should default the activation code to 1234' do
+    bomb.activate('1234')
```
enovajuan: If there is a default code, why are you passing in a 1234?

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    expect(bomb.status).to eq('Active')
+  end
+
+  it 'should default the deactivation code to 0000' do
+    bomb.activate('1234')
+    bomb.deactivate('0000')
```
enovajuan: If it's default you should be able to just call `.deactivate()` without parameters

enovajuan:
Looks pretty good, it needs maybe one more pass. Come by my teamroom(Los Pollos Hermanos/Netcredit) or just find me on HipChat if you need me to clarify any of my comments, and/or need any help.

overlord/spec/bomb_spec.rb
mcdaniel67: I should have been more specific with naming these default tests. What I intended with this test is to ensure that creating a new bomb without specifying parameters will set the activation code to `'1234'` and the deactivation code to `'0000'`

We are passing `'1234'` here in order to ensure that the default activation code is set correctly when creating a new object. 

Would you have any suggestions to make the naming of these tests more clear?

overlord/features/homepage.feature
```diff
@@ -0,0 +1,44 @@
...
+    Then The page should say "Bomb Status: Inactive"
+
+  Scenario: Bomb should accept empty codes (for default values)
+    Given I am on "the starting page" of the bomb
+    When I click the "Submit" button
+    Then I should be on the activation page
+
+  Scenario: Input numeric activation code
+    Given I am on "the starting page" of the bomb
+    And I fill in "activation" with "1111"
```
paulghaddad: General comment for all your features. Stay at a consistent, higher level of abstraction. Something like:

```
Given I am on the starting page of the bomb.
When I activate the bomb
Then I should be on the activation page
```

See this article for reference: http://robots.thoughtbot.com/acceptance-tests-at-a-single-level-of-abstraction

paulghaddad:
@mcdaniel67 We discussed this together. For your third iteration, really focus on writing your features at a high level of abstraction. Nice work on this!

overlord/spec/bomb_spec.rb
enovajuan: I see what you did there. The reason I would say it's misleading is that I could imagine that on the front end I wouldn't be able to just hit-start and have the bomb boot up, but if your intention is to prevent this behavior (aka the evil genius has to KNOW the default code... which he does in this case), then this is actually okay.

same would go for below with deactivate, if this is your intention.

for clarifying it i would maybe add to the end of the test name(s), possibly two more tests:
| activate && deactivate |
1. w/ user interaction
2. w/o user interaction

e.g. Be explicit, write a tests that show the bomb WON'T start (or stop) at all if NO code is entered.

PR #928, opened 2015-08-18
Shopping Cart V2

mcdaniel67:


shopping_cart/add_item.feature
```diff
@@ -0,0 +1,113 @@
...
+  @Happy
+  Scenario: I add item with the same product id to the cart.
+    And I add an item to my cart
+    And I view that item
+    When I try to add 1 item to the cart
+    Then I should have 1 item in my cart
+
+  @Sad
+  Scenario: When I add an item to the cart that is not in stock, it should show an error
+    And I am on the page of an item that it not in stock
```
jvanderz22: There's nothing wrong with the way you have these scenarios split up, but another approach here (and the one I'd probably take) is to get a little more specific with grouping up the features. Like with this and the next, they share the common "Given I am on the page of an item that is not stock," which means they can share that as a background step and might be an indication that they should be part of a separate feature. 

Here that feature could be something like:
```
As a customer
I should not be able to add an out of stock item to my cart
So that I don't attempt to buy something that isn't available
```

 Then you could move that to a separate file as well, which might make everything that the app should be able to do a little easier to follow just by glancing at the layout of the spec directory

shopping_cart/add_item.feature
```diff
@@ -0,0 +1,113 @@
...
+    Then I should have 0 items in my cart
+
+  @Sad
+  Scenario: When I add an item to the cart that is not yet for sale (i.e. not released yet), it should show an error
+    And I am on the page of an unreleased item
+    When I try to add the item to my cart
+    Then I should see an error message
+
+  @Sad
+  Scenario: When I add an item to the cart that is not yet for sale (i.e. not released yet), it shouldn't add to the cart
```
jvanderz22: Again, this and the previous scenario could probably be grouped together as a separate feature

shopping_cart/coupons.feature
```diff
@@ -0,0 +1,68 @@
...
+    When I try to apply an expired coupon code
+    Then I see an error message
+
+
+  @Bad
+  Scenario: The coupon inputs are sanitized
+    Given I have 1 item in my cart which has a valid coupon
+    When I try to apply the coupon code <bad>
+    Then I should see an error message
+
```
jvanderz22: Extra empty lines!

shopping_cart/remove_item.feature
```diff
@@ -0,0 +1,48 @@
+Feature: Remove an Item
+  As a customer
+  I should be able to remove items from the cart
+  In order to buy exactly what I want
+  @Happy
```
jvanderz22: Could use an empty line here

jvanderz22:
This looks good. Tests seem really thorough and cover as many of the plausible happy, bad, and sad scenarios as I can think of

mcdaniel67:
Thanks for the suggestions! I fixed most of them, should I do a third iteration of this exercise?

jvanderz22:
Nah, this is really good and I think you're fine to move on. I'd close the PR if I could but I don't have permission in the repo.

paulghaddad:
@mcdaniel67 We also reviewed this together. Really nice work on writing high level step definitions. You are good to move on.

PR #930, opened 2015-08-19
Refactored Robot Code to slim down methods

Manbearpixel:
Also incorporated name registry checking to ensure multiple names don't exist.

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -10,23 +10,51 @@ class Robot
     @name_generator = args[:name_generator]
 
     if @name_generator
-      @name = @name_generator.call
+      addRobotToRegistry(@name_generator.call)
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+      addRobotToRegistry
+    end
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def self.showRegistered
+    puts @@registry
+  end
+
+  def addRobotToRegistry(robotName="")
+    if robotName.empty?
+      robotName   = "#{generateChar(2)}#{generateNum(3)}"
     end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+    @name = robotName
+    unless robotRegistered?(@name)
+      raise NameCollisionError, "Problem with robot name! [#{@name}]" unless properName?(@name)
+      @@registry << @name
+    end
+  end
+
+  def generateChar(length=1)
+    length.times.collect {('A'..'Z').to_a.sample}.join("")
+  end
+
+  def generateNum(length=1)
+    length.times.collect {rand(10)}.join("")
+  end
+
+  def properName?(name) 
+    name =~ /[[:alpha:]]{2}[[:digit:]]{3}/
+  end
+
+  def robotRegistered?(name)
+    @@registry.include?(name)
   end
 end
 
 robot = Robot.new
 puts "My pet robot's name is #{robot.name}, but we usually call him sparky."
 
-# Errors!
-# generator = -> { 'AA111' }
-# Robot.new(name_generator: generator)
-# Robot.new(name_generator: generator)
+generator = -> { 'AA111' }
+Robot.new(name_generator: generator) 
+Robot.new(name_generator: generator)
+
+puts "Robots Registered:"
+Robot.showRegistered
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -10,41 +10,41 @@ class Robot
     @name_generator = args[:name_generator]
 
     if @name_generator
-      addRobotToRegistry(@name_generator.call)
+      add_robot_to_registry(@name_generator.call)
     else
-      addRobotToRegistry
+      add_robot_to_registry
     end
   end
 
-  def self.showRegistered
+  def self.show_registered
     puts @@registry
   end
 
-  def addRobotToRegistry(robotName="")
+  def add_robot_to_registry(robotName="")
     if robotName.empty?
-      robotName   = "#{generateChar(2)}#{generateNum(3)}"
+      robotName = "#{generate_char(2)}#{generate_num(3)}"
     end
 
     @name = robotName
-    unless robotRegistered?(@name)
-      raise NameCollisionError, "Problem with robot name! [#{@name}]" unless properName?(@name)
+    unless robot_registered?(@name)
+      raise NameCollisionError, "Improper robot name!" unless proper_name?(@name)
       @@registry << @name
     end
   end
 
-  def generateChar(length=1)
-    length.times.collect {('A'..'Z').to_a.sample}.join("")
+  def generate_char(length=1)
+    length.times.map { ('A'..'Z').to_a.sample }.join("")
   end
 
-  def generateNum(length=1)
-    length.times.collect {rand(10)}.join("")
+  def generate_num(length=1)
+    length.times.map { rand(10) }.join("")
   end
 
-  def properName?(name) 
+  def proper_name?(name) 
     name =~ /[[:alpha:]]{2}[[:digit:]]{3}/
   end
 
-  def robotRegistered?(name)
+  def robot_registered?(name)
     @@registry.include?(name)
   end
 end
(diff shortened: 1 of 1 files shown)


robot_name/robot_name.rb
```diff
@@ -10,23 +10,51 @@ def initialize(args = {})
     @name_generator = args[:name_generator]
 
     if @name_generator
-      @name = @name_generator.call
+      add_robot_to_registry(@name_generator.call)
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+      add_robot_to_registry
+    end
```
msabdeljalil: Like we discussed, the above IF block can be refactored as such: 
`add_robot_to_registry(@name_generator && @name_generator.call)`

robot_name/robot_name.rb
```diff
@@ -10,23 +10,51 @@ def initialize(args = {})
...
+
+  def self.show_registered
+    puts @@registry
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def add_robot_to_registry(robotName="")
+    if robotName.empty?
+      robotName = "#{generate_char(2)}#{generate_num(3)}"
     end
```
msabdeljalil: You're checking for a nil argument being passed in, so why not just set a default?
`def add_robot_to_registry(robotName = "#{generate_char(2)}#{generate_num(3)}")`
or if you don't like that, 
```
def add_robot_to_registry(robotName = nil)
  robotName ||= "#{generate_char(2)}#{generate_num(3)}"
```

robot_name/robot_name.rb
```diff
@@ -10,23 +10,51 @@ def initialize(args = {})
...
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+    @name = robotName
+    unless robot_registered?(@name)
+      raise NameCollisionError, "Improper robot name!" unless proper_name?(@name)
```
msabdeljalil: Seperate your tasks, start off with explicitly checking the input, then demonstrate the intention to append the new name provided a certain condition is met. It helps sometimes to literally follow your english thinking pattern when writing ruby code:
```
  def add_robot_to_registry(robotName = "#{generate_char(2)}#{generate_num(3)}")
    raise NameCollisionError, "Improper robot name!" unless proper_name?(robotName)
    
    @@registry << robotName unless robot_registered?(robotName)
  end
```

robot_name/robot_name.rb
```diff
@@ -10,23 +10,51 @@ def initialize(args = {})
...
+      @@registry << @name
+    end
+  end
+
```
msabdeljalil: Everything after here doesn't look like it's part of your public interface - show that the remainder of this code is implementation detail by adding a `private` line here and hiding all methods below it

robot_name/robot_name.rb
```diff
@@ -10,23 +10,51 @@ def initialize(args = {})
     @name_generator = args[:name_generator]
```
msabdeljalil: You could also just assign the name in the initialize and cache it as an instance rather than continuosly passing it around:
`@name = (@name_generator && @name_generator.call) || "#{generate_char(2)}#{generate_num(3)}"`
YOu already have an attr_accessor for :name anyways, why not use it?

robot_name/robot_name.rb
```diff
@@ -10,23 +10,51 @@ def initialize(args = {})
...
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+      add_robot_to_registry
+    end
+  end
+
+  def self.show_registered
+    puts @@registry
+  end
```
msabdeljalil: You're basically exposing only one instance and one class method in your public interface. Until you need access to the other methods, demonstrate intentionality by adding a `private` line here and hiding all methods below it

robot_name/robot_name.rb
```diff
@@ -10,23 +10,51 @@ def initialize(args = {})
...
+
+  def proper_name?(name) 
+    name =~ /[[:alpha:]]{2}[[:digit:]]{3}/
+  end
+
+  def robot_registered?(name)
+    @@registry.include?(name)
   end
 end
```
msabdeljalil: for both `proper_name?` and  `robot_registered?`, use the instance var getter method rather than passing in an arg

robot_name/robot_name.rb
```diff
@@ -1,32 +1,55 @@
 class NameCollisionError < RuntimeError; end
```
paulghaddad: Consider having another error type for validations.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,55 @@
...
 
 class Robot
+  
   attr_accessor :name
 
   @@registry
 
   def initialize(args = {})
     @@registry ||= []
```
paulghaddad: The biggest thing to ask is whether the registry belongs in the robot class. It probably doesn't according to the Single Responsibility Principle. Try to factor out your registry code into a separate class.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,55 @@
 class NameCollisionError < RuntimeError; end
 
 class Robot
+  
   attr_accessor :name
 
   @@registry
```
paulghaddad: Is there a way to avoid the use of the class variable? I'd recommend staying away from them.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,55 @@
...
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
+  def add_robot_to_registry
+    raise NameCollisionError, "Improper robot name!" unless proper_name?
+    @@registry << @name unless robot_registered?
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def proper_name?
```
paulghaddad: I like this method! But do consider making methods private if they are not called from outside the class.

Manbearpixel:
Hey @paulghadad would you mind offering some more advice or a short example of what you'd recommend here? New with Ruby still

robot_name/robot_name.rb
```diff
@@ -1,32 +1,55 @@
...
+  def generate_robot_name
+    "#{generate_char(2)}#{generate_num(3)}"
+  end
+
+  def generate_char(length=1)
+    length.times.map { ('A'..'Z').to_a.sample }.join("")
+  end
+
+  def generate_num(length=1)
+    length.times.map { rand(10) }.join("")
```
paulghaddad: See if there is a simpler, more expressive way. Perhaps using `#concat`.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,55 @@
...
+  def robot_registered?
+    @@registry.include?(@name)
+  end
+
+  def generate_robot_name
+    "#{generate_char(2)}#{generate_num(3)}"
+  end
+
+  def generate_char(length=1)
+    length.times.map { ('A'..'Z').to_a.sample }.join("")
```
paulghaddad: This idiom may simplify your method: `[*'A'..'Z'].sample`

paulghaddad:
@Manbearpixel I suggest you focus on the following for your next revision. This is in a nice place, but I'd like to see one more revision.

* Extract a `Registry` class to adhere to the Single Responsibility Principle.
* Make methods only called inside your class `private`.
* See if there are ways of simplifying your code in the spots I pointed out.

If you'd like to pair on this next week, I'd be glad to! Just let me know.

paulghaddad:
Are you referring to the private methods comment, or one of the others? I
think it would be great to pair next week and go over this together. Let me
know if you would like to do so.

Manbearpixel:
Yeah that would be great! Hands on run through would help out.  My week is free so if you want to set a date / time that works best for you invite me to an event. I'm Vaughn Bullock

paulghaddad:
How about Thursday afternoon around 2 or 3? Would you mind sending a calendar invite?

Manbearpixel:
Hey Paul,

Sorry was OOF yesterday. I have sent the calendar invite to you for this
Thursday at 2pm.

Thanks!

paulghaddad:
That works! I don't believe I received the invite, though.

Manbearpixel:
Hey Paul, did you get the invite yet? I tried re-creating the event.

paulghaddad:
I have not...what's your Enova email? I can send one.

Manbearpixel:
vbullock@enova.com

PR #934, opened 2015-08-21
Yadda Part 2 & 3 v1

ssapra:
Based on feedback from Part 1, I added the beer styles look up table and removed extraneous date column from the ratings table.

### DIFF
--- a/yadda/sql/views.sql
+++ b/yadda/sql/views.sql
@@ -1,0 +1,48 @@
+-- Top beers from a given brewery, according to their total ratings.
+DROP VIEW IF EXISTS top_beers CASCADE;
+CREATE VIEW top_beers AS
+  SELECT
+    beers.name AS beer,
+    beers.brewery_id AS brewery_id,
+    beers.beer_style_id AS beer_style_id,
+    ratings.overall AS score
+  FROM
+    beers
+  INNER JOIN
+    ratings ON beers.id = ratings.beer_id
+  ORDER BY
+    ratings.overall DESC;
+
+-- "Recent score" for a beer, where only ratings within the last six months are counted and the ratings within that period are averaged.
+DROP VIEW IF EXISTS recent_score_for_beer CASCADE;
+CREATE VIEW recent_score_for_beer AS
+  SELECT
+    beer_id,
+    round(AVG(ratings.overall), 2) AS score
+  FROM
+    ratings
+  WHERE
+    ratings.created_on > current_date - INTERVAL '6 months'
+  GROUP BY
+    beer_id;
+
+-- "You might also enjoy", which picks beers of the same style with high average scores, and then sorts them randomly.
+DROP VIEW IF EXISTS recommended_beers CASCADE;
+CREATE VIEW recommended_beers AS
+  SELECT
+    beers.id AS beer_id,
+    beers.name AS beer_name,
+    beer_styles.name AS beer_style,
+    round(AVG(ratings.overall), 2) AS score
+  FROM
+    beers
+  INNER JOIN
+    beer_styles ON beer_styles.id = beers.beer_style_id
+  INNER JOIN
+    ratings ON ratings.beer_id = beers.id
+  GROUP BY
+    beers.id, beers.name, beer_styles.name
+  HAVING
+    AVG(ratings.overall) > 4.0
+  ORDER BY
+    RANDOM();
--- a/yadda/yadda_generator.rb
+++ b/yadda/yadda_generator.rb
@@ -16,9 +16,15 @@ module YaddaGenerator
   end
 
   def self.create_data
+    create_beer_styles
     create_breweries
     create_users
     create_ratings
+  end
+
+  def self.create_beer_styles
+    beer_style_names = File.readlines('data/beer_styles.txt').map(&:chomp)
+    beer_style_names.each { |name| FactoryGirl.create(:beer_style, name: name) }
   end
 
   def self.create_breweries
--- a/yadda/yadda_generator.rb
+++ b/yadda/yadda_generator.rb
@@ -1,0 +1,52 @@
+require 'active_record'
+require 'factory_girl'
+require 'faker'
+Dir['models/*.rb'].each { |f| require_relative f }
+Dir['factories/*.rb'].each { |f| require_relative f }
+
+module YaddaGenerator
+  CONFIG = { adapter: 'postgresql', database: 'yadda' }
+
+  def self.reload_schema
+    `psql -f sql/schema.sql`
+  end
+
+  def self.connect_to_database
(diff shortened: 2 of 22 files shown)


incazteca:
All of the factories can be defined in a single file. There's no need to split out the factories into different files.

yadda/sql/views.sql
```diff
@@ -0,0 +1,48 @@
...
+    ratings ON beers.id = ratings.beer_id
+  ORDER BY
+    ratings.overall DESC;
+
+-- "Recent score" for a beer, where only ratings within the last six months are counted and the ratings within that period are averaged.
+DROP VIEW IF EXISTS recent_score_for_beer CASCADE;
+CREATE VIEW recent_score_for_beer AS
+  SELECT
+    beer_id,
+    round(AVG(ratings.overall), 2) AS score
```
incazteca: Have this column aliased as average score.

yadda/sql/views.sql
```diff
@@ -0,0 +1,48 @@
...
+  INNER JOIN
+    ratings ON beers.id = ratings.beer_id
+  ORDER BY
+    ratings.overall DESC;
+
+-- "Recent score" for a beer, where only ratings within the last six months are counted and the ratings within that period are averaged.
+DROP VIEW IF EXISTS recent_score_for_beer CASCADE;
+CREATE VIEW recent_score_for_beer AS
+  SELECT
+    beer_id,
```
incazteca: Would be nice if were to get the name of the beer. Not that necessary though.

yadda/sql/views.sql
```diff
@@ -0,0 +1,48 @@
...
+CREATE VIEW recommended_beers AS
+  SELECT
+    beers.id AS beer_id,
+    beers.name AS beer_name,
+    beer_styles.name AS beer_style,
+    round(AVG(ratings.overall), 2) AS score
+  FROM
+    beers
+  INNER JOIN
+    beer_styles ON beer_styles.id = beers.beer_style_id
```
incazteca: You can give an alias to the table name. That way you don't have to keep writing out beer_styles. 
Example here assuming `beers` has been aliased as `b`: `INNER JOIN beer_styles bs ON bs.id = b.beer_style_id`

Your SELECT statements could also then use these aliases instead of having to write out the table name.

yadda/yadda_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def self.create_data
+    create_beer_styles
+    create_breweries
+    create_users
+    create_ratings
+  end
+
+  def self.create_beer_styles
+    beer_style_names = File.readlines('data/beer_styles.txt').map(&:chomp)
```
incazteca: Instead of looping through the file to chomp and then once again loop through to make the entry just make a block on `map` that creates the beer_style:
```
File.readlines('beer_styles.txt').map do |beer_style|
  beer_style.chomp!
  FactoryGirl.create :beer_style, name: beer_style
end
```

yadda/yadda_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+  def self.create_breweries
+    100.times do |i|
+      FactoryGirl.create(:brewery)
+      puts "#{i}/100 breweries" if i % 10 == 0
+    end
+  end
+
+  def self.create_users
+    100_000.times do |i|
+      FactoryGirl.create(:user)
```
incazteca: If you're using FactoryGirl just pass in the number of records you want created.

`FactoryGirl.create_list(:user, 100_000)`

yadda/yadda_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+      puts "#{i}/100,000 drinkers" if i % 10_000 == 0
+    end
+  end
+
+  def self.create_ratings
+    all_user_ids = User.pluck(:id)
+    all_beer_ids = Beer.pluck(:id)
+    200_000.times do |i|
+      FactoryGirl.create(:rating, user_id: all_user_ids.sample,
```
incazteca: Use `create_list`

yadda/yadda_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def self.create_beer_styles
+    beer_style_names = File.readlines('data/beer_styles.txt').map(&:chomp)
+    beer_style_names.each { |name| FactoryGirl.create(:beer_style, name: name) }
+  end
+
+  def self.create_breweries
+    100.times do |i|
+      FactoryGirl.create(:brewery)
```
incazteca: Use `create_list`

yadda/yadda_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def self.create_users
+    100_000.times do |i|
+      FactoryGirl.create(:user)
+      puts "#{i}/100,000 drinkers" if i % 10_000 == 0
+    end
+  end
+
+  def self.create_ratings
+    all_user_ids = User.pluck(:id)
```
incazteca: In real life you won't know the ids of users and all. But for this script you already know that all the ids are between 1 and the number created. So you can avoid making this DB call. As well as the call below for the beer ids.

yadda/yadda_generator.rb
incazteca: For `all_user_ids.sample` You could also use `(1..user_count).sample`

yadda/yadda_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+    create_ratings
+  end
+
+  def self.create_beer_styles
+    beer_style_names = File.readlines('data/beer_styles.txt').map(&:chomp)
+    beer_style_names.each { |name| FactoryGirl.create(:beer_style, name: name) }
+  end
+
+  def self.create_breweries
+    100.times do |i|
```
incazteca: Avoid magic numbers. Create a constant above `BREWERY_COUNT = 100`

incazteca:
@ssapra Looks good. It would be nice for the views if they included the beer name or the brewery name instead of just the ids. Most of the times views are created for the benefit of the user and it'll be more helpful for a person to be able to read names instead of just id numbers. You can go ahead with part 4.

Also make some changes to the yadda_generator. I'm not sure but I think the `create_list` function will at least speed up the creation of multiple entries. Though I'll be honest I've never seen anyone before use FactoryGirl for this exercise. The PG gem itself is very fast for adding entries.

yadda/yadda_generator.rb
ssapra: The one benefit of using the times loop is to see the progress, but I'll see if create_list can improve speed

paulghaddad:
@ssapra Nice work on this exercise. As Franco said, you're good to move on to Part 4.

PR #937, opened 2015-08-22
Cyoa V1.5

mcdaniel67:
I thought about the structure of my app since the first submission. I realized that having artists generate the graph might not be the best idea. In the restructured version of the app, graphs have their own model. I split the responsibilities of the artist model into the graph model in order to make the app more logical. I also added the ability to cache a graph of a given depth.

I also renamed the artists controller to graphs. This seems much more logical to me.

There are various other places that I refactored code too.

The app is a related artist graph generator using the spotify api. It is used to visualize the relationships between artists that spotify suggests are related. It can tend to be slow, because any given artist has 20 related artists. This means that graph generation grows exponentially W.R.T. depth. 

There are some pictures of the app in the readme.


spotigraph/app/helpers/graph_helper.rb
```diff
@@ -0,0 +1,2 @@
+module GraphHelper
```
paulghaddad: You can delete unused files.

spotigraph/app/controllers/graph_controller.rb
```diff
@@ -0,0 +1,51 @@
+class GraphController < ApplicationController
+  def index
```
paulghaddad: You shouldn't need a controller action for `index` if you're just rendering the page. Deleting this should work just fine.

spotigraph/app/models/artist.rb
```diff
@@ -0,0 +1,27 @@
+# Graph generation. This recursively builds the tree of graph.
+class Artist < ActiveRecord::Base
+  serialize :related, Array
+
+  def self.lookup_artist(name)
+    artist = Artist.find_by(name: name)
+    return Artist.search_spotify(name) if artist.nil?
```
paulghaddad: This should work just as well to combine lines 7 - 8:

`artist || Artist.search_spotify(name)`

spotigraph/app/models/artist.rb
```diff
@@ -0,0 +1,27 @@
...
+    artist
+  end
+
+  def self.search_spotify(name)
```
paulghaddad: This method appears to be doing more than one thing. Consider extracting functionality to other method(s) to make it do one thing.

spotigraph/app/models/artist.rb
```diff
@@ -0,0 +1,27 @@
...
+    artist = Artist.find_artist_in_array(RSpotify::Artist.search(name), name)
+    return nil if artist.nil?
+    related = Artist.collect_related_artists(artist.related_artists)
+    Artist.create!(name: artist.name, json: artist.to_json, related: related)
+  end
+
+  # Returns nil unless we find a precise match against the searched artist
+  def self.find_artist_in_array(artist_array, name)
```
paulghaddad: Try not to name your methods according in relation to a data structure.

spotigraph/app/models/graph.rb
```diff
@@ -0,0 +1,35 @@
+class Graph < ActiveRecord::Base
+  def self.search(name, depth)
```
paulghaddad: This is an awfully long method. Refactor to simplify by extracting other methods.

spotigraph/app/models/graph.rb
```diff
@@ -0,0 +1,35 @@
...
+    [network, node_depth]
+  end
+
+  # A hash merge with the strategy of selecting the max value with keys in both
+  # hashes. Used to keep track of the maximal depths of nodes.
+  def self.depth_merge!(hash_1, hash_2)
+    hash_1.merge(hash_2) { |_key, old, new| [old.to_i, new.to_i].max }
+  end
+
+  class << self
```
paulghaddad: Use the `self.` syntax. Also, I'm a little surprised you could make these class methods private.

spotigraph/app/models/user.rb
```diff
@@ -0,0 +1,2 @@
+class User < ActiveRecord::Base
```
paulghaddad: Why is this model empty? Do you not have any validations or associations?

spotigraph/app/views/graph/generate_graph.html.erb
```diff
@@ -0,0 +1,13 @@
+<script type="text/javascript">
```
paulghaddad: Your JavaScript code should not be in your views. Only HTML/ERB should live here; the JS goes in `assets/javascripts`.

spotigraph/app/views/layouts/_searchbar.html.erb
```diff
@@ -0,0 +1,18 @@
...
+    </div>
+    <div class="form-group">
+      <label for="recursion-depth">Depth</label>
+      <input type="NUMBER" min="1" max="5" step="1" name="depth" class="form-control" id="recursion-depth" value="<%= params[:depth] %>"><br/>
+    </div>
+    <div class="form-group">
+      <button type="submit" name = "Submit" class="btn btn-default">Submit</button>
+    </div>
+  </form>
+  <% flash.each do |message_type, message| %>
```
paulghaddad: Try extracting the flash portion of this view into its own partial for reuse elsewhere.

spotigraph/config/secrets.yml
```diff
@@ -0,0 +1,22 @@
+# Be sure to restart your server when you modify this file.
```
paulghaddad: This file shouldn't be checked into version control if this was a production app.

spotigraph/app/controllers/graph_controller.rb
mcdaniel67: I left it in there to remind myself that index exists. I can definitely remove it though if you think that makes it cleaner.

paulghaddad: @mcdaniel67 No worries...it is a small thing. I should be done with the review later today.

spotigraph/app/views/graph/generate_graph.html.erb
mcdaniel67: So after extracting this out into a function in `assets/javascripts` would I just make the function call inside of script tags?

spotigraph/spec/controllers/graph_controller_spec.rb
```diff
@@ -0,0 +1,105 @@
+require 'rails_helper'
+
+RSpec.describe GraphController, type: :controller do
+  describe 'GET', '#index' do
```
paulghaddad: There probably isn't too much benefit to a test such as this. Controller tests are useful when testing multiple paths through a controller, such as a successful sign-in, or an unsuccessful sign-in. See this article:

https://robots.thoughtbot.com/how-we-test-rails-applications

spotigraph/spec/controllers/graph_controller_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+        nil,
+        '-1',
+        -1,
+        7,
+        8
+      ]
+    end
+
+    # The next two functions ensure GACC on validate_depth
+    it 'returns 200 on :valid_artists depth: 0 to 2' do
```
paulghaddad: Are these testing an expectation? I'm a little confused as I don't see a `expect`.

paulghaddad: Now I see where the expectation is. It's my opinion that it would be better to have the expectation inside the `it` blocks, that way the person reading the code knows where it is. It is good to extract test setup in methods, but I wouldn't advise it for the expectation. Hence, my comment above.

spotigraph/spec/factories/graphs.rb
```diff
@@ -0,0 +1,8 @@
+FactoryGirl.define do
+  factory :graph do
+    artist "MyString"
```
paulghaddad: I'd give this more meaningful names than "MyString".

spotigraph/spec/lib/graph_json_spec.rb
```diff
@@ -0,0 +1,111 @@
+require_relative('../../lib/graph_json')
```
paulghaddad: You should be able to get this autoloaded by tweaking your the `config.autoload_paths` setting in your configuration file for test (you can also do this for your other environments).

spotigraph/spec/lib/graph_json_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+require('faker')
```
paulghaddad: I'd place this in your `Gemfile`.

spotigraph/spec/lib/graph_json_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+
+describe GraphJSON, '.to_json' do
+  let(:graph_json) { GraphJSON.new }
```
paulghaddad: Good job using `let`.

spotigraph/spec/models/artist_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+    # Note, Faker will not generate names of this form
+    let(:invalid_names) do
+      %w(sPUKslXlnz pdbPxLK7rt w8yUJnC171 mLvr4IaeC8 ZYR9Z5R967)
+    end
+
+    let(:random_array_size_max) { 20 }
+
+    it 'returns nil if we search against an empty array' do
+      valid_names.each do |name|
+        expect(Artist.find_artist_in_array([], name)).to eq(nil)
```
paulghaddad: `be nil` is the correct matcher.

spotigraph/spec/models/artist_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+
+    it 'fails to find an artist not present in a randomly sized array' do
+      size = rand(2...random_array_size_max)
+      artist_array = generate_artist_array(size)
+      invalid_names.each do |invalid|
+        expect(Artist.find_artist_in_array(artist_array, invalid)).to eq(nil)
+      end
+    end
+
+    def generate_artist_array(size, options = {})
```
paulghaddad: I'd place a method supporting your specs at the bottom of the file or in a `support` directory inside your `spec` directory.

spotigraph/spec/models/graph_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+
+    it 'merges correctly when multiple keys coincide (avg use)' do
+      hash_1 = { a: 1, b: 2, c: 3, d: 4, e: 5 }
+      hash_2 = { a: 4, b: 3, c: 2, d: 1, e: 7 }
+      result = { a: 4, b: 3, c: 3, d: 4, e: 7 }
+      expect(Graph.depth_merge!(hash_1, hash_2)).to eq(result)
+    end
+
+    it 'merges correctly when left has nil' do
+      hash_1 = { a: 1, b: 2, c: nil, d: 4, e: 5 }
```
paulghaddad: Consider extracting the setup of creating the two hashes in most of these tests into another method.

paulghaddad:
@mcdaniel67 @djkotowski This is a very good first iteration of the app. There are two things I'd like to see in your second iteration. The implementation of a user authentication system, such as Devise or Clearance. And second, a closer look at your tests. Let's chat next week about the tests, but I'm impressed on what the app does!

spotigraph/app/views/graph/generate_graph.html.erb
paulghaddad: Lets talk next week. We can do it in person.

PR #940, opened 2015-08-25
Refactored a few things.

mostlybadfly:
I changed a few things around, let me know what you think!

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -1,58 +1,60 @@
 # Killer facts about triangles AWW YEAH
 class Triangle
-	attr_accessor :side1,:side2,:side3
+	attr_accessor :side1, :side2, :side3
 
-	def initialize(side1,side2,side3)
-		@side1,@side2,@side3 = side1,side2,side3
+	def initialize(side1, side2, side3)
+		@side1, @side2, @side3 = side1, side2, side3
 	end
 
 	def equilateral()
-		return side1 == side2 && side2 == side3
+		side1 == side2 && side2 == side3
 	end
 
 	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
+		[side1, side2, side3].uniq.length == 2
 	end
 
 	def scalene()
-		unless ! (equilateral || isosceles)
-	    false
-	  else
-	    true
-	  end
+		equilateral || isosceles ? true : false
 	end
 
 	def recite_facts
-		puts 'This triangle is equilateral!' if equilateral
-		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+		if equilateral
+      puts "This triangle is equilateral!"
+    elsif isosceles
+		  puts "This triangle is isosceles! Also, that word is hard to type."
+    else
+      puts "This triangle is scalene and mathematically boring."
+    end
 
-		angles = self.calculate_angles(side1,side2,side3)
-		puts 'The angles of this triangle are ' + angles.join(',')
+		angles = calculate_angles(side1, side2, side3)
 
-		puts 'This triangle is also a right triangle!' if angles.include? 90
+    puts "The angles of this triangle are #{angles.join(', ')}" 
+
+    puts "This triangle is also a right triangle!" if angles.include? 90
+
 		puts ''
 	end
 
-	def calculate_angles(a,b,c)
+	def calculate_angles(a, b, c)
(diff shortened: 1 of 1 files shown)


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,60 @@
 # Killer facts about triangles AWW YEAH
 class Triangle
-	attr_accessor :side1,:side2,:side3
+	attr_accessor :side1, :side2, :side3
 
-	def initialize(side1,side2,side3)
-		@side1,@side2,@side3 = side1,side2,side3
+	def initialize(side1, side2, side3)
+		@side1, @side2, @side3 = side1, side2, side3
```
paulghaddad: I would personally put one instantiation on each line. I know it adds two additional lines, but it is more readable.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,60 @@
...
 	end
 
 	def equilateral()
```
paulghaddad: In Ruby, the parentheses are optional. It is preferred to leave them off if there are no parameters:

```
def equilateral?
...
```

paulghaddad: In Ruby, if a method returns a Boolean, as this one does, it is convention to add a `?` at the end.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,60 @@
...
+		[side1, side2, side3].uniq.length == 2
 	end
 
 	def scalene()
-		unless ! (equilateral || isosceles)
-	    false
-	  else
-	    true
-	  end
+		equilateral || isosceles ? true : false
```
slackjockey: if equilateral and isosceles both return booleans, then just `equilateral || isosceles` should accomplish the same results.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,60 @@
...
+	def initialize(side1, side2, side3)
+		@side1, @side2, @side3 = side1, side2, side3
 	end
 
 	def equilateral()
-		return side1 == side2 && side2 == side3
+		side1 == side2 && side2 == side3
 	end
 
 	def isosceles()
```
paulghaddad: `isosceles?`

triangle_facts/triangle_facts.rb
paulghaddad: I think you may have indented a little too much, although it is hard to tell?

You can also simplify this to:

`!(equilateral? || isosceles?)`

paulghaddad: This will by itself return a Boolean, so the ternary is unneccessary.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,60 @@
...
-	    true
-	  end
+		equilateral || isosceles ? true : false
 	end
 
 	def recite_facts
-		puts 'This triangle is equilateral!' if equilateral
-		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+		if equilateral
```
paulghaddad: It will be more readable to put conditionals at the end of each `puts` statement. For example:

`puts "This triangle is equilateral!" if equilateral?`

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,60 @@
...
+    puts "The angles of this triangle are #{angles.join(', ')}" 
+
+    puts "This triangle is also a right triangle!" if angles.include? 90
 
-		puts 'This triangle is also a right triangle!' if angles.include? 90
 		puts ''
 	end
 
-	def calculate_angles(a,b,c)
+	def calculate_angles(a, b, c)
```
paulghaddad: This method can be made `private` since it is only called within the class.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,60 @@
...
+      puts "This triangle is scalene and mathematically boring."
+    end
 
-		angles = self.calculate_angles(side1,side2,side3)
-		puts 'The angles of this triangle are ' + angles.join(',')
+		angles = calculate_angles(side1, side2, side3)
+
+    puts "The angles of this triangle are #{angles.join(', ')}" 
+
+    puts "This triangle is also a right triangle!" if angles.include? 90
```
paulghaddad: I'd recommend creating a new method to determine if a triangle is right.

paulghaddad:
@mostlybadfly Nice work on this first iteration. Please take a look at our comments and submit another iteration. Let me know if you have any questions.

PR #942, opened 2015-08-25
Supportive v1

abake-enova:


paulghaddad:
@abake-enova Hey Alex...please remove the non-supportive files and resubmit. Thanks!

PR #943, opened 2015-08-25
CYOA v2

ssapra:
I have the individual commits over at github.com/ssapra/good_deeds.


cyoa/README.md
```diff
@@ -1,17 +1,41 @@
...
+SUNLIGHTLABS_APIKEY: 'my-api-key'
+```
+
+# Setup
+```
+git clone https://github.com/ssapra/good_deeds.git
+cd good_deeds
+bin/setup
+rails server
+```
```
jmmastey: Ooh, real README. Nicely done!

cyoa/app/assets/javascripts/animation.js
```diff
@@ -0,0 +1,18 @@
...
+        $('#search').attr('action', '/' + resource.toLowerCase());
+        $("input[type='search']").attr('name', resource.toLowerCase().slice(0,-1) + '[query]');
+    });
+
+    $(document).on('click', '.clickable tr', function() {
+        var url = $(this).attr("data-url");
+        if(url) {
+            window.location = url;
+        }
+    });
```
jmmastey: Minor sin here. If your page requires data-urls and JS to navigate, something's gone a bit wonky. Saved somewhat by the fact that it's part of a table, but even still I might include a link somewhere in the table (first TD maybe) and just have the JS click that link as a handy shortcut.

cyoa/app/assets/javascripts/autocomplete.js
```diff
@@ -0,0 +1,18 @@
+$(document).ready(function() {
+    $('#tag_name').autocomplete({
+        source: $('#tag_name').data('autocomplete-source'),
+        focus: function( event, ui ) {
+            $( "#tag_name" ).val( ui.item.label );
+            return false;
+        },
+        select: function(event, ui){
+            $('#tag_name').val(ui.item.label);
```
jmmastey: Since the lookup is happening four times, may as well throw it into a variable (`$searchField`, maybe). OTOH, lookup by ID is reasonably fast.

Either way, if these call functions get much larger, they need to be real functions with namespacing.

cyoa/app/assets/stylesheets/autocomplete.scss
```diff
@@ -0,0 +1,51 @@
...
+  border-color: #ccc;
+  border-color: rgba(0, 0, 0, 0.2);
+  border-style: solid;
+  border-width: 1px;
+  -webkit-border-radius: 5px;
+  -moz-border-radius: 5px;
+  border-radius: 5px;
+  -webkit-box-shadow: 0 5px 10px rgba(0, 0, 0, 0.2);
+  -moz-box-shadow: 0 5px 10px rgba(0, 0, 0, 0.2);
+  box-shadow: 0 5px 10px rgba(0, 0, 0, 0.2);
```
jmmastey: I'm guessing you didn't write all this. If you did, use SASS helpers instead to do all the vendor prefix stuff.

cyoa/app/assets/stylesheets/autocomplete.scss
```diff
@@ -0,0 +1,51 @@
...
+
+.ui-autocomplete {
+  position: absolute;
+  top: 100%;
+  left: 0;
+  z-index: 1000;
+  float: left;
+  display: none;
+  min-width: 160px;
+  _width: 160px;
```
jmmastey: IE hack? If you wrote this one, leave a comment on it, or someone will "fix" it by accident.

cyoa/app/assets/stylesheets/home.scss
```diff
@@ -0,0 +1,118 @@
...
+  height: 50px;
+  line-height: 50px;
+  display: table-cell;
+  background-color: white;
+  border-radius: 2px;
+}
+
+.search .material-icons{
+  color: white !important;
+}
```
jmmastey: Generally speaking if you need `!important` something's gone wrong. Look into CSS specificity rules and make this not a thing. Also, add a space before your bracket ;)

cyoa/app/assets/stylesheets/home.scss
```diff
@@ -0,0 +1,118 @@
...
+.fb-share-button{
+  display: block !important;
+}
+
+input[type=text]:focus:not([readonly]), input[type=password]:focus:not([readonly]),
+input[type=email]:focus:not([readonly]), input[type=url]:focus:not([readonly]),
+input[type=time]:focus:not([readonly]), input[type=date]:focus:not([readonly]),
+input[type=datetime-local]:focus:not([readonly]), input[type=tel]:focus:not([readonly]),
+input[type=number]:focus:not([readonly]), input[type=search]:focus:not([readonly]),
+textarea.materialize-textarea:focus:not([readonly]){
```
jmmastey: Confused. Which input type is _not_ included here? Is there not a smaller selector for this?

cyoa/app/controllers/bills_controller.rb
```diff
@@ -0,0 +1,48 @@
+class BillsController < ApplicationController
+  before_action :ensure_nonempty_query, only: :index
+
+  def index
+    @pg_results = PgSearch.multisearch(@search_params)
+    @num_bills = @pg_results.count
+    @pg_results = @pg_results.page(params[:page])
+    respond_to do |format|
+      format.html
+      format.json { render 'index', layout: false }
```
jmmastey: The `layout: false` seems extraneous here. Not running the app right now, but I'd look into that at some point.

cyoa/app/controllers/bills_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def random_good_deed
+    find_random_bill && return unless user_signed_in?
+
+    tag_bills = current_user.tags.map(&:bills)
+    legislator_bills = current_user.legislators_by_zipcode.map(&:bills)
+    bill = (tag_bills + legislator_bills).flatten.sample
+    redirect_to bill_path(bill)
+  end
```
jmmastey: The two branches of logic in this method (`user_signed_in?` and not) are masked by the way this method is broken out. Better:

```
def random_good_deed
  if user_signed_in?
    bill = Bill.all.sample
  else
    bill = InterestingBills.for(current_user)
  end
  redirect_to bill_path(bill)
end
```

Also, I think in both cases you're loading a /ton/ of data from the database. Use of scopes here would probably do a great job.

cyoa/app/controllers/bills_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def show
+    @bill = Bill.find_by_bill_id(params[:bill_id])
```
jmmastey: Any reason not to use `@bill = Bill.find(params[:bill_id])`?

cyoa/app/controllers/bookmarks_controller.rb
```diff
@@ -0,0 +1,28 @@
+class BookmarksController < ApplicationController
+  before_action :authenticate_user!
+  before_action :find_bill, only: [:create, :destroy]
+
+  def create
+    Bookmark.create(user_id: current_user.id, bill_id: @bill.id)
+    redirect_to bill_path(@bill), notice: 'Bookmarked!'
+  end
```
jmmastey: Fwiw, I think you can use `redirect_to @bill, notice: "Bookmarked!"` here.

cyoa/app/controllers/legislators_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+      format.json { render 'index', layout: false }
+    end
+  end
+
+  def show
+    @legislator = Legislator.find(params[:id])
+    respond_to do |format|
+      format.html
+      format.json { render 'show', layout: false }
+    end
```
jmmastey: Starting to wonder if these `respond_to` blocks are necessary at all. I think the default rendering code will pick up the correct partial and filetype.

cyoa/app/controllers/user_tags_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+  before_action :authenticate_user!
+
+  def create
+    user_tag = UserTag.new(user_tag_params)
+
+    if user_tag.save
+      redirect_to user_tag.user, notice: 'Updated tags'
+    else
+      flash[:alert] = user_tag.errors.messages.values.join(', ')
+      render 'users/show'
```
jmmastey: Interesting that it's a render and not a `redirect_back`. Any reason?

cyoa/app/controllers/users_controller.rb
```diff
@@ -0,0 +1,32 @@
+class UsersController < ApplicationController
+  before_action :authenticate_user!
+
+  def show
+  end
+
+  def update
+    @user = User.find(params[:user][:id])
```
jmmastey: Ehhh, I think you can edit any user this way.

cyoa/app/helpers/application_helper.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    if params[:controller] == 'legislators'
+      bill_class = disabled_class
+      legislator_class = active_class
+    else
+      bill_class = active_class
+      legislator_class = disabled_class
+    end
+    [bill_class, legislator_class]
+  end
```
jmmastey:
```
def set_classes
  # ...
  if params[:controller] == 'legislators'
    [disabled_class, active_class]
  else
    [active_class, disabled_class]
  end
end    
```

Do the descriptive labels help? Not entirely, and if that's the concern it'd be better to get rid of the multiple return anyway.

cyoa/app/helpers/bill_helper.rb
```diff
@@ -0,0 +1,85 @@
...
+    elsif %w(sconres hconres).include?(@bill.bill_type) && house_action && senate_action
+      'Agreed To Concurrent Resolution'
+    elsif senate_bill_passed_senate_and_house || house_bill_only_passed_house
+      "Passed House on #{house_action.date.strftime('%b %-d, %Y')}"
+    elsif senate_bill_only_passed_senate || house_bill_passed_house_and_senate
+      "Passed Senate on #{senate_action.date.strftime('%b %-d, %Y')}"
+    elsif no_houses_passed
+      "Introduced to #{@bill.chamber.capitalize}"
+    end
+  end
```
jmmastey: I think these are pretty much fine, and I'm certainly glad they're not in the model. The vogue is moving towards Presenters to do this sort of work, or Value Objects, at least partially because these helpers are only available in a view context.

cyoa/app/helpers/bill_helper.rb
```diff
@@ -0,0 +1,85 @@
...
+    senate_action = actions.passed_senate.first
+    determine_bill_state(house_action, senate_action)
+  end
+
+  def determine_bill_state(house_action, senate_action)
+    senate_bill_passed_senate_and_house = (@bill.bill_type == 's' && senate_action && house_action)
+    house_bill_only_passed_house = (%w(hr hconres).include?(@bill.bill_type) && house_action && senate_action.nil?)
+    house_bill_passed_house_and_senate = (@bill.bill_type == 'hr' && senate_action && house_action)
+    senate_bill_only_passed_senate = (%w(s sconres).include?(@bill.bill_type) && senate_action && house_action.nil?)
+    no_houses_passed = house_action.nil? && senate_action.nil?
```
jmmastey: Notwithstanding my note below, there is a _lot_ of logic in this helper method, and the one above it.

cyoa/app/helpers/bill_helper.rb
```diff
@@ -0,0 +1,85 @@
...
+    elsif (vetoed_action = actions.vetoed.first)
+      return vetoed_action.text
+    end
+
+    house_action = actions.passed_house.first
+    senate_action = actions.passed_senate.first
+    determine_bill_state(house_action, senate_action)
+  end
+
+  def determine_bill_state(house_action, senate_action)
```
jmmastey: Ooh, I thought state meant geographical state. Somewhat confusing.

cyoa/app/helpers/bill_helper.rb
```diff
@@ -0,0 +1,85 @@
...
+    elsif (signed_action = actions.signed.first)
+      return signed_action.text
+    elsif (vetoed_action = actions.vetoed.first)
+      return vetoed_action.text
+    end
+
+    house_action = actions.passed_house.first
+    senate_action = actions.passed_senate.first
+    determine_bill_state(house_action, senate_action)
+  end
```
jmmastey: Make the last thing return a simple object that responds to `.text`, then:

```
last_action = (actions.enacted.first || actions.signed.first || actions.vetoed.first || last_bill_state)
last_action.text
```

Saves the multiple return and the needless repetition.

jmmastey: Though the more I think about this one, I think you could probably just sort all the actions by date, pick the top one and save even more text.

cyoa/app/helpers/flash_helper.rb
```diff
@@ -0,0 +1,16 @@
...
+      content_tag(:div, class: %w(col s6 offset-s3)) do
+        content_tag(:div, class: ['card-panel', color_class, 'notice']) do
+          concat icon(icon_value)
+          concat content_tag(:span, notice, class: 'white-text') if notice
+          concat content_tag(:span, alert, class: 'white-text') if alert
+        end
+      end
+    end
+  end
+end
```
jmmastey: This is probably just a partial, not a helper.

cyoa/app/helpers/legislator_helper.rb
```diff
@@ -0,0 +1,61 @@
...
+    party_hash = { 'D' => 'Democrat', 'R' => 'Republican',
+                   'I' => 'Independent' }
+    party_hash[@legislator.party]
+  end
+
+  def legislator_name
+    [:first_name, :middle_name, :last_name].map { |attr| @legislator.send(attr) }
+      .compact.join(' ')
+  end
+end
```
jmmastey: This one looks much more like it should be in a Presenter, since it doesn't do any DOM creation at all.

cyoa/app/helpers/legislator_helper.rb
```diff
@@ -0,0 +1,61 @@
...
+    return '#' unless @legislator.votesmart_id
+    'http://votesmart.org/candidate/' \
+    "#{@legislator.votesmart_id}/#{@legislator.first_name}-#{@legislator.last_name}"
+  end
+
+  def age
+    return 'n/a' unless @legislator.birthday
+    dob = Date.parse(@legislator.birthday)
+    now = Time.zone.now.utc.to_date
+    num_years = now.year - dob.year - ((now.month > dob.month || (now.month == dob.month && now.day >= dob.day)) ? 0 : 1)
```
jmmastey: This is confusing.

cyoa/app/models/bill_action.rb
```diff
@@ -0,0 +1,15 @@
+class BillAction < ActiveRecord::Base
+  belongs_to :bill
+
+  scope :passed_senate, -> { where(result: 'pass', chamber: 'senate') }
+  scope :passed_house, -> { where(result: 'pass', chamber: 'house') }
+  scope :signed, -> { where(text: 'Signed by President.') }
+  scope :vetoed, -> { where(text: 'Vetoed by President.') }
```
jmmastey: It's a little scary that this is how search works, but I believe you.

cyoa/app/models/tag.rb
```diff
@@ -0,0 +1,7 @@
+class Tag < ActiveRecord::Base
+  has_many :user_tags
+  has_many :users
+
+  has_many :bill_tags
+  has_many :bills, through: :bill_tags
+end
```
jmmastey: I think you're missing a `, through: :user_tags`

cyoa/app/models/bill.rb
```diff
@@ -0,0 +1,27 @@
+class Bill < ActiveRecord::Base
+  include PgSearch
+  multisearchable against: [:official_title, :short_title, :summary_short]
+
+  validates :bill_id, uniqueness: true
+  belongs_to :legislator
+
+  has_many :bill_tags, dependent: :destroy
+  has_many :tags, through: :bill_tags
```
jmmastey: The only place I see this relation used is in a single test. Don't be afraid not to declare every relationship symmetrically, it's okay to only access the tags from the other direction. That also makes this class a little simpler.

cyoa/app/views/bill_update_mailer/update.haml
```diff
@@ -0,0 +1,23 @@
+!!!
+%html
+  %head
+    %meta{:content => "text/html; charset=UTF-8", "http-equiv" => "Content-Type"}
+
+  %body
+    %table{style: 'margin-left: 10%; border-collapse: collapse;'}
```
jmmastey: Boooo inline styles. Move it to CSS.

cyoa/app/views/bill_update_mailer/update.haml
```diff
@@ -0,0 +1,23 @@
...
+      %thead
+        %td Bill
+        %td Status Updates
+      %tbody
+        -@bills.each_pair do |bill_id, updates|
+          -bill = Bill.find(bill_id)
```
jmmastey: Loading records in a view is a no-no. Load them in the controller.

cyoa/app/views/bookmarks/index.haml
```diff
@@ -0,0 +1,29 @@
...
+    -else
+      %table.bordered.clickable.hoverable.responsive-table#results
+        %thead
+          .row
+            %td.col.s2 Last Action
+            %td.col.s3 Status
+            %td.col.s7 Summary
+        %tbody
+          .row
+            -@bills.each do |bill|
```
jmmastey: None of these eaches so far have needed it, but make sure you're aware of `render @bills` as a way to render a collection of objects.

cyoa/app/views/users/registrations/edit.html.erb
```diff
@@ -0,0 +1,39 @@
...
+  <div class="actions">
+    <%= f.submit "Update" %>
+  </div>
+<% end %>
+
+<h3>Cancel my account</h3>
+
+<p>Unhappy? <%= button_to "Cancel my account", registration_path(resource_name), data: { confirm: "Are you sure?" }, method: :delete %></p>
+
+<%= link_to "Back", :back %>
```
jmmastey: So, with this and other devise views, try not to generate them unless you have to, or else you have to maintain them. Better to just let devise handle everything it can and have an easier time upgrading later.

cyoa/config/initializers/devise.rb
```diff
@@ -0,0 +1,265 @@
...
+  # confirmation, reset password and unlock tokens in the database.
+  # Devise will use the `secret_key_base` on Rails 4+ applications as its `secret_key`
+  # by default. You can change it below and use your own secret key.
+  # config.secret_key = 'fcbdcc4f358298a1b425ac43aa8d1c90e854f2081653126d42375a8c57e80d6dd62240c44aea7c338d6e2ea346a1647f7d6ce68e772a3f44242c1a03f07c122f'
+
+  # ==> Mailer Configuration
+  # Configure the e-mail address which will be shown in Devise::Mailer,
+  # note that it will be overwritten if you use your own mailer class
+  # with default "from" parameter.
+  config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
```
jmmastey: :(

cyoa/config/initializers/states.rb
```diff
@@ -0,0 +1,58 @@
...
+      "New Mexico" => "NM",
+      "New York" => "NY",
+      "North Carolina" => "NC",
+      "North Dakota" => "ND",
+      "Northern Mariana Islands" => "CM",
+      "Ohio" => "OH",
+      "Oklahoma" => "OK",
+      "Oregon" => "OR",
+      "Pennsylvania" => "PA",
+      "Puerto Rico" => "PR",
```
jmmastey: About time they got representation.

cyoa/db/migrate/20150804191020_change_legislator_attributes.rb
```diff
@@ -0,0 +1,22 @@
...
+    add_column :legislators, :leadership_role, :string
+    add_column :legislators, :in_office, :boolean
+
+    rename_column :legislators, :firstname, :first_name
+    rename_column :legislators, :middlename, :middle_name
+    rename_column :legislators, :lastname, :last_name
+    rename_column :legislators, :birthdate, :birthday
+    rename_column :legislators, :webform, :contact_form
+    rename_column :legislators, :congress_office, :office
+    rename_column :legislators, :youtube_url, :youtube_id
```
jmmastey: I concur.

cyoa/features/bills.feature
```diff
@@ -0,0 +1,39 @@
+Feature: Search and viewing bills
+  As a website visitor
+  I want to find bills
+  Because I want to learn about Congress
```
jmmastey: Said the programming director of CSPAN

cyoa/features/bills.feature
```diff
@@ -0,0 +1,39 @@
...
+  Scenario: Click on search result
+    Given I am a new user
+    When I search and click on a result for "recombinant DNA"
+    Then I see the bill page for "Safe and Accurate Food Labeling Act of 2015"
+
+  @javascript
+  Scenario: No results found for search
+    Given I am a new user
+    When I search bills for "mystery alient act of 2025"
+    Then I see "0 bills found for mystery alient act of 2025"
```
jmmastey: Nice scenarios. New user is probably Background, but otherwise spot on.

cyoa/features/legislators.feature
```diff
@@ -0,0 +1,50 @@
...
+  Scenario: Search legislators by name
+    Given I visit the home page
+    When I search legislators for "Bob"
+    Then I see 1 legislators
+
+  @javascript
+  Scenario: Search legislators by title
+    Given I visit the home page
+    When I search legislators for "Representative"
+    Then I see 1 legislators
```
jmmastey: This feature seems to be of dubious value.

cyoa/features/step_definitions/bookmark_steps.rb
```diff
@@ -0,0 +1,56 @@
...
+
+Then(/^I should see (\d+) bills$/) do |num|
+  expect(page).to have_css('#results')
+  expect(all('tbody tr').count).to eq(num.to_i)
+end
+
+Then(/^I should see 1 bill$/) do
+  expect(page).to have_css('#results')
+  expect(all('tbody tr').count).to eq(1)
+end
```
jmmastey: Can be combined with the above step definition, and probably should be.

cyoa/lib/legislator_query.rb
```diff
@@ -0,0 +1,104 @@
+class LegislatorQuery
```
jmmastey: Given this is just an interface to Postgres, it probably belongs in `app/models` as opposed to `lib`. The latter is for stuff that isn't entirely app specific (like parsing the sunlight API)

cyoa/spec/controllers/bills_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    describe '#show' do
+      before do
+        @bill = create(:bill)
+        create(:bill_action, bill: @bill, result: 'pass', chamber: 'senate')
+        create(:bill_action, bill: @bill)
+        params = default_params.merge(bill_id: @bill.bill_id)
+        get(:show, params)
+      end
```
jmmastey: Don't put the `get` (equivalent to `When`) into the `before` block, even if it's common to each action.

Also, fwiw, this way of creating objects gets out of hand quickly. If you end up with slow tests, this is a big place to refactor.

cyoa/spec/helpers/bill_helper_spec.rb
```diff
@@ -0,0 +1,260 @@
...
+    context 'signed bill' do
+      before do
+        @bill = create(:bill)
+        create(:bill_action, bill_id: @bill.id, text: 'Signed by President.')
+      end
+
+      it 'returns the signed text' do
+        expect(helper.bill_current_status).to eq('Signed by President.')
+      end
+    end
```
jmmastey: So, writing this once but it applies across these helpers. This is the problem with requiring global state (namely the need to create and assign vars elsewhere to test this helper). If it was wrapped in an object somewhere:

```
action = BillAction.new(text: "Signed by President")
bill = Bill.new(bill_actions: [action])

expect(BillStatus.for(bill)).to eq("Signed by President")
```

cyoa/spec/helpers/legislator_helper_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+  end
+
+  describe '#full_party' do
+    context 'expands D and' do
+      before { @legislator = build(:legislator, party: 'D') }
+
+      it 'shows Democrat' do
+        expect(helper.full_party).to eq('Democrat')
+      end
+    end
```
jmmastey: Also if the `before` block only applies to this one test, just merge it into the `it` block. In fact, this is probably overuse of `context` and can be combined into the it block as well, which should increase readability of the tests significantly.

cyoa/spec/lib/sunlightlabs_api_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+        expect(Tag.count).to eq(2)
+        expect(Tag.pluck(:name)).to contain_exactly('Unemployment', 'Finance')
+        expect(Bill.first.tags.count).to eq(2)
+      end
+    end
+
+    context 'pulling minimal bill data' do
+      before do
+        bill_data = { bill_id: 'hr123-114', bill_type: 'hr' }
+        json_response = default_params.merge('results' => [bill_data]).to_json
```
jmmastey: This may not be a very strong test. If the API didn't work, this would still be successful. Something like the VCR gem would probably be better.

cyoa/spec/models/user_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+
+    context 'with invalid attributes' do
+      context 'with invalid email' do
+        let(:user) { build(:user, email: 'bogus') }
+
+        it 'is invalid' do
+          expect(user).to_not be_valid
+        end
+      end
```
jmmastey: Are these devise tests or yours? They're not terribly high value. With this and other model tests, in my experience:

1. Validation tests aren't useful unless the validation is novel (like zipcode below)
2. Scopes don't need much testing unless they're complicated. In which case make them not complicated.

jmmastey:
Hey Shaan, nice job. I added some minor notes, but this is solid. There are some things in your test suite that will slow down pretty quickly, so you may want to focus on getting fast, high quality tests.

PR #946, opened 2015-08-25
Traffic control V1

mcdaniel67:


traffic_control/routes
```diff
@@ -0,0 +1,30 @@
+Prefix                   Verb   URI Pattern                                 Controller#Action
+
+Flight details, passengers, reroute (#update) and search. Also added the ability to destroy, because flights get canceled.
+flight                   GET    /flights/:id                                  flights#show
+                         PUT    /flights/:id                                  flights#update
+                         DELETE /flights/:id                                  flights#destroy
+flight_passengers        GET    /flights/:id/passengers                       flights#passengers
```
katieschilling: for security purposes, I'm sure

traffic_control/routes
```diff
@@ -0,0 +1,30 @@
...
+flight_search            GET    /flights/search/                              flights#search
+
+Find airport details
+airport                  GET    /airports/:id                                 airports#show
+
+Airplane trip taken, when to refuel, and update maintenance records.
+airplane                 GET    /airplanes/:id                                airplanes#show
+airplane_maintenance     GET    /airplanes/:id/maintenance                    airplanes#maintenance
```
katieschilling: this would likely be a separate system as each maintenance record would have an id, worker, date, issue, completed?, etc. I'm guessing this endpoint is only to add a new maintenance record that will then need to be passed over to the real engineers.

traffic_control/routes
```diff
@@ -0,0 +1,30 @@
...
+                         PUT    /airplanes/:id/maintenance                    airplanes#update
+airplane_refuel          GET    /airplanes/:id/refuel                         airplanes#refuel
```
katieschilling: shouldn't there be an endpoint for how much to refuel for a given flight id?

mcdaniel67: I was assuming that when an airplane is refueled it always goes to 100% capacity for safety reasons. It does seem logical that this might not be the case though. I can absolutely change this to allow for partial refueling.

paulghaddad:
@mcdaniel67 Nice work on this. You can go ahead an submit a second iteration with these small changes. You should be good to go after that.

PR #947, opened 2015-08-26
Correct triangle facts Version2

mostlybadfly:
Hi!  Here is my updated iteration.  I was not too sure if calculate_angles could be moved elsewhere as I'm using that In two places, not sure if that matters. Thanks for the feedback.

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -3,49 +3,50 @@ class Triangle
 	attr_accessor :side1, :side2, :side3
 
 	def initialize(side1, side2, side3)
-		@side1, @side2, @side3 = side1, side2, side3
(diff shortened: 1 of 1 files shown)


paulghaddad:
@mostlybadfly Nice work on this second iteration. Before we review it, can you resubmit a pull request with better indenting? If you take a look at the indentation, it is inconsistent. It should be a two space tab.

mostlybadfly:
Will do, sorry about that.  Not sure what happened here.  I recently setup some dotfiles in between iterations so maybe that was it.  I'll get a new one uploaded as soon as I figure it out, hopefully just a matter of a retab, thanks!

paulghaddad:
@mostlybadfly My pleasure!

PR #948, opened 2015-08-26
Deja/dino_catalog (version 2)

DejaAugustine:
- Refactored core logic to drastically simplify matching / filtering logic
- Built interactive CLI interface
- Core class interface remains backward-compatible, so the original examples (such as the one below) will still work without modification:

```
dex = Dinodex.new
dex.import("dinodex.csv")
dex.import("african_dinosaur_export.csv")
dex.filter(continent: "America", op: "include?", negate: true)
   .filter(carnivore: true)
dex.each do |dino|
  puts dino
  puts
end
```

### DIFF
--- a/dino_catalog/dino_catalog.rb
+++ b/dino_catalog/dino_catalog.rb
@@ -1,54 +1,150 @@
 require 'json'
 require_relative 'dinodex'
+require_relative 'criterium'
+require_relative 'errors'
+require_relative 'import_loop'
 
-puts
-puts "***********"
-puts "* Dinodex *"
-puts "***********"
-puts
+class DinodexApp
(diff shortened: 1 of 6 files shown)


paulghaddad:
For my use.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,150 @@
...
+
+One exception to the standard query format is with boolean fields.
+You can test a boolean field is true by preceeding the field with
+"is", "is not", "are", or "are not"'
+e.g. is not carnivore? and period include? Cretaceous and name include? saur
+
+Optionally, the results can be printed in JSON format by appending "as json"
+to the end of a query.
+
+eos
```
paulghaddad: This is great aid to a user of the program.

PR #954, opened 2015-08-27
Yadda Part 1 - V1

mcdaniel67:


yadda/yadda-schema.sql
```diff
@@ -0,0 +1,84 @@
+DROP DATABASE IF EXISTS yadda;
+CREATE DATABASE yadda;
+\connect yadda;
+-- Ensures we have a clean starting state for our db.
+
+
+CREATE TABLE users (
+  user_id     SERIAL PRIMARY KEY,
```
paulghaddad: I'd use `id` for primary keys.

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,84 @@
...
+  name        VARCHAR(50)  NOT NULL,
+  email       VARCHAR(100) NOT NULL,
+  birthday    DATE         NOT NULL,
+  password    VARCHAR(50)  NOT NULL,
+  modified_on TIMESTAMP DEFAULT current_timestamp,
```
paulghaddad: Also see `TIMESTAMP WITH TIME ZONE`.

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,84 @@
...
+-- Store information on users. name, email, birthday, and password can not be set to null.
+-- modified_by and created_by can be set to null if the corresponding user is deleted
+-- A user has many ratings, and may have many modifications.
+
+CREATE TABLE breweries (
+  brewery_id     SERIAL PRIMARY KEY,
+  name           VARCHAR(100) NOT NULL,
+  street_address VARCHAR(100),
+  city           VARCHAR(200) NOT NULL,
+  country        VARCHAR(50),
```
paulghaddad: Some of these fields may warrant a `NOT NULL`.

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,84 @@
...
+  zip_code       VARCHAR(10),
+  description    TEXT,
+  founding_year  INTEGER,
+  modified_on    TIMESTAMP DEFAULT current_timestamp,
+  modified_by    INTEGER REFERENCES users (user_id) ON DELETE SET NULL,
```
paulghaddad: I don't think you want a relation between `breweries` and `users`. That relation should occur in the `ratings` table.

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,84 @@
...
+
+CREATE TABLE beers (
+  beer_id     SERIAL PRIMARY KEY,
+  name        VARCHAR(100) NOT NULL,
+  style       VARCHAR(100),
+  description TEXT,
+  brew_year   INTEGER,
+  brewery_id  INTEGER NOT NULL REFERENCES breweries ON DELETE CASCADE,
+  modified_on TIMESTAMP DEFAULT current_timestamp,
+  modified_by INTEGER NOT NULL REFERENCES users (user_id) ON DELETE SET NULL,
```
paulghaddad: Same here as in `breweries`. You don't need a relation between this table and `users`. A user is someone who rates a beer; I would keep it at that and store the `modified_by` and `create_by` fields as `VARCHARS`.

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,84 @@
...
+);
+COMMENT ON TABLE breweries IS 'Information about each brewery';
+-- Store brewery information. Name and city must not be null. modified_by and created_by
+-- may be set to null if the corresponding user has been deleted.
+-- Zip is varchar in order to support hyphens. Brewery to beers is one to many.
+
+CREATE TABLE beers (
+  beer_id     SERIAL PRIMARY KEY,
+  name        VARCHAR(100) NOT NULL,
+  style       VARCHAR(100),
```
paulghaddad: `style` is an opportunity to create a lookup table of a set number of styles. This will help normalize your database. Let me know if you have any questions about this.

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,84 @@
...
+-- Brewery for a specific beer can not be null, and when a brewery is deleted, all
+-- corresponding beers created by that brewery are deleted as well.
+-- modified_by and created_by may be set to null if the corresponding user is deleted.
+-- beer (singular) to brewery is one to one, though a brewery has many beers.
+
+CREATE TABLE ratings (
+  rating_id   SERIAL PRIMARY KEY,
+  user_id     INTEGER REFERENCES users (user_id) ON DELETE CASCADE,
+  beer_id     INTEGER REFERENCES beers (beer_id) ON DELETE CASCADE,
+  look        SMALLINT,
```
paulghaddad: You could enforce a range of the ratings by using `CHECK`.

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,84 @@
...
+  smell       SMALLINT,
+  taste       SMALLINT,
+  overall     SMALLINT,
+  modified_on TIMESTAMP DEFAULT current_timestamp,
+  modified_by INTEGER REFERENCES users (user_id) ON DELETE SET NULL,
+  created_on TIMESTAMP DEFAULT current_timestamp,
+  created_by INTEGER REFERENCES users (user_id) ON DELETE SET NULL,
+  CHECK (look > 0 AND look < 11),
```
paulghaddad: I like the use of `CHECK`. For conciseness, you can place this on the same lines of the fields they reference (65-68).

paulghaddad:
@mcdaniel67 This is a great start. Just a couple things. The two main ones are eliminating the need to reference the `users` table with other tables' `modified_by` and `updated_by` fields. The second is the use of a beer style lookup table. I think both will help out greatly.

Please take a look and resubmit a second version when ready. Nice work!

yadda/yadda-schema.sql
mcdaniel67: I figured that the only thing that should be necessary would be name and city. The thought was that we would make those the only required entries when setting up a new brewery. Users can supply more information/update if they wish,  but only the name and city would be required. I can change it if you would prefer to have more entries set to `NOT NULL`

PR #957, opened 2015-08-28
Yada v1

abake-enova:


yadda/schema.sql
```diff
@@ -0,0 +1,62 @@
...
+  created_by   INTEGER NOT NULL REFERENCES users,
+  updated_at   TIMESTAMP DEFAULT NOW(),
+  updated_by   INTEGER NOT NULL REFERENCES users
+);
+COMMENT ON TABLE beers IS 'Beer data.';
+
+CREATE TABLE ratings (
+  rating_id  SERIAL PRIMARY KEY,
+  user_Id    INTEGER NOT NULL REFERENCES users ON DELETE CASCADE,
+  beer_id    INTEGER NOT NULL REFERENCES beers ON DELETE CASCADE,
```
paulghaddad: Good job here!

yadda/schema.sql
```diff
@@ -0,0 +1,62 @@
...
+  look       FLOAT(3),
```
paulghaddad: Consider using an `INTEGER` or `SMALLINT`. Also add a condition that it defaults to and checks that it is inside the range of acceptable values. Something like:

`CHECK (look BETWEEN 0 AND 5) DEFAULT 2,`

yadda/schema.sql
```diff
@@ -0,0 +1,62 @@
...
+  created_by     INTEGER NOT NULL REFERENCES users,
+  updated_at     TIMESTAMP DEFAULT NOW(),
+  updated_by     INTEGER NOT NULL REFERENCES users
+);
+COMMENT ON TABLE breweries IS 'Brewery data.';
+
+CREATE TABLE beers (
+  beer_id      SERIAL PRIMARY KEY,
+  name         VARCHAR(255) NOT NULL,
+  style        VARCHAR(50),
```
paulghaddad: `style` is an opportunity to create a lookup table consisting of a set number of styles. This will normalize your data.

yadda/schema.sql
```diff
@@ -0,0 +1,62 @@
...
+  description  TEXT,
+  brewery_id   INTEGER NOT NULL REFERENCES breweries ON DELETE CASCADE,
```
paulghaddad: Good foreign relation.

paulghaddad:
@abake-enova This is very nice. Just a couple comments to take a look at. I would definitely add a `beer_styles` lookup table and ensure ratings stay within a range. But your tables all look good, and you did a nice job considering your field defaults. 

No need to submit a second iteration of this part. I'd like the changes made in Parts 2 and 3. If you could add a comment to the Pull Request that you changed the table per this review, that would be great! Nice work!

PR #959, opened 2015-08-28
Deja/overlord v1.2

DejaAugustine:
First draft of Overlord.

This is AJAX-driven / REST-backed, so I'm particularly interested in feedback on:
* Jasmine coverage for JavaScript
* RSpec coverage for API
* Capybara/cucumber coverage for E2E


overlord/overlord.rb
```diff
@@ -1,14 +1,89 @@
...
+require_relative 'lib/bomb'
+require_relative 'lib/timer'
 
-require 'sinatra'
+class Overlord < Sinatra::Base
+  enable :sessions
 
-enable :sessions
+  helpers do
+    def prepare_bomb(activate_code, disarm_code)
```
jmtai: Any particular reason these are named differently? (i.e. `active/disarm` vs `activate/deactivate` or `arm/disarm`)

overlord/overlord.rb
```diff
@@ -1,14 +1,89 @@
...
+      bomb = Bomb.new(activate_code, disarm_code)
+      bomb.wires = WireBundle.new(3, 2)
+      bomb
+    end
 
-get '/' do
-  "Time to build an app around here. Start time: " + start_time
-end
+    def status_hash(bomb)
```
jmtai: No need to include the return data type in the method name. This can just be named `status`.

overlord/overlord.rb
```diff
@@ -1,14 +1,89 @@
...
+      clock = bomb.timer ? bomb.timer.seconds_remaining : ""
```
jmtai: Why mix data types here? Since this value is being returned as part of your external API, it would be nice if values being returned are consistent and logical. `clock` contains the number of seconds remaining on your bomb timer, so if there's no timer, you could just have this value be `-1` or some other invalid value. That way clock always contains a number instead of a number or an empty string.

overlord/overlord.rb
```diff
@@ -1,14 +1,89 @@
...
+
+  before do
+    @activate_code = "1234"
+    @disarm_code = "0000"
+    session[:bomb] ||= prepare_bomb(@activate_code, @disarm_code)
+    @bomb = session[:bomb]
+  end
+
+  get '/' do
+    session[:bomb] = prepare_bomb(@activate_code, @disarm_code)
```
jmtai: Design question here:  A new bomb will be loaded each time you load the home page, discarding the old bomb. Is that what you want? As a user of the app, does that make sense?

overlord/overlord.rb
```diff
@@ -1,14 +1,89 @@
...
+    @bomb = session[:bomb]
+    erb :bomb
+  end
+
+  get '/BOMB/v1/' do
```
jmtai: +1 for API versioning. -1 for using an unnamed REST endpoint. :smile: 

This endpoint is returning the bomb status, so go ahead and name it `status`.

overlord/overlord.rb
```diff
@@ -1,14 +1,89 @@
...
+    content_type :json
+    status_hash(@bomb).to_json
```
jmtai: Here you can see how including the data type in your method name can unnecessarily clutter your code. Your API always return the status in JSON format, so you could just rename the method `status` and to the `.to_json` conversion within the method itself instead of having to tack it on everywhere you're returning `status_hash`.

overlord/overlord.rb
```diff
@@ -1,14 +1,89 @@
...
+  end
+
+  post '/BOMB/v1/' do
```
jmtai: Unless there's a compelling reason to do so, I'd advise against using mixed case in your REST endpoints. Whether or not case matters in the URL is OS and web-server dependent. There's no point in making your API users have to worry about using the right case if it's not necessary.

jmtai: Again, it's best to name all your endpoints. Ideally, our API should self-documenting but that's really hard to do if we don't assign names to all the endpoints.

overlord/overlord.rb
```diff
@@ -1,14 +1,89 @@
...
+
+    wires = @bomb.wires.wires.each_with_index.map do |wire, i|
+      { type: wire.type,
+        intact: wire.intact?,
+        index: i }
+    end
+    wires.to_json
+  end
+
+  post '/BOMB/v1/wire/cut' do
```
jmtai: Minor nit: This endpoint and the next are all related to the wires but are all named differently. It might be nicer to keep them all logically grouped with the `wires` endpoint like so:

```
/BOMB/v1/wires
/BOMB/v1/wires/cut/[wire_index]
/BOMB/v1/wires/short
```

overlord/overlord.rb
```diff
@@ -1,14 +1,89 @@
...
+    content_type :json
+
+    return { error: "Invalid Wire Selected" }.to_json unless params[:index] &&
```
jmtai: Since these errors aren't changing, you could make this a bit cleaner by defining them as constants.

jmtai: Having an unless condition that splits over multiple lines makes it a little hard to read. Personally, when this happens I prefer to use the less space-efficient but more readable long form:
```ruby
unless params[:index] && params[:index].to_i < @bomb.wires.wires.size
  return { error: "Invalid Wire Selected" }.to_json
end
```

jmtai: Another option is to create a predicate method for the test condition, which has the added benefit of making the logic more understandable:
```
def valid_wire?
  params[:index] && params[:index].to_i < @bomb.wires.wires.size
end

return { error: "Invalid Wire Selected" }.to_json unless valid_wire?
```
Using a constant for your error message makes it even cleaner:
```ruby
return INVALID_WIRE_ERR unless valid_wire?
```

overlord/overlord.rb
```diff
@@ -1,14 +1,89 @@
...
+      params[:index].to_i < @bomb.wires.wires.size
+    return { error: "Bomb is wireless" }.to_json unless @bomb.wires.wires
+
+    @bomb.wires.wires[params[:index].to_i].cut
```
jmtai: You're using `params[:index].to_i` in three different places, so just create a temporary variable to store the result and use that, instead.

overlord/overlord.rb
```diff
@@ -1,14 +1,89 @@
...
+    status_hash(@bomb).to_json
+  end
+
+  post '/BOMB/v1/short' do
+    @bomb.enter_code(@activate_code)
```
jmtai: Conceptually, it seems a bit odd to me that shorting the wires on the bomb actually activate it if it's inactive.

overlord/public/javascripts/detonator.js
```diff
@@ -0,0 +1,170 @@
+// Can't unit test a standard closure
+var $ = jQuery;
+
+function Detonator() {
+  if(this instanceof Detonator){
```
jmtai: Minor nit: Add a space between the `if` and the opening paren, and another space between the closing paren and the curly brace.

overlord/public/javascripts/detonator.js
```diff
@@ -0,0 +1,170 @@
...
+    return new Detonator();
+  }
+};
+
+Detonator.prototype.tick = function() {
+  var detonator = this;
+  $.ajax({
+    url: '/BOMB/v1/',
+    success: function(data) { detonator.tock(data) },
+    error: function(a,b,c) { console.log("Error", a, b, c); },
```
jmtai: `a`, `b`, and `c` aren't great variable names. What exactly is being returned here?

overlord/overlord.rb
DejaAugustine: It's not actually shorting the wires, it's short-circuiting past the detonation mechanism to directly detonate the bomb, which is why it's a child endpoint of bomb, but not of wires.

overlord/public/javascripts/detonator.js
DejaAugustine: A debug function that should have been refactored to display an error.

overlord/overlord.rb
DejaAugustine: Good point.

overlord/overlord.rb
DejaAugustine: It's completely redundant and unnecessary.

overlord/overlord.rb
jmtai: Ah... gotcha.

overlord/features/detonator.feature
```diff
@@ -0,0 +1,115 @@
+Feature: Bomb Detonator
```
jmtai: With TDD, we test units of software (which usually correspond to a single class in Ruby) but with BDD we are testing the behavior of the software from a user perspective. That means our tests should be organized according to related **behaviors**. In this case, you've got all your tests grouped under a single detonator feature but if we look at the behavior of the site, the user can do a number of different things:
- Arm the bomb
- Disarm the bomb
- Explode the bomb
etc.

jmtai: A benefit of moving related behaviors to separate features is that you can extract common behaviors into `background' blocks: https://github.com/cucumber/cucumber/wiki/Background

overlord/features/support/detonator_helper.rb
```diff
@@ -0,0 +1,21 @@
+module DetonatorHelpers
```
jmtai: Are these actually being used?

overlord/lib/bomb.rb
```diff
@@ -1,10 +1,13 @@
+require_relative "timer"
 require_relative "wire_bundle"
 
 class Bomb
   attr_accessor :timer, :wires
   attr_reader :error, :failed_deactivations, :max_failed_deactivations
 
-  def initialize(activation_code, deactivation_code, max_failed_deactivations = 3)
+  def initialize(activation_code,
```
jmtai: My personal preference is to align the parameter names if you're going to split them onto multiple lines:
```ruby
def initialize(activation_code,
               deactivation_code,
               max_failed_deactivations = 3)
```

overlord/features/detonator.feature
```diff
@@ -0,0 +1,115 @@
...
+     And the countdown should be running
+     And an error should be displayed
+
+  @javascript
+  Scenario: Too Many Invalid Disarm Attempts
+    Given the bomb is armed
+      And the countdown is running
+    When I enter an invalid code
+     And I enter an invalid code
+     And I enter an invalid code
```
jmtai: Instead of repeating yourself three times, you can do something like this:
`When I enter an invalid code 3 times`
and in your step definitions:
```
When(/^I enter an invalid code ([0-9]+) times?$/) do |count|
  [do stuff]
end
```

overlord/features/detonator.feature
```diff
@@ -0,0 +1,115 @@
...
+     And I arm the bomb
+    Then the bomb should be armed
+
+  @javascript
+  Scenario: Invalid Activation Code
+    Given the bomb is disarmed
+    When I enter an invalid code
+     And I arm the bomb
+    Then the bomb should be disarmed
+     And an error should be displayed
```
jmtai: The wording of this test sounds a little weird... "when I enter an invalid code and I arm the bomb, then the bomb should be disarmed". To someone looking at this test without seeing the UI, this doesn't make much sense because it's tied to your particular implementation. Conceptually, it's more like this:  "... when I enter an invalid arming code and I submit the code, the bomb remains disarmed".

overlord/lib/bomb.rb
DejaAugustine: That's my personal preference, as well.  However, I changed the alignment to satisfy rubocop's incessant nagging.

overlord/features/detonator.feature
```diff
@@ -0,0 +1,115 @@
...
+
+  @javascript
+  Scenario: Activate Trigger
+    Given the bomb is armed
+    When I enter a disarm code
+    Then the trigger button should be active
```
jmtai: This one also doesn't make sense to me as a user unfamiliar with the implementation. If the bomb is armed and I enter a disarm code, I'd expect the bomb to be disarmed. Is this a valid or invalid disarm code?

overlord/features/support/detonator_helper.rb
DejaAugustine: Yes.  e.g. https://github.com/DejaAugustine/level_up_exercises/blob/deja/overlord/overlord/features/step_definitions/detonator_steps.rb#L28-L30

overlord/features/detonator.feature
DejaAugustine: Backgrounds!  For some reason background blocks consistently slip my mind.

overlord/features/detonator.feature
DejaAugustine: Hm, good point on these two scenarios. They both could use a refactor.

overlord/features/detonator.feature
DejaAugustine: Good point. Thanks.

overlord/features/detonator.feature
```diff
@@ -0,0 +1,115 @@
...
+    Given the bomb is booted
+    When I open the wire panel
+    Then the wires should be exposed
+
+  @javascript
+  Scenario: Cut a Detonating Wire on an Armed Bomb
+    Given the bomb is armed
+      And the wires are exposed
+    When I cut a detonating wire
+    Then the bomb should have exploded
```
jmtai: Nitpicky comment:  No need to mix tenses here. Since your action is in the present tense, the result can be, too. e.g.
```ruby
When I cut a detonating wire
Then the bomb should explode
```

overlord/features/detonator.feature
jmtai: Another good habit to get into when writing your features is to add a user story block:
e.g.
```cucumber
As a supervillain
I want to arm a bomb
So I can blow stuff up
```
This will give a reader the necessary background to understand what user behavior is being tested.

overlord/features/detonator.feature
```diff
@@ -0,0 +1,115 @@
...
+  Scenario: Start Countdown
+    Given the trigger is active
+    When I press the trigger button
+    Then the countdown should be running
+
+  @javascript
+  Scenario: Stop Countdown
+    Given the countdown is running
+    When I enter a disarm code
+     And I disarm the bomb
```
jmtai: You can tighten these scenarios up by divorcing them from your particular implementation. Your implementation requires that a user enter a valid disarming code in an input field and then click a `Disarm` button, but you can move the implementation specific details into the step definitions and just write the scenario at a high level:
```cucumber
When I disarm the bomb
```

overlord/features/detonator.feature
```diff
@@ -0,0 +1,115 @@
...
+    Then the bomb should be disarmed
```
jmtai: If you make the change suggested for the `When` step, this becomes unnecessary, as the positive statement `When I disarm the bomb` already implies that the bomb was successfully disarmed.

overlord/features/detonator.feature
```diff
@@ -0,0 +1,115 @@
...
+     And the countdown should not be running
+
+  @javascript
+  Scenario: Invalid Disarm Code
+    Given the countdown is running
+    When I enter an invalid code
+     And I disarm the bomb
+    Then the bomb should be armed
```
jmtai: Another confusing scenario here. When you disarm the bomb, why should the bomb be armed?

overlord/features/step_definitions/detonator_steps.rb
```diff
@@ -0,0 +1,108 @@
+Given(/^the bomb is (.*)ed$/) do |mode|
+  visit "/"
+
+  step %(I enter a valid code) if mode == "arm"
+  step %(I #{mode} the bomb) unless mode == "boot"
```
jmtai: Nice code reuse here.

overlord/features/step_definitions/detonator_steps.rb
```diff
@@ -0,0 +1,108 @@
...
+end
+
+Given(/^the wires are exposed$/) do
+  step %(I open the wire panel)
+end
+
+Given(/^the bomb exploded$/) do
+  step %(the bomb is booted)
+  page.execute_script(%($(document).trigger('short')))
+  sleep(3)
```
jmtai: In general, try to avoid using sleep in your tests unless absolutely necessary: https://nulogy.com/articles/sleep-is-for-the-weak
sleep() calls are notoriously unreliable and usually lead to sporadic failures.

overlord/features/step_definitions/detonator_steps.rb
```diff
@@ -0,0 +1,108 @@
...
+end
+
+When(/^the countdown ends$/) do
+  step %(the bomb exploded)
+end
+
+Then(/^the (.*) button should( not)? be active$/) do |field, negate|
+  button = find(:css, "[name=#{field}]")
+  op = negate.nil? ? 'not_to' : 'to'
```
jmtai: nil is a `falsey` value in Ruby, so there's no need for the double negative here:
```
op = negate ? 'to' : 'not_to'
```
Although this is a bit confusing since your negate here is referring to the button being active, but the matcher in the next line is checking if the button is disabled.

overlord/features/step_definitions/detonator_steps.rb
DejaAugustine: Sadly, I've found Capybara's ability to cope with the async interface to be lacking.  Shorting the bomb should set the timer to 1 second and trigger the countdown, so sleeping for 3 seconds gives that code time to run and the results to be processed.

It makes me feel awful inside to use that, and it is a fragile test, but I was struggling to find a more reliable solution for detecting when the short completed using either the selenium or webkit drivers.

overlord/lib/bomb.rb
jmtai: Yeah... go ahead and ignore Rubocop when it comes to alignment stuff like this. :smile:

overlord/features/step_definitions/detonator_steps.rb
DejaAugustine: I agree, this is confusing. I decided to go with more complicated negation logic in the step implementations in order to provide clearer natural language in the features.

Having steps like `should not be disabled` and `should not be active` hampered the reusability of the code (both use "not" but the first tests for the absence of a property, while the second tests for the presence of that same property).

It may actually make the most sense to rewrite `the * button should * be active` to `the * button should * be disabled` to be consistent with the negation logic in `the interface should * be disabled`.

Thoughts?

jmtai: :+1: for rewriting to make things consistent. It may not make future maintainers of your code love you, but at least they won't hate you. :smile:

overlord/public/stylesheets/detonator.css
```diff
@@ -0,0 +1,142 @@
+.control-panel {
+  border: 2px solid #AAA;
+  border-radius: 15px;
+  margin: 10% auto;
+  min-width: 300px;
+  padding: 40px;
+  width: 50%;
```
jmtai: I personally like ordering all the properties in alphabetical order, but I don't think there's a general consensus on this one. Whatever you decide is best, keep it consistent through the file. Some of your classes are ordered alphabetically and some aren't.

overlord/public/stylesheets/detonator.css
```diff
@@ -0,0 +1,142 @@
...
+#trigger {
+  display: block;
+  font-size: 30px;
+  margin: 20px auto;
+  padding: 10px 40px;
+}
+#error {
+  position: absolute;
+}
+form {
```
jmtai: Put a blank line in between all your blocks.

overlord/spec/javascripts/detonator_spec.js
```diff
@@ -0,0 +1,181 @@
...
+  });
+
+  it("displays an error message when an error is received", function() {
+    expect($('#error .message')).not.toContainText("abject failure");
+
+    detonator.tock({ error: "abject failure" });
+    expect($('#error .message')).toContainText("abject failure");
+  });
+
+  it("submits the code when a mode different is clicked", function() {
```
jmtai: Weird phrasing here.

overlord/spec/javascripts/detonator_spec.js
```diff
@@ -0,0 +1,181 @@
...
+  });
+
+  it("activates the bomb when trigger is active and clicked", function() {
+    spyOn(detonator, "activateBomb");
+
+    $('[name=trigger]').prop("disabled", false).click();
+    expect(detonator.activateBomb).toHaveBeenCalled();
+  });
+
+  it("does not activate the bomb when trigger is inactive and clicked", function() {
```
jmtai: Since these tests are all grouped together as unit tests for the detonator, it's a little hard to follow all the different states of the other components being tested. As I go down the list I can see the detonator being `connected/disconnected`, the bomb itself switching between `active/inactive` as well as `armed/disarmed` states and then there's the trigger which can also be `active/inactive`.

Since Jasmine is a BDD framework, you really want to organize your tests into behaviors, much like your Cucumber tests. Even though your Javascript code is contained within a single module, we're not doing unit testing here so there's no reason to glob all your tests into a single file.

overlord/spec/overlord_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+  let(:invalid_code) { "Get outta here" }
+  let(:disarm_code) { "0000" }
+  let(:ghost_wire) { 999 }
+  let(:wire_index) { 1 }
+  let(:short_time) { 1 }
+  let(:status) do
+    get '/BOMB/v1/'
+    parse_response(last_response.body)
+  end
+  let(:armed_bomb) do
```
jmtai: Add a blank line between your let blocks.

overlord/spec/overlord_spec.rb
```diff
@@ -0,0 +1,153 @@
+require_relative "spec_helper"
+require_relative "overlord_helper"
+
+describe Overlord do
+  let(:valid_code) { "1234" }
```
jmtai: Good use of let statements here for constants.

overlord/views/bomb.erb
```diff
@@ -0,0 +1,48 @@
...
+    </div>
+
+    <div class="btn-group" data-toggle="buttons">
+      <label class="btn btn-primary mode disarmed active">
+        <input type="radio" name="mode" value="disarmed" checked="checked"> Disarm
+      </label>
+      <label class="btn btn-primary mode armed">
+        <input type="radio" name="mode" value="armed"> Arm
+      </label>
+    </div>
```
jmtai: From a new user perspective, this was a bit confusing to me, since I normally expect radio buttons to be simple option selectors. The fact that clicking the button actually submitted the codes to the bomb threw me off initially just because the buttons were visually represented as a radio button group.

It was also confusing that after arming the bomb, I had to manually activate it by clicking another button. Looking at it with no prior knowledge of the implementation details, I assumed armed == activated. Since there's no documentation anywhere, I didn't know how the system was structured until I looked at the source code.

overlord/spec/overlord_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+  end
+
+  it "allows accessing the home page" do
+    get '/'
+    expect(last_response).to be_ok
+  end
+
+  describe "/BOMB/v1/" do
+    it "provides the status as valid json" do
+      expect(status).not_to eq(false)
```
jmtai: `be_false`?

overlord/spec/overlord_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+
+    it "provides a clock value" do
+      expect(status).to have_key(:clock)
+    end
+
+    it "provides an error value" do
+      expect(status).to have_key(:error)
+    end
+
+    it "can be posted to" do
```
jmtai: This condition is implicitly tested by your "provides status on post" test, so it's not really necessary.

overlord/spec/overlord_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+    end
+  end
+
+  describe "/BOMB/v1/trigger" do
+    let(:triggered_bomb) do
+      post '/BOMB/v1/trigger'
+      parse_response(last_response.body)
+    end
+
+    it "does nothing with an inactive bomb" do
```
jmtai: The wording here says "does nothing" but your check is looking for a clock so this could probably be reworded to make it a little clearer.

jmtai:
Really nice job! The implementation is solid and cleanly written and I like how much coverage you've got with your tests.

From a design standpoint, I'd recommend reading over the requirements again and thinking about whether or not your implementation meets them. One requirement listed is that the activation/deactivation codes should be configurable on boot. Although "boot" can be interpreted in different ways, there should be a way for the end user to configure these codes without having to modify the source code. The other big thing is to determine what should happen if the page gets reloaded - should the current bomb be discarded?

Read over the comments, refactor a bit (especially your Cucumber tests) and resubmit when ready.

paulghaddad:
@DejaAugustine Really nice job as Josh said. Please take a look at the comments and resubmit a second iteration when ready.

PR #960, opened 2015-08-28
deja/shopping_cart v1.0

DejaAugustine:
I set a timer for 10 minutes and this is the result.

I intentionally kept it to higher level behaviors (much like I did with Overlord), but tried to use consistent syntax that could actually be converted directly into steps.

### DIFF
--- a/shopping_cart/shopping_cart.feature
+++ b/shopping_cart/shopping_cart.feature
@@ -1,0 +1,61 @@
+Feature: Shopping Cart
+  Scenario: Add Item to Cart
+    Given I am on an item page
+    When I add the item to the cart
+    Then the cart should contain the item
+
+  Scenario: Remove Item from Cart
+    Given I am viewing the cart
+      And the cart contains an item
+    When I remove the item from the cart
+    Then the cart should not contain the item
+
+  Scenario: Change Quantity
+    Given I am viewing the cart
+      And the cart contains an item
+    When I set the quantity of the item to 3
+    Then the cart should contain 3 of the item
+
+  Scenario: View Item from Cart
+    Given I am viewing the cart
+      And the cart contains an item
+    When I click the item link
+    Then I should be on the item page
+
+  Scenario: Shipping Estimate
+    Given I am on the checkout page
+    When I enter my address
+     And I request a shipping estimate
+    Then I should see a shipping estimate line item
+
+  Scenario: Add Coupon
+    Given I am on the checkout page
+    When I enter a valid coupon code
+    Then I should see a discount line item
+
+  Scenario: Add Expired Coupon
+    Given I am on the checkout page
+    When I enter an expired coupon code
+    Then I should see an error message
+     And I should not see a discount line item
+
+  Scenario: Add Existing Item to Cart
+    Given I am on an item page
+      And the cart contains 1 of an item
+    When I add the item to the cart
+    Then the cart should contain 2 of the item
+
+  Scenario: Log In With Items Already in Cart
+    Given I am viewing the cart
+      And the cart contains an item
+    When I log in
+     And I had the same item saved
+    Then the cart should contain 2 of the item
+
+  Scenario: Log In With Same Items Already In Cart
+    Given I am viewing the cart
+      And the cart contains an item
+    When I log in
+     And I had a different item saved
+    Then the cart should contain the item
+     And the cart should contain the other item


shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,61 @@
...
+    Given I am viewing the cart
+      And the cart contains an item
+    When I remove the item from the cart
+    Then the cart should not contain the item
+
+  Scenario: Change Quantity
+    Given I am viewing the cart
+      And the cart contains an item
+    When I set the quantity of the item to 3
+    Then the cart should contain 3 of the item
```
paulghaddad: Perhaps rephrase to `Then the cart should contain 3 items`? I'm being super nit-picky here, though. :)

shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,61 @@
...
+
+  Scenario: Shipping Estimate
+    Given I am on the checkout page
+    When I enter my address
+     And I request a shipping estimate
+    Then I should see a shipping estimate line item
+
+  Scenario: Add Coupon
+    Given I am on the checkout page
+    When I enter a valid coupon code
```
paulghaddad: You do a really good job staying at a constant, high-level of abstraction. This `Given` step encapsulates a lot of logic behind it, but this is hidden from the reader. This is exactly how feature steps should be written.

shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,61 @@
...
+  Scenario: Add Expired Coupon
+    Given I am on the checkout page
+    When I enter an expired coupon code
+    Then I should see an error message
+     And I should not see a discount line item
+
+  Scenario: Add Existing Item to Cart
+    Given I am on an item page
+      And the cart contains 1 of an item
+    When I add the item to the cart
```
paulghaddad: Perhaps rephrase in a way to say `When you add an existing item to the cart`?

shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,61 @@
...
+    Then the cart should contain 2 of the item
+
+  Scenario: Log In With Items Already in Cart
+    Given I am viewing the cart
+      And the cart contains an item
+    When I log in
+     And I had the same item saved
```
paulghaddad: I'm a little confused on this one. Are you adding a second item in this step?

shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,61 @@
...
+     And I should not see a discount line item
+
+  Scenario: Add Existing Item to Cart
+    Given I am on an item page
+      And the cart contains 1 of an item
+    When I add the item to the cart
+    Then the cart should contain 2 of the item
+
+  Scenario: Log In With Items Already in Cart
+    Given I am viewing the cart
```
paulghaddad: Perhaps you can create a step that combines lines 49-50 into one step. This can then be reused in two other steps.

`Given I am viewing the cart with 1 item`

shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,61 @@
...
+    When I log in
+     And I had the same item saved
+    Then the cart should contain 2 of the item
+
+  Scenario: Log In With Same Items Already In Cart
+    Given I am viewing the cart
+      And the cart contains an item
+    When I log in
+     And I had a different item saved
+    Then the cart should contain the item
```
paulghaddad: Perhaps combine 60-61 into `Then the cart should contain both items`.

paulghaddad:
@DejaAugustine I think your features are excellent. As I said in a comment, you stay at a consistent, high level of abstraction. There may be a little room to combine steps, as I also point out, but overall, I think you're right on.

Go ahead and add scenarios for happy and bad paths, and you should be good to go!

shopping_cart/shopping_cart.feature
DejaAugustine: I tried to keep a consistent vocabulary for writing the steps.  For example, "of the item" could easily be used to designate a specific item in the tests.

shopping_cart/shopping_cart.feature
DejaAugustine: Again, I'm using a consistent "an item" in given steps and "the item" in when/then steps to refer to a specific item.

shopping_cart/shopping_cart.feature
DejaAugustine: This scenario was a little awkward given the consistent vocabulary I was striving for. I'm not in love with it, but with the 10-minute stricture I didn't dawdle too long.  What would you think of rewriting line 52 to read `And the item was already in my cart`?  Would that be too abstract?

shopping_cart/shopping_cart.feature
DejaAugustine: Again, "an item" refers to a specific testable item in given steps, so to remain consistent it would likely need to be written as `I am viewing the cart, which contains an item`.

That said, I was also envisioning a scenario where it might be necessary to view an empty cart. I suppose that I could write an `I am viewing an empty cart` step and just reference the step from the `I am viewing a cart, which...` to solve that particular problem, though.

shopping_cart/shopping_cart.feature
paulghaddad: I see...the only thing from my vantage point is that it is not immediately clear that you're adding an existing item to the cart.

shopping_cart/shopping_cart.feature
paulghaddad: Hmmm...this one is a tough call. See if you restructure this scenario for additional clarity.

shopping_cart/shopping_cart.feature
DejaAugustine: The purpose for this was favoring cleanliness and consistency of the implementation over plain English legibility.

`the cart should contain both items` is contextual to previous steps; it has to know what items had previously been added to check if they are both in the cart. This means it either has to be hard-coded in the implementation (that step can only be used in scenarios with that exact setup), or the implementation would have to use shared variables to track the cart (basically a parallel shopping cart in the test rig)

`the cart should contain (.*)` however, can be fully atomic. It doesn't need to know about anything that's happened in any other steps in the scenario giving it far greater simplicity and utility.

shopping_cart/shopping_cart.feature
DejaAugustine: Hm, would it be more clear if it read `And the cart contains 1 of the item` instead?

PR #961, opened 2015-08-30
Lets Be Rational

zsyed1:
Removed HTTP component
Removed test application
Updated ReadMe


lets_be_rational/fraction.rb
```diff
@@ -0,0 +1,67 @@
+require "net/http"
```
paulghaddad: We should be able to delete line 1 now.

lets_be_rational/fraction.rb
```diff
@@ -0,0 +1,67 @@
...
+  def to_s
+    "#{num}/#{den}"
+  end
+
+  def from_json(body)
+    parsed = JSON.parse(body)
+    set(parsed["num"], parsed["den"])
+  end
+
+  def load(url)
```
paulghaddad: We should be able to get rid of this method, too.

lets_be_rational/fraction.rb
```diff
@@ -0,0 +1,67 @@
...
+    http = Net::HTTP.new(uri.host, uri.port)
+    request = Net::HTTP::Get.new(uri.request_uri)
+    response = http.request(request)
+    from_json(response.body)
+  end
+
+  private
+
+  def set(num, den)
+    @num = num
```
paulghaddad: What do you think about moving the assignment for `@num` and `@den` to the `initialize` method. Until I saw the `set` method down here, I was unsure how the instance variables were getting set. Then the `initialize` method could look like the following:

```
def initialize(num, den)
    @num = num
    @den = den
    set(num, den)
 end
```

I think it would make it a little more clear.

lets_be_rational/fraction.rb
```diff
@@ -0,0 +1,67 @@
+require "net/http"
+require "json"
+
+class Fraction
+  attr_accessor :num, :den
+
+  def initialize(num, den)
+    set(num, den)
```
paulghaddad: What are your thoughts about renaming this method. In my opinion, it is a little too similar to the data structure. Perhaps `fraction`?

lets_be_rational/fraction.rb
```diff
@@ -0,0 +1,67 @@
+require "net/http"
+require "json"
+
+class Fraction
+  attr_accessor :num, :den
+
+  def initialize(num, den)
```
paulghaddad: We could introduce them to keyword arguments here. I think this would be a good spot:

`def initialize(num:, den:)`

lets_be_rational/fraction.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def normalize
+    make_denominator_positive
+    reduce_fraction
+  end
+
+  def reduce_fraction
+    divisor = @num.abs.gcd(@den)
+    @num /= divisor
```
paulghaddad: What are your thoughts on calling the accessor method in the instance methods, such as here, instead of using the `@` sign?

paulghaddad:
@zsyed1 I think this is pretty good. It is simple and sticks to the basics. Adding the functionality you specify with TDD shouldn't be challenging. Let me know what you think about my comments. I definitely like it though. We can send it over to Dan and Joe shortly.

zsyed1:
Yeah, I accidentally left that `load` method in there.
About the style changes, we can change them, or just leave them to the learner to fix up - kind of like the `Supportive` exercise. Merely using active-support wasn't enough. The reviewers made sure the code was fixed up as well.

lets_be_rational/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+Your task is to complete the implementation of `Fraction` by doing the following:
+
+1. Implement additional functionality
+  1. Add other operators: subtraction(`-`), multiplication(`*`), division(`/`)
+  2. Allow for zero, one or two arguments in the initializer.
+  3. It should raise an exception if initialized with non-integer arguments or if the denominator is zero.
+  4. Add `to_json` that conforms with the `from_json` method already implemented
+  5. Add `to_f`
+2. Implement tests
```
djkotowski: Since we're pushing TDD, this should be the first step.

lets_be_rational/fraction.rb
djkotowski: I'm not sure keyword args in a fraction are the best.

Also, we'll want to remove abbreviation to improve clarity.

lets_be_rational/fraction.rb
```diff
@@ -0,0 +1,67 @@
...
+    @num = num
+    @den = den
+    validate
+    normalize
+  end
+
+  def validate
+    raise "Can't have zero denominator" if @den == 0
+    raise "Numerator must be an integer" unless @num.is_a? Integer
+    raise "Denominator must be an integer" unless @den.is_a? Integer
```
djkotowski: If we validate these before the assignment to instance variables, we could raise `ArgumentError` for these.

lets_be_rational/fraction.rb
djkotowski: Definitely.  This is too much and violates SRP.

lets_be_rational/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Requirements
+
+Your task is to complete the implementation of `Fraction` by doing the following:
+
+1. Implement additional functionality
+  1. Add other operators: subtraction(`-`), multiplication(`*`), division(`/`)
+  2. Allow for zero, one or two arguments in the initializer.
+  3. It should raise an exception if initialized with non-integer arguments or if the denominator is zero.
+  4. Add `to_json` that conforms with the `from_json` method already implemented
```
djkotowski: I'm not a big fan of having a JSON representation.  Maybe a `.parse` class method to convert String -> Fraction?  Then we can get rid of the JSON dependency.

djkotowski:
One thing to keep in mind is they're no longer getting unit testing practice on Overlord, so we'll want this to be substantial enough to let them get good practice with TDD.

zsyed1:
@djkotowski @paulghaddad I recommend forking this branch, make the changes you want, and resubmit - that will be a lot faster than PR dialogue. You can make the comments here, but I recommend _Operate As Owner_ to move this along.

paulghaddad:
@zsyed1 Will close and resubmit with a PR for this repo and Level Up. Thanks for your hard work on this!

PR #962, opened 2015-08-31
deja/shopping_cart v2.0

DejaAugustine:
Incorporating feedback and adding additional sad/bad paths.


paulghaddad:
@DejaAugustine This looks good! The only possible comment could be to break up this feature file into smaller, more focused files. Regardless, this is very good!

PR #964, opened 2015-08-31
Yadda Part 1 - V2

mcdaniel67:
Fixed what was recommended to be fixed in V1.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+.idea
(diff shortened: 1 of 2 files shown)


paulghaddad:
@mcdaniel67 We reviewed this together. You refactored well from V1: removed foreign reference for Users; added a check for the rating number; and added time zone for your created_on and updated_on fields. Nice work; you can move on!

PR #965, opened 2015-08-31
DinoDex v1

Manbearpixel:


dino_catalog/print_test.rb
```diff
@@ -0,0 +1,35 @@
+require_relative "DinoDex"
+
+dino_dex   = DinoDex.new
```
dustinhansen: You might already know this, but these warnings that come from houndci stem from rubocop. I personally use this tool to lint in real time with rubocop and it helped to eliminate lots of these annoyances: https://github.com/SublimeLinter/SublimeLinter-rubocop

dino_catalog/DinoDex.rb
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  def initialize
+    @dinosaurs = []
+  end
+
+  def import_csv_file(csv_file_path)
+    @csv_rows = []
+    CSV.foreach(csv_file_path, {headers: true, header_converters: :symbol}) do |row|
+      normalize_headers(row)
```
dustinhansen: normalize_headers should only run once. There's a few ways we could go about this. One way might be:

csv_contents = CSV.read(csv_file_path)
@csv_rows << normalize_headers(csv_contents.shift)
csv_contents.each do |row|

dino_catalog/DinoDex.rb
```diff
@@ -0,0 +1,93 @@
...
+  def normalize_headers(csv_row)
+    HEADER_NORMALIZE.each do |old_header, new_header|
+      next unless csv_row.header?(old_header.to_sym)
+      csv_row.each do |row_header, row_value|
+        next unless row_header.to_s == old_header
+        csv_row.delete(old_header.to_sym)
+        csv_row[new_header.to_sym] = row_value
+      end
+    end
+  end
```
dustinhansen: Methods like this one that are only referenced from within its own class should be marked private or protected.

dino_catalog/DinoDex.rb
```diff
@@ -0,0 +1,93 @@
...
+  def import_csv_file(csv_file_path)
+    @csv_rows = []
+    CSV.foreach(csv_file_path, {headers: true, header_converters: :symbol}) do |row|
+      normalize_headers(row)
+      @csv_rows << row
+    end
+
+    @csv_rows.each_index do |index|
+      dinosaurs << generate_dinosaur(@csv_rows[index])
+    end
```
dustinhansen: This is mostly a style issue, but when you have a code block that's one line it's considered a best practice to use alernate syntax:

@csv_rows.each_index { |index| dinosaurs << generate_dinosaur(@csv_rows[index]) }

dino_catalog/DinoDex.rb
```diff
@@ -0,0 +1,93 @@
...
+
+class DinoDex
+
+  attr_accessor :dinosaurs
+
+  HEADER_NORMALIZE = {
+    "genus"         => "name",
+    "carnivore"     => "diet",
+    "weight_in_lbs" => "weight",
+  }
```
dustinhansen: Symbols are typically preferred for hashes. It's better for memory usage. This thread does a pretty good job of explaining: http://stackoverflow.com/questions/8189416/why-use-symbols-as-hash-keys-in-ruby

dino_catalog/DinoDex.rb
```diff
@@ -0,0 +1,93 @@
...
+    dinosaur = Dinosaur.new
+    dinosaur.diet         = csv_row[:diet]
+    dinosaur.name         = csv_row[:name]
+    dinosaur.period       = csv_row[:period]
+    dinosaur.weight       = csv_row[:weight]
+    dinosaur.walk_type    = csv_row[:walking]
+    dinosaur.location     = csv_row[:continent]
+    dinosaur.description  = csv_row[:description]
+    dinosaur
+  end
```
dustinhansen: This could be shortened a bit by using assign_attributes.

dino_catalog/DinoDex.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  def select_smaller_than(value)
+    DinoCatalog.new(dinosaurs.select { |dinosaur| dinosaur.smaller_than?(value) })
+  end
+
+  def select_hash(args)
+    catalog = DinoCatalog.new(dinosaurs)
+    args.each do |key, value|
+      catalog.filter(key, value)
+    end
```
dustinhansen: Another example of where we can lop off a few more lines:

args.each { |key, value| catalog.filter(key, value) }

dino_catalog/Dinosaur.rb
```diff
@@ -0,0 +1,54 @@
...
+    @diet = str
+  end
+
+  def has?(key)
+    instance_variables.include?("@#{key}".to_sym) && !empty?(key)
+  end
+
+  def empty?(key)
+    instance_variable_get("@#{key}".to_sym).nil?
+  end
```
dustinhansen: Is this method necessary? I think we could just use .nil?

dino = Dinosaur.new
dino[:name].nil? #should return true or false based on whether it exists or not

dino_catalog/FileFinder.rb
```diff
@@ -0,0 +1,38 @@
...
+  def directory_path=(path)
+    return puts ERROR_NO_DIRECTORY if Dir.exists?(path) == false
+    Dir.chdir(path)
+    @directory_path = set_proper_path(Dir.pwd)
+    return true
+  end
+
+  def exists?(path)
+    Dir.exists?(path)
+  end
```
dustinhansen: This method seems unnecessary. We're not doing anything differently in this method than if we just called Dir.exists?(path) directly.

dino_catalog/cli_test.rb
```diff
@@ -0,0 +1,217 @@
...
+    header_line(@header_line_length, @header_line_padding, "Version 4.0.5, Alpha E")
+    header_line(@header_line_length)
+  end
+
+  def header_line(line_length = 40, line_padding = 0, line_str = "")
+    return puts "#" * line_length if line_str == ""
+    empty_space = line_length - line_str.length - (line_padding*2)
+    return puts "#" * line_length if empty_space < 0
+    line_str = "##" + ( " " * (empty_space/2)) + line_str + (" " * (empty_space/2))
+    unless empty_space%2 == 0
```
dustinhansen: This will save you some lines. :)

dino_catalog/cli_test.rb
```diff
@@ -0,0 +1,217 @@
...
+    print_header
+    gets_option
+  end
+
+  def print_header
+    header_line(@header_line_length)
+    header_line(@header_line_length, @header_line_padding, "Jurassic Park, DinoDex Interface")
+    header_line(@header_line_length, @header_line_padding, "Version 4.0.5, Alpha E")
+    header_line(@header_line_length)
+  end
```
dustinhansen: It appears we could privatize this method.

dustinhansen:
Overall, great job. I like the design of this solution and each class is organized very well. I'd like to see how many lines we can shave off with a refactor. Take a look at my notes and apply those tips to try and slim down the code a bit more.

dino_catalog/DinoDex.rb
```diff
@@ -0,0 +1,93 @@
+require_relative "FileFinder"
```
paulghaddad: Ruby files should be named with snake case: `file_finder.rb`. So this `require_relative` would be: `file_finder`.

dino_catalog/DinoDex.rb
```diff
@@ -0,0 +1,93 @@
...
+    "genus"         => "name",
+    "carnivore"     => "diet",
+    "weight_in_lbs" => "weight",
+  }
+
+  def initialize
+    @dinosaurs = []
+  end
+
+  def import_csv_file(csv_file_path)
```
paulghaddad: This method is doing two things, so I'd extract it into two separate methods, each one with a single responsibility.

dino_catalog/DinoDex.rb
```diff
@@ -0,0 +1,93 @@
...
+    dinosaur.period       = csv_row[:period]
+    dinosaur.weight       = csv_row[:weight]
+    dinosaur.walk_type    = csv_row[:walking]
+    dinosaur.location     = csv_row[:continent]
+    dinosaur.description  = csv_row[:description]
+    dinosaur
+  end
+
+  def empty?
+    return dinosaurs.length == 0
```
paulghaddad: You don't need this `return` as it is the last line of the method.

dino_catalog/Dinosaur.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def smaller_than?(size)
+    return false unless has?("weight")
+    !larger_than?(size)
+  end
+    
+  def print_facts
+    instance_variables.each do |instance_var|
+      unless self.instance_variable_get(("#{instance_var}")).nil?
```
paulghaddad: You can just use `if self.instance_variable_get(("#{instance_var}"))`

paulghaddad:
@Manbearpixel Please take a look at Dustin's comments, refactor and submit a second iteration.

PR #966, opened 2015-08-31
Robot Name v2

Manbearpixel:


robot_name/RobotRegistry.rb
```diff
@@ -0,0 +1,22 @@
+class RobotRegistry
```
paulghaddad: You should name your file, `robot_registry.rb`

robot_name/Robot.rb
```diff
@@ -0,0 +1,25 @@
+require_relative "RobotRegistry"
```
paulghaddad: By naming your Ruby files correctly (see later comment), this would be `require_relative robot_registry`

robot_name/Robot.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  private
+
+  def generate_robot_name
+    "#{generate_char(2)}#{generate_num(3)}"
+  end
+
+  def generate_char(length=1)
+    length.times.map { ('A'..'Z').to_a.sample }.join("")
```
paulghaddad: The method `String#concat` could simplify this. Try it out.

robot_name/Robot.rb
```diff
@@ -0,0 +1,25 @@
+require_relative "RobotRegistry"
+
+class Robot
+  
+  attr_accessor :name
+
+  def initialize(args = {})
+    @name = (args[:name_generator] && args[:name_generator].call) || generate_robot_name
```
paulghaddad: This is confusing to me. I would suggest the part before the `||` to be a well named method that basically creates a name if supplied with a name generator.

robot_name/Robot.rb
paulghaddad: This idiom may work better instead of converting the range to an array: `[*'A'..'Z']'. Try it out in IRB!

robot_name/RobotRegistry.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  @@registry
```
paulghaddad: Try to refactor this class to avoid the use of the class variable. You should be able to create an instance of a registry that holds the names of the robot names.

robot_name/RobotRegistry.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def self.add_robot_to_registry(name)
+    @@registry ||= []
+    raise NameCollisionError, "Improper robot name!" unless proper_name?(name)
+    @@registry << name unless robot_registered?(name)
+  end
+
+  def self.proper_name?(name)
+    name =~ /[[:alpha:]]{2}[[:digit:]]{3}/
```
paulghaddad: Consider extracting this regex to a well-named constant, perhaps `VALID_ROBOT_NAME`.

robot_name/RobotRegistry.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  @@registry
+
+  def self.get_registered_robots
+    @@registry
+  end
+
+  def self.add_robot_to_registry(name)
+    @@registry ||= []
+    raise NameCollisionError, "Improper robot name!" unless proper_name?(name)
```
paulghaddad: I think a better error message is "Invalid name. Name already exists.". Or something along those lines.

robot_name/RobotRegistry.rb
```diff
@@ -0,0 +1,22 @@
...
+    @@registry << name unless robot_registered?(name)
```
paulghaddad: Good use of the `<<` method.

paulghaddad:
@Manbearpixel This is a nice implementation. I really liked your Robot class. I do suggest refactoring the `Registry` class to avoid the class variable. I'd like you to refactor this for the third iteration, and then you should be good to go! Nice work!

PR #969, opened 2015-09-01
Overlord - v2

oharace:


overlord/public/scripts/overlord.js
```diff
@@ -0,0 +1,163 @@
...
+function activatedUpdates(jsonObj) {
+  deactivated = false;
+  var bombState = "<span style=\"color:red\">Activated!</span>";
+  var deactivateLabel = "To deactivate, enter your deactivation code: ";
+  document.getElementById("bombState").innerHTML = bombState;
+  document.getElementById("formLabel").innerHTML = deactivateLabel;
+  if (jsonObj.attempts_remaining <= 0) {
+    explodedUpdates();
+  } else if (jsonObj.attempts_remaining < 3) {
+    var attemptsRemaining = "(" + jsonObj.attempts_remaining;
```
oharace: That is from a JSON string. It is not a variable.

overlord/bomb.rb
```diff
@@ -0,0 +1,64 @@
+class Bomb
+  TWO_MINUTES = 120
+  attr_reader :state, :attempts_remaining, :timer_end
+
+  def initialize(custom_activation_code = nil, custom_deactivation_code = nil)
+    @default_activation_code = "1234"
+    @default_deactivation_code = "0000"
+
+    @custom_activation_code = custom_activation_code
+    @custom_deactivation_code = custom_deactivation_code
```
johnmcconnell: More code isn't always better code. You do not need to store both a `custom_[de]activation_code` and a `default_[de]activation_code`
In the initializer
```ruby
@activation_code = activation_code || "1234"
@deactivation_code = deactivation_code || "000"
```

overlord/features/manage_bomb.feature
```diff
@@ -0,0 +1,56 @@
...
+  Scenario: Activate Already-Activated Bomb
+    Given the bomb is activated
+    When I enter an activation code
+    Then nothing happens
+
+  @javascript
+  Scenario: Enter Incorrect Activation Code
+    Given the bomb is deactivated
+    When I enter an incorrect activation code
+    Then the bomb should be deactivated
```
johnmcconnell: Personal preference, you are not asserting something *should* happen, you are asserting something *does* happen. I would remove should and make it present tense.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,106 @@
+Given(/^I am on the bomb page$/) do
+  visit '/'
+  sleep(0.5)
```
johnmcconnell: There better be a good explanation like this. A system which uses sleep in the unit test does not scale well. Think about it.

When the number of functions using sleep is < 10. Okay, we have an extra 5 seconds on testing. Now, years go on, things get bigger and the number of Functions using sleep is > 1000. Boom! eight minutes to run unit tests.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,106 @@
...
+  fill_in('code', with: '1111')
+  click_button('Submit')
+  sleep(0.3)
+  expect(page).to have_content("(2 attempts remaining!)")
+end
+
+When(/^I enter an incorrect deactivation code three times$/) do
+  fill_in('code', with: '1111')
+  click_button('Submit')
+  sleep(0.3)
```
johnmcconnell: No, no, no!

overlord/overlord.rb
```diff
@@ -1,14 +1,88 @@
...
+    deactivate_bomb
+  else
+    invalid_deactivate_attempt
+  end
+end
+
+def create_response(code)
+  string = "{\"status\":#{code}, \"state\":\"#{session[:bomb].state}\","
+  string << "\"attempts_remaining\":#{session[:bomb].attempts_remaining},"
+  string << "\"timer_end\":#{session[:bomb].timer_end}}"
```
johnmcconnell: Wow, isn't that some nice fault tolerant javascript serializing...

How about?
```ruby
require 'json'
{ "anything" => "I want" }.to_json # => "{\"anything\":\"I want\"}"

Comment (johnmcconnell) [inline overlord/overlord.rb]: Example escape characters
```ruby
{ 'hello \ test' => 2 }.to_json
# => "{\"hello \\\\ test\":2}"
```

overlord/public/scripts/overlord.js
```diff
@@ -0,0 +1,163 @@
...
+  var code = document.forms[0].elements[0].value;
+  var json = {"code":code};
+  $.ajax({
+    url: "/",
+    type: "POST",
+    data: json,
+    success: function(data) {
+      var jsonData = JSON.parse(data);
+      analyzeData(jsonData);
+      $("#codeForm")[0].reset();
```
johnmcconnell: Okay so you use `jQuery` here...

overlord/public/scripts/overlord.js
```diff
@@ -0,0 +1,163 @@
...
+      hideWarning();
+    }
+  } else { // 400 - not numerical format
+    showInputWarning();
+  }
+}
+
+function activatedUpdates(jsonObj) {
+  deactivated = false;
+  var bombState = "<span style=\"color:red\">Activated!</span>";
```
johnmcconnell: But you do not use `jQuery` here

overlord/public/scripts/overlord.js
```diff
@@ -0,0 +1,163 @@
...
+  var deactivateLabel = "To deactivate, enter your deactivation code: ";
+  document.getElementById("bombState").innerHTML = bombState;
```
johnmcconnell: And you do not use `jQuery` here

overlord/bomb.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    @state = "Deactivated"
+    @timer_end = 0
+    @attempts_remaining = 3
+  end
+
+  def activate
+    @state = "Activated"
```
johnmcconnell: Why are you letting any outside caller activate the bomb with out checking an activation code?
Your controller will be a lot smaller, for example: https://github.com/jmmastey/level_up_exercises/pull/969/files#diff-6a91898037a35198546f25decf722dbbR33

You are the forcing the caller to know business logic about the bomb. I would rewrite the method as.
```ruby
def activate(code)
  valid = correct_activation_code?(code) && @state == "Activated"
  if valid
    @state = "Activated"
    start_timer
    @attempts_remaining = 3
  end
end
```

overlord/overlord.rb
```diff
@@ -1,14 +1,88 @@
...
+  end
+end
+
+def attempt_activation(code)
+  return create_response(400) unless numeric_input?(code)
+  if session[:bomb].correct_activation_code?(code)
+    activate_bomb
+  else
+    create_response(401)
+  end
```
johnmcconnell: If you rewrote your bomb logic like above this becomes as simple as:
```ruby
if session[:bomb].activate(code)
  create_response(200)
else
  create_response(401)
end
```

johnmcconnell: There is no need to create another helper method in the controller

overlord/bomb.rb
```diff
@@ -0,0 +1,64 @@
...
+  def correct_activation_code?(code)
+    default_activation_code?(code) || custom_activation_code?(code)
+  end
+
+  def correct_deactivation_code?(code)
+    default_deactivation_code?(code) || custom_deactivation_code?(code)
+  end
+
+  def decrement_attempt
+    @attempts_remaining -= 1
```
johnmcconnell: You really do not need a method for something like this

overlord/bomb.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def activate
+    @state = "Activated"
+    start_timer
+    @attempts_remaining = 3
+  end
+
+  def deactivate
+    @state = "Deactivated"
```
johnmcconnell: And here:
```ruby
def deactivate
  if @state == "Activated"
    if correct_deactivation_code?(code)
      @state == "Deactivated"
    else
       @attempts_remaining -= 1
    end
  end
end
```

overlord/overlord.rb
```diff
@@ -1,14 +1,88 @@
...
+def deactivate_bomb
+  session[:bomb].deactivate
+  create_response(200)
+end
+
+def timer_ran_out?
+  session[:bomb].timer_end - Time.now.to_i <= 0
+end
+
+def invalid_deactivate_attempt
```
johnmcconnell: This logic belongs in the model not the controller

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      expect(@bomb.state).to eq("Deactivated")
+    end
+
+    it "can be exploded" do
+      @bomb.explode
+      expect(@bomb.state).to eq("Exploded")
+    end
+
+    it "accepts the default deactivation code" do
+      expect(@bomb.correct_deactivation_code?("0000")).to be true
```
johnmcconnell: This and the one below are the only tests that actually test any business logic.

overlord/spec/bomb_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      expect(@bomb.state).to eq("Activated")
+      expect(@bomb.attempts_remaining).to eq(3)
+      expect(@bomb.timer_end).to be <= (time + 125)
+    end
+  end
+
+  context "when activated" do
+    it "can be deactivated" do
+      @bomb.activate
+      expect(@bomb.state).to eq("Activated")
```
johnmcconnell: These tests are the same as testing setters work. That is not good.

```ruby
x.a = 3
expect(x.a).to eq 3
```

johnmcconnell:
To-Do:
1. Try to remove sleeps. I understand if cucumber does not work will with AJAX but try to find a better way. There are javascript drivers such as selenium and poltergeist.
2. Move the logic out of the controller and into the bomb
3. Remove unnecessary tests. More code is not always better.

overlord/features/manage_bomb.feature
```diff
@@ -0,0 +1,56 @@
...
+    And the timer should be stopped
+
+  @javascript
+  Scenario: Enter Incorrect Deactivation Code 3 Times
+    Given the bomb is activated
+    When I enter an incorrect deactivation code three times
+    Then the bomb should explode
+    And the buttons should be disabled
```
paulghaddad: Is it absolutely necessary to have the steps on lines 41 and 42? I'd argue your test is checking to see if the bomb explodes, which is covered by line 40. Second, if these are steps are necessary, and they always occur after the bomb explodes, then just place them in the step definition on line 40. This will make your features more concise and easier to read.

paulghaddad:
@oharace Nice overall work on this. Your Cucumber tests are overall in good shape, although they could use consolidation in some spots as noted. Please take a look at John's comments, and submit a third iteration. You should be good to go after that.

PR #971, opened 2015-09-01
Robot Name v3

Manbearpixel:
Changed class names, added better exception handling, refactored Registry class to not incorporate instance variable


robot_name/print_test.rb
```diff
@@ -0,0 +1,20 @@
+require_relative "Robot"
```
paulghaddad: The Ruby convention is to name files lower-snake case. So you should use `robot.rb`, which would mean this line is `require_relative "robot"`.

robot_name/print_test.rb
```diff
@@ -0,0 +1,20 @@
...
+
+registry  = RobotRegistry.new
```
paulghaddad: Consider restructuring your `Robot` class to create a new `Registry` object.

paulghaddad:
@Manbearpixel Discussed that a pairing would be good, and a resubmit after that.

PR #972, opened 2015-09-01
Robot Name v4

Manbearpixel:
Refactored Robot class to hold class variable of a Registry which will be used for all Robots until changed


robot_name/Robot.rb
```diff
@@ -0,0 +1,51 @@
+require_relative "errors"
+require_relative "robot_registry"
+
+class Robot
+  attr_reader :name
+
+  ERROR_INVALID_REGISTRY = "Registry is invalid. Must be type RobotRegistry."
```
paulghaddad: Good use of a constant!

robot_name/Robot.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  class << self
+    attr_reader :registry
+
+    def registry=(input_registry)
+      raise ArgumentError, ERROR_INVALID_REGISTRY unless valid_registry?(input_registry)
```
paulghaddad: I like this.

robot_name/Robot.rb
```diff
@@ -0,0 +1,51 @@
+require_relative "errors"
+require_relative "robot_registry"
+
+class Robot
+  attr_reader :name
+
+  ERROR_INVALID_REGISTRY = "Registry is invalid. Must be type RobotRegistry."
+
+  class << self
```
paulghaddad: It's generally better to create class methods with `self.method_name` instead of using `class << self`. But it's the same thing.

robot_name/Robot.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def self.valid_registry?(input_registry)
+    input_registry.instance_of?(RobotRegistry)
+  end
+
+  private
+
+  def parse_name_generator(name_generator)
+    return false if name_generator.nil?
```
paulghaddad: An alternative syntax for this check is `unless name_generator`.

robot_name/errors.rb
```diff
@@ -0,0 +1,3 @@
+class InvalidNameError < Exception; end
```
paulghaddad: Excellent extraction of your errors into their own classes/file.

robot_name/robot_registry.rb
```diff
@@ -0,0 +1,30 @@
+require_relative "errors"
+
+class RobotRegistry
+  attr_accessor :robots
+
+  ERROR_INVALID_NAME  = "Invalid robot name."
+  ERROR_ROBOT_EXISTS  = "Robot name already registered."
```
paulghaddad: +1

robot_name/robot_registry.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  VALID_ROBOT_NAME_REGEX = /[[:alpha:]]{2}[[:digit:]]{3}/
```
paulghaddad: Great use of a constant.

robot_name/robot_registry.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def initialize
+    @robots = []
```
paulghaddad: Consider using `robot_name`, or another variable name, as you are technically adding the robot's name to the registry, not the robot object itself.

robot_name/Robot.rb
```diff
@@ -0,0 +1,51 @@
...
+    attr_reader :registry
+
+    def registry=(input_registry)
+      raise ArgumentError, ERROR_INVALID_REGISTRY unless valid_registry?(input_registry)
+      @registry = input_registry
+    end
+  end
+
+  def initialize(args = {})
+    self.class.registry ||= args[:use_registry] if self.class.valid_registry?(args[:use_registry])
```
paulghaddad: i'm not entirely sure you need `self.class` here. Does `self.registry` work? Let me know; I am curious.

paulghaddad:
@Manbearpixel I think you did a great job on this exercise. You used some good patterns, used constants, had well-defined classes with one job, and had easy to understand methods. All around excellent work. Keep up the nice work! You're good on this exercise.

robot_name/Robot.rb
Manbearpixel: Actually needed to define it with `self.class` to work properly. Ran into some odd issues yesterday implementing this new idea for Robot to contain a class variable of the Registry it is associated with when I was pairing with Deja.

PR #974, opened 2015-09-04
Shopping Cart - v2

oharace:


shopping_cart/features/manage_cart.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Manage Cart
+  In order to not lose any items
+  As a customer
+  I want to save my items if I log in or out
+
+  Background:
```
ochagata: You don't need to give a background in a feature file if you only have 1 scenario in it.
Use backgrounds in feature files where there are multiple scenarios with the same setup

shopping_cart/features/manage_cart.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Manage Cart
```
ochagata: This could be turned into a feature describing logging in/out since it has little to do with actually managing a cart

shopping_cart/features/manage_shipping.feature
```diff
@@ -0,0 +1,15 @@
+Feature: Manage Shipping
```
ochagata: This one could use a few more scenarios to make sure that different destinations are getting different rates.

paulghaddad:
@oharace Nice work, Chris. You're good to move on.

PR #975, opened 2015-09-04
Cyoa V2

mcdaniel67:
Second iteration of my CYOA.


spotigraph/app/models/user.rb
```diff
@@ -0,0 +1,6 @@
+class User < ActiveRecord::Base
+  # Include default devise modules. Others available are:
```
paulghaddad: Nice job adding Devise.

paulghaddad:
@mcdaniel67 As we discussed, the primary feature to add was user authentication. You did a good job of this and are good to move on.

PR #977, opened 2015-09-07
Yadda 2 3 & 4 V1

mcdaniel67:
Parts 2, 3, and 4 for yadda.


yadda/Database.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+  end
+
+  def random_id(table, depth = 0)
+    @db.with do |conn|
+      conn.exec("SELECT * FROM #{table} ORDER BY RANDOM() LIMIT 1").first['id']
+    end
+  rescue
+    rescue_random_id_failure(table, depth)
+  end
```
ssapra: This seems somewhat complicated. An easier method could be to pick a random id between 1 and the constants you have in data_generation.rb.

yadda/Database.rb
```diff
@@ -0,0 +1,57 @@
+require 'pg'
+require 'connection_pool'
+require 'colorize'
```
ssapra: Are you using colorize in this file?

yadda/data_generation.rb
```diff
@@ -0,0 +1,127 @@
...
+RATINGS_COUNT = 10_000
+
+def insert_fake_user
+  @database.insert(
+    'users',
+    name: Faker::Name.name[0, MAX_USER_NAME],
+    email: Faker::Internet.email[0, MAX_USER_EMAIL],
+    birthday: Faker::Time.between(DateTime.now - 30, DateTime.now),
+    password: Faker::Internet.password(8, 49)[0, MAX_USER_PASSWORD],
+    modified_by: Faker::Internet.user_name[0, MAX_CREATED],
```
ssapra: Technically should use MAX_MODIFIED. Is there no created_by column for users?

yadda/data_generation.rb
```diff
@@ -0,0 +1,127 @@
...
+  @database.insert(
+    'ratings',
+    user_id: @database.random_id('users'),
+    beer_id: @database.random_id('beers'),
+    look: rand(1..10),
+    smell: rand(1..10),
+    taste: rand(1..10),
+    overall: rand(1..10),
+    modified_by: Faker::Internet.user_name,
+    created_by: Faker::Internet.user_name,
```
ssapra: Missing the [0, MAX_] for these columns

yadda/data_generation.rb
```diff
@@ -0,0 +1,127 @@
...
+
+def percent(count, multiplier)
+  ((count * 100) / (count * multiplier)).to_i
+end
+
+def function_to_s(function)
+  name = function.to_s.split('_').last
+  return 'breweries' if name == 'brewery'
+  name << 's'
+end
```
ssapra: This looks brittle to me. If a new table like addresses was added, this would get messy. I would change the fake method to take in the table name as the first argument and then you can add what you need to the argument to call the right insert function.

```ruby
def fake(table, count, multiplier)
  ...
  count.times { send("insert_fake_#{table}") }
  ...
end
```

yadda/data_generation.rb
```diff
@@ -0,0 +1,127 @@
...
+
+def fake_it
+  @threads = []
+  fake(:insert_fake_user, USER_COUNT, THREAD_MULTIPLIER)
+  fake(:insert_fake_brewery, BREWERIES_COUNT, THREAD_MULTIPLIER)
+  fake(:insert_fake_style, STYLES_COUNT, THREAD_MULTIPLIER)
+  fake(:insert_fake_beer, BEERS_COUNT, THREAD_MULTIPLIER)
+  fake(:insert_fake_rating, RATINGS_COUNT, THREAD_MULTIPLIER)
+  @threads.each(&:join)
```
ssapra: Using threads is awesome to make this performant!

yadda/sql/all.sql
```diff
@@ -0,0 +1,13 @@
...
+SELECT * FROM breweries;
+SELECT * FROM styles;
+SELECT * FROM beers;
+SELECT * FROM ratings;
+
+DELETE FROM beers;
+DELETE FROM users;
+DELETE FROM breweries;
+DELETE FROM styles;
+DELETE FROM ratings;
```
ssapra: What's the purpose of this file?

yadda/sql/views.sql
```diff
@@ -0,0 +1,41 @@
...
+
+-- Top beers.
+CREATE OR REPLACE VIEW top_beers AS
+  SELECT
+    breweries.id AS brewery_id,
+    breweries.name AS brewery_name,
+    beers.name AS beer_name,
+    COUNT(ratings.overall) AS rating_count
+  FROM breweries
+  LEFT JOIN beers ON beers.brewery_id = breweries.id
```
ssapra: A left join on beers will include breweries that have no beers. You might want to do an inner join instead.

yadda/sql/views.sql
```diff
@@ -0,0 +1,41 @@
...
+CREATE OR REPLACE VIEW recent_score AS
+  SELECT
+    breweries.id AS brewery_id,
+    breweries.name AS brewery_name,
+    beers.name AS beer_name,
+    AVG(ratings.overall) AS average_rating
+  FROM breweries
+  LEFT JOIN beers ON beers.brewery_id = breweries.id
+  LEFT JOIN ratings ON ratings.beer_id = beers.id
+  WHERE ratings.created_on >= (NOW() - interval '6 months')
```
ssapra: I would use INTERVAL since all the other keywords are uppercase just to be consistent.

yadda/sql/explain
```diff
@@ -0,0 +1,162 @@
...
+                     ->  Index Scan using breweries_pkey on breweries  (cost=0.28..8.29 rows=1 width=22) (actual time=0.009..0.009 rows=1 loops=1)
+                           Index Cond: (id = 483)
+                     ->  Nested Loop  (cost=4.61..68.52 rows=5 width=28) (actual time=0.022..0.043 rows=5 loops=1)
+                           ->  Bitmap Heap Scan on beers  (cost=4.33..22.61 rows=6 width=26) (actual time=0.011..0.016 rows=6 loops=1)
+                                 Recheck Cond: (brewery_id = 483)
+                                 ->  Bitmap Index Scan on beers_brewery_id  (cost=0.00..4.33 rows=6 width=0) (actual time=0.007..0.007 rows=6 loops=1)
+                                       Index Cond: (brewery_id = 483)
+                           ->  Index Scan using ratings_beer_id_index on ratings  (cost=0.28..7.64 rows=1 width=6) (actual time=0.004..0.004 rows=1 loops=6)
+                                 Index Cond: (beer_id = beers.id)
+                                 Filter: (created_on >= (now() - '6 mons'::interval))
```
ssapra: Since you are doing a filter using the created_on column, it's a good candidate for an index.

yadda/sql/yadda-schema.sql
```diff
@@ -0,0 +1,88 @@
...
+-- Store beer information. name and style must be under 100 characters.
+-- Brewery for a specific beer can not be null, and when a brewery is deleted, all
+-- corresponding beers created by that brewery are deleted as well.
+-- beer (singular) to brewery is one to one, though a brewery has many beers.
+
+CREATE TABLE ratings (
+  id          SERIAL PRIMARY KEY,
+  user_id     INTEGER REFERENCES users (id) ON DELETE CASCADE,
+  beer_id     INTEGER REFERENCES beers (id) ON DELETE CASCADE,
+  look        SMALLINT CHECK (look > 0 AND look < 11),
```
ssapra: Alternatively, the constraint could be `CHECK (look BETWEEN 0 AND 11) DEFAULT 0`

yadda/sql/yadda-schema.sql
```diff
@@ -0,0 +1,88 @@
...
+  created_by     VARCHAR(50)
+);
+COMMENT ON TABLE breweries IS 'Information about each brewery';
+-- Store brewery information. Name and city must not be null.
+-- Zip is varchar in order to support hyphens. Brewery to beers is one to many.
+
+
+CREATE TABLE styles (
+  id          SERIAL PRIMARY KEY,
+  style       VARCHAR(50) NOT NULL
```
ssapra: I would suggest a column name like `name` or `title`

yadda/sql/yadda-schema.sql
```diff
@@ -0,0 +1,88 @@
...
+-- overall are integers between 1 and 10.
+-- A user has many ratings, a beer has many ratings.
+
+CREATE INDEX beers_brewery_id on beers(brewery_id);
+CREATE INDEX beer_style_id on beers(style_id);
+CREATE INDEX ratings_beer_id on ratings(beer_id);
+CREATE INDEX ratings_id_index ON ratings(id);
+CREATE INDEX ratings_user_id_index ON ratings(user_id);
+CREATE INDEX ratings_beer_id_index ON ratings(beer_id);
+CREATE INDEX ratings_overall_index ON ratings(overall);
```
paulghaddad: Good indices, and like Shaan said, ratings.create_on would be another good candidate.

paulghaddad:
@mcdaniel67 Nice work on Yadda! I recommend you do a quick resubmit to complete this exercise.

yadda/Database.rb
mcdaniel67: I was but decided to take it out. Got sidetracked and forgot I left it in here.

yadda/data_generation.rb
mcdaniel67: I assumed that user creation would be done by the user via a signup page.

yadda/sql/all.sql
mcdaniel67: I was using this as a way to quickly view all tables then delete everything. It was useful when I was testing `data_generation.rb` to ensure that everything was inserted properly. I'll remove it from the branch now that it's not likely to end up being used (it'd be better for me to just use psql console in another screen to check everything)

yadda/Database.rb
mcdaniel67: This led me to run into an id problem when using `sql/all.sql` to view/clean the tables. (The ids would continue counting above the constraints in data_generation.rb) I decided that this was better, as it always guaranteed that the random id was present in the table.

PR #979, opened 2015-09-08
Yadda pt 1

ochagata:


yadda/yadda-schema.sql
```diff
@@ -0,0 +1,59 @@
+DROP DATABASE IF EXISTS yadda CASCADE;
+CREATE DATABASE yadda;
+\connect yadda;
+
+CREATE TABLE users (
+  id              SERIAL                    PRIMARY KEY,
+  name            VARCHAR(40)               NOT NULL,
```
jmmastey: There are folks at Enova with names longer than 40. Any reason for this limit?

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,59 @@
+DROP DATABASE IF EXISTS yadda CASCADE;
+CREATE DATABASE yadda;
+\connect yadda;
+
+CREATE TABLE users (
+  id              SERIAL                    PRIMARY KEY,
+  name            VARCHAR(40)               NOT NULL,
+  email           VARCHAR(100)              NOT NULL,
```
jmmastey: Same question as email.

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,59 @@
...
+CREATE DATABASE yadda;
+\connect yadda;
+
+CREATE TABLE users (
+  id              SERIAL                    PRIMARY KEY,
+  name            VARCHAR(40)               NOT NULL,
+  email           VARCHAR(100)              NOT NULL,
+  birthday        DATE                      NOT NULL,
+  password        VARCHAR(50)               NOT NULL,
+  modified_on     TIMESTAMP WITH TIME ZONE  NOT NULL DEFAULT current_time,
```
jmmastey: Typically fields that are suffixed with `_at` are understood to be DateTime objects, whereas `_on` means a Date. These are timestamps, so they should be `modified_at` (more typically `updated_at`) and `created_at`. 

It may seem pedantic, but these kinds of conventions make it easier to understand the intent of a system. So, unless there's a good reason to depart from them, it's better to stick to them.

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,59 @@
...
+  modified_by     VARCHAR(50),
```
jmmastey: What are the contents of this field? Enova has issues where there's a `created_by`, but because the field is too small they have to truncate data, which is a cardinal sin. Consider either a larger field or a proper relation to some other table.

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,59 @@
...
+  created_by      VARCHAR(50)
+);
+
+CREATE TABLE breweries (
+  id              SERIAL                    PRIMARY KEY,
+  name            TEXT                      NOT NULL,
+  address_id      INTEGER                   NOT NULL,
+  description     TEXT                      NOT NULL,
+  founding_year   INTEGER                   NOT NULL,
+  modified_on     TIMESTAMP WITH TIME ZONE  NOT NULL DEFAULT current_time,
```
jmmastey: With default times, I think this is probably a good idea ;) There's no way to be 100% consistent between every machine (lord knows webservers are often way off) so using the DB time may be the safest option.

yadda/yadda-schema.sql
ochagata: It was an arbitrary limit.
I was wondering: I could set this to just VARCHAR, and it would be as much as you like, but is there a limit in postgres or can you pass in infinite length strings?

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,59 @@
...
+  id              SERIAL                    PRIMARY KEY,
+  style           VARCHAR(50)               NOT NULL
+);
+
+CREATE TABLE beers (
+  id              SERIAL                    PRIMARY KEY,
+  name            VARCHAR(100)              NOT NULL,
+  style_id        INTEGER                   NOT NULL REFERENCES styles (id) ON DELETE SET NULL,
+  description     TEXT                      NOT NULL,
+  brew_year       INTEGER                   NOT NULL,
```
jmmastey: If I have partial information about an older beer, I might not want to include year. Be careful not to just mark everything as `NOT NULL` unless you really need it.

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,59 @@
...
+
+CREATE TABLE styles (
+  id              SERIAL                    PRIMARY KEY,
+  style           VARCHAR(50)               NOT NULL
+);
+
+CREATE TABLE beers (
+  id              SERIAL                    PRIMARY KEY,
+  name            VARCHAR(100)              NOT NULL,
+  style_id        INTEGER                   NOT NULL REFERENCES styles (id) ON DELETE SET NULL,
```
jmmastey: I'm not sure _how_ I would handle this. Do we restrict deletion of styles? This is a reasonable way to handle cascading.

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,59 @@
...
+  created_on      TIMESTAMP WITH TIME ZONE  NOT NULL DEFAULT current_time,
+  created_by      VARCHAR(50)
+);
+
+CREATE TABLE ratings (
+  id              SERIAL                    PRIMARY KEY,
+  user_id         INTEGER                   NOT NULL REFERENCES users (id) ON DELETE CASCADE,
+  beer_id         INTEGER                   NOT NULL REFERENCES beers (id) ON DELETE CASCADE,
+  look            SMALLINT                  NOT NULL CHECK ( look > 0 AND look < 11),
+  smell           SMALLINT                  NOT NULL CHECK ( smell > 0 AND smell < 11),
```
jmmastey: Our smell goes up to 11!

jmmastey:
Looks good. Let's see Part 2 ;)

yadda/yadda-schema.sql
ochagata: Noted

yadda/yadda-schema.sql
```diff
@@ -0,0 +1,59 @@
+DROP DATABASE IF EXISTS yadda CASCADE;
+CREATE DATABASE yadda;
+\connect yadda;
+
+CREATE TABLE users (
+  id              SERIAL                    PRIMARY KEY,
+  name            VARCHAR(40)               NOT NULL,
+  email           VARCHAR(100)              NOT NULL,
+  birthday        DATE                      NOT NULL,
+  password        VARCHAR(50)               NOT NULL,
```
jmmastey: Btw, Bcrypt hashes (the correct way to store a hashed password) are 60 characters long.

yadda/yadda-schema.sql
ochagata: I imagined that this field would basically be a signature of the last person who changed it. I would expect this to be automated and be filled in with the username of the last person to edit this table

PR #983, opened 2015-09-10
initial pass

yuvalroth:


supportive/supportive.rb
```diff
@@ -38,63 +36,30 @@ def to_s
...
-    if categories.length > 1
-      last_category = categories.pop
-      suffix = " and #{as_title(last_category)}"
-    else
-      suffix = ""
-    end
-
-    label + ": " + categories.map { |cat| as_title(cat) }.join(", ") + suffix
+    return '' if categories.empty?
+    "Category".pluralize(categories.length) + ": " + categories.to_sentence
```
djkotowski: Prefer `<<` over `+` to avoid creating a new string per concatenation.

supportive/supportive.rb
```diff
@@ -38,63 +36,30 @@ def to_s
   private
 
   def byline
-    if author.nil?
-      ""
-    else
-      "By #{author.name}, at #{author.url}"
-    end
+    author.nil? ? '' : "By #{author.name}, at #{author.url}"
```
djkotowski: You can use `#try` to simplify this and avoid the ternary:

```ruby
author.try { |a| "By #{a.name}, at #{a.url}" } || ""
```

supportive/supportive.rb
```diff
@@ -9,12 +9,10 @@ class BlagPost
...
   def initialize(args)
-    args = args.inject({}) do |hash, (key, value)|
-      hash[key.to_sym] = value
-      hash
-    end
 
-    if args[:author] != '' && args[:author_url] != ''
+    args.symbolize_keys!
+
+    if args[:author].present? && args[:author_url].present?
```
djkotowski: Should you need to refactor to ensure more params are present, you could use a DRYer pattern:
```ruby
if [:author, :author_url].all? { |key| args[key].present? }
```
or
```ruby
if [args[:author], args[:author_url]].all?(&:present?)
```

supportive/supportive.rb
```diff
@@ -27,7 +25,7 @@ def initialize(args)
     end
```
djkotowski: You could simplify L20-22 with
```ruby
@categories = args[:categories] - DISALLOWED_CATEGORIES
```

supportive/supportive.rb
```diff
@@ -38,63 +36,30 @@ def to_s
...
-    words = string.gsub('_', ' ').split(' ')
-
-    words.map!(&:capitalize)
-    words.join(' ')
+    string.to_s.titleize
   end
 
   def commenters
     return '' unless comments_allowed?
     return '' unless comments.length > 0
```
djkotowski: This condition can be simplified to `comments.any?`.  Also, since you're returning the same thing, just combine L52-53.

```ruby
return '' unless comments_allowed? || comments.any?
```

djkotowski:
Solid work.  Just a few comments throughout.  No need for a refactor.

PR #985, opened 2015-09-11
DinoCatalog v2

Manbearpixel:
Refactored code for v2


dino_catalog/Dinosaur.rb
```diff
@@ -0,0 +1,44 @@
+class Dinosaur
+  attr_accessor :name, :location, :period, :weight, :walk_type, :description
+  attr_reader :diet
+
+  def diet=(str)
+    str = ("Carnivore" if str == "Yes") || (nil if str == "No")
```
djkotowski: Assume herbivore if not carnivore.

dino_catalog/Dinosaur.rb
```diff
@@ -0,0 +1,44 @@
...
+    return false if weight.nil?
+    Float(weight) > Float(size)
+  end
+
+  def smaller_than?(size)
+    return false if weight.nil?
+    !larger_than?(size)
+  end
+
+  def print_facts
```
djkotowski: Dinosaur should not be responsible for console output.  You could, however, return a string with the formatted data.

dino_catalog/Dinosaur.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def export_hash
+    hash = {}
+    instance_variables.each do |instance_var|
+      fact_header  = "#{clean_attr_name(instance_var)}"
+      fact_value   = "#{instance_variable_get(("#{instance_var}"))}"
+      hash[fact_header] = fact_value
+    end
+    hash
```
djkotowski: You can remove L30 and L36 by using `#each_with_object` instead.

dino_catalog/Dinosaur.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def smaller_than?(size)
+    return false if weight.nil?
+    !larger_than?(size)
+  end
+
+  def print_facts
+    instance_variables.each do |instance_var|
+      next unless instance_variable_get(("#{instance_var}"))
+      fact_header  = "[#{clean_attr_name(instance_var).gsub(/[A-Za-z']+/, &:capitalize)}]"
```
djkotowski: This header formatting looks like it belongs in a different method.

dino_catalog/Dinosaur.rb
```diff
@@ -0,0 +1,44 @@
...
+      fact_value   = "#{instance_variable_get(("#{instance_var}"))}"
```
djkotowski: `("#{instance_var}")` should just be `instance_var`

dino_catalog/Dinosaur.rb
```diff
@@ -0,0 +1,44 @@
...
+      printf("%-15s %s\n", fact_header, fact_value)
+    end
+  end
+
+  def export_hash
+    hash = {}
+    instance_variables.each do |instance_var|
+      fact_header  = "#{clean_attr_name(instance_var)}"
+      fact_value   = "#{instance_variable_get(("#{instance_var}"))}"
```
djkotowski: This method looks very similar to `print_facts` above.

dino_catalog/cli_test.rb
```diff
@@ -0,0 +1,215 @@
...
+  private
+
+  def print_main
+    clear_screen
+    print_header
+  end
+
+  def state_info
+    printf("Dinosaurs:\t\t(%s)\nWorking Path:\t\t(%s)\nCSV Files Found:\t(%s)\n\n",
+      @dex.dinosaurs.length.to_s,
```
djkotowski: Align your arguments.

djkotowski: Also, calling `#to_s` on the arguments is redundant.

dino_catalog/cli_test.rb
```diff
@@ -0,0 +1,215 @@
...
+  end
+
+  def print_header
+    header_line(@header_line_length)
+    header_line(@header_line_length, @header_line_padding, "Jurassic Park, DinoDex Interface")
+    header_line(@header_line_length, @header_line_padding, "Version 4.0.5, Alpha E")
+    header_line(@header_line_length)
+  end
+
+  def header_line(line_length = 40, line_padding = 0, line_str = "")
```
djkotowski: Why have a variable line length here and a constant line length for `#print_line`?

dino_catalog/cli_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    return puts "#" * line_length if line_str == ""
+    empty_space = line_length - line_str.length - (line_padding * 2)
+    return puts "#" * line_length if empty_space < 0
+    line_str = "##" + (" " * (empty_space / 2)) + line_str + (" " * (empty_space / 2))
+    return printf(line_str + " ##\n") unless empty_space.even?
+    printf(line_str + "##\n")
+  end
+
+  def print_line
+    puts "*" * 80
```
djkotowski: The line length should be a constant, not a magic number.

dino_catalog/cli_test.rb
```diff
@@ -0,0 +1,215 @@
+require_relative 'dino_dex'
+
+class DinodexShell
```
djkotowski: Why the filename/class name mismatch?

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,29 @@
+class DinoCatalog
+  attr_accessor :dinosaurs
+
+  def initialize(dinosaurs = [])
+    @dinosaurs = dinosaurs.clone
+  end
+
+  def filter(key, value)
+    if key.to_s == "weight" && value.include?("<")
+      dinosaurs.select! { |dinosaur| dinosaur.smaller_than?(value.to_s[1, value.length]) }
```
djkotowski: Rarely will you see destructive filters used.  Usually, filters will return a new collection with the filtered data.

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  def empty?
+    dinosaurs.length == 0
+  end
+
+  def select(key, value)
+    DinoCatalog.new(dinosaurs).filter(key, value)
+  end
+
+  def select_larger_than(value)
```
djkotowski: Why do you have both a Dinodex and DinoCatalog?  They look like they should be the same class.

dino_catalog/dino_catalog.rb
Manbearpixel: This is setup like that because it allows for multiple "catalogs" to be created from one original source of dinosaurs. Each catalog can be filtered to their own liking.

dino_catalog/dino_dex.rb
Manbearpixel: As stated in the above comment, yes they -could- be combined into just Dinodex, but I set this up so that there would always be an original list of dinosaurs, with select statements returning an altered list of dinosaurs which can be filtered even more.

dino_catalog/cli_test.rb
Manbearpixel: Not sure what you mean here... The file requirement is referencing the class DinoDex which is in "dino_dex.rb"
The class DinodexShell is only for this Command Line Interface

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def select_smaller_than(value)
+    DinoCatalog.new(dinosaurs.select { |dinosaur| dinosaur.smaller_than?(value) })
+  end
+
+  def select_hash(args)
+    catalog = DinoCatalog.new(dinosaurs)
+    args.each { |key, value| catalog.filter(key, value) }
+    catalog
```
djkotowski: Generally in this case, `#tap` is preferred.
```ruby
value.tap do |var|
  # do stuff
end
```
instead of
```ruby
var = value
# do stuff
var
```

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,92 @@
...
+      HEADER_NORMALIZE.each do |old_header, new_header|
+        next unless row.header?(old_header)
+        row_value = row[old_header]
+        row.delete(old_header)
+        row[new_header] = row_value
+      end
+    end
+  end
+
+  def create_dinosaur(csv_row)
```
djkotowski: This should be the responsibility of Dinosaur, not DinoDex.

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,92 @@
...
+  def select_dinosaur(dinosaur_name)
+    dinosaurs.each do |dinosaur|
+      return dinosaur if dinosaur.name.downcase == dinosaur_name.downcase
+    end
+    nil
+  end
+
+  def export_json
+    {
+      "dinosaurs" => dinosaurs.map(&:export_hash),
```
djkotowski: Since JSON automatically parses strings _and_ symbols, I'd prefer a symbol key here.  That way, you aren't creating a new temp string "dinosaurs" each time export_json is called.

Additionally, to be consistent, `#to_json` would be a better method name.

dino_catalog/Dinosaur.rb
Manbearpixel: This should be stated in the README then... Personally I didn't want to assume a dinosaur was a herbivore just because it was not a Carnivore. Never trust information from Pirate Bay :P

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def import_csv_file(csv_file_path)
+    extract_rows(csv_file_path)
+    normalize_headers(@csv_rows)
+    @csv_rows.each_index { |index| dinosaurs << create_dinosaur(@csv_rows[index]) }
+  end
+
+  def empty?
+    dinosaurs.length == 0
```
djkotowski: Prefer `dinosaurs.empty?`

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,92 @@
...
+    catalog = DinoCatalog.new(dinosaurs)
+    args.each { |key, value| catalog.filter(key, value) }
+    catalog
+  end
+
+  def select_dinosaur(dinosaur_name)
+    dinosaurs.each do |dinosaur|
+      return dinosaur if dinosaur.name.downcase == dinosaur_name.downcase
+    end
+    nil
```
djkotowski: Use `#find` instead.

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,92 @@
...
+    {
+      "dinosaurs" => dinosaurs.map(&:export_hash),
+    }.to_json
+  end
+
+  private
+
+  def extract_rows(csv_file_path)
+    CSV.foreach(csv_file_path, headers: true, header_converters: :symbol) do |row|
+      (@csv_rows ||= []) << row
```
djkotowski: This is doing too much.  Why are you rechecking if you need to initialize for each row?  It would be cleaner to put `@csv_rows ||= []` before the `CSV.foreach`.

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  def initialize
+    @dinosaurs = []
+  end
+
+  def import_csv_file(csv_file_path)
+    extract_rows(csv_file_path)
+    normalize_headers(@csv_rows)
+    @csv_rows.each_index { |index| dinosaurs << create_dinosaur(@csv_rows[index]) }
```
djkotowski: This should be a map.

dino_catalog/Dinosaur.rb
djkotowski: Yep.  Joe did some research and apparently, omnivores were extremely rare back then.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+    self
+  end
+
+  def print_data(requested_data)
+    requested_data = Array(requested_data)
+    dinosaurs.each do |dinosaur|
+      dino_data ||= ''
+      requested_data.each do |header_name|
+        dino_data << "#{dinosaur.instance_variable_get(("@#{header_name}"))} | "
```
djkotowski: This is getting a bit arrowhead-y.

dino_catalog/file_finder.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def search
+    @files = []
+    Dir.glob(file_type) { |file_name| @files.push(@directory_path + file_name) }
+    files.length
+  end
+
+  def set_proper_path(directory_path)
+    return directory_path << '/' unless directory_path.end_with?('/')
```
djkotowski: If this were run on Windows, it would fail.  Use `File.join` to help with directory delimiters and joining absolute and relative paths.

dino_catalog/file_finder.rb
```diff
@@ -0,0 +1,33 @@
...
+    @file_type        = file_type
+    @directory_path   = directory_path
+
+    Dir.chdir(@directory_path)
+    @directory_path = set_proper_path(Dir.pwd)
+    search
+  end
+
+  def directory_path=(path)
+    return puts ERROR_NO_DIRECTORY if Dir.exist?(path) == false
```
djkotowski: Shouldn't this be
```ruby
unless Dir.exist?(path)
```

dino_catalog/file_finder.rb
```diff
@@ -0,0 +1,33 @@
...
+    Dir.chdir(path)
+    @directory_path = set_proper_path(Dir.pwd)
+    true
+  end
+
+  def search
```
djkotowski: I would expect `search` to return `@files`, not the length.

dino_catalog/cli_test.rb
djkotowski: The filename for DinodexShell should be `dinodex_shell.rb` not `cli_test.rb`

dino_catalog/Dinosaur.rb
jmmastey: Basically omnivorousness hadn't been invented yet. Weird, huh? Still, more particularly, this contains CSV parsing logic within the model class, which is not a great choice. If I added a second CSV (which I did consider) that had an Herbivore column, this code would break unnecessarily.

dino_catalog/cli_test.rb
```diff
@@ -0,0 +1,215 @@
...
+  end
+
+  def state_options
+    puts "COMMANDS:\t\t(P): Set Path, (L): Load CSV, (S): Search, (A): List Dinosaurs"
+    puts "\t\t\t(E): Export JSON Dinosaurs, (Q): Quit DinoDex\n\n"
+  end
+
+  def print_header
+    header_line(@header_line_length)
+    header_line(@header_line_length, @header_line_padding, "Jurassic Park, DinoDex Interface")
```
jmmastey: image

dino_catalog/cli_test.rb
```diff
@@ -0,0 +1,215 @@
...
+      when "a"
+        list_dinosaurs
+      when "q"
+        stop_shell
+      when "e"
+        export_dinosaurs
+      else
+        gets_option
+      end
+  end
```
jmmastey: Small note. While I almost always try to refactor case statements out, this can be a little more succinct as:

```
case input
  when "p" then print_prompt
  when "l" then prompt_load_csv
  when "s" then prompt_search_dinodex
  when "a" then list_dinosaurs
  when "q" then stop_shell
  when "e" then export_dinosaurs
  else gets_option
end
```

dino_catalog/dino_dex.rb
jmmastey: At the least the naming is odd. Without reading them, which is a Catalog and which is a Codex?

dino_catalog/cli_test.rb
djkotowski: :+1:

dino_catalog/cli_test.rb
```diff
@@ -0,0 +1,215 @@
...
+
+  def prompt_set_path
+    printf "Please specify the CSV search path or (!Q)uit: "
+    input = gets.strip
+    case input.downcase
+      when "!q"
+        print_main
+      else
+        if !set_working_dir(input)
+          prompt_set_path
```
djkotowski: This also is a bit arrowhead-y.

Manbearpixel:
Will refactor using noted comments

PR #987, opened 2015-09-15
DinoCatalog v3

Manbearpixel:


djkotowski:
This is the same as V2.  Resubmit when ready.

PR #989, opened 2015-09-17
Supportive - V1

oharace:


supportive/supportive.rb
```diff
@@ -32,77 +32,45 @@ def initialize(args)
...
-    end
-    "You will be the #{comments.length}#{ordinal} commenter"
+    return '' unless comments_allowed? && comments.empty?
+
+    "You will be the #{comments.length.ordinalize} commenter"
   end
 
   def comments_allowed?
-    publish_date + (365 * 3) > Date.today
+    publish_date.years_since(3) > Date.today
```
ssapra: This works fine, but I personally like `publish_date > 3.years.ago`

supportive/supportive.rb
```diff
@@ -32,77 +32,45 @@ def initialize(args)
...
-    return '' unless comments.length > 0
-
-    ordinal = case comments.length % 10
-      when 1 then "st"
-      when 2 then "nd"
-      when 3 then "rd"
-      else "th"
-    end
-    "You will be the #{comments.length}#{ordinal} commenter"
+    return '' unless comments_allowed? && comments.empty?
```
ssapra: I think you mean `comments.present?`.

ssapra:
In the initialize method, there are a few neat methods that can help simplify the code.

Check out: 
* [with_indifferent_access](http://apidock.com/rails/HashWithIndifferentAccess/ActiveSupport/CoreExtensions/Hash/IndifferentAccess/with_indifferent_access) so that you can access a value in the hash using only a key
* [squish](http://apidock.com/rails/ActiveSupport/CoreExtensions/String/Filters/squish) to filter strings
* [array subtraction](http://ruby-doc.org/core-2.1.2/Array.html#method-i-2D) to remove elements from an array
* [Array.wrap](http://apidock.com/rails/Array/wrap/class) to convert nil values and arrays into arrays
* [try](http://apidock.com/rails/Object/try): This is useful in the above case when you want to convert a value to a date, but you're not sure if the value will parse correctly

Everything else looks good to me!

paulghaddad:
@oharace This is good work! You are good to move on.

PR #992, opened 2015-09-28
Update Arrowhead

mostlybadfly:
I was initially able to separate out into two If/else statements, but it made the method more than 5 lines. I have created some private methods for each check.  I get the same response, but not sure if it adds unneeded complexity. I'd love your thoughts!

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -15,18 +15,22 @@ class Arrowhead
     },
   }
 
-  # FIXME: I don't have time to deal with this.
   def self.classify(region, shape)
-    if CLASSIFICATIONS.include? region
-      shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
+    shapes = CLASSIFICATIONS[region] unless check_region(region)
+    arrowhead = shapes[shape] unless check_shape(shapes, shape)
+    "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
+  end
+
+  private
+  def self.check_region(region)
+    unless CLASSIFICATIONS.include? region
+      raise "Uknown region, please provide a valid region."
+    end
+  end
+
+  def self.check_shape(shapes, shape)
+    unless shapes.include? shape
+      raise "Uknown shape value. Are you sure you know what you're talking about?"
     end
   end
 end
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -16,20 +16,24 @@ class Arrowhead
   }
 
   def self.classify(region, shape)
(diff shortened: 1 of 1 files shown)


mostlybadfly:
Hello, I was wondering if I could get some feedback on this.  I see the Hound reply but I believe it is on the line that is a longer String. This was an interesting exercise so I was hoping to get more progess on it, thanks!

arrowhead/arrowhead.rb
```diff
@@ -15,18 +15,26 @@ class Arrowhead
...
   }
 
-  # FIXME: I don't have time to deal with this.
   def self.classify(region, shape)
+    shapes = CLASSIFICATIONS[region] if check_region(region)
+    arrowhead = shapes[shape] if check_shape(shapes, shape)
+    "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
+  end
+
+  private
```
incazteca: There's really no need for a private modifier since the methods are defined with `self`. This allows them to be called in a manner such as `Arrowhead.check_region(region)`

arrowhead/arrowhead.rb
```diff
@@ -15,18 +15,26 @@ class Arrowhead
...
+  def self.check_region(region)
     if CLASSIFICATIONS.include? region
```
incazteca: This check could be changed into one line to raise if the classification is not included:

`raise 'message' unless CLASSIFICATIONS.include? region`

arrowhead/arrowhead.rb
```diff
@@ -15,18 +15,26 @@ class Arrowhead
     },
   }
 
-  # FIXME: I don't have time to deal with this.
   def self.classify(region, shape)
+    shapes = CLASSIFICATIONS[region] if check_region(region)
```
incazteca: This line could just be `shapes = check_region(region)`. You could have the `check_region` method do the check and then return the shapes.

arrowhead/arrowhead.rb
```diff
@@ -15,18 +15,26 @@ class Arrowhead
...
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
+      true
+    else
+      raise "Uknown region, please provide a valid region."
+    end
+  end
+
+  def self.check_shape(shapes, shape)
+    if shapes.include? shape
```
incazteca: Similar note to the one in `check_region`. Have the condition to raise be a one liner.

arrowhead/arrowhead.rb
```diff
@@ -15,18 +15,26 @@ class Arrowhead
     },
   }
 
-  # FIXME: I don't have time to deal with this.
   def self.classify(region, shape)
+    shapes = CLASSIFICATIONS[region] if check_region(region)
+    arrowhead = shapes[shape] if check_shape(shapes, shape)
```
incazteca: This line could just become `arrowhead = check_shape(shapes, shape)`. If you do just want to make `check_shape` return a boolean value have it become `check_shape?`. However, the line is simpler if `check_shape` were to return the value.

incazteca:
@mostlybadfly Apologies for the late review. Most of the methods could be shorter once you get rid of the `if` blocks and use one line conditionals.

mostlybadfly:
@incazteca thanks for all this great feedback!  I'll fix these up, much appreciated :)

PR #995, opened 2015-10-02
Shopping Cart v1.0

Manbearpixel:
30 min timer for a quick Cucumber write up of Shopping cart and it's features.

### DIFF
--- a/shopping_cart/shopping_cart.feature
+++ b/shopping_cart/shopping_cart.feature
@@ -1,0 +1,123 @@
+Feature: Shopping Cart
+  As a potential customer
+  I want a shopping cart to save items I'm interested in
+  So I can review and purchase everything I'm interested in
+
+  @HappyTask
+  Scenario: Should be able to add product to Shopping Cart
+    Given I view the "Lego Set" product page
+    When I click "Add to cart"
+    Then I should see I have 1 item in my cart
+
+  @HappyTask
+  Scenario: Should be able to remove product from Shopping Cart
+    Given I added "Lego Set" to my cart
+    When I am on the shopping cart page
+    And I click "Remove Lego Set" on the page
+    Then "Lego Set" should no longer be in my cart
+
+  @HappyTask
+  Scenario: Should be able to adjust quanities of a product in Shopping Cart
+    Given I am on the shopping cart page with "Lego Set" in my cart
+    When I set the "Lego Set" quantity to "2" and click "Update"
+    Then I should see "Quantity: 2" in the "Lego Set" row
+
+  @HappyTask
+  Scenario: Should remove item if quantity is set to 0
+    Given I am on the shopping cart page with "Lego Set" in my cart
+    When I set the "Lego Set" quantity to "0" and click "Update"
+    Then "Lego Set" should no longer be in my cart
+
+  @HappyTask
+  Scenario: Should be able to visit product page from shopping cart
+    Given I am viewing my cart's contents
+    When I click on the "Lego Set" row
+    Then I should be on a page that says "Product: Lego Set"
+
+  @HappyTask
+  Scenario: Should be able to get shipping estimates
+    Given I am on the checkout page
+    When I enter my Shipping Address "175 W. Jackson"
+    And I enter my City "Chicago"
+    And I choose my state "IL"
+    And I enter my zipcode "60606"
+    Then I should see my shipping estimate to have "3.00"
+
+  @HappyTask
+  Scenario: Should be able to enter a coupon code
+    Given I am on the checkout page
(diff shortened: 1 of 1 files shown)


shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,123 @@
+Feature: Shopping Cart
+  As a potential customer
+  I want a shopping cart to save items I'm interested in
+  So I can review and purchase everything I'm interested in
+
+  @HappyTask
+  Scenario: Should be able to add product to Shopping Cart
+    Given I view the "Lego Set" product page
```
ochagata: You shouldn't use specific items when testing, unless the test case is for that item.
Make it random, so that when you run this test, maybe it catches an item that causes an error

shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,123 @@
...
+    When I click "Add to cart"
+    Then I should see I have 1 item in my cart
+
+  @HappyTask
+  Scenario: Should be able to remove product from Shopping Cart
+    Given I added "Lego Set" to my cart
```
ochagata: Personal preference: I'd probably go with `Given I have an item in my cart` instead

shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,123 @@
...
+    When I am on the shopping cart page
```
ochagata: This is more of an assumptive test, so it would make more sense for it to be a given
If it were a true When, it would be an action: `When I navigate to the shopping cart page`

shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,123 @@
...
+    And I click "Remove Lego Set" on the page
```
ochagata: You don't need `on the page` here

shopping_cart/shopping_cart.feature
ochagata: Also, this item might not always be there, thus the test is not future-proof

shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,123 @@
...
+
+  @HappyTask
+  Scenario: Should remove item if quantity is set to 0
+    Given I am on the shopping cart page with "Lego Set" in my cart
+    When I set the "Lego Set" quantity to "0" and click "Update"
+    Then "Lego Set" should no longer be in my cart
+
+  @HappyTask
+  Scenario: Should be able to visit product page from shopping cart
+    Given I am viewing my cart's contents
```
ochagata: You should be consistent with your `Given` statements, otherwise you'll have to reimplement the same steps, use the one used above: `Given I am on the shopping cart page`

shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,123 @@
...
+    When I click on the "Lego Set" row
```
ochagata: You shouldn't need `row` here

shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,123 @@
...
+    Then I should be on a page that says "Product: Lego Set"
```
ochagata: Try to use the common steps more:
`Then I should see "foobar"`

shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,123 @@
...
+
+  @HappyTask
+  Scenario: Should be able to get shipping estimates
+    Given I am on the checkout page
+    When I enter my Shipping Address "175 W. Jackson"
+    And I enter my City "Chicago"
+    And I choose my state "IL"
+    And I enter my zipcode "60606"
+    Then I should see my shipping estimate to have "3.00"
```
ochagata: You should use a more common step here too:
`Then I should see "Shipping estimate: 3.00"`

shopping_cart/shopping_cart.feature
```diff
@@ -0,0 +1,123 @@
...
+
+  @HappyTask
+  Scenario: Should be able to enter a coupon code
+    Given I am on the checkout page
+    When I enter my coupon "FREESHIP"
+    Then I should see "FREESHIP" under my list of active coupons
```
ochagata: You should also test that the coupon is doing what it should be doing:
`Then I should see "Shipping fee: 0.00"`
Also, test different kinds of coupons, discount, discount for specific item, discount for $n or more

ochagata:
Needs Resubmission @paulghaddad

paulghaddad:
@Manbearpixel Please take a look at the comments and submit a second iteration.

PR #997, opened 2015-10-05
Robot

arpanp1986:


robot_name/robot_name.rb
```diff
@@ -1,32 +1,52 @@
...
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  def generate_num
+    rand(10)
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
+  def duplication_check
+    raise NameCollisionError, ERROR if incorrect_name_pattern? || name_taken?
```
djkotowski: Is it accurate to call this a collision error if the name is invalid, rather than a duplicate?

robot_name/robot_name.rb
```diff
@@ -1,32 +1,52 @@
+require 'pry'
 class NameCollisionError < RuntimeError; end
 
+ERROR = 'There was a problem generating the robot name!'
+
 class Robot
   attr_accessor :name
 
   @@registry
```
djkotowski: Avoid class variables in favor of [class instance variables](http://www.railstips.org/blog/archives/2006/11/18/class-and-instance-variables-in-ruby/).

djkotowski:
Solid submission!  Only a couple minor comments.  Nice work.

PR #1008, opened 2015-10-09
Arrowhead

athiercelin:


arrowhead/arrowhead.rb
```diff
@@ -15,19 +15,17 @@ class Arrowhead
...
-    if CLASSIFICATIONS.include? region
-      shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
+    if CLASSIFICATIONS.include? region == false
```
djkotowski: Wrap with parentheses, and never compare to false.  Prefer `unless condition` instead.

arrowhead/arrowhead.rb
```diff
@@ -15,19 +15,17 @@ class Arrowhead
...
       raise "Unknown region, please provide a valid region."
```
djkotowski: If you extract the error-raising out to methods, you can reduce the arrowhead-ness even further.

djkotowski:
Good work so far.  Take a look at the comments and refactor where needed.  No need for a resubmission.

PR #1010, opened 2015-10-09
Robot name

athiercelin:


robot_name/robot_name.rb
```diff
@@ -3,23 +3,43 @@ class NameCollisionError < RuntimeError; end
...
+      return true
+    # redundancy matching
+    elsif @registry.include?(new_name)
+      return true
+    end
+    false
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def initialize(args = {})
```
djkotowski: `#initialize` should be the first method defined.

robot_name/robot_name.rb
```diff
@@ -3,23 +3,43 @@ class NameCollisionError < RuntimeError; end
...
+    @registry ||= []
+    @name_generator = args[:name_generator]
+    @name = ''
+
+    while new_name_is_invalid(@name)
+      if @name_generator
+        @name = @name_generator.call
```
djkotowski: This is a bit arrowhead-y.  You also have an unnecessary decision block here.  What if you just assigned the `@name_generator` to the default if none is provided.

robot_name/robot_name.rb
```diff
@@ -3,23 +3,43 @@ class NameCollisionError < RuntimeError; end
...
+    false
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def initialize(args = {})
+    @registry ||= []
+    @name_generator = args[:name_generator]
+    @name = ''
+
+    while new_name_is_invalid(@name)
```
djkotowski: What if the generator never provides a valid name?  Should the program run indefinitely?

robot_name/robot_name.rb
```diff
@@ -3,23 +3,43 @@ class NameCollisionError < RuntimeError; end
...
+    # redundancy matching
+    elsif @registry.include?(new_name)
+      return true
+    end
+    false
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def initialize(args = {})
+    @registry ||= []
```
djkotowski: This creates a new name registry for every Robot.  You should have one registry for all robots.

robot_name/robot_name.rb
```diff
@@ -3,23 +3,43 @@ class NameCollisionError < RuntimeError; end
...
+    "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  end
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  def new_name_is_invalid(new_name)
+    # garbage sorting
```
djkotowski: You really don't need this decision block.  First, if you return, the `elsif` isn't necessary.  Also, when you have several conditions that all return `true`, why not just combine the conditions with 'or's?

robot_name/robot_name.rb
```diff
@@ -3,23 +3,43 @@ class NameCollisionError < RuntimeError; end
...
+    if new_name == '' || new_name.nil?
```
djkotowski: The first comparison is unnecessary, as it will be caught by the Regex below.

robot_name/robot_name.rb
```diff
@@ -3,23 +3,43 @@ class NameCollisionError < RuntimeError; end
...
+    "#{generate_char.call}#{generate_char.call}"\
+    "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  end
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  def new_name_is_invalid(new_name)
```
djkotowski: This is a predicate method, and should end with `?`

djkotowski:
These exercises seemed a bit rushed.  Take your time and really get the most out of them.  Each exercise has a purpose and a lesson to teach.  The more thorough you are the first time, the less you'll have to refactor in future submissions.

For this one, take a look at the comments, refactor and resubmit when ready.

PR #1019, opened 2015-10-12
Triangle facts

athiercelin:
Updated based on Dan’s comments

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -1,21 +1,21 @@
 # Killer facts about triangles AWW YEAH
 class Triangle
-	attr_accessor :side1,:side2,:side3
+	attr_accessor :side1, :side2, :side3
 
-	def initialize(side1,side2,side3)
-		@side1,@side2,@side3 = side1,side2,side3
+	def initialize(side1, side2, side3)
+	  @side1, @side2, @side3 = side1, side2, side3
 	end
 
 	def equilateral()
-		return side1 == side2 && side2 == side3
+	  return side1 == side2 && side2 == side3
 	end
 
 	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
+	  return [side1, side2, side3].uniq.length == 2
 	end
 
 	def scalene()
-		unless ! (equilateral || isosceles)
+	  unless !(equilateral || isosceles)
 	    false
 	  else
 	    true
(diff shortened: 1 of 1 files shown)


triangle_facts/triangle_facts.rb
```diff
@@ -1,21 +1,23 @@
...
-	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
+	def isosceles?()
+	  [side1, side2, side3].uniq.length == 2
 	end
 
-	def scalene()
-		unless ! (equilateral || isosceles)
+	def scalene?()
+	  if !(equilateral || isosceles)
```
djkotowski: Now this is invalid.  This should be true if not equilateral or isosceles.

As mentioned in the previous PR, just replace the `if` with the condition.

triangle_facts/triangle_facts.rb
```diff
@@ -1,21 +1,23 @@
...
-		@side1,@side2,@side3 = side1,side2,side3
+	def initialize(side1, side2, side3)
+	  @side1 = side1
+	  @side2 = side2
+	  @side3 = side3
 	end
 
-	def equilateral()
-		return side1 == side2 && side2 == side3
+	def equilateral?()
```
djkotowski: No need for parentheses if the method has no params.

triangle_facts/triangle_facts.rb
```diff
@@ -23,34 +25,34 @@ def scalene()
 	end
 
 	def recite_facts
-		puts 'This triangle is equilateral!' if equilateral
-		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+	  puts 'This triangle is equilateral!' if equilateral
+	  puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles
```
djkotowski: Did you test this?

triangle_facts/triangle_facts.rb
```diff
@@ -23,34 +25,34 @@ def scalene()
...
 
-		angles = self.calculate_angles(side1,side2,side3)
-		puts 'The angles of this triangle are ' + angles.join(',')
+	  angles = self.calculate_angles(side1,side2,side3)
+	  puts 'The angles of this triangle are ' << angles.join(',')
 
-		puts 'This triangle is also a right triangle!' if angles.include? 90
-		puts ''
+	  puts 'This triangle is also a right triangle!' if angles.include?(90)
+	  puts ''
```
djkotowski: Not a big issue, but there's no need to pass an empty string to `puts`.  Just call the method without args.

djkotowski:
Better job.  As I said in Robot Name, don't rush.  Take your time.  Take a look at the comments and refactor.  No resubmission needed.

PR #1021, opened 2015-10-12
Robot name

athiercelin:
Fixed begin redundant

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -1,9 +1,7 @@
 class NameCollisionError < RuntimeError; end
 
-class Robot
-  attr_accessor :name
-
-  def new_name_using_default_generator
+class DefaultNameGenerator
+  def call
     generate_char = -> { ('A'..'Z').to_a.sample }
     generate_num = -> { rand(10) }
 
@@ -11,39 +9,68 @@ class Robot
     "#{generate_char.call}#{generate_char.call}"\
     "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
   end
+end
 
-  def new_name_is_invalid(new_name)
-    # garbage sorting
-    if new_name == '' || new_name.nil?
-      return true
-    # pattern matching
-    elsif !(new_name =~ /[[:alpha:]]{2}[[:digit:]]{3}/)
-      return true
-    # redundancy matching
-    elsif @registry.include?(new_name)
-      return true
-    end
-    false
+class Registry
+  attr_reader :registry
+
+  def initialize(_args = {})
+    @registry = []
   end
 
-  def initialize(args = {})
-    @registry ||= []
-    @name_generator = args[:name_generator]
-    @name = ''
+  def add_entry(new_entry)
+    validate_entry(new_entry)
+    @registry.push(new_entry)
+  end
 
-    while new_name_is_invalid(@name)
-      if @name_generator
-        @name = @name_generator.call
-      else
-        @name = new_name_using_default_generator
-      end
+  def validate_entry(new_entry)
+    # garbage sorting
+    if new_entry == '' || new_entry.nil?
+      raise "GarbageError"
+    # pattern matching
+    elsif !(new_entry =~ /[[:alpha:]]{2}[[:digit:]]{3}/)
+      raise "PatternMatchingError"
+    # redundancy matching
+    elsif @registry.include?(new_entry)
+      raise NameCollisionError
     end
-
-    @registry << @name
   end
 end
 
-robot = Robot.new
+class Robot
+  attr_accessor :name
+
+  def initialize(args = {})
+    @name_generator = assign_name_generator(args[:name_generator])
+    @registry = args[:registry]
+    @max_attempts = 10
+
+    produce_valid_name
+  end
+
+  def assign_name_generator(name_generator)
+    name_generator.nil? ? DefaultNameGenerator.new : name_generator
+  end
+
+  def generate_self_name
+    @name = @name_generator.call
+  end
+
+  def produce_valid_name
+    begin
+      generate_self_name
+      @registry.add_entry(@name)
+    rescue NameCollisionError
+      max_attempts -= 1
+      retry unless @max_attempts == 0
+    rescue StandardError => error
+      puts "Error: " << error.to_s
+    end
+  end
+end
+
+registry = Registry.new
+robot = Robot.new(registry: registry)
 puts "My pet robot's name is #{robot.name}, but we usually call him sparky."
 
 # Errors!
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -3,23 +3,43 @@ class NameCollisionError < RuntimeError; end
 class Robot
   attr_accessor :name
 
-  @@registry
+  def new_name_using_default_generator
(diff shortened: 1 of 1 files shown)


robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
 class NameCollisionError < RuntimeError; end
```
djkotowski: One file per class.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
...
+    # pattern AB123
+    "#{generate_char.call}#{generate_char.call}"\
+    "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  end
+end
+
+class Registry
+  attr_reader :registry
+
+  def initialize(_args = {})
```
djkotowski: If you don't need args, don't include them.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
...
+
+  def add_entry(new_entry)
+    validate_entry(new_entry)
+    @registry.push(new_entry)
+  end
+
+  def validate_entry(new_entry)
+    # garbage sorting
+    if new_entry == '' || new_entry.nil?
+      raise "GarbageError"
```
djkotowski: Either raise a detailed error message or a class that represents the error (preferably both).

robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
...
+    # pattern matching
+    elsif !(new_entry =~ /[[:alpha:]]{2}[[:digit:]]{3}/)
+      raise "PatternMatchingError"
+    # redundancy matching
+    elsif @registry.include?(new_entry)
+      raise NameCollisionError
+    end
```
djkotowski: Again, I might split these up to avoid the `if/elsif` syntax.  By moving each check into its own method, you also clarify what the condition is checking.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
...
   attr_accessor :name
 
-  @@registry
-
   def initialize(args = {})
-    @@registry ||= []
-    @name_generator = args[:name_generator]
+    @name_generator = assign_name_generator(args[:name_generator])
+    @registry = args[:registry]
+    @max_attempts = 10
```
djkotowski: `10` here is a magic number.  It should be defined as a constant.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
 class NameCollisionError < RuntimeError; end
 
+class DefaultNameGenerator
+  def call
```
djkotowski: Interesting way to conform to the syntax.  While this isn't the strategy I would use, it works well enough.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
...
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def produce_valid_name
+    generate_self_name
+    @registry.add_entry(@name)
+  rescue NameCollisionError
+    max_attempts -= 1
+    retry unless @max_attempts == 0
+  rescue StandardError => error
+    puts "Error: " << error.to_s
```
djkotowski: Your robot should not be responsible for console output.  That's the responsibility of the runner.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
...
   end
 end
 
-robot = Robot.new
+registry = Registry.new
+robot = Robot.new(registry: registry)
```
paulghaddad: Nice use of keyword arguments.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
...
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+    produce_valid_name
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
+  def assign_name_generator(name_generator)
+    name_generator.nil? ? DefaultNameGenerator.new : name_generator
```
paulghaddad: Could you alternatively use:

`name_generator ||= DefaultNameGenerator.new`

robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
...
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def produce_valid_name
+    generate_self_name
+    @registry.add_entry(@name)
+  rescue NameCollisionError
+    max_attempts -= 1
+    retry unless @max_attempts == 0
```
paulghaddad: `@max_attemps.zero?`

robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
...
+    validate_entry(new_entry)
+    @registry.push(new_entry)
+  end
+
+  def validate_entry(new_entry)
+    # garbage sorting
+    if new_entry == '' || new_entry.nil?
+      raise "GarbageError"
+    # pattern matching
+    elsif !(new_entry =~ /[[:alpha:]]{2}[[:digit:]]{3}/)
```
paulghaddad: I think making this regex as a well-named constant would convey its meaning better.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
...
+  def generate_self_name
+    @name = @name_generator.call
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def produce_valid_name
+    generate_self_name
+    @registry.add_entry(@name)
+  rescue NameCollisionError
```
paulghaddad: I personally feel the Registry class should handle errors raised by its methods.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
 class NameCollisionError < RuntimeError; end
 
+class DefaultNameGenerator
+  def call
+    generate_char = -> { ('A'..'Z').to_a.sample }
```
paulghaddad: You can also make lines 5 and 6 their own methods instead of using lambdas.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,74 @@
...
+    generate_num = -> { rand(10) }
+
+    # pattern AB123
+    "#{generate_char.call}#{generate_char.call}"\
```
paulghaddad: If you have `generate_char` and `generate_num` as their own methods, you avoid using `call` on both the lambdas and within this class as a separate method. I find this a little confusing. Then, if you switch the methods, you can do the following to simplify building the name:

new_name = ""
2.times { new_name.concat(('A'..'Z').to_a.sample) }
3.times { new_name.concat(rand(10).to_s) }

djkotowski:
Take a look at the comments, refactor and resubmit when ready.  Keep these in mind for your other exercises.  Let us know if you have any questions.

PR #1022, opened 2015-10-12
Robot name

athiercelin:
Edits based on Dan and Paul’s comments.


robot_name/lib/registry.rb
```diff
@@ -0,0 +1,40 @@
+require './lib/pattern_error'
```
paulghaddad: Since you are not using outside files for your errors, these `require` statements aren't needed. But extracting the errors into separate files is preferable.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,40 @@
-class NameCollisionError < RuntimeError; end
+require './lib/name_collision_error'
+require './lib/default_name_generator'
+require './lib/registry'
 
 class Robot
```
paulghaddad: A better file name is `robot.rb` to match your class name.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,40 @@
...
   attr_accessor :name
-
-  @@registry
+  MAX_ATTEMPTS = 10
 
   def initialize(args = {})
```
paulghaddad: Another pattern, preferable over an options hash in my opinion, is the use of keyword arguments. You can use default values and specify whether any are required:

Just an option (no pun intended!) for future exercises.

robot_name/lib/default_name_generator.rb
```diff
@@ -0,0 +1,16 @@
+class DefaultNameGenerator
+  def alpha_gen
+    ('A'..'Z').to_a.sample
```
paulghaddad: Another idiom you can use here is the splat to generate the array of characters:

`[*'A'..'Z']`

robot_name/lib/default_name_generator.rb
```diff
@@ -0,0 +1,16 @@
+class DefaultNameGenerator
```
paulghaddad: This class is very clean now. Nice improvements.

robot_name/robot_name.rb
```diff
@@ -1,29 +1,40 @@
...
+
+  def generate_self_name
+    @name = @name_generator.call
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def produce_valid_name
+    generate_self_name
+    @registry.add_entry(@name)
```
paulghaddad: Since `name` has an `attr_accessor` method defined on it, you don't need the `@` sign.

robot_name/lib/registry.rb
```diff
@@ -0,0 +1,40 @@
...
+  def initialize
+    @registry = []
+  end
+
+  def add_entry(new_entry)
+    validate_entry(new_entry)
+    @registry.push(new_entry)
+  end
+
+  def test_garbage_name(new_entry)
```
paulghaddad: These `test_*` methods can be private since they are called within the `Registry` class.

robot_name/lib/registry.rb
```diff
@@ -0,0 +1,40 @@
...
+    raise GarbageError, "Junk Name Error" if new_entry == '' || new_entry.nil?
```
paulghaddad: A good method for future use is ActiveSupports `#blank?` to test for nil and empty strings at the same time.

http://guides.rubyonrails.org/active_support_core_extensions.html#blank-questionmark-and-present-questionmark

robot_name/lib/registry.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  FORMAT_REGEXP = /[[:alpha:]]{2}[[:digit:]]{3}/
+
+  def initialize
+    @registry = []
+  end
+
+  def add_entry(new_entry)
+    validate_entry(new_entry)
+    @registry.push(new_entry)
```
paulghaddad: Alternatively, you can use:

`@registry << new_entry`

robot_name/lib/registry.rb
```diff
@@ -0,0 +1,40 @@
...
+  class GarbageError < RuntimeError; end
+  class CollisionError < RuntimeError; end
+  class PatternError < RuntimeError; end
+
+  attr_reader :registry
+
+  FORMAT_REGEXP = /[[:alpha:]]{2}[[:digit:]]{3}/
+
+  def initialize
+    @registry = []
```
paulghaddad: You can also use a `Set` for keeping a data structure with unique items.

robot_name/lib/registry.rb
djkotowski: However, as this is not a Rails app, it's a good idea to avoid reliance on Rails methods, including those from ActiveSupport.

paulghaddad:
@athiercelin @djkotowski Really like rework. I feel this is a great refactor. There are just a couple changes I'd like you to make before closing out this exercise.

PR #1023, opened 2015-10-12
Dinodex

athiercelin:


dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
+class Dinobase
+  attr_reader :dinosaurs
+  @debug = true
```
jmtai: Why is this variable being initialized here? It just gets overridden in the initializer later.

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
...
+  
+  def initialize(_args = {})
```
jmtai: If no arguments are being used, there's no need to include them in the method definition.

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
...
+    @dinosaurs = []
+    @debug = false
+  end
+
+  def open_and_read_file(file_path)
+    file = File.open(file_path, "r")
+    data = file.read
+    file.close
+    data
```
jmtai: Ruby's block notation gives us an easy way to eliminate the need to manually manage the file object:
```
File.open(file_path, "r") do |file|
  file.read
end
```
In this case, the file is automatically closed after the block is executed. Even better, since all we're doing is performing a file read, we can use the curly bracket block notation and write it all in a single line:
```
File.open(file_path, "r") { |file| file.read }
```

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def if_not_number_then_not_avail(potential_number)
```
jmtai: Avoid describing logic in the method name. In this case you've explicitly encoded an if-then condition in the method name itself, which means if you want to refactor the code later, you'll have to refactor the method name as well.

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
...
+    (potential_number =~ /\A[-+]?\d+\z/) ? potential_number : 'N/A'
```
jmtai: Although this regex is relatively simple, they can easily become large and difficult to understand for maintainers that did not originally write them. An easy way to improve readability is to define the regex as a constant and give it a descriptive name.
e.g.
```ruby
(define this as a class-level constant)
VALID_INTEGER_REGEX = /\A[-+]?\d+\z/
```
Then your code looks like this:
```ruby
(potential_number =~ VALID_INTEGER_REGEX) ? potential_number : 'N/A'
```

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
...
+    new_entry = {}
+    new_entry['name'] = cells[0]
+    new_entry['period'] = cells[1]
+    new_entry['continent'] = cells[2]
+    new_entry['diet'] = cells[3]
+    new_entry['weight'] = if_not_number_then_not_avail(cells[4])
+    new_entry['walking'] = cells[5]
+    new_entry['description'] = cells[6]
+
+    new_entry
```
jmtai: You can tighten this up by simply declaring the hash directly:
```ruby
  { 'name' => cells[0],
    'period' => cells[1],
    'continent' => cells[2],
    'diet' => cells[3],
    'weight' => if_not_number_then_not_avail(cells[4]),
    'walking' => cells[5],
    'description' => cells[6] }
```

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def parse_dinodex
+    csv_path = './dinodex.csv'
```
jmtai: Since this isn't being modified, it seems like a good candidate for being declared as a constant.

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
...
+    csv_content = open_and_read_file(csv_path)
+    line_count = 0
```
jmtai: Is `line_count` being used?

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
...
+    skipped_first_line = false
+    csv_content.each_line do |row|
+      # skip the header.
```
jmtai: If you make your variables a bit more descriptive, you can avoid the need for clarifying comments. How about something like `is_header_row` instead of `skipped_first_line`? Then you could do something like this:
```ruby
is_header_row = true
csv_content.each_line do |row|
  if is_header_row
    is_header_row = false
    next
  end
```

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
...
+      if skipped_first_line == false
```
jmtai: The if-not condition can be replaced with `unless` to make the code a bit more readable:
```ruby
unless skipped_first_line
  skipped_first_line = true
  next
end
```

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
...
+        skipped_first_line = true
+        next
+      end
+      new_entry = new_entry_for_dinodex_row(row)
+
+      @dinosaurs.push(new_entry)
+      line_count += 1
+    end
+    print "Just loaded " + line_count.to_s +
```
jmtai: If line count is just being used for debugging, there's no need to create a temporary variable to store it. You could just count the number of items in the @dinosaurs array.

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
...
+      " Dinosaurs from dinodex\n" if @debug
+  end
+
+  def new_entry_for_african_dinos_row(row)
```
jmtai: This method looks awfully similar to your `new_entry_for_dinodex_row` method. Perhaps there's a way to DRY (Don't Repeat Yourself) this up?

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
...
+    new_entry['diet'] = cells[2] == 'Yes' ? 'Carnivore' : 'Herbivore'
+    new_entry['weight'] = if_not_number_then_not_avail(cells[3])
+    new_entry['walking'] = cells[4]
+    new_entry['description'] =
+    'Not a lot is known about ' + cells[0] + ' at this point.'
+
+    new_entry
+  end
+
+  def parse_african_dinos
```
jmtai: Again, this method is almost identical to your `parse_dinodex` method.

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,92 @@
...
+    csv_path = './african_dinosaur_export.csv'
+    csv_content = open_and_read_file(csv_path)
+
+    line_count = 0
+    skipped_first_line = false
+    csv_content.each_line do |row|
```
jmtai: One thing you might want to take a look at is Ruby's built-in CSV library: http://apidock.com/ruby/CSV. It can handle header rows and also allows you to also define custom header and data converters.

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,245 @@
+require './dinobase.rb'
+
+class Dinocalls
```
jmtai: The Rails naming convention is for class names to be written in CamelCase and file names with underscores. So a class name of `DinoCalls` would be stored in a file named `dino_calls.rb`. We adopt this convention for plain old Ruby code as well, even though it's not strictly necessary.

In Rails, the class loader uses this convention to determine what files need to be loaded, so deviating from this can cause problems when you start writing Rails code.

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,245 @@
...
+  def filter_bipeds(dinosaurs)
+    new_dino_list = []
+
+    dinosaurs.each do |dino|
+      dino_biped_status = dino['walking']
+
+      new_dino_list.push(dino) if dino_biped_status == 'Biped'
+    end
+
+    new_dino_list
```
jmtai: Any time you see the pattern:
```
list = []
other_list.each { do stuff and store in new list }
return list
```
it's a sign that you probably want some other operation. In this case, you're selecting bipeds from the dinosaurs array, so you can use `select`:
```ruby
dinosaurs.select { |dino| dino['walking'] == 'Biped' }
```

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,245 @@
+require './dinobase.rb'
+
+class Dinocalls
+  LARGE_SIZE = 2000
```
jmtai: Good use of a constant here.

djkotowski:
Reopened so @jmtai can complete review.

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def african_dino_recipe
+    {
+      'name' => 'african',
+      'file_path' => './african_dinosaur_export.csv',
+      'recipe_proc' => proc do |row|
+        { 'name' => row[0],
+          'period' => row[1],
+          'continent' => 'N/A',
```
jmtai: Since the file is called `african_dinosaur_export.csv` it might be safe to assume that the continent is Africa.

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,64 @@
...
+          'diet' => row[2] == 'Yes' ? 'Carnivore' : 'Herbivore',
+          'weight' => number_or_na(row[4]),
+          'walking' => row[4],
```
jmtai: Hmmm... should both of these be using `row[4]`?

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,64 @@
...
+          'Not a lot is known about ' + row[0] + ' at this point.' }
+      end,
+    }
+  end
+
+  def number_or_na(potential_number)
+    (potential_number =~ VALID_INTEGER_REGEX) ? potential_number : 'N/A'
+  end
+
+  def parse(source_name:)
```
jmtai: Why is there a colon here?

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,64 @@
...
+  def dinodex_recipe
+    {
+      'name' => 'dinodex',
+      'file_path' => './dinodex.csv',
+      'recipe_proc' => proc do |row|
+        { 'name' => row[0],
+          'period' => row[1],
+          'continent' => row[2],
+          'diet' => row[3],
+          'weight' => number_or_na(row[4]),
```
jmtai: I'm not a huge fan of the naming of this method since `number_or_na` would imply to me that it could return an actual number, not just a string representing a number.

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def parse(source_name:)
+    @parsing_recipes.each do |recipe|
+      if recipe['name'] == source_name
+        CSV.foreach(recipe['file_path']) do |row|
+          @dinosaurs.push(recipe['recipe_proc'].call(row))
+        end
+        break
+      end
+    end
```
jmtai: This is getting a bit arrowhead-y. See if you can find a way to clean this up a bit.

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def parse_all
+    parse(source_name: 'african')
+    parse(source_name: 'dinodex')
```
jmtai: You could also do something like this:
```ruby
['african', 'dinodex'].each { |name| parse(source_name: name) }
```

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
+require './dinobase.rb'
+
+class Dinocatalog
+  LARGE_SIZE = 2000
+
+  def filter_bipeds(dinosaurs)
+    dinosaurs.select { |dino| dino['walking'] == 'Biped' }
```
jmtai: What happens if the source file didn't have consistent formatting? If the value is "biped" should your filter still work?

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+  end
+
+  def filter_carnivors(dinosaurs)
```
jmtai: I'm gonna assume this is a typo and you weren't trying to save space by deleting the 'e'. :smile:

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+    dinosaurs.select { |dino| dino['diet'] != 'Herbivore' }
+  end
+
+  def filter_small(dinosaurs)
+    dinosaurs.select { |dino| dino['weight'].to_i <= LARGE_SIZE }
+  end
+
+  def filter_big(dinosaurs)
+    dinosaurs.select { |dino| dino['weight'].to_i >= LARGE_SIZE }
```
jmtai: Hmmm... `filter_small` will match dinosaurs that equal `LARGE_SIZE`, and `filter_big` will also match dinosaurs that equal `LARGE_SIZE`. Does that sound right?

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+  end
+
+  def filter_period(dinosaurs, period_string)
```
jmtai: No need to include a data type in your variable names. Ruby is dynamically typed so this is a good way to confuse someone if you pass something in that behaves like a string but isn't actually a `String` object.

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+    dinosaurs.select do |dino|
+      dino['period'].downcase.include? period_string.downcase
```
jmtai: :+1: for the case insensitive filter.

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+    end
+  end
+
+  def test_if_value_matches_search(field, search_string)
```
jmtai: "Predicate methods" are methods that return `true` or `false`. In Ruby, the convention is to name these methods with an ending question mark. You can see this with the `include?` method being called on the field string.

In this case, you could rename your method something like `value_matches?` which has the additional bonus of reading more like natural language when used. Which of these looks better to you?
```ruby
return if test_if_value_matches_search(field, search_value)
```
or
```ruby
return if value_matches?(field, search_value)
```

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+    field.downcase.include?(search_string.downcase)
+  end
+
+  def test_if_dino_matches_search(dino, search_string)
+    keys_to_search = %w(name description period diet continent
```
jmtai: I'd argue that this variable name is unnecessarily long, since `keys_to_search` doesn't really provide much more information than `keys`, and since we're keeping our method short, the next line of code makes it obvious that we're iterating over the array of keys.

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+    dinosaurs.select do |dino|
+      dino['period'].downcase.include? period_string.downcase
+    end
+  end
+
+  def test_if_value_matches_search(field, search_string)
+    field.downcase.include?(search_string.downcase)
+  end
+
+  def test_if_dino_matches_search(dino, search_string)
```
jmtai: Same comment here about predicate methods.

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+    keys_to_search = %w(name description period diet continent
+                        walking weight)
+    keys_to_search.each do |key|
```
jmtai: Since you want to return true if **any** of the keyed values match the search_string, you can use the `any?` method instead. http://ruby-doc.org/core-2.2.0/Enumerable.html#method-i-any-3F

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+      return true if test_if_value_matches_search(dino[key], search_string)
+    end
+    false
+  end
+
+  def filter_search(dinosaurs, search_string)
+    dinosaurs.select { |dino| test_if_dino_matches_search(dino, search_string) }
+  end
+
+  def text_export(dinosaurs)
```
jmtai: The method name `text_export` is a little confusing. Just reading the name would make me assume that some sort of text file is being generated but it's really printing the dinosaurs to the console. Naming is hard (and often subjective) but I'd argue that this one is a bit ambiguous. Why not something like "print_info"?

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+      puts "Diet: " + dino['diet']
+      puts "Weight: " + dino['weight']
+      puts "Walking: " + dino['walking']
+      puts "Description: " + dino['description']
+      puts "---"
+    end
+    puts "Showed #{dinosaurs.count} Dinos"
+  end
+
+  def json_export(dinosaurs)
```
jmtai: Again here, I'd expect an export method to output a file or some sort of output object instead of just printing to the console.

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+    is_first_line = true
+    print "["
+    dinosaurs.each do |dino|
+      unless is_first_line
+        print ","
+        is_first_line = false
+      end
```
jmtai: Instead of using a flag here, you could do something like
```ruby
print "," if dino == dinosaurs.first
```

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+      print "{"
+      print "\"name\":" + "\"" + dino['name'] + "\","
```
jmtai: If you use single quotes for your strings, you don't have to escape the double quote characters. e.g.
```ruby
print '"name:"'
```

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+      print "\"diet\":" + "\"" + dino['diet'] + "\","
+      print "\"weight\":" + "\"" + dino['weight'] + "\","
+      print "\"walking\":" + "\"" + dino['walking'] + "\","
+      print "\"description\":" + "\"" + dino['description'] + "\""
+      print "}"
+    end
+    print "]"
+  end
+
+  def arg_does_not_exist(args)
```
jmtai: This is a predicate method, so it should probably be named something like `valid_arg?`

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+    list_of_valid_arg = ["-b",
```
jmtai: No need to prepend `list_of` to your variable name. Something like `valid_args` would indicate that the variable contains multiple items.

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+      puts "Period: " + dino['period']
+      puts "Continent: " + dino['continent']
+      puts "Diet: " + dino['diet']
+      puts "Weight: " + dino['weight']
+      puts "Walking: " + dino['walking']
+      puts "Description: " + dino['description']
+      puts "---"
+    end
+    puts "Showed #{dinosaurs.count} Dinos"
+  end
```
jmtai: For displaying all the known facts for a particular dinosaur, the requirements state the following:
```
For a given dino, I'd like to be able to print all the known facts about that dinosaur. If there are facts missing, please don't print empty values, just skip that heading. Make sure to print Early / Late etc for the periods.
```
Does this output fit the requirements?

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+    puts "-c         Sort to show only carnivores dinosaurs"
+    puts "--period period      Sort to show dinosaurs which lived during period"
+    puts "-s         Sort to only show small dinosaurs"
+    puts "-B         Sort to only show big dinosaurs"
+    puts "--search pattern    Search for pattern in the database"
+    puts "--help        Shows this usage/help message"
+    puts "--json         Export results in JSON"
+  end
+
+  def handle_posture(args, final_dinosaurs_list)
```
jmtai: The `final_dinosaurs_list` param seems oddly named, since if it can be modified by the method, it's not **really** the final list, is it?

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+  def handle_params(args, final_dinosaurs_list)
+    final_dinosaurs_list = handle_posture(args, final_dinosaurs_list)
+    final_dinosaurs_list = handle_diet(args, final_dinosaurs_list)
+    final_dinosaurs_list = handle_size(args, final_dinosaurs_list)
+    final_dinosaurs_list = handle_period(args, final_dinosaurs_list)
+    final_dinosaurs_list = handle_search(args, final_dinosaurs_list)
+    final_dinosaurs_list
+  end
+
+  def handle_usage_and_decide_if_continue(args)
```
jmtai: This is a predicate method. Can you think of a better (and more succinct) name?

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+  end
+
+  def handle_params(args, final_dinosaurs_list)
+    final_dinosaurs_list = handle_posture(args, final_dinosaurs_list)
+    final_dinosaurs_list = handle_diet(args, final_dinosaurs_list)
+    final_dinosaurs_list = handle_size(args, final_dinosaurs_list)
+    final_dinosaurs_list = handle_period(args, final_dinosaurs_list)
+    final_dinosaurs_list = handle_search(args, final_dinosaurs_list)
+    final_dinosaurs_list
+  end
```
jmtai: Each of your handler methods is independently examining the entire argument list to decide whether or not to perform a specific filtering operation. So really what you want is a mapping between the input argument and the associated method to call. Maybe something like this:
```ruby
HANDLERS = {
  '-b' => handle_posture,
  '-s' => handle_size,
  etc...
}
```
Unfortunately, this doesn't work because `handle_posture` and `handle_size` are method calls. However, Ruby lets us do this:
```ruby
HANDLERS = {
  '-b' => :handle_posture,
  '-s' => :handle_size,
  etc...
}
```
Using symbols allows us to then use another cool Ruby trick:
```ruby
handler = HANDLERS[arg]
send(handler, final_dinosaur_list)
```
Now we can just iterate through the argument list, and use the handler lookup table to call the appropriate method:
```ruby
args.each do |arg|
  handler = HANDLERS[arg]
  dinos = send(handler, dinos) if handler
end
```

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+    true
+  end
+
+  def main(args)
+    return unless handle_usage_and_decide_if_continue(args)
+    dinodatabase = Dinobase.new
+
+    dinodatabase.parse_all
+
+    final_dinosaurs_list = Array.new(dinodatabase.dinosaurs)
```
jmtai: Again here, try to use descriptive and accurate variable names. This is really the full dinosaur list, not the **final** dinosaur list, since the full list may get filtered depending on the command line parameters.

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,196 @@
...
+    final_dinosaurs_list = handle_params(args, final_dinosaurs_list)
+    handle_export(args, final_dinosaurs_list)
+  end
+end
+
+# creating a new dinocall and passing args
+dinocall = Dinocatalog.new
+dinocall.main(ARGV)
```
jmtai: So this class is doing a LOT of stuff. It contains a runner for the app, parses command line input, filters the dinosaur data, outputs results to the console, etc. We want our classes to follow the Single Responsibility Principle, so from that description we can see that this principle is being violated.

If you haven't already, skim through (or read, it's a short book) Practical Object Oriented Design in Ruby by Sandi Metz. After you've done that, take a look at this class and try breaking it up into appropriate pieces. Additionally, think about creating a `Dinosaur` class instead of manipulating raw data. Ruby is an object oriented language, so take advantage of that fact to make your code more readable and maintainable.

jmtai:
Nice first pass! Your code is very clear and easy to follow. Read over the comments and take a stab at refactoring the classes to adhere to the Single Responsibility Principle and resubmit when ready.

paulghaddad:
@athiercelin Take a look at Josh's comments and submit a second iteration when ready. Nice work!

PR #1027, opened 2015-10-14
Triangle facts

drosholt:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+  def equilateral?
+    @side1 == @side2 && @side2 == @side3
+  end
+
+  def isosceles?
+    [side1, side2, side3].uniq.length == 2
+  end
+
+  def scalene?
+    !equilateral? && !isosceles?
```
paulghaddad: Or: `!(equilateral? || isosceles?)`

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+  end
+
+  def recite_facts
+    puts 'This triangle is equilateral!' if equilateral?
+    puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles?
+    puts 'This triangle is scalene and mathematically boring.' if scalene?
+
+    angles = calculate_angles
```
paulghaddad: If you rename `calculate_angles` to `angles`, you can remove this line altogether.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+    puts "The angles of this triangle are #{angles.join(', ')}."
+
+    puts 'This triangle is also a right triangle!' if angles.include? 90
+    puts ''
```
paulghaddad: A simple `puts` is equivalent.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+
+  def recite_facts
+    puts 'This triangle is equilateral!' if equilateral?
+    puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles?
+    puts 'This triangle is scalene and mathematically boring.' if scalene?
+
+    angles = calculate_angles
+    puts "The angles of this triangle are #{angles.join(', ')}."
+
+    puts 'This triangle is also a right triangle!' if angles.include? 90
```
paulghaddad: I'd personally be consistent on using parentheses for method parameters. If this case, I'd recommend enclosing the `90` in parentheses.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+
+  def isosceles?
+    [side1, side2, side3].uniq.length == 2
+  end
+
+  def scalene?
+    !equilateral? && !isosceles?
+  end
+
+  def recite_facts
```
paulghaddad: The only method that truly needs to be public for this exercise is `recite_facts`, but I agree that `equilateral?`, `isosceles?` and `scalene` should be public-facing.

paulghaddad:
@drosholt This is a very nice first iteration. Just a couple minor points to rework in a second iteration and you'll be good. Also, if you could name your PRs in the following fashion, that helps me out: Triangle Facts - V1.

triangle_facts/triangle_facts.rb
drosholt: Yes, they are mathematically equivalent.  I opted for the separated approach simply because it reads better in English, and so is a little faster to parse for someone who may be new to the code.  There is a performance difference as well.

triangle_facts/triangle_facts.rb
drosholt: Though that would have an equivalent functional result.  But since calculate_angles does work (and sqrt and acos are not cheap), we avoid recalculating the results for lines 29 and 31 if we store the result in a variable.

triangle_facts/triangle_facts.rb
drosholt: The way that the code is written makes it extremely unclear what the actual public interface should be.  I essentially left it the way it was (except for moving calculate_angles to private) so that I wasn't changing the interface at all.

It doesn't make sense to overthink this, though, because if I was actually writing this code, I would separate recite_facts and and it's private methods from the base definition of a Triangle.  If I think about what the definition of a Triangle is, recite_facts doesn't fit into that definition, so it belongs outside of a Triangle data structure.

triangle_facts/triangle_facts.rb
paulghaddad: Great point there.

triangle_facts/triangle_facts.rb
paulghaddad: Yes, indeed...it should in another class altogether. Good point.

PR #1028, opened 2015-10-15
Dinodex

athiercelin:
Changes based on comments.
Kept the filtering/handling in catalog as it made sense to me that the
‘smartness’ of the data processing remained there

### DIFF
--- a/dino_catalog/dinobase.rb
+++ b/dino_catalog/dinobase.rb
@@ -1,4 +1,5 @@
 require 'csv'
+require './dinosaur'
 
 class Dinobase
   attr_reader :dinosaurs
@@ -14,13 +15,13 @@ class Dinobase
       'name' => 'dinodex',
       'file_path' => './dinodex.csv',
       'recipe_proc' => proc do |row|
-        { 'name' => row[0],
-          'period' => row[1],
-          'continent' => row[2],
-          'diet' => row[3],
-          'weight' => number_or_na(row[4]),
-          'walking' => row[5],
-          'description' => row[6] ? row[6] : 'N/A' }
+        Dinosaur.new(name: row[0],
+          period: row[1],
+          continent: row[2],
+          diet: row[3],
+          weight: evaluated_weight(row[4]),
+          walking: row[5],
+          description: row[6] ? row[6] : 'N/A')
       end,
     }
   end
@@ -30,35 +31,33 @@ class Dinobase
       'name' => 'african',
       'file_path' => './african_dinosaur_export.csv',
       'recipe_proc' => proc do |row|
-        { 'name' => row[0],
-          'period' => row[1],
-          'continent' => 'N/A',
-          'diet' => row[2] == 'Yes' ? 'Carnivore' : 'Herbivore',
-          'weight' => number_or_na(row[4]),
-          'walking' => row[4],
-          'description' =>
-          'Not a lot is known about ' + row[0] + ' at this point.' }
+        Dinosaur.new(name: row[0],
+          period: row[1],
+          continent: 'Africa',
+          diet: row[2] == 'Yes' ? 'Carnivore' : 'Herbivore',
+          weight: evaluated_weight(row[3]),
+          walking: row[4],
+          description: 'N/A')
       end,
     }
   end
 
-  def number_or_na(potential_number)
+  def evaluated_weight(potential_number)
     (potential_number =~ VALID_INTEGER_REGEX) ? potential_number : 'N/A'
   end
 
-  def parse(source_name:)
-    @parsing_recipes.each do |recipe|
-      if recipe['name'] == source_name
-        CSV.foreach(recipe['file_path']) do |row|
-          @dinosaurs.push(recipe['recipe_proc'].call(row))
-        end
-        break
-      end
+  def parse(source_name)
+    recipe = @parsing_recipes.select do |stored_recipe|
+      stored_recipe['name'] == source_name 
+    end
+    return if recipe.empty?
+
+    CSV.foreach(recipe[0]['file_path'], :headers => :first_row) do |row|
+      @dinosaurs.push(recipe[0]['recipe_proc'].call(row))
     end
   end
 
   def parse_all
-    parse(source_name: 'african')
-    parse(source_name: 'dinodex')
+    %w(african, dinodex).each { |name| parse(name) }
   end
 end
--- a/dino_catalog/dinobase.rb
+++ b/dino_catalog/dinobase.rb
@@ -1,92 +1,64 @@
+require 'csv'
+
 class Dinobase
   attr_reader :dinosaurs
-  @debug = true
-  
-  def initialize(_args = {})
+  VALID_INTEGER_REGEX = /\A[-+]?\d+\z/
+
+  def initialize
     @dinosaurs = []
-    @debug = false
+    @parsing_recipes = [dinodex_recipe, african_dino_recipe]
   end
 
-  def open_and_read_file(file_path)
-    file = File.open(file_path, "r")
-    data = file.read
-    file.close
-    data
+  def dinodex_recipe
+    {
+      'name' => 'dinodex',
+      'file_path' => './dinodex.csv',
+      'recipe_proc' => proc do |row|
+        { 'name' => row[0],
+          'period' => row[1],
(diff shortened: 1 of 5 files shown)


dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,121 @@
...
+    end
+    dinosaurs
+  end
+
+  def args_invalid?(args)
+    if args.include?("--help") || !valid_arg?(args)
+      Dinoshow.print_usage
+      return true
+    end
+    false
```
djkotowski: You could remove the if-block by reversing the process:
```ruby
return false unless <condition>

Dinoshow.print_usage
true
```

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  def main(args)
+    return if args_invalid?(args)
+    dinodatabase = Dinobase.new
```
djkotowski: Naming this `dino_database` could improve readability a little.

dino_catalog/dinoshow.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def self.valid_data?(data)
+    return false if data.nil? || data == 'N/A' || data.empty?
+    true
+  end
+
+  def self.print_text(dinosaurs)
+    dinosaurs.each do |dino|
+      puts "Name: " + dino.name if valid_data?(dino.name)
```
djkotowski: Prefer `<<` over `+` when concatenating _unless_ the leftmost string should not be modified.

dino_catalog/dinoshow.rb
```diff
@@ -0,0 +1,52 @@
...
+      puts "Diet: " + dino.diet if valid_data?(dino.diet)
+      puts "Weight: " + dino.weight if valid_data?(dino.weight)
+      puts "Walking: " + dino.walking if valid_data?(dino.walking)
+      puts "Description: " + dino.description if valid_data?(dino.description)
+      puts "---"
+    end
+    puts "Showed #{dinosaurs.count} Dinos"
+  end
+
+  def self.print_json(dinosaurs)
```
djkotowski: This is dangerous, as you do not escape characters.  Take a look at using the JSON module within the standard library.

dino_catalog/dinoshow.rb
```diff
@@ -0,0 +1,52 @@
...
+    puts "ruby dinocatalog.rb <options>"
+    puts "Options are: "
+    puts "-b         Sort to show only bipeds dinosaurs"
+    puts "-c         Sort to show only carnivores dinosaurs"
+    puts "--period period      Sort to show dinosaurs which lived during period"
+    puts "-s         Sort to only show small dinosaurs"
+    puts "-B         Sort to only show big dinosaurs"
+    puts "--search pattern    Search for pattern in the database"
+    puts "--help        Shows this usage/help message"
+    puts "--json         Export results in JSON"
```
paulghaddad: Excellent guidance for users.

dino_catalog/dinoshow.rb
paulghaddad: Agree, the JSON library will make this a simple call.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,29 @@
+class Dinosaur
+  attr_accessor :name
+  attr_accessor :period
+  attr_accessor :continent
+  attr_accessor :diet
+  attr_accessor :weight
+  attr_accessor :walking
+  attr_accessor :description
+
+  def initialize(name: 'Trollosaurus Rex',
```
paulghaddad: Nice job with the keyword arguments, but I don't think the defaults are necessary.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,29 @@
+class Dinosaur
```
paulghaddad: Nice contained class.

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,63 @@
+require 'csv'
+require './dinosaur'
+
+class Dinobase
+  attr_reader :dinosaurs
+  VALID_INTEGER_REGEX = /\A[-+]?\d+\z/
+
+  def initialize
+    @dinosaurs = []
+    @parsing_recipes = [dinodex_recipe, african_dino_recipe]
```
paulghaddad: Interesting way to do this. I haven't seen this before. :)

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,63 @@
+require 'csv'
+require './dinosaur'
+
+class Dinobase
+  attr_reader :dinosaurs
+  VALID_INTEGER_REGEX = /\A[-+]?\d+\z/
```
paulghaddad: Good use of a constant for this regex. Quick style nit-pick: a line break between the `attr_reader` and the regex would be nice.

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def parse(source_name)
+    recipe = @parsing_recipes.select do |stored_recipe|
+      stored_recipe['name'] == source_name 
+    end
+    return if recipe.empty?
+
+    CSV.foreach(recipe[0]['file_path'], :headers => :first_row) do |row|
+      @dinosaurs.push(recipe[0]['recipe_proc'].call(row))
```
paulghaddad: You can also use `<` instead of `push`. Either works, but the former is preferred.

dino_catalog/dinobase.rb
```diff
@@ -0,0 +1,63 @@
...
+          description: 'N/A')
+      end,
+    }
+  end
+
+  def evaluated_weight(potential_number)
+    (potential_number =~ VALID_INTEGER_REGEX) ? potential_number : 'N/A'
+  end
+
+  def parse(source_name)
```
paulghaddad: Consider splitting this up into two methods.

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,121 @@
...
+                "-c",
+                "--period",
+                "-s",
+                "-B",
+                "--search",
+                "--help",
+                "--json"]
+
+  def filter_bipeds(dinosaurs, _args)
+    dinosaurs.select { |dino| dino['walking'].casecmp('Biped') == 0 }
```
paulghaddad: You can use `.zero?` in place of `==  0`.

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,121 @@
...
+  def filter_small(dinosaurs, _args)
+    dinosaurs.select { |dino| dino['weight'].to_i < LARGE_SIZE }
+  end
+
+  def filter_big(dinosaurs, _args)
+    dinosaurs.select { |dino| dino['weight'].to_i >= LARGE_SIZE }
+  end
+
+  def filter_period(dinosaurs, period)
+    return [] if period.nil?
```
paulghaddad: Or: `return [] unless period`

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,121 @@
...
+    end
+  end
+
+  def value_matches?(field, search_terms)
+    field.downcase.include?(search_terms.downcase)
+  end
+
+  def dino_matches?(dino, search_terms)
+    return [] if search_terms.nil?
+    keys = %w(name description period diet continent
```
paulghaddad: Nice way to generate an array.

dino_catalog/dinocatalog.rb
```diff
@@ -0,0 +1,121 @@
...
+    args.each do |arg|
+      if arg.start_with?("-") && !VALID_ARGS.include?(arg)
+        print "*** Invalid Option " + arg + " ***\n"
+        return false
+      end
+    end
+    true
+  end
+
+  def handle_period(dinosaurs, args)
```
paulghaddad: I like your way in taking in the args and calling the appropriate method.

paulghaddad:
@athiercelin Nice job on this iteration. Your code is easy to follow, which is great. The only big comment I have is to make methods private if they aren't called from outside the class.

PR #1030, opened 2015-10-16
Arrowhead - V1

drosholt:


arrowhead/arrowhead.rb
```diff
@@ -1,33 +1,30 @@
...
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+    raise "Unknown region #{region}, please provide a valid region." unless CLASSIFICATIONS.include?(region)
+
+    shapes = CLASSIFICATIONS[region]
+    raise "Unknown shape #{shape}, please provide a valid shape." unless shapes.include?(shape)
```
drosholt: The community seems to be consolidating around a longer line length (in my opinion, rightfully so).  https://github.com/bbatsov/ruby-style-guide/issues/207.  I can make this line shorter, but it's at the cost of readability.  I'd rather adhere to the spirit of the law instead of the letter.

arrowhead/arrowhead.rb
```diff
@@ -1,33 +1,32 @@
...
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
+    unless CLASSIFICATIONS.include?(region)
+      raise "Unknown region #{region}, please provide a valid region."
     end
+
+    shapes = CLASSIFICATIONS[region]
+    raise "Unknown shape #{shape}, please provide a valid shape." unless shapes.include?(shape)
```
drosholt: The community seems to be consolidating around a longer line length (in my opinion, rightfully so). bbatsov/ruby-style-guide#207. I can make this line shorter, but it's at the cost of readability. I'd rather adhere to the spirit of the law instead of the letter.

djkotowski: One possible solution would be to extract the error message out to a constant.  However, when the line length warning is a result of string length rather than complexity, I'm not too concerned.

djkotowski:
Nice work.  This is done.

arrowhead/arrowhead.rb
drosholt: I'm actually a big fan of moving strings to constants, though constants are uglier in Ruby than in other languages, so my opinion there may change.  The issue here is that it's an interpolated string, so I'd either need to do 2 separate string constants, use a proc/lambda, or run an class_eval statement against the string.  Those are all worse options.  :(

PR #1031, opened 2015-10-19
Robot name - V1

drosholt:


robot_name/default_robot_name_generator.rb
```diff
@@ -0,0 +1,23 @@
+class DefaultRobotNameGenerator < Proc
```
djkotowski: Inheriting from a proc this way is bad practice.

robot_name/default_robot_name_generator.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def self.new
+    super() do
+      "#{random_char}#{random_char}#{random_num}#{random_num}#{random_num}"
+    end
+  end
+
+  class << self
```
djkotowski: Prefer `def self.method` to `class << self` style.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,40 @@
-class NameCollisionError < RuntimeError; end
+require './robot_name_registry'
```
djkotowski: Prefer `require_relative`.

robot_name/default_robot_name_generator.rb
drosholt: Can you point me to some documentation as to why?  And what alternatives are there besides defining a class variable that's a proc?

robot_name/default_robot_name_generator.rb
drosholt: There doesn't seem to be a clear consensus on this discussion, and the style guide explicitly states that both approaches are okay.  Can you point me to some documentation that clearly states why one approach is better than the other?

robot_name/default_robot_name_generator.rb
djkotowski: You could define the generator in your Robot as a method, then retrieve the method via `method(:test)`.

*Some docs:*
- http://learnrubythehardway.org/book/ex44.html
- http://ducktypo.blogspot.com/2010/08/why-inheritance-sucks.html

robot_name/default_robot_name_generator.rb
djkotowski: It's the preferred approach at Enova.  Also:  https://github.com/bbatsov/ruby-style-guide

The primary reason is it helps you spot right away what is a class method and what is an instance method without relying on indentation or searching for `class << self`.

robot_name/default_robot_name_generator.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    def random_num
+      rand(10)
+    end
+  end
+
+  private_class_method :random_char, :random_num
+
+  self
```
djkotowski: This is unnecessary.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,40 @@
...
-    @name_generator = args[:name_generator]
+  private_constant :VALID_NAME_REGEXP
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  def initialize(name_generator = DefaultRobotNameGenerator.new,
+                 registry = RobotNameRegistry.instance)
```
djkotowski: Using a class instance variable would also be acceptable.

robot_name/default_robot_name_generator.rb
paulghaddad: Agree with Dan: https://github.com/thoughtbot/guides/tree/master/style/ruby

`self.method` is preferred.

robot_name/default_robot_name_generator.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def self.new
+    super() do
+      "#{random_char}#{random_char}#{random_num}#{random_num}#{random_num}"
+    end
+  end
+
+  class << self
+    def random_char
+      [*'A'..'Z'].sample
```
paulghaddad: Good!

robot_name/default_robot_name_generator.rb
paulghaddad: Agree.

robot_name/robot_name_registry.rb
```diff
@@ -0,0 +1,23 @@
...
+require 'singleton'
+
+class NameCollisionError < RuntimeError
+  def initialize(name)
+    super("Robot name #{name} is already in use.")
+  end
+end
+
+class RobotNameRegistry
+  include Singleton
```
paulghaddad: Why are you including this?

robot_name/robot_name_registry.rb
```diff
@@ -0,0 +1,23 @@
...
+  def initialize
+    @registry = Set.new
+  end
+
+  def <<(name)
+    fail NameCollisionError, name if @registry.include?(name)
+    @registry << name
+  end
+
+  self
```
paulghaddad: Is this necessary?

robot_name/robot_name.rb
```diff
@@ -1,32 +1,40 @@
...
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
+    @name = name_generator.call
```
paulghaddad: See if you can refactor the `DefaultRobotNameGenerator` class to simply return an instance of a method. This will eliminate you using `#call` here. Then you can use instance methods instead of class methods.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,40 @@
...
-  attr_accessor :name
+  attr_reader :name
 
-  @@registry
+  VALID_NAME_REGEXP = /\A[[:alpha:]]{2}[[:digit:]]{3}\Z/
 
-  def initialize(args = {})
-    @@registry ||= []
-    @name_generator = args[:name_generator]
+  private_constant :VALID_NAME_REGEXP
```
paulghaddad: I don't believe this is necessary.

robot_name/robot_name_registry.rb
```diff
@@ -0,0 +1,23 @@
+require 'set'
+require 'singleton'
+
+class NameCollisionError < RuntimeError
```
paulghaddad: I'd extract this class into its own file.

robot_name/robot_name.rb
djkotowski: There isn't a big benefit to including it.  It was introduced in Ruby 1.9.3 and provides little protection.

robot_name/default_robot_name_generator.rb
paulghaddad: See my comment below. I suggest refactoring this class to create instances of a name instead of calling a proc. It will allow you to use instance methods instead and simplify your code.

paulghaddad:
@drosholt @djkotowski Please take a look at our comments and refactor. I really think getting away from calling a proc to generate your names will simplify your code throughout.

robot_name/robot_name_registry.rb
drosholt: So that there can be only one version of that Registry.  People can go ahead and create other registries if they want, but I figured that since the original registry was a class variable, this was the equivalent correct OO approach.

robot_name/robot_name_registry.rb
drosholt: No,  not at all.  I was just being persnickety because I'm not a big fan of ruby evaluating class definitions to the last declared method.  I fully expected to get called out on it.  :)  My only excuse is that it was a Monday.

robot_name/robot_name.rb
drosholt: I'm creating a proc on purpose.  I'd like the interface of the default generator to match the interface that the Robot class expects: something that it can execute 'call' on.  This way, the naming logic is very simple: either provide your own proc, or the default one will be used.

robot_name/robot_name.rb
drosholt: I realize that it may not be in widespread use, but I kind of think that it should be.  Regardless of how often we look at the constants inside of classes, classes should still only expose the absolute bare minimum that we need to.  In this particular case, for example, we would expose outside of the class that we are leveraging regular expressions in order to do our validation.  Not the biggest deal here, but nobody outside the class should know or care about it.

robot_name/robot_name.rb
djkotowski: As an alternative, you could also have a PORO that has a `#call` method that does this same thing.  Then, it doesn't even need to be a proc.

djkotowski: As a side note, it's often considered an anti-pattern to inherit from Ruby core classes.  See this [Ruby Rogues podcast](https://devchat.tv/ruby-rogues/032-rr-ruby-antipatterns)

drosholt: We just talked about this, but I'll post the comment anyway.

That's a good idea, and I like it.

I wanted to understand the performance impacts of duck typing a class, and directly using a proc, so I ran some benchmarks against those two approaches.  Interestingly enough, using procs is 50% slower than this duck-typed approach, so I think that's a clear winner.

I think that this highlights some of the challenges with ruby - that the internals can be pretty funky because of the way it evolved.  For example, https://gist.github.com/mislav/4508988.

PR #1032, opened 2015-10-20
Robot name - V2

drosholt:


robot_name/default_robot_name_generator.rb
```diff
@@ -0,0 +1,20 @@
+class DefaultRobotNameGenerator
+
```
ssapra: Consider using the ruby gem rubocop locally so you can catch these small warnings before you commit.

robot_name/default_robot_name_generator.rb
```diff
@@ -0,0 +1,20 @@
...
+  CHARACTER_ARRAY = [*'A'..'Z']
+
+  private_constant :CHARACTER_ARRAY
+
+  def call
+    "#{random_char}#{random_char}#{random_num}#{random_num}#{random_num}"
```
ssapra: This is fine as is, but if the rules changed the frequency of characters and digits, this could get unwieldy. Maybe a cleaner approach could be to specify the count of each type of character that's needed. Just something to think about.

robot_name/default_robot_name_generator.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  private
+  def random_char
+    CHARACTER_ARRAY.sample
```
ssapra: Another possibility could be `(65 + rand(26)).chr`. This feels and looks more like the random_num method below. Or to make the random_num method more like this, I would make the 10 a constant as well to show that these are values that are not supposed to be tweaked carelessly.

robot_name/robot.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  VALID_NAME_REGEXP = /\A[[:alpha:]]{2}[[:digit:]]{3}\Z/
+
+  private_constant :VALID_NAME_REGEXP
+
+  def initialize(name_generator = DefaultRobotNameGenerator.new,
+                 registry = RobotNameRegistry.instance)
+
+    @name = name_generator.call
+    fail InvalidNameFormatError, @name unless @name =~ VALID_NAME_REGEXP
```
ssapra: Personally, I think it's easier to read the error message right next to the error instead of needing the initialize method to describe the error, especially one as simple as this.

ssapra:
Great work refactoring the code with Paul and Dan's feedback! You're already demonstrating a strong command of the language.

paulghaddad:
@drosholt Great work on this refactoring. You are good to go!

PR #1033, opened 2015-10-23
Dino catalog

drosholt:
Run dino_dex.rb


dino_catalog/operation_choice.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  private
+
+  def print_choices
+    @operations.each_with_index { |f, i| puts "#{i + 1}. #{f.message}" }
+  end
+
+  def valid_operation_index
+    Kernel.loop do
```
johnmcconnell: So `Kernel` is a module which is included in every ruby object. Including the "main" object, so you do not need to call `Kernel.loop`. Using `loop` will be just fine. I guess with this code you are ensuring the if anyone defined `loop` in global scope you are using the `Kernel.loop`.

dino_catalog/custom_filter.rb
```diff
@@ -0,0 +1,55 @@
+class CustomFilter
+  HASH_SPLIT_REGEXP = /\s*(?:,|:)\s/
```
johnmcconnell: That is a really well defined regex!

johnmcconnell: Non-capture group and everything

dino_catalog/custom_filter.rb
```diff
@@ -0,0 +1,55 @@
...
+  private_constant :HASH_SPLIT_REGEXP
+
+  def initialize(description)
```
johnmcconnell: Okay, the responsibility of this class is to filter the items. Why are we hard coding `stdin`? Let the user pass the buffer they want into the class.

dino_catalog/custom_filter.rb
```diff
@@ -0,0 +1,55 @@
...
+      reject_item_for_field?(field_value, filter_value)
+    end
+  end
+
+  def reject_item_for_field?(field_value, filter_value)
+    if field_value.kind_of?(Array)
+      !field_value.include?(filter_value)
+    else
+      field_value.nil? || field_value.to_s != filter_value.to_s
+    end
```
johnmcconnell: This can all be shorten to.
```ruby
def reject_item_for_field?(field_value, filter_value)
  !Array(field_value).include?(filter_value)
end
```

johnmcconnell: On a side not defining methods based on positives and using the `!` operator is easier to read then defining methods for negation.

For example:
```ruby
def valid
   true
end

if !valid
  put 'yay!'
end
```
is easier to read than:
```ruby
```ruby
def invalid
   false
end

if invalid
  put 'yay!'
end
```

johnmcconnell: Maybe that is not the best example, but trust me

johnmcconnell: Or don't trust me, up to you

dino_catalog/dino_dex.rb
```diff
@@ -0,0 +1,23 @@
...
+
+class DinoDex
+  def initialize
+    dinodex_dinosaurs = DinosaurDeserializer::CSV.deserialize('dinodex.csv')
+
+    african_dinosaurs = DinosaurDeserializer::CSV.deserialize('african_dinosaur_export.csv',
+                                                               :african)
+
+    @dinosaurs = dinodex_dinosaurs + african_dinosaurs
+  end
```
johnmcconnell: Why are we hard coding the file location and the parsing logic inside the object initialization?

johnmcconnell: You forcing any consumer of this class to specifically use the dinosaurs located in the files are that location.
Why not let anyone pass in the dinosaurs they have currently in the program?
I suspect you already know this, but not having a simple initializer breaks dependency injection.

johnmcconnell:
```ruby
def initialize(dinos1, dinos2)
   @dinosaurs = dinos1
   @african_dinos = dinos2
end
```

dino_catalog/dinosaur_operations.rb
```diff
@@ -0,0 +1,70 @@
...
+  CUSTOM_QUERY_DESCRIPTION = 'Custom query (e.g. "diet: Carnivore, weight_in_lbs: 2200").'
+  private_constant :TWO_TONS_IN_LBS, :CUSTOM_QUERY_DESCRIPTION
+
+  def self.operations
+    OperationChoice.new(*basic_filters, *nested_operations, *io_operations)
+  end
+
+  def self.basic_filters
+    [
+      Filter.new('Filter by bipeds.', proc { |d| d.walking == Dinosaur::WALKING_BIPED }),
```
johnmcconnell: FYI, I don't really care if you use proc or lambda. There are some differences but I forget. To use lambdas:
```ruby
->() { puts 'hello' }
```

```ruby
->() do
  put 'hello'
end
```
Just FYI

johnmcconnell: Okay I found one difference, with how return statements are evaluated:
http://stackoverflow.com/questions/626/when-to-use-lambda-when-to-use-proc-new

dino_catalog/dinosaur_operations.rb
```diff
@@ -0,0 +1,70 @@
...
+require_relative 'dinosaur'
+require_relative 'filter'
+require_relative 'operation'
+require_relative 'operation_choice'
+require 'json'
+require 'pp'
+
+class DinosaurOperations
+  TWO_TONS_IN_LBS = 4000
+  CUSTOM_QUERY_DESCRIPTION = 'Custom query (e.g. "diet: Carnivore, weight_in_lbs: 2200").'
```
johnmcconnell: okay if you want to have a bunch of methods on a class you can save yourself some typing with:
```ruby
class Hello
   class << self
      def a
          2
       end
       def b
           3
        end
    end
end

Hello.a #=> 2
Hello.b #=> 3
```

Although, I strongly recommend not using class methods unless you want some "helper" methods. How much harder is it to make those methods normal instance methods and pass a instance of your class? Now, if those methods now need to track state, you don't have global methods mutating state because everything is encapsulated in an instance!

johnmcconnell: But yes pointer lookup is slow

dino_catalog/operation_choice.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'operation'
+
+class OperationChoice
```
johnmcconnell: Nice... Kind of like a weird version of the interpreter design pattern.

dino_catalog/dinosaur_deserializer.rb
```diff
@@ -0,0 +1,39 @@
...
+  #
+  # This Hash holds the built-in dinosaur formats that can be accessed by name.
+  # You can select the formats through the parameter passed to CSV::new().
+  #
+  # This Hash is intentionally left unfrozen and users should feel free to add
+  # values to it.
+  #
+  # <b><tt>:dinodex</tt></b>::      Parses data in the dinodex format.
+  # <b><tt>:african</tt></b>::      Parses data in the african format.
+  #
```
johnmcconnell: Ahhh... I see you went enterprise with this initial comment and then quickly dropped the habit lol

johnmcconnell: I do like it. And you mention you intentionally stated you are leaving it unfrozen, so you are thinking of the details.

johnmcconnell:
Okay so this is a really neat PR and I think it has a nice design. I feel like you are really close to making code which is just the library and code which consumes that library! I would like to see you finish that.

What do I mean by that?
You have this FORMATS constant here https://github.com/jmmastey/level_up_exercises/pull/1033#discussion-diff-42884034R17. You know only your app cares about it and it is pure business logic. And you have those static file names.

Move all those things into variables that get passed in at instantiation and you will have a really decoupled application.

For example:

```ruby
FORMATS = { dinodex: proc { |array| Dinosaur.new(array.to_hash) } } // blah...
def my_app
  deserializer = DinosaurDeserializer.new(FORMATS, 'african_dinosaurs.csv')
  choices = OperationChoice.new(*basic_filters, *nested_operations, *io_operations)
  filter = Filter.new($stdin.readline.chomp, choices)
  filter.call(deserializer.dinos)
end
```

I know I am not using your current API 100% correctly, but I think you know where I am going with that.

dino_catalog/dinosaur_operations.rb
drosholt: That is my philosophy as well, but my time working with Spark, which (relatively) strictly enforces functional, immutable behavior, has led me to believe that there are times when it is appropriate to leverage class (static) variables and methods.  My general rules for when to use class variables/methods are:
* Is it an immutable fact of the universe?  There's no need to create an instance method for Math.cos.
* Is it a first class function that will be passed around?  These have DI capabilities built in, and so don't need to be classes.  (This is at the heart of why FP advocates claim that most design patterns aren't necessary in purely functional languages.)
* Does it not contain business logic?
* Is it essentially a CONSTANT, but just a little bit too complex to make an actual constant?

Where I had class variables/methods, I felt that it adhered to one of these, and so I was comfortable with the tradeoff of DI, testability, and ease of access.

Also, as a general rule, your code should not have state (I'd like to say never, but that's a little too draconian, of course.)

dino_catalog/custom_filter.rb
drosholt: If we were at the testing portion of level-up, then I would have had to abstract away some of the input/output logic so that I could do appropriate testing (because testing against puts is not the best way to do it).  Fortunately, we're not there yet.  :)

Ultimately, I'm not sure that your point actually changes that much.  No matter what, some class would have needed to do the output and the input gathering, and something would have needed to   understand that we're dealing with stdin and stdout.  I could have extracted the string-> hash conversion, but it wasn't really that complicated, and for a problem of this size, I wasn't too worried about coupling it to stdin and stdout.  For a more complex project, or one that I had to support long-term (or that I had to test), then I would do so, but it felt like overkill here.

dino_catalog/custom_filter.rb
drosholt: Ruby is a strange language.  I would have sworn that my code would have better performance than your yours, but it's not the case when I benchmarked it.  Something to do with the inherent C code, I'm sure.

I'm going to respond to your other points with a question.  Is there an inherent performance advantage to doing the reject my way, rather than yours?

dino_catalog/dino_dex.rb
drosholt: That's exactly what I'm doing, and it was actually a conscious choice.  This class is the endpoint, and I wanted it to be a clean API from the command-line to the execution of the code.  Clean in the consumer-driven design sense.

If someone already has their dinosaurs, then they can just execute `operations.call(dinosaurs) while dinosaurs.any?` themselves.

Where I would actually take this, in a larger project, is to use command-line arguments.  For example, `ruby dino_dex.rb -a african_dinosaur_export.csv -d dinodex.csv`.  That way, someone could pass in as many files as they wanted, from wherever they wanted, and it would handle it just fine.

dino_catalog/dinosaur_deserializer.rb
drosholt: I actually stole much of this text directly from the ::CSV class and how it manages converters.  I was wondering how whoever was going to review this was going to interpret it.  :)

drosholt:
I had something similar in my original solution.  I changed it purely to see how the reviewer would respond to the code that I cribbed from ::CSV.  I can change it back, but I'm not sure that it's that valuable.

dino_catalog/dinosaur_operations.rb
drosholt: It is worth noting that the behavior of the "proc" method changed from being equivalent to "lambda" to being equivalent to "Proc.new"

johnmcconnell: lol, I did not know that! I'm glad it is no longer lambda.

dino_catalog/dinosaur_operations.rb
johnmcconnell: Hmmm... interesting

dino_catalog/custom_filter.rb
johnmcconnell: I was definitely not thinking of the performance when I wrote it. I suspect it has to do with the way it has to follow the inheritance tree. I really have no idea though (the Array method is a C binding). This is rubinius's implementation https://github.com/rubinius/rubinius/blob/4eec62a07d6b3cbdfee3c4ce0b6e6a18498641a4/kernel/common/kernel.rb#L2

paulghaddad:
@drosholt Really nice job on this exercise. No need to resubmit. I like the debate with @johnmcconnell :)

I learned some things reviewing this.

dino_catalog/custom_filter.rb
drosholt: I took a look at the C code.  There's a couple of reasons behind the performance difference.  Ruby keeps a bitmask of each object that identifies it as one of Ruby's "primitives": Object, Float, Array, Hash, nil, etc.  The Kernel#Array method determines directly (much like it does in the Rubinius code) if it is an Array, and then proceeds accordingly.  Using kind_of? iterates through the set of primitives before it can identify which one it is.  It does this, like you expected, so that it can find the correct ancestor information for each class.  Then it does the ancestor comparison and goes from there.

So, I think that it's a combination of a two things:
1. Kernel#Array only does one type comparison, and it's for the raw c metadata
2. kind_of? is also comparing raw metadata, but it's iterating through a series of possible options first, and only then does it make the comparison.

PR #1034, opened 2015-10-26
Dinodex

athiercelin:
Multiple fixes based on comments.


paulghaddad:
@athiercelin Thanks for submitting this again. It was a good submission and this makes it even better! :)

PR #1038, opened 2015-11-02
Data science

athiercelin:
Sorry for the past injection of dinodex stuff.


data_science/data_science.rb
```diff
@@ -0,0 +1,135 @@
...
+  attr_reader :a_conv_count
+  attr_reader :avg_rate_a
+  attr_reader :rate_a_low
+  attr_reader :rate_a_high
+  attr_reader :b_count
+  attr_reader :b_conv_count
+  attr_reader :avg_rate_b
+  attr_reader :rate_b_low
+  attr_reader :rate_b_high
+
```
zsyed1: Lines 7 - 17: Notice lines 7-11 and 12-16 are identical except for the cohort tag `a` or `b`. This suggests the need to encapsulate those attributes into a struct or a class:

```
class CohortAttributes
attr_accessor :count, conversions, avg_rate, rate_low, rate_high 
end
```

Also, you can use a single helper line and put all the attributes into a single line. I also did it the way you did (one attribute per line but learned that the ruby convention is to through them all on one line!)

data_science/data_science.rb
```diff
@@ -0,0 +1,135 @@
...
+  def parse_json_file(file_path)
+    json_data = open_and_read_file(file_path)
+    @visits = JSON.parse(json_data)
+  end
+
+  def successfully_loaded?
+    return false if @visits.nil? || @visits.empty?
+    true
+  end
```
zsyed1:
```
def successully_loaded?
  !@visits.nil? && !@visits.empty?
end
```

data_science/data_science.rb
```diff
@@ -0,0 +1,135 @@
...
+  attr_reader :rate_b_high
+
+  def initializer(_args = {})
+    @visits = []
+  end
+
+  def parse_json_file(file_path)
+    json_data = open_and_read_file(file_path)
+    @visits = JSON.parse(json_data)
+  end
```
zsyed1: You should probably do all this in the initializer. Remove the `parse_json_file` and pass `file_path` into the initializer and call the parse method there as well.

data_science/data_science.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  def find_counts_per_cohort
+    @a_count, @a_conv_count = find_counts_for_cohort('A')
+    @b_count, @b_conv_count = find_counts_for_cohort('B')
+  end
+
+  def able_to_find_cohort_counts?
+    return true if @a_count != 0 && @b_count != 0
+    false
+  end
```
zsyed1:
```
def able_to_find_cohort_counts?
  @a_count > 0 && @b_count > 0
end
```

data_science/data_science.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  def calculate_average_conversion_rate
+    @avg_rate_a = 0.0
+    @avg_rate_b = 0.0
+
```
zsyed1: Lines 43-44 are not needed.

data_science/data_science.rb
```diff
@@ -0,0 +1,135 @@
...
+    @avg_rate_a = @a_conv_count.to_f / @a_count.to_f
+    @avg_rate_b = @b_conv_count.to_f / @b_count.to_f
+  end
+
+  def able_to_calculate_avg_conv_rate?
+    return false if @avg_rate_a == 0 || @avg_rate_b == 0
+    true
+  end
```
zsyed1: Lines 51-52: Replace with one-liner like above

data_science/data_science.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  def calculate_confidence_level
+    values = {}
+    values['a'] = { 'conved' => @a_conv_count,
+                    'unconved' => @a_count - @a_conv_count }
+    values['b'] = { 'conved' => @b_conv_count,
+                    'unconved' => @b_count - @b_conv_count }
+
```
zsyed1: Move this logic into a separate method and call that method twice within here:

```
def confidence_interval(cohort)
  { converted: @a_conv_count, unconverted: @a_count - @a_conv_count }
end
```

data_science/data_science.rb
```diff
@@ -0,0 +1,135 @@
...
+    analyzer = ABAnalyzer::ABTest.new values
+
+    @confidence_level = 0
```
zsyed1: No need for line 77

data_science/data_science.rb
```diff
@@ -0,0 +1,135 @@
...
+    @confidence_level = analyzer.chisquare_p
+  end
+
+  def able_to_produce_confidence_level?
+    return true if @confidence_level != 0
+    false
+  end
```
zsyed1: Make into one-liner

data_science/data_science.rb
```diff
@@ -0,0 +1,135 @@
...
+  def find_counts_for_cohort(name)
+    count = 0
+    conv_count = 0
+    @visits.each do |visit|
+      if visit['cohort'] == name
+        count += 1
+        conv_count += visit['result']
+      end
+    end
+    return count, conv_count
```
zsyed1: Lines 98 - 103: Use `select` and `inject` methods:

```
cohort_visits = @visits.select { |v| v['cohort'] == name }
conversions = cohort_visits.inject(0) { |sum, v| sum += v['result'] }
return cohort.size, conversions
```

Also try to use the full names: `conversion_count` or `conversions` instead of abbreviating. That is a style feature of the ruby community.

data_science/data_science.rb
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+  def calc_95_conv_rate_for(count, conv_count)
+    result = ABAnalyzer.confidence_interval(conv_count, count, 0.95)
+    low = result[0]
+    high = result[1]
+    return low, high
```
zsyed1:
```
result = ABAnalyzer ....
return result[0], result[1]
```

data_science/data_science_rspec.rb
```diff
@@ -0,0 +1,68 @@
+require './data_science'
+
+describe "data_science" do
+  it "should parse the json file of the experiment", :happy do
+    data_science = DataScience.new
+
```
zsyed1: You can remove these blank lines - ruby-community style is not to have blank lines.

data_science/data_science_rspec.rb
```diff
@@ -0,0 +1,68 @@
...
+    data_science.parse_json_file('./data_export_2014_06_20_15_59_02.json')
```
zsyed1: Move the setup-logic into a let statement:

```
describe "data_science" do
  let(:data_science) { DataScience.new }
  ...
```

data_science/data_science_rspec.rb
```diff
@@ -0,0 +1,68 @@
...
+    expect { data_science.parse_json_file(f) }.to raise_error(JSON::ParserError)
+  end
+
+  it "should find visits and conversions count for each cohort", :happy do
+    data_science = DataScience.new
+
+    data_science.parse_json_file('./data_export_2014_06_20_15_59_02.json')
+    data_science.find_counts_per_cohort
+
+    expect(data_science).to be_able_to_find_cohort_counts
```
zsyed1: The method name `able_to_find_cohort_counts` may not be what you are looking for. After all, a data set may contain zero data for a particular cohort. If you really want to test for non-zero, you can check the at the count itself is not zero:

```
expect(data_science.count('a')).not_to be_zero
```

data_science/data_science_rspec.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  it "should give the conversion rate for each cohort", :happy do
+    data_science = DataScience.new
+
+    data_science.parse_json_file('./data_export_2014_06_20_15_59_02.json')
+    data_science.find_counts_per_cohort
+
+    data_science.calculate_average_conversion_rate
+    data_science.calculate_95_perc_conversion_rate_range
```
zsyed1: Lines 47 - 50 should be done in a let statement:

```
let(:calculated) do
  data_science.parse_json_file('./data_export_2014_06_20_15_59_02.json')
  data_science.find_counts_per_cohort
  data_science.calculate_average_conversion_rate
  data_science.calculate_95_perc_conversion_rate_range
  data_science
end
```

Then this test becomes:

```
it "should give the conversion rate for each cohort", :happy do
  expect(data_science).to be_able_to_calculate_avg_conv_rate
  expect(data_science).to be_able_to_calculate_95_conv_rate_ranges
end
```

data_science/data_science_rspec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    data_science.calculate_confidence_level
+
+    expect(data_science).to be_able_to_produce_confidence_level
+  end
```
zsyed1: See comment above about moving into `let`s

zsyed1: Your tests check that non-zero/non-empty values are returned by your `DataScience` objects. However, you need to also test that the values themselves are correct.

This exercise makes TDD nearly impossible :person_frowning: so most learners never really practice TDD when implementing the solution. In any case, you should still add tests like:

```
it "should compute the correct confidence interval for cohort A" do
  expect(confidence_a_lower).to be_within(0.0001).of(0.455456)
  expect(confidence_a_upper).to be_within(0.0001).of(0.534783)
end
```

data_science/data_print.rb
```diff
@@ -0,0 +1,36 @@
...
+    puts "Confidence Level: " << conf_pres.round(2).to_s << "%"
+  end
+
+  def self.print_avg_rate(name, avg_rate, count, conv_count)
+    avg_rate_pres = avg_rate * 100
+    conv_count_pres = conv_count * 100
+    count_pres = count * 100
+
+    puts "Avg " << name << " Rate: " << avg_rate_pres.round(2).to_s << "% (" +
+      conv_count_pres.to_s << "/" << count_pres.to_s << ")"
```
zsyed1: Normally interpolation is used to print data:
```
  puts "Avg #{name} Rate: #{avg_rate_pres.round(2)}% ( #{conv_count_pres}/#{count_pres})"
```
Notice you no longer need the `to_s` call since interpolation invokes it automatically.

data_science/data_print.rb
```diff
@@ -0,0 +1,36 @@
...
+    # presentable values
+    conf_pres = conf_level * 100
+
+    puts "Confidence Level: " << conf_pres.round(2).to_s << "%"
+  end
+
+  def self.print_avg_rate(name, avg_rate, count, conv_count)
+    avg_rate_pres = avg_rate * 100
+    conv_count_pres = conv_count * 100
+    count_pres = count * 100
```
zsyed1: Why do you need to multiple by 100? Are these stored as proportions?

data_science/data_print.rb
```diff
@@ -0,0 +1,36 @@
...
+    puts "Avg " << name << " Rate: " << avg_rate_pres.round(2).to_s << "% (" +
+      conv_count_pres.to_s << "/" << count_pres.to_s << ")"
+  end
+
+  def self.print_rate(name, low, high)
+    low_pres = low * 100
+    high_pres = high * 100
+
+    puts name << " Rates: [" << low_pres.round(2).to_s << "% - " +
+      high_pres.round(2).to_s << "%]"
```
zsyed1: Use interpolation to display.

data_science/data_print.rb
```diff
@@ -0,0 +1,36 @@
...
+  def self.print_rates_ranges(a_low, a_high, b_low, b_high)
+    print_rate('A', a_low, a_high)
+    print_rate('B', b_low, b_high)
+  end
+
+  def self.print_confidence_level(conf_level)
+    # presentable values
+    conf_pres = conf_level * 100
+
+    puts "Confidence Level: " << conf_pres.round(2).to_s << "%"
```
zsyed1: Use interpolation to display.

data_science/data_print.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def self.print_rates_ranges(a_low, a_high, b_low, b_high)
+    print_rate('A', a_low, a_high)
+    print_rate('B', b_low, b_high)
+  end
+
+  def self.print_confidence_level(conf_level)
+    # presentable values
+    conf_pres = conf_level * 100
+
```
zsyed1: Remove whitespace (Ruby community style)

data_science/data_print.rb
```diff
@@ -0,0 +1,36 @@
...
+    print_avg_rate('B', avg_rate_b, b_count, b_c_count)
+  end
+
+  def self.print_rates_ranges(a_low, a_high, b_low, b_high)
+    print_rate('A', a_low, a_high)
+    print_rate('B', b_low, b_high)
+  end
+
+  def self.print_confidence_level(conf_level)
+    # presentable values
```
zsyed1: Remove comment

data_science/data_science.rb
```diff
@@ -0,0 +1,135 @@
...
+  attr_reader :rate_a_low
+  attr_reader :rate_a_high
+  attr_reader :b_count
+  attr_reader :b_conv_count
+  attr_reader :avg_rate_b
+  attr_reader :rate_b_low
+  attr_reader :rate_b_high
+
+  def initializer(_args = {})
+    @visits = []
```
zsyed1: Interesting - almost all learners store the entire data set in here. But imagine if there were TONS of data. It's probably better to slurp the data point by point and store the calculations instead of storing the entire data set. Do NOT change this, the way you did it is fine for the purposes of this exercise.

data_science/data_science_rspec.rb
zsyed1: @athiercelin Nice work implementing the features! I would recommend making revisions and resubmitting. If anything is unclear, or you just want to chat about this stuff (or anything for that matter), please feel free to come by or hit me up on hipchat (@zamir)

paulghaddad:
@athiercelin Nice work on this! Please take a look at Zamir's comments and submit a second iteration.

data_science/data_science_rspec.rb
athiercelin: Will do, but I loose my fresh state from one test to the other, am I not?

data_science/data_print.rb
athiercelin: Yes, we never know when we would want to use these values for future calculation

data_science/data_science.rb
athiercelin: Memory vs CPU battle ;)

PR #1043, opened 2015-11-05
Data science - V1

drosholt:
Created as a gem.

Test run using `ruby -Ilib exe/data-science data/data_export_2014_06_20_15_59_02.json` .

Build, install, and run by:
```bash
gem build data_science.gemspec
gem install data_science-0.1.0.pre.gem
data-science data/data_export_2014_06_20_15_59_02.json
```


data_science/Gemfile
```diff
@@ -0,0 +1,3 @@
+source 'https://rubygems.org'
```
paulghaddad: Really nice job making this a gem.

data_science/bin/setup
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
paulghaddad: Nice use of a `bin/setup`. More people should use this.

data_science/exe/data-science
```diff
@@ -0,0 +1,23 @@
...
+EOS
+end
+
+Trollop::with_standard_exception_handling p do
+  raise Trollop::HelpNeeded if ARGV.size != 1 # show help screen
+  p.parse ARGV
+end
+
+filename = ARGV[0]
+DataScience.print_statistics(filename)
```
paulghaddad: Really clean file. I'll have to take a look at `trollop` myself.

data_science/lib/data_science/base.rb
```diff
@@ -0,0 +1,28 @@
+require 'json'
+require 'pp'
+
+require 'data_science/cohort_factory'
+require 'data_science/ab_test'
+
+module DataScience
+  class Base
+    def initialize(cohort_factory = CohortFactory.new, ab_test = ABTest.new)
```
paulghaddad: Do you think keyword arguments would be a good choice here? Just want to pose the question.

data_science/lib/data_science/cohort_factory.rb
```diff
@@ -0,0 +1,25 @@
+require 'data_science/cohort'
+
+module DataScience
+  class CohortFactory
+    CONVERSION_RESULT = 1
+    private_constant :CONVERSION_RESULT
+
+    def from_hash_array(results)
+      aggregate_results = results
+        .group_by { |r| r.fetch(:cohort) }
```
paulghaddad: Nice use of `group_by`.

data_science/lib/data_science/cohort_factory.rb
```diff
@@ -0,0 +1,25 @@
...
+        .map { |k, v| [k.to_sym, cohort_for(v)] }
+
+      Hash[aggregate_results]
+    end
+
+    private
+
+    def cohort_for(hash_array)
+      size = hash_array.size
+      conversions = hash_array.count { |h| h.fetch(:result) == CONVERSION_RESULT }
```
paulghaddad: It is good you are only concerned with the number of conversions and not making each data point an object.

data_science/lib/data_science/cohort.rb
```diff
@@ -0,0 +1,18 @@
+module DataScience
+  class Cohort
+    attr_reader :no, :yes
+
+    def initialize(no, yes)
```
paulghaddad: Perhaps keyword arguments will make this simpler to build objects so you don't have to remember the positional order of the `#initialize` method.

data_science/lib/data_science/cohort.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    def size
+      no + yes
+    end
+
+    def to_h
+      { no: no, yes: yes }
+    end
+  end
+end
```
paulghaddad: I also really like your small, simple classes. This really makes it easy to understand.

data_science/spec/data_science/ab_test_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  it 'provides the a conversion rate in the results' do
+    expect(ab_analyzer).to receive(:confidence_interval).with(a_cohort.yes, a_cohort.size, 0.95) { :a_conversion }
+    expect(ab_analyzer).to receive(:confidence_interval).with(anything, anything, anything)
+
+    results = ab_test.parse_test_results(a_cohort, b_cohort)
+    expect(results.a_conversion_rate).to equal :a_conversion
+  end
+
+  it 'provides the b conversion rate in the results' do
+    expect(ab_analyzer).to receive(:confidence_interval).with(anything, anything, anything)
```
paulghaddad: What are your thoughts in using spies here instead of mocks. That was you can use keep the setup, expect and expectations phases of the test organized.

https://github.com/rspec/rspec-mocks

https://robots.thoughtbot.com/four-phase-test

data_science/spec/data_science/base_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    expect(base.instance_variable_get(:@cohort_factory)).to be_kind_of DataScience::CohortFactory
+  end
+
+  it 'uses ABTest by default' do
+    expect(base.instance_variable_get(:@ab_test)).to be_kind_of DataScience::ABTest
+  end
+
+  private
+
+  def expect_file_to_be_read
```
paulghaddad: Good method extraction here that makes your test easier to understand.

data_science/spec/data_science/cohort_factory_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'data_science/cohort_factory'
+require 'data_science/cohort'
+
+describe DataScience::CohortFactory do
+  class DataScience::Cohort
```
paulghaddad: This makes me think a little...is it good to redefine the `eql?` method?

data_science/spec/data_science/cohort_factory_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  let(:factory) { DataScience::CohortFactory.new }
+
+  let(:cohort_json_array) do
+    [ build(:cohort_json), build(:b_cohort_json), build(:converted_cohort_json) ]
+  end
+
+  it 'groups by cohort' do
+    expect(factory.from_hash_array(cohort_json_array).keys).to eql [:A, :B]
```
paulghaddad: The contain-exactly matcher may be a good option here:

https://www.relishapp.com/rspec/rspec-expectations/v/3-3/docs/built-in-matchers/contain-exactly-matcher

data_science/spec/factories/cohort/cohort_json.rb
```diff
@@ -0,0 +1,19 @@
+FactoryGirl.define do
```
paulghaddad: Awesome job using FactoryGirl for this exercise!

data_science/spec/data_science/ab_test_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  let(:a_cohort) { build(:a_cohort) }
+  let(:b_cohort) { build(:b_cohort) }
+
+  let(:ab_analyzer) { class_double('ABAnalyzer').as_stubbed_const(transfer_nested_constants: true) }
+
+  let(:ab_test) { DataScience::ABTest.new }
+
+  it 'provides the a cohort in the results' do
+    results = ab_test.parse_test_results(a_cohort, b_cohort)
+    expect(results.a_cohort).to equal a_cohort
```
paulghaddad: Could you possibly combine this spec and the next by using the `have_attributes` matcher?

https://www.relishapp.com/rspec/rspec-expectations/v/3-3/docs/built-in-matchers/have-attributes-matcher

data_science/spec/data_science/cohort_factory_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    cohort_hash = factory.from_hash_array(cohort_json_array)
+
+    expect(cohort_hash[:A].size).to equal 2
+    expect(cohort_hash[:B].size).to equal 1
+  end
+
+  it 'provides the conversions' do
+    cohort_hash = factory.from_hash_array(cohort_json_array)
+
+    expect(cohort_hash[:A].yes).to equal 1
```
paulghaddad: You probably don't need `equal` here: `eql` will suffice as we're only concerned with values here. Now this is nit-picking! :)

paulghaddad:
@drosholt This is the most clear data science I have reviewed. Excellent use of small classes, easy to understand methods and building a gem. I really, really enjoyed reviewing it, and learned a lot myself. Good tests as well. Take a look at my comments (I'd love to hear what you think about my questions), but no need to resubmit. This is excellent work.

data_science/lib/data_science/base.rb
drosholt: Could be.  I'd have to get some more intuition on how to balance them with normal parameters.  As I'm thinking about it, maybe a good rule of thumb could be passing in parameters for the business logic normally, and passing in DI attributes using keyword parameters?  It provides a clear separation between the two, and for "things that do other things", we are clearly calling out that we are overriding the default behavior?

paulghaddad: I think the tradeoffs cited at the bottom of this article make a good argument: 

data_science/spec/data_science/cohort_factory_spec.rb
drosholt: Good call - the dependency on the ordering did bother me.

data_science/spec/data_science/cohort_factory_spec.rb
drosholt: It's a great question.  I definitely wouldn't have wanted to override it in the actual class (but in testing, I was okay with it).  Is there a preferred way to answer the question: These two classes have the exact same attributes, and those attributes have the exact same values?

data_science/spec/data_science/cohort_factory_spec.rb
drosholt: Actually, I had eql, but mutant complained about it.  Technically, mutant was right (though I agree with you on the pedantics), so I changed it.

data_science/spec/data_science/ab_test_spec.rb
drosholt: Hmm, I'll want to think about this one.  I like the purity of checking for the a_cohort separately from the b_cohort, in that if one breaks, the other test still passes.  But I also recognize that that may be overkill.   ...

data_science/spec/data_science/cohort_factory_spec.rb
paulghaddad: I meant `eq` instead of `eql`

PR #1044, opened 2015-11-06
Refactor Ruby code to reflect Style Guide; Add Rspec test

creativisionist:


triangle_facts/triangle_test.rb
```diff
@@ -0,0 +1,22 @@
+require 'rspec'
```
paulghaddad: Simply great that you added tests! :)

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
-
-		return [angleA, angleB, angleC]
-	end
-
-	def radians_to_degrees(rads)
-		return (rads * 180 / Math::PI).round
-	end
-end
+  attr_accessor :side1, :side2, :side3, :all
+  def initialize(side1, side2, side3)
```
paulghaddad: Have a line break between lines 6 and 7.

paulghaddad: Also check out keyword arguments; just another option. See:

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+    @all = [side1, side2, side3]
```
paulghaddad: Interesting option here. I haven't seen it before, but it's quite nice. But do consider renaming to something more descriptive, such as `triangle_side`.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
+
+  def scalene?
+    !(equilateral? || isosceles?)
+  end
+
+  def rads_to_degs(rads)
+    (rads * 180 / Math::PI).round
+  end
 
+  def calculate(first, second, third)
```
paulghaddad: Consider naming your parameters to something more descriptive.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
+  end
+
+  def calculate_angles
+    angle_a = calculate(side1, side2, side3)
+    angle_b = calculate(side2, side1, side3)
+    angle_c = calculate(side3, side1, side2)
+    [angle_a, angle_b, angle_c]
+  end
+
+  def prints(triangle_type, secondary = nil)
```
paulghaddad: Nice way of extracting some of this code from the `print_triangle_facts` method.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
-		angleC = radians_to_degrees(Math.acos((a**2 + b**2 - c**2) / (2.0 * a * b)))
-
-		return [angleA, angleB, angleC]
-	end
-
-	def radians_to_degrees(rads)
-		return (rads * 180 / Math::PI).round
-	end
-end
+  attr_accessor :side1, :side2, :side3, :all
```
paulghaddad: As a principle, only expose as much access to your class as needed. In this particular exercise, you only need to get the instance variables, so I'd recommend using `attr_reader` here.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
+
+  def isosceles?
+    all.uniq.length == 2
+  end
+
+  def scalene?
+    !(equilateral? || isosceles?)
+  end
+
+  def rads_to_degs(rads)
```
paulghaddad: There are a number of methods in this class that are only called from inside the class. I'd recommend to make this methods `private`.

paulghaddad:
@creativisionist This is a great job! I liked your thinking about including tests even though it wasn't required. Take a look at my comments and submit a second revision. There shouldn't be too many changes, but do note my comments on access levels (private methods and attrs).

PR #1045, opened 2015-11-06
move methods to private, change to attr, descriptive variable

creativisionist:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,74 @@
...
+  end
+
+  def calculate_angles
+    angle_a = calc_angle(side1, side2, side3)
+    angle_b = calc_angle(side2, side1, side3)
+    angle_c = calc_angle(side3, side1, side2)
+    [angle_a, angle_b, angle_c]
+  end
+
+  def prints(triangle_type, secondary = nil)
```
djkotowski: Not a big fan of the method name here.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,74 @@
...
+    prints("equilateral!") if equilateral?
+    prints("isosceles!", "Also, that word is hard to type.") if isosceles?
+    prints("scalene", "and mathematictriangle_sidesy boring.") if scalene?
+  end
+
+  def print_triangle_angles
+    angles = calculate_angles
+    puts "The angles of this triangle are #{angles.join(',')}"
+    puts "This triangle is also a right triangle!" if angles.include? 90
+    puts ""
```
djkotowski: No need for the empty string here.  `puts` doesn't require an argument if you only want to add a newline.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,74 @@
...
+  end
+
+  def recite_facts
+    print_triangle_type
+    print_triangle_angles
+  end
+
+  private :rads_to_degs, :calc_angle, :calculate_angles
```
djkotowski: We prefer using `private` as a separator to divide public and private methods.  It makes it clearer which methods are private vs public before you get to the end of the class.

triangle_facts/triangle_test.rb
```diff
@@ -0,0 +1,22 @@
+require 'rspec'
+require './triangle_facts.rb'
+
+describe Triangle do
+  sides_1 = [5, 12, 13]
+  sides_2 = [5, 5, 5]
+
+  it 'test for scalene' do
+    expect(Triangle.new(*sides_1).scalene?).to be true
```
djkotowski: This is getting ahead of the lesson a bit, but for these tests, prefer using the predicate matchers.  This one could be:
```ruby
expect(Triangle.new(*sides_1)).to be_scalene
```

triangle_facts/triangle_test.rb
```diff
@@ -0,0 +1,22 @@
+require 'rspec'
+require './triangle_facts.rb'
+
+describe Triangle do
+  sides_1 = [5, 12, 13]
```
djkotowski: Prefer `let` statements to variable assignments.

djkotowski:
Solid work here.  Some comments throughout, but no need for resubmission.  Nice job!

PR #1046, opened 2015-11-09
Refactoring to conform to thoughtbot Ruby style guide

jrade-enova:
Changes based on https://github.com/thoughtbot/guides/tree/master/style/ruby

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -1,37 +1,45 @@
 # Killer facts about triangles AWW YEAH
 class Triangle
-	attr_accessor :side1,:side2,:side3
 
 	def initialize(side1,side2,side3)
-		@side1,@side2,@side3 = side1,side2,side3
+		@side1 = side1
+		@side2 = side2
+		@side3 = side3
 	end
 
 	def equilateral()
-		return side1 == side2 && side2 == side3
+		side1 == side2 && side2 == side3
 	end
 
 	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
+		[side1,side2,side3].uniq.length == 2
 	end
 
 	def scalene()
-		unless ! (equilateral || isosceles)
-	    false
-	  else
+		if !equilateral && !isosceles
 	    true
+	  else
+	    false
 	  end
 	end
 
 	def recite_facts
-		puts 'This triangle is equilateral!' if equilateral
-		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+		if equilateral
+			puts "This triangle is equilateral!"
+		elsif isosceles
+			puts "This triangle is isosceles! Also, that word is hard to type." 
+		elsif scalene 
+			puts "This triangle is scalene and mathematically boring."
+		end
 
 		angles = self.calculate_angles(side1,side2,side3)
-		puts 'The angles of this triangle are ' + angles.join(',')
+		angles_prettified = angles.join(",")
+		puts "The angles of this triangle are #{angles_prettified}"
 
-		puts 'This triangle is also a right triangle!' if angles.include? 90
-		puts ''
+		if angles.include? 90
+				puts "This triangle is also a right triangle!"
+		end
+		puts ""
 	end
 
 	def calculate_angles(a,b,c)
@@ -39,11 +47,19 @@ class Triangle
 		angleB = radians_to_degrees(Math.acos((a**2 + c**2 - b**2) / (2.0 * a * c)))
 		angleC = radians_to_degrees(Math.acos((a**2 + b**2 - c**2) / (2.0 * a * b)))
 
-		return [angleA, angleB, angleC]
+		[angleA, angleB, angleC]
 	end
+	
+	protected
+	
+	attr_accessor :side1
+	attr_accessor :side2
+	attr_accessor :side3
+	
+	private
 
 	def radians_to_degrees(rads)
-		return (rads * 180 / Math::PI).round
+		(rads * 180 / Math::PI).round
 	end
 end
 
@@ -52,7 +68,7 @@ triangles = [
 	[5,5,5],
 	[5,12,13],
 ]
-triangles.each { |sides|
+triangles.each do |sides|
 	tri = Triangle.new(*sides)
 	tri.recite_facts
-}
+end


ochagata:
According to the ruby style guide, we use spaces not tabs
Also, we use 2 spaces not 4.
You can configure your text editor to do this for you. :)

triangle_facts/triangle_facts.rb
```diff
@@ -1,37 +1,45 @@
 # Killer facts about triangles AWW YEAH
 class Triangle
-	attr_accessor :side1,:side2,:side3
 
 	def initialize(side1,side2,side3)
```
ochagata: Use spaces between the variables here: (side1, side2, side3)

triangle_facts/triangle_facts.rb
```diff
@@ -1,37 +1,45 @@
...
-		@side1,@side2,@side3 = side1,side2,side3
+		@side1 = side1
+		@side2 = side2
+		@side3 = side3
 	end
 
 	def equilateral()
-		return side1 == side2 && side2 == side3
+		side1 == side2 && side2 == side3
```
ochagata: You'll have to use the '@' sign in front of these since they're local variables and not instance variables.

triangle_facts/triangle_facts.rb
```diff
@@ -1,37 +1,45 @@
...
-	attr_accessor :side1,:side2,:side3
 
 	def initialize(side1,side2,side3)
-		@side1,@side2,@side3 = side1,side2,side3
+		@side1 = side1
+		@side2 = side2
+		@side3 = side3
 	end
 
 	def equilateral()
```
ochagata: You also don't need the parens when you aren't passing in arguments

triangle_facts/triangle_facts.rb
```diff
@@ -1,37 +1,45 @@
...
-		return side1 == side2 && side2 == side3
+		side1 == side2 && side2 == side3
 	end
 
 	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
+		[side1,side2,side3].uniq.length == 2
 	end
 
 	def scalene()
```
ochagata: This method can be just the conditional, no need for 'if' and 'else' if you're returning a true/false response

triangle_facts/triangle_facts.rb
```diff
@@ -1,37 +1,45 @@
...
+	  else
+	    false
 	  end
 	end
 
 	def recite_facts
-		puts 'This triangle is equilateral!' if equilateral
-		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+		if equilateral
```
ochagata: You can use a case statement here. If you do, indent the 'when' statements as deep as the 'case'

triangle_facts/triangle_facts.rb
```diff
@@ -39,11 +47,19 @@ def calculate_angles(a,b,c)
 		angleB = radians_to_degrees(Math.acos((a**2 + c**2 - b**2) / (2.0 * a * c)))
```
ochagata: Why are you using a 2.0 instead of a 2?

ochagata: Also, the calculations for these angles can be a method.
You can have three methods that can calculate the angles.
Or you can have a method that calculates the first argument with the other 2

triangle_facts/triangle_facts.rb
```diff
@@ -39,11 +47,19 @@ def calculate_angles(a,b,c)
...
 		angleC = radians_to_degrees(Math.acos((a**2 + b**2 - c**2) / (2.0 * a * b)))
 
-		return [angleA, angleB, angleC]
+		[angleA, angleB, angleC]
 	end
+	
+	protected
+	
+	attr_accessor :side1
```
ochagata: These don't have to be on individual lines

triangle_facts/triangle_facts.rb
```diff
@@ -52,7 +68,7 @@ def radians_to_degrees(rads)
 	[5,5,5],
 	[5,12,13],
 ]
-triangles.each { |sides|
+triangles.each do |sides|
```
ochagata: You could write a test for this so that instead of running the actual file, you can run the test file which will call the Triangle class.

ochagata:
Will need resubmission

triangle_facts/triangle_facts.rb
jrade-enova: 2.0 is a float type and will keep precision when dividing. (e.g. 3 / 2 => 1, but 3 / 2.0 => 1.5)

triangle_facts/triangle_facts.rb
jrade-enova: I did it that way since it is seen that way in the thoughbot guide https://github.com/thoughtbot/guides/blob/master/style/ruby/sample.rb#L88

triangle_facts/triangle_facts.rb
paulghaddad: You can also define a `attr_reader` or `attr_accessor` method for these instance variables if you prefer not using the `@`sign (which I recommend).

triangle_facts/triangle_facts.rb
paulghaddad: If a method returns a boolean, you can add a `?` after it to signify it.

triangle_facts/triangle_facts.rb
```diff
@@ -1,37 +1,45 @@
...
+		@side2 = side2
+		@side3 = side3
 	end
 
 	def equilateral()
-		return side1 == side2 && side2 == side3
+		side1 == side2 && side2 == side3
 	end
 
 	def isosceles()
```
paulghaddad: `def isoceles?`

triangle_facts/triangle_facts.rb
```diff
@@ -1,37 +1,45 @@
...
+	    false
 	  end
 	end
 
 	def recite_facts
-		puts 'This triangle is equilateral!' if equilateral
-		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+		if equilateral
+			puts "This triangle is equilateral!"
```
paulghaddad: I'd actually recommend keeping with the original syntax; I think it reads a little more concisely.
 i.e.: `puts 'This triangle is equilateral!' if equilateral?`

triangle_facts/triangle_facts.rb
```diff
@@ -39,11 +47,19 @@ def calculate_angles(a,b,c)
 		angleB = radians_to_degrees(Math.acos((a**2 + c**2 - b**2) / (2.0 * a * c)))
 		angleC = radians_to_degrees(Math.acos((a**2 + b**2 - c**2) / (2.0 * a * b)))
 
-		return [angleA, angleB, angleC]
+		[angleA, angleB, angleC]
 	end
+	
+	protected
```
paulghaddad: Is there a reason to make these `attr` methods protected? Conventionally these aren't private or protect in almost all cases. For this case, I think it is reasonable that an outside caller would call these methods.

triangle_facts/triangle_facts.rb
```diff
@@ -39,11 +47,19 @@ def calculate_angles(a,b,c)
...
+	
+	attr_accessor :side1
+	attr_accessor :side2
+	attr_accessor :side3
+	
+	private
```
paulghaddad: Really nice job using `private` here. See if there are other methods that can be private.

triangle_facts/triangle_facts.rb
ochagata: Or you can have an unprotected reader and a protected writer, in the case that you dont want someone to edit just to view

triangle_facts/triangle_facts.rb
```diff
@@ -1,37 +1,45 @@
...
 
-		puts 'This triangle is also a right triangle!' if angles.include? 90
-		puts ''
+		if angles.include? 90
+				puts "This triangle is also a right triangle!"
+		end
+		puts ""
 	end
 
 	def calculate_angles(a,b,c)
```
paulghaddad: Spaces are needed between the parameters. Also, consider the naming of these parameters. Can you make them more clear?

triangle_facts/triangle_facts.rb
ochagata: The example did that because they were 3 separate methods: attr_reader, attr_writer, attr_accessor. If you're just calling attr_accessor, you can use the same line

triangle_facts/triangle_facts.rb
```diff
@@ -1,37 +1,45 @@
...
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+		if equilateral
+			puts "This triangle is equilateral!"
+		elsif isosceles
+			puts "This triangle is isosceles! Also, that word is hard to type." 
+		elsif scalene 
+			puts "This triangle is scalene and mathematically boring."
+		end
 
 		angles = self.calculate_angles(side1,side2,side3)
```
paulghaddad: You shouldn't need to use `self` when calling this instance method. Inside an instance method, the value of self is the object, so this should call `calculate_angles` on the triangle object.

paulghaddad:
@jrade-enova Nice work on this first iteration. Please take a look at the comments and submit a second iteration.

triangle_facts/triangle_facts.rb
ochagata:
```
case
when equilateral?
  puts 'This triangle is equilateral!'
when isosceles?
  puts 'This triangle is isosceles! Also, that word is hard to type.'
when scalene?
  puts 'This triangle is scalene and mathematically boring.'
end
```
Why not use this syntax instead of multiple ifs?

jrade-enova: thoughtbot says, "Avoid conditional modifiers (lines that end with conditionals)."  Does that mean to NOT do something like this?
```
puts "Print Me!" if printable?
```

triangle_facts/triangle_facts.rb
jrade-enova: Ah, I see that now.  Will change to:

```
protected attr_accessor :side1,  :side2, :side3
```

triangle_facts/triangle_facts.rb
ochagata: https://github.com/bbatsov/ruby-style-guide#if-as-a-modifier

jrade-enova: Does bbatsov trump thoughtbot in style?  For instance, one says favor the ternary and the other does not.

ochagata: thoughtbot decided to go with bbatsov besides 2 points:

Avoid trailing conditionals.
Don't use ternary expressions.

jrade-enova: Good to know, thanks!

PR #1048, opened 2015-11-09
Refactor Arrowhead exercise to consolidate error conditions

cstavitsky:


arrowhead/arrowhead.rb
```diff
@@ -17,14 +17,10 @@ class Arrowhead
...
+    if CLASSIFICATIONS[region]
       shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
+      return "You have a(n) '#{shapes[shape]}' arrowhead. Probably priceless." if shapes[shape]
+      raise "Unknown shape value. Are you sure you know what you're talking about?"
```
incazteca: Have this raise closer to the code that checks if a shape is invalid. You should also have the other error message be closer to the code that checks for regions instead of tucking it a few lines below. Preferably we should see:
`raise 'an error message' if an_error?`

incazteca:
Try breaking this code up into methods so that the error messages are closer to the code that causes it. Putting the code into methods would also keep the code from being too indented.

paulghaddad:
@cstavitsky Please take a look at Franco's comment, and resubmit.

paulghaddad:
@cstavitsky I see you already resubmitted this, so I'll get it reviewed.

PR #1050, opened 2015-11-09
Robot Name - Iteration 1

creativisionist:


robot_name/robot_name.rb
```diff
@@ -1,6 +1,8 @@
+require 'pry'
```
ochagata: Remove debugger when you submit :goat:

robot_name/robot_name.rb
```diff
@@ -1,6 +1,8 @@
...
+
 class NameCollisionError < RuntimeError; end
```
ochagata: Might want to move your errors into it's own file :goat:

arrowhead/arrowhead.rb
```diff
@@ -1,33 +1,28 @@
...
-      bifurcated: "Cody",
-    },
-    northern_plains: {
-      notched: "Besant",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Humboldt Constricted Base",
-      bifurcated: "Oxbow",
-    },
-  }
+require 'pry'
```
ochagata: Remove pry when submitting :goat:

arrowhead/arrowhead.rb
```diff
@@ -1,33 +1,28 @@
...
+
+# This seriously belongs in a database.
+## CSV or JSON -- overkill?
+CLASSIFICATIONS = {
```
ochagata: Multi-line hash/arrays should be indented 2 spaces beyond the opening bracket(and have the key -> aligned), so more like:
```
CLASSIFICATIONS = {
                    far_west: {
                                notched:    "Archaic Side Notch",
                                stemmed:    "Archaic Stemmed",
                                lanceolate: "Agate Basin",
```

arrowhead/arrowhead.rb
```diff
@@ -1,33 +1,28 @@
...
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+    raise "Unknown region." unless CLASSIFICATIONS.include? region
```
ochagata: Standardize your errors into custom error classes.
That way you can reuse the same exceptions/errors without having to rewrite the string every time

arrowhead/arrowhead.rb
```diff
@@ -1,33 +1,28 @@
...
+    shapes = CLASSIFICATIONS[region]
+    raise "Unknown shape value." unless shapes.include? shape
```
ochagata: same as above

arrowhead/arrowhead.rb
```diff
@@ -1,33 +1,28 @@
...
+    "You have a(n) '#{shapes[shape]}' arrowhead.  Probably priceless."
```
ochagata: You can do a check here for if the first letter of the shape is a vowel and decide if you should use 'an' or 'a'

robot_name/robot_name.rb
```diff
@@ -1,6 +1,6 @@
-class NameCollisionError < RuntimeError; end
+require_relative 'robot_name_errors.rb'
```
ochagata: Shut up hound

arrowhead/arrowhead_test.rb
```diff
@@ -0,0 +1,20 @@
+require 'rspec'
+require './arrowhead.rb'
+
+describe Arrowhead do
+  it 'is Oxbow' do
```
ochagata: You can do a describe for the classify method here and then expand on the different things it should do, that way when you get a failing test, the name of the test will give you a good indicator of what was the intended functionality and that way you can debug.

robot_name/robot_name.rb
```diff
@@ -8,25 +8,36 @@ class Robot
   def initialize(args = {})
     @@registry ||= []
     @name_generator = args[:name_generator]
+    generate_name
```
ochagata: You should assign name to the output here.
`@name = generate_name`
I saw that you had it inside of the `generate_name` method, but it would make more sense to make the `generate_name` method spit out a newly generated name instead of assigning it to the robot. Otherwise it'd be a `generate_and_assign_name`

robot_name/robot_name.rb
```diff
@@ -8,25 +8,36 @@ class Robot
...
+  end
+
+  def alpha_char
+    ('A'..'Z').to_a.sample
```
ochagata: Interesting implementation of this. But this excludes lowercase letters.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,74 @@
-# Killer facts about triangles AWW YEAH
+require 'pry'
```
ochagata: I won't do the triangle_facts since this is a submission for arrowhead and robot_name

robot_name/robot_name_errors.rb
```diff
@@ -0,0 +1 @@
+class NameCollisionError < RuntimeError; end
```
ochagata: :goat: :+1:

robot_name/robot_name.rb
```diff
@@ -8,25 +8,36 @@ class Robot
...
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
+  def generate_name
+    @name = @name_generator ? name_generator.call : rand_name_generator
     @@registry << @name
   end
 end
 
-robot = Robot.new
+robot = RobotName.new
```
ochagata: Once you have tests, you can remove this from the base file

arrowhead/arrowhead.rb
creativisionist: Although stylistically it may read better to a degree it could go over the 80 character limit, and the cop gives all the offenses.  I can see how that would read better for a larger hash but if it was a huge hash of data (say, an API) it would likely be getting pulled in and still not as spaced out.  Since there isn't a large amount of data in this hash I will likely keep it as is but will definitely think about that for future!

robot_name/robot_name.rb
```diff
@@ -8,25 +8,36 @@ class Robot
...
-    end
+  def rand_name_generator
+    @name = "#{alpha_char}#{alpha_char}#{digit}#{digit}#{digit}"
+    return @name unless invalid_name?
+    raise NameCollisionError, "There was a problem generating the robot name!"
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
+  def generate_name
+    @name = @name_generator ? name_generator.call : rand_name_generator
```
ochagata: I mentioned above that you can let this method just spit out the final result
You can replace `@@registry << @name` with `@@registry.push(@name)`, which returns `@@registry` after the push happens, so then you can do `@@registry.push(@name).last` which will return the last thing in the array which will be the last thing you sent in which would be `@name`
But since you're moving `@name` to the initialize method, then you can do something like:
`@@registry.push(@name_generator ? name_generator.call : rand_name_generator).last`

ochagata: And you can move the logic from `@name_generator ? name_generator.call : rand_name_generator` into either a method or a line above into a variable

ochagata:
Robot name needs resubmission, arrowhead should be fine :goat:

robot_name/robot_name.rb
paulghaddad: You can omit the file extension:

`require_relative 'robot_name_errors'`

robot_name/robot_name.rb
paulghaddad: Also try this: `[*'A'..'Z'].sample

robot_name/robot_name.rb
```diff
@@ -8,25 +8,36 @@ class Robot
...
+    rand(10)
+  end
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  def invalid_name?
+    !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
```
paulghaddad: Consider extracting this regex as a constant and naming what it does.

robot_name/robot_name.rb
```diff
@@ -8,25 +8,36 @@ class Robot
...
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
+  def rand_name_generator
+    @name = "#{alpha_char}#{alpha_char}#{digit}#{digit}#{digit}"
+    return @name unless invalid_name?
```
paulghaddad: Consider using a `rescue` clause. Ozzy will show you how it works.

paulghaddad:
@creativisionist Nice work on this! Please take a look at the comments and resubmit Robot Name. You are good to go on Triangle Facts.

robot_name/robot_name.rb
creativisionist: I'll comment out those 2 lines so it's there for reference since that was in the original output of the program (exercise).

PR #1051, opened 2015-11-09
Robot Name - Iteration 1

cstavitsky:


magicalbanana:
First off, address houndcl comments. As of late we are throughout all applications here in Enova extensively use Rubocop which is similar to houndcl. So getting used to this at the get go will prove beneficial for you in the long run.

Look into running rubocop or houndcl locally prior to uploading your PRs to avoid this flags.

robot_name/robot_name.rb
```diff
@@ -3,30 +3,69 @@ class NameCollisionError < RuntimeError; end
...
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  private
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def add_to_registry(registry)
+    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || registry.include?(name)
```
magicalbanana: You can move your error handler to a separate class and have a default message.

Also, the you're conditional here (if) should be converted to a full if block rather than single line. Because when you are on a team and you will be writing code, when our code coverage runs despite not writing the code nor writing a different scenario this line will get flagged to have been run.

I realize that you were told to be more "clean" with your code and limit the number of lines but that's not always the case.

A clean code that has good tests that are reflective of a good coverage is better than a clever way of writing it.

robot_name/robot_name.rb
```diff
@@ -3,30 +3,69 @@ class NameCollisionError < RuntimeError; end
...
+    registry.list << self.name
```
magicalbanana: No need for self here.

robot_name/robot_name.rb
paulghaddad: @cstavitsky To add to Kareem's comment, you have an opportunity to extract the conditional:

`if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || registry.include?(name)`

into a well named method. Also, you can make the regex a constant with a well-considered name.

robot_name/robot_name.rb
```diff
@@ -3,30 +3,69 @@ class NameCollisionError < RuntimeError; end
...
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+    name
+  end
+
+  private
+
+  def generate_char
+    ('A'..'Z').to_a.sample
```
paulghaddad: Also, try: `[*'A'..'Z].sample`

robot_name/robot_name.rb
magicalbanana: Another method is just overkill for me.

I prefer to see this as it is than digging through the entire code just for a method that does this check. But to each his own, I just prefer direct code that I can easily test against and refactor.

robot_name/robot_name.rb
```diff
@@ -3,30 +3,69 @@ class NameCollisionError < RuntimeError; end
...
+
+  attr_accessor :list
+  
+  def initialize(args = {})
+    @type = args[:type]
+    @list = args[:list] || []
+  end
+
+  def include?(string)
+    @list.include?(string)
```
paulghaddad: Since you defined an accessor method on `list`, you don't need the `@` symbol.

robot_name/robot_name.rb
```diff
@@ -3,30 +3,69 @@ class NameCollisionError < RuntimeError; end
...
+  end
+
 end
 
-robot = Robot.new
+class Registry
+
+  attr_accessor :list
+  
+  def initialize(args = {})
```
paulghaddad: Consider using Ruby's keyword arguments:

paulghaddad:
@cstavitsky Nice work on this first iteration! Take a look at our comments, and please submit a second iteration.

robot_name/robot_name.rb
magicalbanana: Also since you're defining this as an empty array just incase there's nothing passed on the args, I would check this if this is empty and perhaps raise an error if no list was passed, that is if you're code requires a list to be passed at all times.

PR #1053, opened 2015-11-09
Paulo dyno

pdgonzalez872:
Thank you for looking at this. This is an incredible resource.

Thank you for any feedback you guys could give. The testing uses RSpec.

Paulo


paulghaddad:
@pdgonzalez872 Our pleasure! If you could submit your exercises as one PR per exercise, that would be great. There are other exercises mixed in here. Let me know if you have any problems, but essentially, create a new branch off a untouched `master` branch for each exercise.

Thanks!

PR #1054, opened 2015-11-10
Arrowhead - Iteration 1

jrade-enova:
Refactored arrowhead.rb and added test cases in test_arrowhead.rb


incazteca:
Arrowhead exercise looks good. Minor note regarding testing. I know it's not required for this assignment but in future testing assignments remember to test for cases where the code will fail. In this case, for example, testing for a region or shape that is not in `CLASSIFICATIONS`.

jrade-enova:
@incazteca Will do, thank you!

paulghaddad:
@jrade-enova Nice work! You can move on!

PR #1055, opened 2015-11-10
Triangle Facts - Iteration 2

cstavitsky:
If you look at the commit titled "Revised after Zamir's comments on the first iteration pull request" it shows the most relevant changes from Iteration 1 to Iteration 2 of the exercise. Don't understand why the most recent commit on Nov 10 compares against the original version of the file, any idea?

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -17,33 +17,38 @@ class Triangle
 	end
 
 	def scalene?
-		return true unless (equilateral? || isosceles?)
-	    false
+		!equilateral? && !isosceles?
+	end
+
+	def right_triangle?
+		angles.include?(90)
 	end
 
 	def recite_facts
 		puts 'This triangle is equilateral!' if equilateral?
 		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles? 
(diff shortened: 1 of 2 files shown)


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+    angles.include?(90)
+  end
+
+  def recite_facts
+    puts 'This triangle is equilateral!' if equilateral?
+    puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles? 
+    puts 'This triangle is scalene and mathematically boring.' if scalene?
+    puts 'The angles of this triangle are ' + angles.join(',')
+    puts 'This triangle is also a right triangle!' if right_triangle?
+    puts ''
```
paulghaddad: You can simply call `puts` here without the quotes. It will print a blank line.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,64 @@
...
+  end
+
+  private
```
paulghaddad: Good job making these methods private.

paulghaddad:
@cstavitsky Not sure on the PR issues, but I was able to review it easily. This looks really nice and clean all around. Good work! You can move on.

PR #1056, opened 2015-11-10
Data science

athiercelin:
Fixes based on latest comments.


data_science/data_science.rb
```diff
@@ -0,0 +1,110 @@
...
+require 'ABAnalyzer'
+require './data_print'
+require './cohort_attributes'
+
+class DataScience
+  attr_reader :confidence_level
+  attr_reader :cohort_a, :cohort_b
+
+  def initialize(file_path)
+    @visits = []
```
zsyed1: Do you need line 11 given line 15?

data_science/data_science.rb
```diff
@@ -0,0 +1,110 @@
...
+    @cohort_a = CohortAttributes.new
+    @cohort_b = CohortAttributes.new
+    json_data = open_and_read_file(file_path)
+    @visits = JSON.parse(json_data)
+  end
+
+  def successfully_loaded?
+    !@visits.nil? || !@visits.empty?
```
zsyed1: `!@visits.nil?` is the same as `@visits`.

data_science/data_science.rb
```diff
@@ -0,0 +1,110 @@
...
+      calc_95_conv_rate_for(@cohort_b.count, @cohort_b.conversions)
+  end
+
+  def able_to_calculate_95_conv_rate_ranges?
+    if @cohort_a.rate_low == 0 || @cohort_a.rate_high == 0 ||
+       @cohort_b.rate_low == 0 || @cohort_b.rate_high == 0
+      return false
+    end
+    true
+  end
```
zsyed1: Lines 44-48
```
def able_to_calculate...
!(@cohort_a.rate_low == 0 || @cohort_a.rate_high == 0 ||
  @cohort_b.rate_low == 0 || @cohort_b.rate_high == 0)
end
```

data_science/data_science.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  def calculate_confidence_level
+    values = {}
+    values['a'] = confidence_interval(cohort_a)
+    values['b'] = confidence_interval(cohort_b)
+
+    analyzer = ABAnalyzer::ABTest.new values
+
+    @confidence_level = 0
```
zsyed1: Line 62 can be removed

data_science/data_science.rb
```diff
@@ -0,0 +1,110 @@
...
+                          data_science.cohort_a.count,
+                          data_science.cohort_a.conversions,
+                          data_science.cohort_b.avg_rate,
+                          data_science.cohort_b.count,
+                          data_science.cohort_b.conversions)
+DataPrint.print_rates_ranges(data_science.cohort_a.rate_low,
+                             data_science.cohort_a.rate_high,
+                             data_science.cohort_b.rate_low,
+                             data_science.cohort_b.rate_high)
+DataPrint.print_confidence_level(data_science.confidence_level)
```
zsyed1: From the user's perspective, it would be easier if you had only one `calculate` method that invoked all your other calculate methods. The user will most likely want to perform all the calculations.

```
def calculate
  data_science.find_counts_per_cohort
  data_science.calculate_average_conversion_rate
  data_science.calculate_95_perc_conversion_rate_range
  data_science.calculate_confidence_level
end
```

Then you can make all the sub-calculations private and expose only the `calculate` method. The `let` statement in your tests will become shorter and in fact you may want to just move the `calculate` into the initializer and make it private so that it is always called.

data_science/data_science_rspec.rb
```diff
@@ -0,0 +1,77 @@
...
+    expect(data_science.cohort_b.rate_high).to be_within(0.0001).of(0.0622)
+  end
+
+  it "should give the confidence level ", :happy do
+    confidence_calculated
+    expect(data_science).to be_able_to_produce_confidence_level
+  end
+
+  it "should compute the correct confidence level" do
+    confidence_calculated
```
zsyed1: See my comment about making a `calculate` method that is called automatically in the initializer. Would a user ever create a `DataScience` instance and not call the calculation methods?

data_science/data_science_rspec.rb
```diff
@@ -0,0 +1,77 @@
...
+    expect(data_science.confidence_level).to be_within(0.0001).of(0.031564)
+  end
+end
```
zsyed1: Nice work with the revision!

I would recommend doing the `calculate` refactor but don't think this needs a resubmission.

paulghaddad:
@athiercelin Nice work on this second iteration. You are good to move on!

athiercelin:
Thanks for the feedback! I've done the changes on my side.

paulghaddad:
Excellent! :+1:

PR #1057, opened 2015-11-11
Robot Name - Iteration 2

cstavitsky:
Should I generally be making a new branch when I submit iteration 2? Or just keep committing on the same branch I did the first iteration?

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -1,32 +1,78 @@
-class NameCollisionError < RuntimeError; end
-
-class Robot
-  attr_accessor :name
-
-  @@registry
-
-  def initialize(args = {})
-    @@registry ||= []
-    @name_generator = args[:name_generator]
-
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
-
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+class NameCollisionError < RuntimeError
+  def self.failure
+    raise 'There was a problem generating the robot name!'
   end
 end
 
-robot = Robot.new
+class Robot
+  attr_accessor :name
+  attr_reader :registry
+
+  def initialize(registry:, name: NameGenerator.new.generate_name(2, 3))
+    @registry = registry
+    @name = name
+
+    add_to_registry(@registry)
+  end
+
+  private
+
+  def add_to_registry(registry)
+    if invalid_name?
+      NameCollisionError.failure
+    end
+    registry.list << name
+  end
+
+  def invalid_name?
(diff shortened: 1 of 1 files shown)


paulghaddad:
@cstavitsky Yes, you can keep committing on the same branch as the first iteration.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,78 @@
-class NameCollisionError < RuntimeError; end
+class NameCollisionError < RuntimeError
```
paulghaddad: I don't think this is completing necessary. When you raise the exception, you can pass an optional message:

http://ruby-doc.org/core-2.2.0/Kernel.html#method-i-raise

robot_name/robot_name.rb
```diff
@@ -1,32 +1,78 @@
...
+  def self.failure
+    raise 'There was a problem generating the robot name!'
+  end
+end
 
 class Robot
```
paulghaddad: I'd recommend splitting each class into its own file, such as `robot.rb` going forward. Not a big deal for this exercise, though.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,78 @@
...
     end
+    registry.list << name
+  end
+
+  def invalid_name?
+    !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || registry.include?(name)
+  end
+end
+
+class NameGenerator
```
paulghaddad: Overall, nice work on this class!

robot_name/robot_name.rb
```diff
@@ -1,32 +1,78 @@
...
+      name += generate_char
+    end
+    num_nums.times do
+      name += generate_num
+    end
+
+    name
+  end
+
+  private
```
paulghaddad: Nice work making these methods private.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,78 @@
...
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def add_to_registry(registry)
+    if invalid_name?
+      NameCollisionError.failure
     end
+    registry.list << name
+  end
+
+  def invalid_name?
+    !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || registry.include?(name)
```
paulghaddad: Consider extracting this regex into a well-named constant.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,78 @@
...
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  private
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def add_to_registry(registry)
+    if invalid_name?
```
paulghaddad: I'd personally switch this logic around to `valid_name?`. That way, you can avoid the `!` in the `invalid_name?` method, which makes it more clear. Then you can just flip the order of the `if-else`.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,78 @@
...
+      NameCollisionError.failure
     end
+    registry.list << name
```
paulghaddad: Good use of the shovel operator.

paulghaddad:
@cstavitsky Nice overall work on this! Take a look at my comments and please submit a final iteration. Good progress on this.

PR #1058, opened 2015-11-11
Arrowhead - Iteration 2

cstavitsky:


arrowhead/arrowhead.rb
```diff
@@ -15,20 +15,38 @@ class Arrowhead
...
-    if CLASSIFICATIONS.include? region
-      shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
+    if valid_region(region) && valid_shape(region, shape)
+      "You have a(n) '#{valid_shape(region, shape)}' arrowhead. Probably priceless."
```
cstavitsky: Is this something I would need to fix in real life, or an acceptable style violation? (slightly over 80 chars, mostly due to method name and arguments)

paulghaddad: Good point; this is a one of those times where the 80 character limit may not make sense. However, I see you are calling `valid_shape` twice in the `.classify` method. You could extract this out into one method call by storing the shape as a variable.

arrowhead/arrowhead.rb
```diff
@@ -15,20 +15,37 @@ class Arrowhead
...
+    elsif !(valid_region(region))
+      region_error
     else
-      raise "Unknown region, please provide a valid region."
+      shape_error
     end
   end
+
+  def self.valid_region(region)
```
paulghaddad: I'd recommend renaming this. By using "valid", you are implying it returns a Boolean. Consider having `region` and `shape` methods and raising an exception in the method if it isn't a valid region/shape.

johnmcconnell: Agreed. Chris, I think I see what you are trying to do. There is a method on hash called: `fetch` that will clean this up.

arrowhead/arrowhead.rb
```diff
@@ -15,20 +15,37 @@ class Arrowhead
     },
   }
 
-  # FIXME: I don't have time to deal with this.
   def self.classify(region, shape)
```
johnmcconnell: Okay, so you are putting a lot of methods on the class. You can save yourself sometime by:
Doing 
```ruby
class A
  class << self
    def a
      2
    end
    def b
      3
    end
  end
end

A.a #=> 2
A.b #=> 3
```

Instead of:
```ruby
class A
  def a
    2
  end
  def b
    3
  end
end

A.a #=> 2
A.b #=> 3
```

arrowhead/arrowhead.rb
```diff
@@ -15,20 +15,37 @@ class Arrowhead
...
+      shape_error
     end
   end
+
+  def self.valid_region(region)
+    CLASSIFICATIONS[region]
+  end
+
+  def self.valid_shape(region, shape)
+    CLASSIFICATIONS[region][shape] if valid_region(region)
```
johnmcconnell: Okay so instead of making 2 methods just to grab something in a hash. You may want to know about `fetch`:
```ruby
irb(main):001:0> X = { a: 2 }
=> {:a=>2}
irb(main):002:0> X.fetch(:a)
=> 2
irb(main):003:0> X.fetch(:b)
KeyError: key not found: :b
	from (irb):3:in `fetch'
	from (irb):3
	from /Users/jmcconnell1/.rubies/ruby-2.2.3/bin/irb:11:in `<main>'
irb(main):004:0> X.fetch(:b) do
irb(main):005:1*   puts 'Oh no!'
irb(main):006:1> end
Oh no!
=> nil
irb(main):007:0>
```

paulghaddad: Great suggestion!

paulghaddad:
@cstavitsky You are free to move on. Just be aware of the methods that John suggested. Let me know if you have any questions.

PR #1059, opened 2015-11-11
Robot Name - Iteration 2

creativisionist:
• Chose to leave the splat out and leave as is -- what's the advantage to using this?  when is it more appropriate than others, or is it just preference?
• Will ask about previous rescue consideration
• Note (in case): Wasn't included in previous comment but told Arrowhead is good to go


robot_name/robot_name.rb
```diff
@@ -1,32 +1,41 @@
...
+  end
+
+  def rand_name_generator
+    @name = "#{alpha_char}#{alpha_char}#{digit}#{digit}#{digit}"
+    return @name unless invalid_name?
+    raise NameCollisionError, "There was a problem generating the robot name!"
+  end
+
+  def generate_name
+    @name_assign = @name_generator ? name_generator.call : rand_name_generator
```
ochagata: This doesn't need to be a local variable

arrowhead/arrowhead.rb
```diff
@@ -1,33 +1,26 @@
-class Arrowhead
```
paulghaddad: Arrowhead? :)

robot_name/robot_name.rb
```diff
@@ -1,32 +1,41 @@
-class NameCollisionError < RuntimeError; end
-
-class Robot
+class RobotName
   attr_accessor :name
 
-  @@registry
+  VALID_NAME_REGEX = /[[:alpha:]]{2}[[:digit:]]{3}/
```
paulghaddad: Nice, well-named constant.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,41 @@
...
 
   def initialize(args = {})
```
paulghaddad: Nicely organized `#initialize` method.

robot_name/robot_name_errors.rb
```diff
@@ -0,0 +1 @@
+class NameCollisionError < RuntimeError; end
```
paulghaddad: Excellent.

robot_name/robot_name_test.rb
```diff
@@ -0,0 +1,41 @@
...
+  describe '#alpha_char' do
+    it "generates random alpha character" do
+      expect(robot_name.alpha_char).not_to eq(robot_name.alpha_char)
+    end
+  end
+
+  describe '#digit' do
+    it "generates random digits 0..10" do
+      digit = robot_name.digit
+      expect(digit).to be >= 0
```
paulghaddad: You can use  the `be_within` matcher to keep this to one line.

robot_name/robot_name_test.rb
```diff
@@ -0,0 +1,41 @@
...
+describe RobotName do
+  before(:all) do
+    @robot_name = RobotName.new
+  end
+
+  let (:robot_name) { @robot_name }
+
+  describe '#alpha_char' do
+    it "generates random alpha character" do
+      expect(robot_name.alpha_char).not_to eq(robot_name.alpha_char)
```
paulghaddad: Will this test always pass? What if the name turns out to be the same for both. Also, at the unit test level, I'd only test the functionality of this method, which is to return a character, not to generate different characters. A better RSpec matcher would be `to_match` against a regex:

https://www.relishapp.com/rspec/rspec-expectations/v/3-4/docs/built-in-matchers/match-matcher

paulghaddad:
@creativisionist This is very nice work! I'm impressed with your tests, which will suit you very well in Module 3. You are good to move on!

PR #1060, opened 2015-11-12
Robot Name - Iteration 1

jrade-enova:
Added a while loop to continue creating a new generator if there is a name collision in the registry.
Added rescue in order for client code to continue on.
Added test cases.

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -9,24 +9,35 @@ class Robot
     @@registry ||= []
     @name_generator = args[:name_generator]
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+    begin
+      @name = generate_name(@name_generator)
+      raise NameCollisionError, 'There was a problem generating the robot name!' if is_invalid_name?(@name)
+    rescue Exception => e
+      puts e.message
+      puts e.backtrace.inspect
     end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
     @@registry << @name
   end
+
+  def generate_name(name_generator)
+    name = name_generator.call if name_generator
+
+    while is_invalid_name?(name) do
+      name_generator = create_random_generator
+      name = name_generator.call
+    end
+
+    name
+  end
+
+  def create_random_generator
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
+    name_generator = -> { "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}" }
+  end
+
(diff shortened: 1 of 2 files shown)


robot_name/robot_name.rb
```diff
@@ -9,24 +9,41 @@ def initialize(args = {})
...
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+    begin
+      @name = generate_name(@name_generator)
+      raise NameCollisionError, 'There was a problem ' \
+        'generating the robot name!' if invalid_name?(@name)
+    rescue StandardError => e
```
zbright: Since you have a custom error already, I think it would be cleaner if you added some fields to it. For example you could add a name field and a to_s method instead of doing a begin...rescue

```
class NameCollisionError < StandardError
  attr_accessor :name
  def initialize(name)
    @name = name
  end

  def to_s
    "The name #{@name} is already taken!"
  end
end
```

And then you could do `raise NameCollisionError(@name)` which would print out the string you specified in the `to_s` method

robot_name/robot_name.rb
```diff
@@ -9,24 +9,41 @@ def initialize(args = {})
...
+  end
 
-# Errors!
-# generator = -> { 'AA111' }
-# Robot.new(name_generator: generator)
-# Robot.new(name_generator: generator)
+  def create_random_generator
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
+    lambda do
```
zbright: You could replace these things with something like 

```
robot_chars = 2.times.collect { generate_char }.join
robot_numbers = 3.times.collect { generate_num }.join
"#{robot_chars}#{robot_numbers}"
```

robot_name/robot_name.rb
```diff
@@ -9,24 +9,41 @@ def initialize(args = {})
...
+      "#{generate_char.call}#{generate_char.call}" \
+      "#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+    end
+  end
+
+  def invalid_name?(name)
+    name.nil? ||
```
zbright: flip this around and it would remove the need to check for nil

```
!(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
```

This way, the regex would fail if name is nil

zbright:
All in all this looks pretty good. Well done!

robot_name/robot_name.rb
```diff
@@ -9,24 +9,41 @@ def initialize(args = {})
...
+
+    name
+  end
 
-# Errors!
-# generator = -> { 'AA111' }
-# Robot.new(name_generator: generator)
-# Robot.new(name_generator: generator)
+  def create_random_generator
+    generate_char = -> { ('A'..'Z').to_a.sample }
```
paulghaddad: You actually don't need to use lambdas here. Instance methods will work just fine.

paulghaddad: You can replace `('A'..'Z').to_a` with `[*'A'..'Z'].

paulghaddad:
@jrade-enova Nice work on this. Consider the comments for future exercises, but there is no need to submit a second iteration.

jrade-enova:
Thank you @zbright and @paulghaddad

PR #1062, opened 2015-11-16
Robot Name - Iteration 3

cstavitsky:


robot_name/robot.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'name_generator'
+require_relative 'registry'
+require_relative 'name_collision_error'
+
+class Robot
+  THREE_LETTERS_TWO_DIGITS = /[[:alpha:]]{2}[[:digit:]]{3}/
```
cstavitsky: Thanks- that makes a lot of sense.

PR #1064, opened 2015-11-17
Overlord - V1

drosholt:
App is available for use at overlord.elasticbeanstalk.com


overlord/features/deactivation.feature
```diff
@@ -0,0 +1,33 @@
+Feature: Bomb Deactivation
+
+  Scenario: Bomb is deactivated on boot
+    When the bomb is first booted
```
magicalbanana: Try having a Given/Setup step. Because When is normally used to state the key action or the action followed by the outcome or result.

overlord/features/step_definitions/initialize_steps.rb
```diff
@@ -0,0 +1,30 @@
+CODE_WITH_INVALID_FORMAT = 'ABCD'
```
magicalbanana: Instead of putting this here on your tests try to convey this on your cucumber's feature steps using tables or examples.

overlord/overlord.rb
```diff
@@ -1,14 +1,77 @@
...
+      errors
+    end
+
+    def validate_code(code)
+      code =~ /\A\d+\Z/
+    end
+
+    def set_timer_on_bomb
+      if session[:bomb].active?
+        session[:bomb].timer ||= Timer.new(FIVE_MINUTES_IN_SECONDS)
```
magicalbanana: Why not just right 300?

overlord/features/deactivation.feature
```diff
@@ -0,0 +1,33 @@
...
+    When I enter an invalid deactivation code
+    Then I should see a "Deactivation Code must be a 4-digit number" error
+
+  Scenario: Deactivate bomb with custom key
+    Given that the bomb is booted with a custom deactivation key
+    And the bomb has been activated
+    When I enter the custom deactivation code
+    Then the bomb should be in a Deactivated state
+
+  Scenario: Explosion
```
magicalbanana: Since this is the deactivation feature it would make more sense if there was an explosion or disarming feature to accommodate explosion since it's really not worthy of a scenario given that other steps contribute to that outcome.

magicalbanana:
First off, please do not commit everything on one commit and write better commit titles/summaries.

I would like to see the commits broken apart and better detailed. Also, address the comments as well I would like to see your unit tests revisited. It seems your it block statements or summaries are more in line of a context than an expectation.

Also, I want to see the Timer class rewritten. I'm not to fancy about the to_s method you have. Instead of doing it so, why not appropriate it to a method that does what it's written to do. You're method is not just transforming it to a string, it's reformatting it and defining it altogether.

And keep in mind, #to_s is a method already in a Ruby object.

Also, try revisiting the bomb class. The way you wrote the states are a good chance to learn on states for a Ruby object.

I also want to see coverage for your app, use the simplecov gem to determine your test coverage so I can gauge better if I require to see more tests written or not.

So address my comments and resubmit. Thanks!

overlord/features/activation.feature
```diff
@@ -0,0 +1,48 @@
+Feature: Bomb Activation
+
+  Scenario: Activate bomb with default key
```
paulghaddad: These are fantastic feature tests. High and single level of abstraction.

overlord/features/step_definitions/navigation_steps.rb
```diff
@@ -0,0 +1,3 @@
+Then /^(?:|I )should be redirected to (.+)$/ do |page_name|
```
paulghaddad: I like how you break out your step definitions into separate files. Good organization.

PR #1067, opened 2015-11-19
Dinosaur Catalog - Iteration 1

cstavitsky:
Please don't hold back! Would like as much feedback as you are willing to give.

Sorta tried to start writing tests and decided to wait till I go through the LevelUp section on testing to really focus on that. So tests are definitely limited.


dino_catalog/lib/dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,65 @@
+class DinoCatalog::Dinosaur
```
cstavitsky: Does this mean,

module DinoCatalog
     class Dinosaur
     end
end

?

dino_catalog/README.md
```diff
@@ -1,28 +1,70 @@
...
-2. I have friends who ask me a lot of questions about dinosaurs (I'm kind of a big deal). Please make sure the dinodex is able to answer these things for me:
-    * Grab all the dinosaurs that were bipeds.
-    * Grab all the dinosaurs that were carnivores (fish and insects count).
-    * Grab dinosaurs for specific periods (no need to differentiate between Early and Late Cretaceous, btw).
-    * Grab only big (> 2 tons) or small dinosaurs.
-    * Just to be sure, I'd love to be able to combine criteria at will, even better if I can chain filter calls together.
-3. For a given dino, I'd like to be able to print all the known facts about that dinosaur. If there are facts missing, please don't print empty values, just skip that heading. Make sure to print Early / Late etc for the periods.
-4. Also, I'll probably want to print all the dinosaurs in a given collection (after filtering, etc).
+```
+importer = DinoCatalog::DinoImporter.new('path_to_csv_file.csv') #DinoImporter.dinosaur_list contains a collection of dinosaurs importer from the csv file
```
drosholt: Text that's this long doesn't fit inside Github's code block window, so it's hard to parse.  You can always look at the README in your private branch to see how it will look once merged.

dino_catalog/dino_catalog.gemspec
```diff
@@ -0,0 +1,19 @@
+Gem::Specification.new do |s|
+  s.name = 'dino_catalog'
+  s.version = '0.0.0'
```
drosholt: You should probably use a valid version number.  If you want to refer to something as pre-release, you can append a string to the version number.  "-pre" is a common identifier.

dino_catalog/dino_catalog.gemspec
```diff
@@ -0,0 +1,19 @@
...
+  s.date = '2015-11-15'
+  s.summary = 'Dino Catalog'
+  s.description = 'Import Dinosaur data and filter by parameters.'
+  s.authors = ['Chris Stavitsky']
+  s.email = 'cstavitsky@enova.com'
+  s.files = [
+    "lib/dino_catalog.rb",
```
drosholt: You should be consistent in using either single-quotes or double-quotes for your non-interpolated strings.

dino_catalog/lib/dino_catalog.rb
```diff
@@ -0,0 +1,9 @@
+module DinoCatalog
+end
+
+require_relative './dino_catalog/dinosaur'
```
drosholt: There's nothing really wrong with this code, but the traditional place for require statements is at the top of the file.  This is where others will typically look for any referenced files.  http://stackoverflow.com/questions/605261/do-ruby-require-statements-go-inside-or-outside-the-class-definition.

dino_catalog/lib/dino_catalog/dino_importer.rb
```diff
@@ -0,0 +1,55 @@
+require_relative '../dino_catalog'
```
drosholt: If you are creating a gem, your lib directory is placed on the library path.  So, you could use ```require 'dino_catalog/dino_importer'``` and ```require 'dino_catalog'``` instead of basing everything off of the current directory.

dino_catalog/dino_catalog.gemspec
```diff
@@ -0,0 +1,19 @@
+Gem::Specification.new do |s|
```
drosholt: I'm guessing that you created this gemspec by hand?  Take a look at bundler's tools for creating gems: http://bundler.io/rubygems.html.  It sets up a few conventions for you: versioning, Gemfiles, etc.  One of the things that you are missing here, for example, is something that defines rspec as a dependency.  If I didn't have rspec installed, there would be no way of knowing from your project metadata that I need to install it, or which version I need to install.

dino_catalog/lib/dino_catalog/dino_importer.rb
```diff
@@ -0,0 +1,55 @@
...
+  def parse_csv(file, format = nil)
+    CSV.foreach(file, headers: true) do |row|
+      format_pirate_bay_data(row) if format == "pirate_bay"
+      dinosaur = convert_to_dinosaur(row)
+      add_to_list(dinosaur)
+    end
+    dinosaur_list
+  end
+
+  def convert_to_dinosaur(row_data)
```
drosholt: Is there a way to make this simpler?  Your parameter names and your hash names are virtually identical.

dino_catalog/lib/dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,35 @@
+require_relative '../dino_catalog'
+
+class DinoCatalog::Dinodex
+  attr_reader :dinosaurs
+
+  def initialize(dinosaurs)
+    @dinosaurs = dinosaurs
+  end
+
+  def carnivores
```
drosholt: A simpler way to write this is ```["carnivore", "piscivore", "insectivore"].include?(dinosaur.diet.downcase)```.  Are there further optimizations that you could make?

dino_catalog/lib/dino_catalog/dinosaur.rb
drosholt: That's exactly what it means.  I wouldn't worry about it too much.

dino_catalog/lib/dino_catalog/dino_importer.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def import_from_csv(file, format = nil)
+    parse_csv(file, format)
+  end
+
+  private
+
+  def parse_csv(file, format = nil)
+    CSV.foreach(file, headers: true) do |row|
+      format_pirate_bay_data(row) if format == "pirate_bay"
```
drosholt: Options in ruby are usually represented as symbols, so :pirate_bay would be a better choice.

```format: :pirate_bay``` is very clear: use the pirate_bay format.  What does ```format: nil``` mean?

File: dino_catalog/lib/dino_catalog/dino_importer.rb [hunk 11]
```diff
@@ -0,0 +1,55 @@
...
+    )
+  end
+
+  def format_pirate_bay_data(row_data)
+    row_data["NAME"] = row_data["Genus"]
+    row_data["PERIOD"]	= row_data["Period"]
+    row_data["CONTINENT"] =     "Africa"
+    row_data["WEIGHT_IN_LBS"] = row_data["Weight"]
+    row_data["WALKING"] =       row_data["Walking"]
+    if row_data["Carnivore"] == "Yes"
```
drosholt: A ternary might be more concise here.

dino_catalog/lib/dino_catalog/dino_serializer.rb
```diff
@@ -0,0 +1,15 @@
+class DinoCatalog::DinoSerializer
+  def self.export_json(dino_collection)
+    dino_collection.map(&:to_json)
+  end
+
+  # private
```
drosholt: You should remove commented-out code before committing.

dino_catalog/lib/dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,35 @@
...
+  def bipeds
+    filter(attribute: "walking", value: "biped")
+  end
+
+  def filter(attribute:, value:)
+    dinos = dinosaurs.select do |dinosaur|
+      val_of_attribute = dinosaur.send(attribute)
+      val_of_attribute.downcase == value.downcase unless val_of_attribute.nil?
+    end
+    self.class.new(dinos)
```
drosholt: It's usually a better idea to use Dinodex.new(dinos).  That way, if this code gets refactored, you are still guaranteed that it will mean the same thing.

dino_catalog/lib/dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,65 @@
...
+    @size	= calculated_size(weight_in_lbs.to_i)
+    @walking	= walking
+    @description	= description
+  end
+
+  def print_facts
+    puts formatted_facts
+  end
+
+  def to_json
```
drosholt: This is totally fine for dinodex, but for future reference, this feels a little brittle.  What if I also need to output as CSV, BSV, CORBA, EBCDIC, etc? Would I have a to_xxx method for each one of those output types?  Or is there a better design pattern that will allow me to vary how I externalize this information>

dino_catalog/lib/dino_catalog/dino_importer.rb
drosholt: Come to think of it, do you even need the default value here?

dino_catalog/lib/dino_catalog/dino_importer.rb
```diff
@@ -0,0 +1,55 @@
+require_relative '../dino_catalog'
+
+class DinoCatalog::DinoImporter
+  attr_reader :dinosaur_list
+
+  def initialize(csv_file)
+    @dinosaur_list = []
```
drosholt: If this class is holding on to the current set of dinosaurs, is it just a dino importer?

dino_catalog/spec/unit/dino_importer_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    @importer = DinoCatalog::DinoImporter.new('lib/dinodex.csv')
+  end
+
+  describe '#import_from_csv' do
+    it 'returns a collection of Dinosaur objects' do
+      @importer.import_from_csv('lib/dinodex.csv')
+      expect(@importer.dinosaur_list).to all(be_a DinoCatalog::Dinosaur)
+    end
+
+    it 'changes data format for pirate bay format'
```
drosholt: ?

dino_catalog/lib/dino_catalog/dinodex.rb
drosholt: Is the desired behavior for somebody that wants to chain multiple filters together, using multiple lines of code, that they would have to assign each result to a variable?
```
d1 = dinodex.filter(attribute: "walking", value: "biped")
d2 = d1.filter(attribute: "diet", value: "carnivore")
d3 = d2.filter(attribute: "period", value: "Early Cretaceous")
etc
```

dino_catalog/lib/dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def print_collection(dino_collection = dinosaurs)
```
drosholt: What is the reason behind being able to print any collection from dinodex?  Why would a particular dinodex instance need to be aware of other instances?

dino_catalog/dino_catalog.gemspec
cstavitsky: Thanks for this resource! I did create the file by hand. Appreciate you pointing out that rspec wouldn't have been included. I had spoken at some point to someone who was surprised to see I didn't have a Gemfile so this makes sense.

dino_catalog/dino_catalog.gemspec
cstavitsky: Thanks. I had a pairing session today and spoke to my pair about this comment, who reinforced what the 3 different digits mean.

dino_catalog/lib/dino_catalog/dino_importer.rb
cstavitsky: Do you mean, if you are creating a gem using the skeleton generator that you linked to before? In the gemspec file of a test gem I generated, it shows 

```
lib = File.expand_path('../lib', __FILE__)
$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
```

Is that what you're referring to? Is this a thing with gems that are manually created as well?

dino_catalog/lib/dino_catalog/dino_importer.rb
cstavitsky: Good point, thanks.

dino_catalog/lib/dino_catalog/dinodex.rb
cstavitsky: Thoughts on giving Dinosaur the instance method 

```
class Dinosaur 
     ...
     def carnivore?
          ["carnivore", "piscivore", "insectivore"].include?(dinosaur.diet.downcase)
     end
end
``` 

to hold your suggested code? And then, in Dinodex:

```
class Dinodex
     ...
     def carnivores
          dinosaurs.select { |dinosaur| dinosaur.carnivore? }
     end
end
```

? Would like to hear what you think of this, and then what else you might have been thinking of as a further optimization.

dino_catalog/lib/dino_catalog/dinodex.rb
cstavitsky: "It's usually a better idea to use Dinodex.new(dinos). That way, if this code gets refactored, you are still guaranteed that it will mean the same thing."

Question about this one, would self.class.new would be more stable against refactoring if someone were to change the name of the Dinodex class?

Also confused about what you are saying with this comment "Is the desired behavior for somebody that wants to chain multiple filters together, using multiple lines of code, that they would have to assign each result to a variable?". I think the desired behavior is to chain filters so you can say, 

```
dinodex.filter(attribute: "walking", value: "biped").filter(attribute: "diet", value: "carnivore").filter(attribute: "period", value: "Early Cretaceous")
```

dino_catalog/lib/dino_catalog/dinodex.rb
cstavitsky: Good point!

dino_catalog/lib/dino_catalog/dinosaur.rb
cstavitsky: Turns out that nesting the classes properly under the modules allows the require_relative files in the dino_catalog.rb to be put at the top of the file (I had listed them at the bottom of the code because it was breaking when I put them on top!).

dino_catalog/lib/dino_catalog/dinosaur.rb
cstavitsky: My first instinct is to say something like

```
FORMATS = [:json, :csv, ...]

def export(format)
     if FORMATS.include?(format)
          send(format)
     end
end
```

And then have methods called "csv", "json", etc. But that doesn't really seem like it solves the problem. I'm unsure how to proceed, doesn't each file format need to have its own way of processing somewhere in the code?

dino_catalog/lib/dino_catalog/dino_importer.rb
cstavitsky: ```format: nil``` currently means that the parser will default to the format found in dinodex.csv. Would it be clearer to call the default format ```:default``` ?

dino_catalog/lib/dino_catalog.rb
djkotowski: As we discussed, the way around the undefined constant error is to wrap the class with the namespace module.

Example:
```ruby
module DinoCatalog
  class DinoImporter
    # Your stuff here
  end
end

# instead of

class DinoCatalog::DinoImporter
  # Your stuff here
end
```

dino_catalog/lib/dino_catalog/dinosaur.rb
djkotowski: I will say that `#to_json` and `#to_h` are accepted ruby-like methods.  I'm happy with this format.

dino_catalog/lib/dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,65 @@
...
+    json_hash = {}
+    instance_variables.each do |variable|
+      json_hash[variable] = instance_variable_get(variable)
+    end
+    json_hash.to_json
```
djkotowski: Why not extract everything except `json_hash.to_json` to a method `to_h` (the standard method for creating hashes).  Then you can reuse the hash creation method if you need, and just call `to_h.to_json` in this method.

dino_catalog/lib/dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,65 @@
...
+    @continent	= continent
+    @diet	= diet
+    @weight_in_lbs	= weight_in_lbs
+    @size	= calculated_size(weight_in_lbs.to_i)
+    @walking	= walking
+    @description	= description
+  end
+
+  def print_facts
+    puts formatted_facts
```
djkotowski: Should Dinosaur be used for console output?

dino_catalog/lib/dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def bipeds
+    filter(attribute: "walking", value: "biped")
+  end
+
+  def filter(attribute:, value:)
+    dinos = dinosaurs.select do |dinosaur|
+      val_of_attribute = dinosaur.send(attribute)
+      val_of_attribute.downcase == value.downcase unless val_of_attribute.nil?
```
djkotowski: This will fail if `dinosaur.send(attribute)` is not a String.

dino_catalog/lib/dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def initialize(dinosaurs)
+    @dinosaurs = dinosaurs
+  end
+
+  def carnivores
+    dinosaurs.select do |dinosaur|
+      dinosaur.diet.downcase == "carnivore" ||
+        dinosaur.diet.downcase == "piscivore" ||
+        dinosaur.diet.downcase == "insectivore"
```
djkotowski: This could be simplified:

```ruby
%w( carnivore piscivore insectivore).include?(dinosaur.diet.downcase)
```

dino_catalog/lib/dino_catalog/dinodex.rb
djkotowski: Also, generally the preferred method for defining an array of strings is:

```ruby
%w( carnivore piscivore insectivore )
```

drosholt: Actually, it's %w[carnivore, etc]

dino_catalog/lib/dino_catalog/dinodex.rb
djkotowski: Also, should Dinodex be responsible for determining whether or not a Dinosaur is a biped, or should Dinosaur handle that?

dino_catalog/lib/dino_catalog/dinodex.rb
djkotowski: https://github.com/bbatsov/ruby-style-guide#collections

dino_catalog/lib/dino_catalog/dinosaur.rb
djkotowski: As a side note, if you did want to move `#to_json` out, still having `#to_h` in this class would help.

djkotowski:
Looking good.  Take a look at the comments, refactor and resubmit.  Let me know if you have any questions.

PR #1069, opened 2015-11-23
Dinodex - Iteration 1

jrade-enova:


dino_catalog/dinodex/test_library.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def test_create_hash_from_mapping
+    column_mapping = {"name" => "genus", "weight_in_lbs" => "weight"}
+
+    hash = {"genus" => "foo", "weight" => "10", "bar" => "42"}
+    actual_hash = Library.create_hash_from_mapping(hash, column_mapping)
+    expected_hash = {"name" => "foo", "weight_in_lbs" => "10"}
+    assert_equal(expected_hash, actual_hash, "Expected hash is not equal")
+
```
magicalbanana: Plus on using unit test and testing this early. Right on.

dino_catalog/dinodex/test_library.rb
```diff
@@ -0,0 +1,59 @@
...
+    expected = ["key", "<=", "2"]
+    assert_equal(expected, actual, "Actual is not equal to expected")
+
+    string = "=~bar"
+    actual = Library.parse_search_values(key, string)
+    expected = ["key", "=~", "bar"]
+    assert_equal(expected, actual, "Actual is not equal to expected")
+  end
+
+end
```
magicalbanana: I know it's not really a unit test thing but you could split this into different test categories or context. And add more scenarios to your test.

dino_catalog/README.md
```diff
@@ -26,3 +26,55 @@ Go check out the CSVs and come back. Done? Cool, I've just got a few features I
 2. CSV isn't may favorite format in the world. Can you implement a JSON export feature?
 
 Happy Hunting. (Giganotosaurus was the largest hunting dinosaur, at 46 feet long and up to 8 tons! Suh-weet.)
+
+## Dinodex Implementation ##
+
+##### _Note: The approach taken here is to accept any CSV file of any size_ #####
+
+### Prerequisites ###
```
DejaAugustine: If external libraries are required for this application to run, you should provide an appropriate Gemfile.

dino_catalog/dinodex/csv_utility.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+  end
+
+  def self.read_csv(path)
+    raise "You need to specify a path and file" unless valid_path?(path)
+    csv_options = {:encoding => "UTF-8", :skip_blanks => true, :headers => true}
+    data_objects = process_csv_rows(path, csv_options)
+  end
+
+  def self.process_csv_rows(path, csv_options)
```
DejaAugustine: Consider making this method private

dino_catalog/dinodex/csv_utility.rb
```diff
@@ -0,0 +1,45 @@
...
+    data_objects = []
+
+    CSV.foreach(path, csv_options) do |row|
+      data_objects << process_row(row)
+    end
+
+    data_objects
+  end
+
+  def self.process_row(row)
```
DejaAugustine: Consider making this method private

dino_catalog/dinodex/data/hbase_service.rb
```diff
@@ -0,0 +1,48 @@
+require 'stargate'
+
+module HbaseService
```
DejaAugustine: What is the benefit of having a module full of static methods and class variables over a standard class with instance variables?

dino_catalog/dinodex/data/hbase_service.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  TABLE = "dinodex"
+  CF = "cf"
+  QUAL = "data"
+  URL = "http://localhost:8777"
+
+  @@client = Stargate::Client.new(URL, { :proxy => URL })
+
+  def self.init
```
DejaAugustine: When is this called?

dino_catalog/dinodex/data/hbase_service.rb
```diff
@@ -0,0 +1,48 @@
+require 'stargate'
+
+module HbaseService
+
+  TABLE = "dinodex"
+  CF = "cf"
+  QUAL = "data"
+  URL = "http://localhost:8777"
+
+  @@client = Stargate::Client.new(URL, { :proxy => URL })
```
DejaAugustine: I'm not a big fan of setting variables outside of methods.  Would it be better to place this in an initialize method using the `||=` syntax?

dino_catalog/dinodex/data/hbase_service.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def self.init
+    @@client.create_table(TABLE, CF)
```
DejaAugustine: Rather than relying on `@@` notation, try using an accessor method

dino_catalog/dinodex/data/hbase_service.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def self.put(key, data)
+    raise "You must provide key and data" if key.nil? || data.nil?
+    @@client.create_row(TABLE, key, Time.now.to_i, {:name => "#{CF}:#{QUAL}", :value => Marshal.dump(data)})
```
DejaAugustine: Since `QUAL` is never used outside of the `"#{CF}:#{QUAL}"` context, why not make that CF:QUAL pairing a constant rather than re-evaluating the string interpolation every time?

dino_catalog/dinodex/library.rb
```diff
@@ -0,0 +1,137 @@
+require './csv_utility'
+require './data/hbase_service.rb'
+
+module Library
```
DejaAugustine: In the same vein as my question about `module HbaseService` why is this a module of static methods rather than a class?

dino_catalog/README.md
jrade-enova: Added Gemfile and Gemfile.loc and also updated README.md

dino_catalog/dinodex/library.rb
```diff
@@ -0,0 +1,137 @@
...
+  def self.build_search_params(search_options)
+    search_keys = []
+    search_criteria = []
+
+    search_options.each do |search_key, search_value|
+      search_keys << search_key
+      search_criteria << parse_search_values(search_key, search_value)
+    end
+
+    return search_keys, search_criteria
```
jrade-enova: Removing 'return' gives error: syntax error, unexpected '\n', expecting '=' (SyntaxError)

magicalbanana: Well yeah because you're returning two items. Ignore the hounds nag,

dino_catalog/dinodex/csv_utility.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+  end
+
+  def self.read_csv(path)
+    raise "You need to specify a path and file" unless valid_path?(path)
+    csv_options = { :encoding => "UTF-8", :skip_blanks => true, :headers => true }
+    process_csv_rows(path, csv_options)
+  end
+
+  def self.process_csv_rows(path, csv_options)
```
DejaAugustine: Consider making this method private

dino_catalog/dinodex/data/hbase_service.rb
```diff
@@ -0,0 +1,47 @@
...
+    row = @@client.show_row(TABLE, key)
+    value = row.columns[0].value
+    Marshal.load(value)
+  end
+
+  def self.delete(key)
+    @@client.delete_row(TABLE, key, nil, "#{CF}:#{QUAL}")
+  end
+
+  def self.scan_values
```
DejaAugustine: This method doesn't appear to be used anywhere

dino_catalog/dinodex/csv_utility.rb
jrade-enova: Done

dino_catalog/dinodex/data/hbase_service.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  @@client = Stargate::Client.new(URL, :proxy => URL)
+
+  def self.init
+    @@client.create_table(TABLE, CF)
+  end
+
+  def self.put(key, data)
+    raise "You must provide key and data" if key.nil? || data.nil?
+    @@client.create_row(TABLE, key, Time.now.to_i, :name => "#{CF}:#{QUAL}", :value => Marshal.dump(data))
```
DejaAugustine: Since you're using a whole databasing library, is there a more effective use for it than simply storing serialized CSV files?

dino_catalog/dinodex/data/hbase_service.rb
jrade-enova: Deleted

dino_catalog/dinodex/data/hbase_service.rb
jrade-enova: So I initially thought of that when starting, but decided to avoid it since it would have added complexity that is outside the scope of the exercise.  It was also suggested to avoid persistence all together but wanted to at least have a mechanism in place for files too large to keep in memory.

dino_catalog/dinodex/data/hbase_service.rb
jrade-enova: From what I have read, I am under the impression that classes are good for object creation and keeping state.  Modules are for libraries and singleton-like functionality.

dino_catalog/dinodex/library.rb
```diff
@@ -0,0 +1,138 @@
...
+
+  def self.remove_file(name)
+    HbaseService.delete(name)
+  end
+
+  def self.get_data_columns(name)
+    get_data(name).first.keys
+  end
+
+  def self.get_all_files_loaded
```
DejaAugustine: Since `scan_rowkeys` is included as part of the `include HbaseService` on line 5, why are you providing a wrapper method?  (Also applies to `add_data()` and `delete()`)

dino_catalog/dinodex/data/hbase_service.rb
DejaAugustine: I fear that this may not actually solve that problem.  If the file is too large to keep in memory, serializing it wholesale won't solve that problem, since you need to deserialize it to actually parse it anyway.

dino_catalog/dinodex/library.rb
```diff
@@ -0,0 +1,138 @@
...
+    end
+  end
+
+  def self.parse_search_values(key, string)
+    return [key, "==", ""] if string.nil? || string.length == 0
+    return [key, "=~", string.split(/=~/).last] if string.include?("=~")
+    foo(key, string)
+  end
+
+  def self.foo(key, string)
```
DejaAugustine: Can you come up with something a little more descriptive than `foo()`?

dino_catalog/dinodex/library.rb
```diff
@@ -0,0 +1,138 @@
...
+    result_data
+  end
+
+  def self.build_search_params(search_options)
+    search_keys = []
+    search_criteria = []
+
+    search_options.each do |search_key, search_value|
+      search_keys << search_key
+      search_criteria << parse_search_values(search_key, search_value)
```
DejaAugustine: You pass `key` through the entire `parse_search_values()` chain without actually referencing it once.  Consider refactoring this line to something more like `search_criteria << [search_key, parse_search_values(search_value)]` to keep the rest of the execution chain simpler.

dino_catalog/dinodex/library.rb
```diff
@@ -0,0 +1,138 @@
...
+    filter_data_objects(data, search_keys, search_criteria)
+  end
+
+  def self.filter_data_objects(data, search_keys, search_criteria)
+    result_data = []
+
+    data.each do |data_object|
+      data_object.each do |obj_key, obj_value|
+        if search_keys.include?(obj_key) && !obj_value.nil?
+          matches = build_match_lambda(obj_key, obj_value, search_criteria).call
```
DejaAugustine: At this point, you have everything you need to make the comparison.  What benefit is there to building a lambda over simply doing something like `obj_value.send(search_criteria[1], search_criteria[0])`?

dino_catalog/dinodex/library.rb
DejaAugustine: As an added note, why are you returning `search_key` in both `search_keys` and `search_criteria`?  Seems redundant.

dino_catalog/dinodex/data/hbase_service.rb
jrade-enova: Done

dino_catalog/dinodex/data/hbase_service.rb
jrade-enova: Updated with [this](https://github.com/jrade-enova/level_up_exercises/commit/c674660adc5d34311d776fad7a45a6fd3b13b4e2) commit

dino_catalog/dinodex/library.rb
jrade-enova: Same idea I think.  Library are a set of functions to be used by other classes as opposed to state-ful objects to be created

dino_catalog/dinodex/library.rb
jrade-enova: That was on my todo list and forgot to rename it :)

dino_catalog/dinodex/test_csv_utility.rb
```diff
@@ -0,0 +1,33 @@
+require "./csv_utility"
+require "test/unit"
+
+class TestCsvUtility < Test::Unit::TestCase
```
DejaAugustine: :+1: for unit tests!

As you'll find out in the Test Engineering, we use RSpec at Enova, so you may want to consider using it in future projects.

dino_catalog/dinodex/test_csv_utility.rb
```diff
@@ -0,0 +1,33 @@
...
+    expected = "a,b\n1,2\n2,4\n3,6\n"
+    assert_equal(expected, actual, "CSV String does not equal expected")
+  end
+
+  def test_valid_path
+    actual = CsvUtility.valid_path?("")
+    assert_equal(false, actual, "Empty String is an invalid file path")
+
+    actual = CsvUtility.valid_path?("foo/bar.txt")
+    assert_equal(false, actual, ".txt is an invalid file path")
```
DejaAugustine: As a best practice, try to stick to one assert per test.

dino_catalog/dinodex/library.rb
jrade-enova: I use `key` on line 112 when adding to the return array.

jrade-enova: I see now the redundancy with `search_keys` array.  It looks like I can probably get rid of it and write a method that will look for the `search_key` in the `search_criteria` array.

DejaAugustine: My point was that you pass `key` in through several methods just to prepend it to the return array that gets passed right back out. The `key` argument could be refactored away by the example (`search_criteria << [search_key, parse_search_values(search_value)]`)

dino_catalog/dinodex/data/hbase_service.rb
```diff
@@ -0,0 +1,42 @@
...
+
+module HbaseService
+  TABLE = "dinodex"
+  CF = "cf"
+  QUAL = "data"
+  URL = "http://localhost:8777"
+
+  attr_accessor :hbase_client
+
+  def self.init
```
DejaAugustine: `init` implies the initialization of an object. Considering this is a Module and looking at how this method is used, consider simply calling it `client`.  That will allow you to condense the following code from:
```(ruby)
client = init
client.delete_row(TABLE, key, nil, "#{CF}:#{QUAL}")
```
down to
```(ruby)
client.delete_row(TABLE, key, nil, "#{CF}:#{QUAL}")
```

DejaAugustine: It also makes it very clear that the method is an accessor.

dino_catalog/dinodex/data/hbase_service.rb
```diff
@@ -0,0 +1,42 @@
+require 'stargate'
+
+module HbaseService
+  TABLE = "dinodex"
+  CF = "cf"
+  QUAL = "data"
+  URL = "http://localhost:8777"
+
+  attr_accessor :hbase_client
```
DejaAugustine: Since the only time you reference hbase_client is on Line 12, and you are setting the variable directly, there isn't actually any need to have an accessor for it.

dino_catalog/dinodex/dinodex.thor
```diff
@@ -0,0 +1,61 @@
+require 'json'
+require './csv_utility'
+require './library'
+
+class Dinodex < Thor
+  include Library
```
DejaAugustine: Because you `include Library` in the `Dinodex` class, all of the methods described in the `Library` module (and by inclusion the `HbaseService` module) are considered methods of `Dinodex`.

This means that the following is completely valid:
```(ruby)
d = Dinodex.new
d.add_data(file, CsvUtility.read_csv(file))
d.init.delete_row(HbaseService::TABLE, key, nil, "#{HbaseService::CF}:#{HbaseService::QUAL}")
```

dino_catalog/dinodex/dinodex.thor
```diff
@@ -0,0 +1,61 @@
...
+
+  desc "load_file FILE", "This task will load a single file with path provided"
+  def load_file(file)
+    data = CsvUtility.read_csv(file)
+    Library.add_data(file, data)
```
DejaAugustine: Because Library is included on Line 6, you don't actually need to prepend the module.  This could be rewritten as simply `add_data(file, data)`.

DejaAugustine: N/m, all of those methods are static.

dino_catalog/dinodex/dinodex.thor
DejaAugustine: Self-correction: the module methods are all static, so they could be used as follows instead:

```(ruby)
Dinodex.init.delete_row(HbaseService::TABLE, key, nil, "#{HbaseService::CF}:#{HbaseService::QUAL}")
```

dino_catalog/dinodex/data/hbase_service.rb
DejaAugustine: While all of that is very true, it misses the key drawback with Modules: code coupling and its impact on reusability.

While the module itself can easily be re-used with other modules/classes, any code that includes the module becomes tightly coupled to it. Consider what code you'll have to rewrite when it becomes necessary for Library to be backed by redis instead of hbase.

jrade-enova: I know how to write Interfaces in other languages, but how do I do that with Ruby?  Should I have instead wrote some generic "data access" module that sits between `hbase_service.rb` and `library.rb`?

DejaAugustine: A data access module doesn't really solve the problem, as Library is still coupled to DataAccess, and DataAccess is coupled to HbaseService, meaning that replacing HbaseService would require refactoring DataAccess.

Using classes helps alleviate that coupling a bit. Library would still be tethered to the interface provided by HbaseService, but instead of having to refactor every wrapper function (`add_data`, `get_data`, `remove_file`, `get_data_columns`, and `get_all_files_loaded`) you can simply refactor a single accessor or initialize method to use a different data store instance.

dino_catalog/dinodex/library.rb
jrade-enova: Fixed.  Removed lambda evaluation and now using .send()

dino_catalog/dinodex/data/hbase_service.rb
jrade-enova: Done

dino_catalog/dinodex/library.rb
magicalbanana: @jrade-enova start addressing the hounds comments about syntax and what not, we use rubocop here so formatting your code is also essential.

dino_catalog/dinodex/library.rb
jrade-enova: Got it.  Fixed.

DejaAugustine:
@jrade-enova, while this is an interesting approach, it feels over-engineered for what you get.

Your singleton approach using modules and static methods is valid in many ways, but it doesn't really feel very ruby, and the way that you've implemented it here leads to strong code coupling and inter-dependence that inhibits the benefits of object-oriented programming (such as encapsulation and reusability).  What you've essentially done is created one large atomic Dinodex object with code that's distributed across several code files.

Further, your use of the hbase library adds complexity without actually providing a benefit. Since you are serializing and deserializing the entire file's contents, it actually uses more memory than if you were to simply read the files straight off of the disk. Rather than solving the problem of working with large datasets, it exacerbates it.

I'd like you to resubmit this exercise after refactoring to take one of the following two approaches:

1. Rewrite the project to work without hbase
2. Refactor the use of hbase to provide a specific benefit

I'd also like to see the following overall changes:

# Use rubocop to resolve any syntax errors prior to review
# Leverage Ruby's encapsulation to make the project more modular (e.g. be able to swap out the CSV parser for a JSON parser or an XML parser with a minimum amount of refactor required.  Note, you don't actually have to implement JSON or XML parsers, but develop the CSV parser and attendant Library object to support different data formats)

Once you've refactored, create a new PR and hipchat me when it's ready for review.

PR #1070, opened 2015-11-23
Traffic Control API - V1

drosholt:
Run ```rake routes``` to generate the routes.  A couple of comments:
* There isn't any referential integrity built into this.  Want to have your plane go from ORD -> LAX, magically transport to SFO, and then go from SFO -> HKG?  Go for it.
* Refueling is assumed to happen at each destination.
* Airports do have a numeric primary id, but are expected to be looked up by code.  Linkages are done via the airport code, but relationships aren't hard-coded from a json-api standpoint.


traffic_control/db/migrate/20151106205445_add_boeing737_to_plan_type.rb
```diff
@@ -0,0 +1,5 @@
+class AddBoeing737ToPlanType < ActiveRecord::Migration
+  def change
+    PlaneType.create!(name: 'Boeing 737', seat_count: 143)
+  end
```
ssapra: I would rather see all database seeding happen in the seeds file instead of individual migrations

drosholt: Thanks.  I actually only learned about the seed file a few days after I submitted this.  Will definitely do that.

ssapra:
It's a bit difficult to see the individual routes that match the requirements of the API since there are so many routes.

I believe you are using the plane has many flights relationship to cover the "See the trip an airplane will make regardless of individual itineraries". How do you identify a trip which might have a few stop overs? Do you store something to identify which stop overs are refuel stops?

While I think using the jsonapi-resources gem is great for building such an API, I would have liked to see the raw work in just designing the API by defining only the routes instead of building an entire rails app.

Great work overall!

paulghaddad:
@drosholt Nice work, David. You've satisfied the requirements of the exercise. Feel free to move on!

drosholt:
Thanks.  I agree - the API doesn't print as well as I would have liked.  I went with that approach anyway because I was already familiar with HATEOAS, and was looking for a more concrete learning opportunity in ruby.

This documentation challenges represent global flaws in web APIs.  Given the lack of context provided by JSON and HATEOAS and the challenges in communicating the intent of an API, people have spent time putting together specifications such as RAML and WADL.  The community is, in general, against these documentation standards (for good reason), but they at least attempt to address some of the gaps in our purer implementations.

I'm not sure what you mean by "a few stopovers".  I imagined that the only thing a plane does is go from one airport to another, so understanding a plane's path is represented by accessing its flights, filtering by date, and then sorting in date ascending.

For refueling, I cheated and assumed that planes would refuel at each stop.  In general, at least for non-Cessna-type planes, this is true because it is so much cheaper than carting extra fuel around.  There are some exceptions, particularly for extremely small airports which still require some regional connections.  If I were to design that, I would identify airports that have refueling capabilities and then put a flag on stopovers that require refueling.

PR #1071, opened 2015-11-24
Yadda Part 1 - V1

drosholt:
I went a little overboard with the comments, but it was in good fun.


yadda/yadda.sql
```diff
@@ -0,0 +1,134 @@
...
+CREATE ROLE yadda_ui WITH LOGIN PASSWORD 'yaddayaddayadda' IN ROLE yadda_rw;
+COMMENT ON ROLE yadda_rw IS 'Inheritable role for reading/writing/deleting data on the system.';
+COMMENT ON ROLE yadda_ui IS 'Login role for the user interface to interact with the system.';
+
+CREATE SCHEMA yadda;
+GRANT USAGE ON SCHEMA yadda TO yadda_ro;
+GRANT USAGE ON SCHEMA yadda TO yadda_rw;
+COMMENT ON SCHEMA yadda IS 'Contains all read-writable data maintained by via the Yadda web site.';
+
+CREATE OR REPLACE FUNCTION yadda.set_updated_timestamp()
```
incazteca: function name could just be `update_timestamp()`

yadda/yadda.sql
```diff
@@ -0,0 +1,134 @@
...
+END;
+$$;
+
+CREATE TABLE yadda.breweries (
+  id SERIAL PRIMARY KEY,
+  name TEXT NOT NULL,
+  address TEXT NOT NULL,
+  description TEXT NOT NULL,
+  founding_year INTEGER CONSTRAINT valid_founding_year CHECK (founding_year >= 725) NOT NULL,
+  insert_user_id INTEGER NOT NULL,
```
incazteca: Rename this field to `created_by` and the update to `updated_by`. It's easier to understand and used in many other tables.

yadda/yadda.sql
drosholt: I was a little less concise because I wanted to make it clear that the field that I was updating was the "update_timestamp" (oops typo, it's update, not udpated)

yadda/yadda.sql
```diff
@@ -0,0 +1,134 @@
...
+  update_user_id INTEGER NOT NULL,
+  insert_timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
```
incazteca: Rename this field to `created_at` and `updated_at`. These are field names that are commonly used for these situations. Also they're rails defaults.

yadda/yadda.sql
```diff
@@ -0,0 +1,134 @@
...
+  RETURN NEW;
+END;
+$$;
+
+CREATE TABLE yadda.breweries (
+  id SERIAL PRIMARY KEY,
+  name TEXT NOT NULL,
+  address TEXT NOT NULL,
+  description TEXT NOT NULL,
+  founding_year INTEGER CONSTRAINT valid_founding_year CHECK (founding_year >= 725) NOT NULL,
```
incazteca: Should probably have the validation of the founding year be done in the application and not in the database.

yadda/yadda.sql
```diff
@@ -0,0 +1,134 @@
...
+  insert_user_id INTEGER NOT NULL,
+  update_user_id INTEGER NOT NULL,
+  insert_timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
+  update_timestamp TIMESTAMP NOT NULL
+);
+GRANT SELECT ON yadda.breweries TO yadda_ro;
+GRANT ALL PRIVILEGES ON yadda.breweries TO yadda_rw;
+REVOKE TRUNCATE ON yadda.breweries FROM yadda_rw;
+
+COMMENT ON CONSTRAINT valid_founding_year ON yadda.breweries IS 'Brewery must not be older than Weihenstephan, the oldest brewery.  As an aside, they have an excellent Hefeweissbier.';
```
incazteca: Let the application deal with this.

yadda/yadda.sql
drosholt: That's fair.  We should be clear that you're stating ruby standards.  Other systems use other, but similar names (these names are some that I've used in the past, because it avoided certain kinds of name conflicts).

yadda/yadda.sql
```diff
@@ -0,0 +1,134 @@
...
+
+CREATE TRIGGER test_table_update_timestamp
+  BEFORE UPDATE OR INSERT ON yadda.breweries
+  FOR EACH ROW EXECUTE PROCEDURE yadda.set_updated_timestamp();
+GRANT USAGE ON yadda.breweries_id_seq TO yadda_rw;
+
+CREATE TABLE yadda.brews (
+  id SERIAL PRIMARY KEY,
+  brewery_id INTEGER NOT NULL REFERENCES yadda.breweries,
+  style TEXT NOT NULL,
```
incazteca: Look into normalizing the styles.

yadda/yadda.sql
```diff
@@ -0,0 +1,134 @@
...
+  description TEXT NOT NULL,
+  year INTEGER NOT NULL CONSTRAINT valid_brewing_year CHECK (year >= -7000),
```
incazteca: If you're checking against negative years you probably should have to include a table to specify what calendar era the year belongs to.

yadda/yadda.sql
```diff
@@ -0,0 +1,134 @@
...
+GRANT USAGE ON yadda.brews_id_seq TO yadda_rw;
+
+
+CREATE TABLE yadda.users (
+  id SERIAL PRIMARY KEY,
+  first_name TEXT NOT NULL,
+  last_name TEXT NOT NULL,
+  address TEXT NOT NULL,
+  email_address TEXT NOT NULL,
+  birth_date DATE NOT NULL CONSTRAINT valid_birth_date CHECK (birth_date >= '1899-07-06') CONSTRAINT can_drink CHECK (birth_date <= now() - interval '21 years'),
```
incazteca: We'll allow Susannah to become a member but not 18 year olds who are of legal drinking age, at least outside of the US. Also though we can do these constraints in the application.

yadda/yadda.sql
```diff
@@ -0,0 +1,134 @@
...
+  BEFORE UPDATE OR INSERT ON yadda.brews
+  FOR EACH ROW EXECUTE PROCEDURE yadda.set_updated_timestamp();
+GRANT USAGE ON yadda.brews_id_seq TO yadda_rw;
+
+
+CREATE TABLE yadda.users (
+  id SERIAL PRIMARY KEY,
+  first_name TEXT NOT NULL,
+  last_name TEXT NOT NULL,
+  address TEXT NOT NULL,
```
incazteca: Could probably normalize this field in future iterations if we want to allow a user to have more than one address or address type. However, this is not needed now.

yadda/yadda.sql
drosholt: Of course.  All of these constraints belong in the application.  But that wouldn't have been any fun.

yadda/yadda.sql
```diff
@@ -0,0 +1,134 @@
...
+  insert_user_id INTEGER NOT NULL,
+  update_user_id INTEGER NOT NULL,
+  insert_timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
+  update_timestamp TIMESTAMP NOT NULL
+);
+GRANT SELECT ON yadda.users TO yadda_ro;
+GRANT ALL PRIVILEGES ON yadda.users TO yadda_rw;
+REVOKE TRUNCATE ON yadda.users FROM yadda_rw;
+
+COMMENT ON COLUMN yadda.users.address IS 'User''s full address, including their state, zip code, country, etc - whatever is applicable for their region.';
```
incazteca: `COMMENT` is not a command in SQL standard so just in the future keep in mind that if you want to comment your SQL files to just use `--`.

yadda/yadda.sql
```diff
@@ -0,0 +1,134 @@
...
+
+CREATE TRIGGER test_table_update_timestamp
+  BEFORE UPDATE OR INSERT ON yadda.users
+  FOR EACH ROW EXECUTE PROCEDURE yadda.set_updated_timestamp();
+GRANT USAGE ON yadda.users_id_seq TO yadda_rw;
+
+
+CREATE TABLE yadda.ratings (
+  id SERIAL PRIMARY KEY,
+  user_id INTEGER NOT NULL REFERENCES yadda.users (id),
```
incazteca: The `(id)` portion is not necessary since it'll assume that you're referencing based on the primary key.

yadda/yadda.sql
drosholt: But that has entirely different semantics.  Using -- is a comment in the file and has no impact on the database.  COMMENT actually records something in the database.  I definitely had fun with the comments, but I only put them in because the requirements were specifically for a PostgreSQL database.

incazteca:
@drosholt Please clean up the PR so that only the relevant files are in the diff. Also please update the field names for who created and updated the record as well as when it was created or updated. Also a lot of those validations such as those for valid years and drinking age can be handled by the application. The validations for the ratings though are fine so those can stay.

I would like to see this resubmitted with a version that addresses the issues commented on. You'll also need to set cascades so that there is proper handling of deletion of records.

paulghaddad:
@drosholt Please take a look at Franco's comments and submit a second iteration.

PR #1072, opened 2015-11-25
init solution

Dwin357:
Hey jmm,

Here is an initial baseline solution to the problem.  I still need to fiddle around with it to implement syntactic matching (weight == weight_in_pounds; carnivore true for insectivore and piscivore; et cetera) and probably clean up the display a bit (also go over the read me again to make sure I'm not missing anything user requests).

Please let me know if you want me to push this exercise in any particular direction,
Thanks!
-dwin


jmmastey:
Mostly mechanical notes added inline :)

dino_catalog/dasController.rb
```diff
@@ -0,0 +1,45 @@
+
```
jmmastey: Filenames in ruby should be underscore cased. `class DasController` => `das_controller.rb`.

jmmastey: This is because autoloading in frameworks like Rails does a bunch of standard conversions, including caps-to-underscores.

dino_catalog/dasController.rb
```diff
@@ -0,0 +1,45 @@
...
+
```
jmmastey: Don't leave blank lines at the beginning of files.

dino_catalog/dasController.rb
```diff
@@ -0,0 +1,45 @@
...
+require_relative "dasLogic"
+require_relative "dasView"
+require_relative "dasTable"
+
+class DasController
```
jmmastey: This classname is somewhat ambiguous. I keep reading this as german ("the controller", though wouldn't that be der controller? what is the gender of a controller?), but I'd expect this to either just be called `Controller`, or even better, something more particular.

dino_catalog/dasController.rb
```diff
@@ -0,0 +1,45 @@
...
+	def initialize
+		@table = DasTable.new
```
jmmastey: Use spaces, not tabs. Your editor should be configurable for this.

dino_catalog/dasController.rb
```diff
@@ -0,0 +1,45 @@
...
+		@view = DasView.new
+	end
+
+	def run
+		@view.greeting_display
+		command = ["help"]
```
jmmastey: ?

Oh, this is the part you're still working on. Make sure the grabbing of user input is in its own method.

dino_catalog/dasController.rb
```diff
@@ -0,0 +1,45 @@
...
+		end
+	end
+
+
+
+
+end
+
+defacto = DasController.new
+defacto.run
```
jmmastey: Good choice of separating the runner from the library itself.

dino_catalog/dasLogic.rb
```diff
@@ -0,0 +1,16 @@
+module DasLogic
+
+	def select(list, column_name, value)
+		list.select{ |dino| dino[column_name] == value}
+	end
+
+	def search(args)
```
jmmastey: This is probably best as a `.map`, `.select`, or `.all?`. Some structure that doesn't involve holding extra local variables (`list` in this case).

dino_catalog/dasLogic.rb
```diff
@@ -0,0 +1,16 @@
...
+		list = @table
+		args.each do |arg|
+			search = arg.split(":")
```
jmmastey: This logic (that search parameters are `field:value` is probably only necessarily relevant in the runner (above). That means that it probably doesn't belong in the non-console-driven code, such as here. By the time we call search, we probably already should have converted into `key => value` form. Also, not sure how this will work with `.big?` type queries.

dino_catalog/dasLogic.rb
```diff
@@ -0,0 +1,16 @@
...
+			list = select(list, search[0], search[1])
+		end
+		return list
```
jmmastey: Explicit return is not necessary here, skip it. Even better, if you can convert the entire body of this method into `@table.select do`, it will even easier to read:

```
def search(args)
  @table.select do |record|
    # match all filters
  end
end
```

dino_catalog/dasLogic.rb
```diff
@@ -0,0 +1,16 @@
+module DasLogic
+
+	def select(list, column_name, value)
+		list.select{ |dino| dino[column_name] == value}
+	end
+
+	def search(args)
+		list = @table
```
jmmastey: I'm conflicted about having the `@table` ivar here. Generally ivars become problematic quickly (so for instance you couldn't use this module in any other class that doesn't implement `@table`. If there were an obvious way to get rid of it I'd suggest that.

dino_catalog/dasScanner.rb
```diff
@@ -0,0 +1,32 @@
+require 'csv'
```
jmmastey: Again, good that this class is just loading, and the previous class is just searching. Good composition.

dino_catalog/dasScanner.rb
```diff
@@ -0,0 +1,32 @@
...
+module DasScanner
```
jmmastey: Same comments on naming as above.

dino_catalog/dasScanner.rb
```diff
@@ -0,0 +1,32 @@
...
+
+	def set_table
```
jmmastey: This method does modify data, and so I think a verb-ish name makes sense. But I'm not sure `set_table` captures it accurately. If I read that name without much context, I would definitely not guess that it was loading several CSVs. What name more accurately captures the work that happens in here?

dino_catalog/dasScanner.rb
```diff
@@ -0,0 +1,32 @@
...
+		Dir['./dino_fact_csvs/*'].each do |csv_file_path|
+    	i = 0
+    	headers = ""
```
jmmastey: CSV includes an option for `include_headers`, use that instead. The `i = 0` etc stuff is not wrong, or expensive, but it muddies up what this method is actually doing -- loading records.

dino_catalog/dasScanner.rb
```diff
@@ -0,0 +1,32 @@
...
+	    	if i == 0
+	    		headers = line
+	    		self.update_headers(line)
+	    	else
+	    		self.add_line(line, headers)
+	    	end
+	    	i += 1
+	    end
+		end
+	end
```
jmmastey: Note here that the CSV headers are not just different in capitalization. There are several places where the data is incompatible. Handling and standardizing that data is probably the only big logic change that's necessary right now.

dino_catalog/dasLogic.rb
```diff
@@ -0,0 +1,16 @@
+module DasLogic
+
+	def select(list, column_name, value)
+		list.select{ |dino| dino[column_name] == value}
+	end
```
jmmastey: You may eventually want to move towards a model for the dinosaurs themselves, if they start to gain more behavior than just values.

dino_catalog/dasScanner.rb
```diff
@@ -0,0 +1,32 @@
...
+
+	def update_headers(new_headers)
+		@headers = new_headers.map(&:upcase) | @headers
```
jmmastey: I'm not sure what you're using unary-or for here. To grab the super-set of all headers you receive? You should probably be controlling what headers exist.

dino_catalog/dasScanner.rb
```diff
@@ -0,0 +1,32 @@
...
+	end
+
+	def add_line(line, headers)
+		dino = {}
+		headers.each_with_index do |header, i|
+			dino[header] = line[i]
+		end
```
jmmastey: The CSV `include_headers` flag will subsume all this code. Yay!

dino_catalog/dasScanner.rb
```diff
@@ -0,0 +1,32 @@
...
+		@table << dino
```
jmmastey: This is the third file between which the `@table` ivar is split so far. If you changed the name of that var, you'd have to change at least three files so far. That's dangerous amounts of coupling.

jmmastey: For a better approach, have the `load_table` method _return_ an array of dinosaurs, and have the controlling class assign that into variables as it sees fit.

dino_catalog/dasTable.rb
```diff
@@ -0,0 +1,19 @@
+require_relative "dasScanner"
```
jmmastey: Not sure why this is required but not `logic`

dino_catalog/dasView.rb
```diff
@@ -0,0 +1,30 @@
+class DasView
+
+	def header_display(headers)
+		headers.each do |header|
+			puts header
+		end
+	end
```
jmmastey: Probably just `puts headers.join("\n")`

dino_catalog/dasView.rb
```diff
@@ -0,0 +1,30 @@
...
+
+	def dinoFullDisplay(dinos)
+		dinos.each do |dino|
+			dino.each do |k, v|
+				print "#{k}: #{v} "
+			end
+			puts
+		end
+	end
```
jmmastey: This may be hard to read (and include lots of empty values). If you do end up with a dino class, add a `to_s` to that class that returns the string representation of that dino.

dino_catalog/dasView.rb
```diff
@@ -0,0 +1,30 @@
...
+
+	def options_display
+		puts "-h or help: access this help menue"
+		puts "-e or exit: exit the program"
+		puts "-l or list: view a list of all dinos"
+		puts "-c or categories: view a list of all search categories"
+		puts "-s or search category:<term> category:<term>... : view a filtered list of dinos"
+	end
```
jmmastey: This seems to duplicate logic that lives in the driver itself. It probably belongs in that file.

dino_catalog/dasView.rb
```diff
@@ -0,0 +1,30 @@
...
+
+	def greeting_display
+		puts "Welcome to Di-Facto, your presumed source for prehistoric points of interest"
+	end
```
jmmastey: <3 this title :D

Dwin357:
hey joe,
funny how the more you do the less your imagination is able to prop up your ego with honey-talk of what "might be".  this is definitely a stronger composition, but there are a lot of near replications which I can't seem to factor out; particularly in the class methods of the dinosaur obj.

please let me know what you think; feedback from last time resulted in quite a jump and I am curious to see what happens next.
-dwin

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,65 @@
+require 'csv'
+require 'json'
+require_relative "dino_saur"
+
+class DinoCatalog
+  attr_reader :dino_collection
```
jmmastey: Tend to prefer names that don't include the datatype of the thing. So, `dinoaurs` > `dino_collection`.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,65 @@
...
+                  converters: :numeric, 
+                  headers: true,
+                  :header_converters => lambda {|h| h.downcase}
+                  ) do |row|
+        args = DinoSaur.construct_arguments(row)
+        rtn << DinoSaur.new(args)
+      end
+    end
+    rtn
+  end
```
jmmastey: Generally, if you see a method that initializes a variable, then does a loop, then returns that variable, it can be refactored into `inject`, `map`, or `each_with_object`. After considering for a moment, I think the right one here is `each_with_object`:

```
def build_collection(files = Dir['./dino_fact_csvs/*'])
  files.each_with_object([]) do |file, dinos|
    CSV.foreach(file, CSV_HEADERS) do |row|
      dinos << Dinosaur.from_row(row)
    end
  end
end

File: dino_catalog/dino_catalog.rb [hunk 27]
```diff
@@ -0,0 +1,65 @@
...
+
+  def big_dinosaurs(collection = dino_collection)
+    collection.select(&:is_big?)
+  end
```
jmmastey: Generally speaking, if you see a bunch of methods with a common prefix or suffix, it's an indication that you need to extract an object. In this case, I think Dinodex _is_ that object, so instead you can probably just drop the suffix. Good use of .select, btw.

As a nitpicky point, it'd be nice if I could chain these (`.carnivorous.bipeds`), which would require that you return another dinodex instance.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+    end
+    collection
+  end
+
+  ####  class methods  ####
+
+  def self.collection_to_json(collection)
+    collection.map {|d| DinoSaur.hash_dino(d) }.to_json
+  end
```
jmmastey: Change the `Dinosaur#hash_dino` method so that instead Dinosaur defines `.to_h`. It's more idiomatic, and you can change this entire method to:

```
def self.to_json(collection = dinosaurs)
  collection.to_json
end

File: dino_catalog/dino_catalog.rb [hunk 29]
```diff
@@ -0,0 +1,65 @@
...
+        collection = collection.select{ |d| d.has_attribute_value?(katagory, value)}
+      elsif DinoSaur::RECOGNIZED_CSV_COLUMN_HEADERS.include?(katagory)
+        attribute = DinoSaur.look_up_attribute_by_csv_header(katagory)
+        collection = collection.select{ |d| d.has_attribute_value?(attribute, value)}
+      else
+        collection = collection.select{ |d| d.has_additional_info_value?(katagory, value)}
+      end
+    end
+    collection
+  end
```
jmmastey: This method is more complex than I'd like to see. Overall, I think what's happening here is that a lot of this logic belongs in Dinosaur. I know this because six lines in this method call stuff on the Dinosaur model (three of them by classname explicitly).

If we pulled the search matching of a single criterion into Dinosaur, the method would look like this:

```
def search(terms)
  collections.select do |dinosaur|
    terms.all? { |field, value| dinosaur.match(field, value) }
  end
end
```

I think this makes sense because this method then just controls the logic of matching the entire collection. I'm tempted to wonder whether even the `.all?` stuff belongs in the Dinosaur class, but I think it would make way less difference.

jmmastey:
Overall comment on this last round: explicit `self.` is fine. Some folks don't like it, I think it actually removes some bugs. If there were a counter-argument, it would be that it makes the Extract Method pattern harder, and I do love that pattern lots.

dino_catalog/dino_command_line_interface.rb
```diff
@@ -0,0 +1,93 @@
...
+          end
+          if filter_criteria.include?("-b") || filter_criteria.include?("big")
+            collection = dino_catalog.big_dinosaurs(collection)
+          end
+          if filter_criteria.include?("-2") || filter_criteria.include?("biped")
+            collection = dino_catalog.biped_dinosaurs(collection)
+          end
+        end
+        display_collection(collection)
+
```
jmmastey: This chunk is big. Pull it into a method. Ditto the one below.

dino_catalog/dino_command_line_interface.rb
```diff
@@ -0,0 +1,93 @@
...
+      when "-s", "search"
+        search_terms = parse_search_terms(command[1..-1])
+        display_collection(
+          dino_catalog.search(search_terms)
+        )
+      end
+      print "Dino Request: "
+      command = gets.chomp.split(" ")
+    end
+  end
```
jmmastey: There might be some library to do these command line args, but this is at least very clear, and that counts for a lot. If you wanted to break it down further (since it's a big method), the way to do it would be to write a configuration block at the top:
```
COMMANDS = [
  { flag: "2", label: "biped", filter: :biped_dinosaurs },
  ...

]
```

And then your game loop basically just looks up the relevant line, applies the filter, and then displays. I don't think you need to here, though, just food for thought.

dino_catalog/dino_command_line_interface.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  def parse_search_terms(search_terms)
+    rtn = {}
+    search_terms.each do |term|
+      search_pair = term.split(":")
+      rtn[search_pair[0].downcase] = search_pair[1].downcase
+    end
+    rtn
+  end
```
jmmastey: See above about initialize-loop-return. See if you can break this one into another `.each_with_object`.

dino_catalog/dino_command_line_interface.rb
```diff
@@ -0,0 +1,93 @@
...
+    puts "-h or help: access this help menue"
+    puts "-e or exit: exit the program"
+    puts "-l or list: view a list of all dinos"
+    puts "-b or big: view a list of dino >= 4000lb"
+    puts "-c or carnivorus: view a list of all non-plant eating dinosaurs"
+    puts "-2 or biped: view a list of all bipedal dinosaurs"
+    puts "-p or period <period>: view a list of dinosaurs from the <period> period"
+    puts "-f or filter <big/biped/carnivorus or their abbreviations>: view a listed filtered by all included criteria"
+    puts "-s or search category:<term> category:<term>... : view a list of dinos filtered by search criteria"
+  end
```
jmmastey: In the world where you had a configuration block, this would probably roll into it. Seems fine as-is. The counter-argument would be that if you changed any arg, you'd probably forget to update it here.

dino_catalog/dino_command_line_interface.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  def greeting_display
+    puts "Welcome to Di-Facto, your presumed source for prehistoric points of interest"
+  end
+
+  private
+  attr_reader :dino_catalog
```
jmmastey: I'm not sure this has an effect. Have you tried it? Anyway, it might be okay to have the catalog readable from the outside.

dino_catalog/dino_saur.rb
```diff
@@ -0,0 +1,143 @@
...
+    self.period && self.period.downcase.include?(period.downcase)
+  end
+
+  def has_attribute_value?(dino_attribute, value)
+    self.send(dino_attribute).downcase == value.downcase
+  end
+
+  def has_additional_info_value?(additional_info_catagory, additional_info_value)
+    self.additional_info && self.additional_info.has_key?(additional_info_catagory) && self.additional_info[additional_info_catagory].include?(additional_info_value)
+  end
```
jmmastey: I find this method a bit confusing. Probably worth refactoring.

dino_catalog/dino_saur.rb
```diff
@@ -0,0 +1,143 @@
...
+          if v
+            rtn << "#{k}: #{v}"
+          end
+        end
+      else
+        rtn << "#{attribute.to_s[1..-1]}: #{self.instance_variable_get(attribute)}"
+      end
+    end
+    rtn.compact.join(", ") + "\n\n"
+  end
```
jmmastey: Define `.to_h` and then just use `.to_h.map` here.

dino_catalog/dino_saur.rb
```diff
@@ -0,0 +1,143 @@
...
+    when "weight", "weight_in_lbs"
+      "weight"
+    when "carnivore", "diet"
+      "diet"
+    when "walking"
+      "locomotion"
+    when "period"
+      "period"
+    end
+  end
```
jmmastey: So remember at the top where I said that a repeated prefix often meant that there's an object to be extracted? This is that case. The Dinosaur's logic should be concerned in this case with storing attributes and matching them. The logic for standardizing incoming data belongs elsewhere. By these method names, in some sort of parser, though I'm not sure that's the right term.

One way to think of it is this: If I give you another CSV file, or a JSON one, there's no good reason why the logic of the `Dinosaur` class should change. On the other hand, a Parser class would make sense to change in this situation.

dino_catalog/dino_saur.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  def self.parse_name(csv_table_row)
+    possible_headers = ["name", "genus"]
+    DinoSaur.general_parser(possible_headers, csv_table_row).downcase
```
jmmastey: Since these methods are static, go back to using self rather than the class name.

Also, since several methods look like this, I would expect to see something closer to:

```
self.extract_value(["name", "genus"]).downcase
```

Though, I think it might be possible to do:

```
(row["name"] || row["genus"] || "no data").downcase
```

and remove extra indirection from these methods.

dino_catalog/dino_saur.rb
```diff
@@ -0,0 +1,143 @@
...
+  def self.parse_additional_info(csv_table_row)
+    unrecognized_fields = csv_table_row.headers - RECOGNIZED_CSV_COLUMN_HEADERS
+    if !unrecognized_fields.empty?
+      additional_info = {}
+      unrecognized_fields.each do |header|
+        additional_info[header] = csv_table_row[header].downcase if csv_table_row[header] 
+      end  
+      additional_info
+    end
+  end
```
jmmastey: Ohhh, now I get the unrecognized thing. This is pretty clever :)

If you're wrapping most of the method in an if, use a short-circuit return instead:

`return {} if unrecognized_fields.empty?`

Otherwise, you have another initialize-loop-return sequence that's asking for refactoring.

dino_catalog/dino_saur.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  def self.hash_dino(dino_instance)
+    dino_hash = {}
+    dino_instance.instance_variables.each do |attribute|
+      next if !dino_instance.instance_variable_get(attribute)
+      dino_hash[attribute.to_s[1..-1]] = dino_instance.instance_variable_get(attribute)
+    end
+    dino_hash
+  end
```
jmmastey: See above, the idiomatic way to do this in ruby is an instance method `.to_h`

jmmastey:
To your note on `is_foo?` versus `foo?`, I think in this case they read a little better. Rubocop wants the other form because when you do testing you can use magical matchers if they're formatted the other way (`expect(dinosaur).to be_carnivore`).

jmmastey:
Closing this one since it has tons of closed comments on it. Feel free to re-open.

PR #1073, opened 2015-11-25
Dinosaur Catalog - Iteration 2

cstavitsky:
New look, same great taste!


dino_catalog/lib/dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,77 @@
...
+      :name,
+      :description,
+      :walking,
+      :diet,
+      :period,
+      :weight_in_lbs,
+      :continent,
+      :size,
+    ]
+    attr_reader *(ATTRIBUTES)
```
djkotowski: This means prefer
```ruby
attr_reader(*ATTRIBUTES)
```
rather than
```ruby
attr_reader *ATTRIBUTES
```
since it's easy to confuse the splat operator with the method `#*`

dino_catalog/README.md
```diff
@@ -1,28 +1,82 @@
...
+small = dinodex.filter(attribute: "size", value: "small")
+small.filter(attribute: "diet", value: "carnivore").dinosaurs
+#=> collection of Dinosaurs that weigh less than 2000 lbs AND are also carnivores.
+```
+
+To print data about a collection of dinosaurs (whether all the 'saurs in your Dinodex, or just some that have been filtered):
+```
+quadrupeds = dinodex.filter(attribute: "walking",value: "quadruped").dinosaurs
+
+dinodex.print_collection(quadrupeds) #prints a formatted list of quadruped Dinosaurs.
```
jmtai: +1 for the excellent documentation. However, this line looks like it was written before your refactor and is no longer accurate. Make sure you review your documentation to ensure it's still accurate after you change the code.

dino_catalog/dino_catalog.gemspec
```diff
@@ -0,0 +1,21 @@
+Gem::Specification.new do |s|
+  s.name = 'dino_catalog'
+  s.version = '0.0.0'
```
jmtai: This should be set to something non-zero. If you were searching RubyGems for a dino catalog gem to use, would you trust something with a version number of `0.0.0`?

dino_catalog/lib/dino_catalog/dino_importer.rb
```diff
@@ -0,0 +1,21 @@
+require_relative '../dino_catalog'
```
jmtai: Put a blank line between your `require` statements and the following code.

dino_catalog/lib/dino_catalog/dino_importer.rb
```diff
@@ -0,0 +1,21 @@
...
+module DinoCatalog
+  class DinoImporter
+    class << self
```
jmtai: The Ruby style guide (https://github.com/styleguide/ruby) recommends avoiding `class << self`. Instead, use the form
```ruby
def self.import_from_csv
end
```

dino_catalog/lib/dino_catalog/dino_printer.rb
```diff
@@ -0,0 +1,14 @@
+module DinoCatalog
+  class DinoPrinter
+    def self.print_collection(dinosaurs)
+      dinosaurs.each do |dinosaur|
+        puts dinosaur
+        puts ""
```
jmtai: You could avoid having to add the additional `puts` statement by simply adding an explicit newline to the output:
```ruby
puts dinosaur + "\n\n"
```

dino_catalog/lib/dino_catalog/dino_printer.rb
```diff
@@ -0,0 +1,14 @@
...
+      end
+    end
+
+    def self.print_dinosaur(dinosaur)
+      puts dinosaur
```
jmtai: This seems of limited value. If it were doing something other than just calling straight `puts`, it might have a reason to exist. What if you included the printing of the extra newline in this method? You could then call `print_dinosaur` in your `print_collection` method and make it even more readable:
```ruby
def self.print_collection(dinosaurs)
  dinosaurs.each { |dino| print_dinosaur(dino) }
end

def self.print_dinosaur(dinosaur)
  puts dinosaur + "\n\n"
end
```

dino_catalog/lib/dino_catalog/dino_serializer.rb
```diff
@@ -0,0 +1,16 @@
+module DinoCatalog
+  class DinoSerializer
+    def self.serialize_collection(dinosaurs)
+      dinosaurs.map do |dinosaur|
+        binding.pry
```
jmtai: Hound is definitely right.

dino_catalog/lib/dino_catalog/dino_importer.rb
```diff
@@ -0,0 +1,21 @@
...
+        parse_csv(file: file, formatter: formatter)
+      end
+
+      private
+
+      def parse_csv(file:, formatter: DinoCatalog::DinodexFormatter)
+        dinosaur_list = []
+        CSV.foreach(file, headers: true) do |row|
+          dinosaur = DinoCatalog::Dinosaur.new(formatter.convert(row))
+          dinosaur_list << dinosaur
```
jmtai: Since you're not actually using the temporary variable `dinosaur` for anything, you could just eliminate it:
```ruby
dinosaur _list << DinoCatalog::Dinosaur.new(formatter.convert(row))
```

dino_catalog/lib/dino_catalog/dino_serializer.rb
```diff
@@ -0,0 +1,16 @@
...
+        DinoSerializer.serialize(dinosaur.to_h)
+      end
+    end
+
+    private
+
+    def self.serialize(dino_hash)
```
jmtai: In general, avoid naming variables and parameters with specific data types. It can lead to confusion if the code changes in the future and some other data type gets passed in.

dino_catalog/lib/dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    def initialize(
+      name: nil,
```
jmtai: Keep your spacing consistent.

dino_catalog/lib/dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,77 @@
...
+      description:   nil)
+
+      @name	          = name
+      @period	        = period
+      @continent = continent
+      @diet = diet
+      @weight_in_lbs	= weight_in_lbs
+      @size = calculated_size(weight_in_lbs.to_i)
+      @walking	      = walking
+      @description	  = description
```
jmtai: Woah, what's up with the formatting in this section?

dino_catalog/lib/dino_catalog/formatters/dinodex_formatter.rb
```diff
@@ -0,0 +1,16 @@
+module DinoCatalog
+  class DinodexFormatter
+    def self.convert(row_data)
+      {
+        name:             row_data["NAME"],
+        period:           row_data["PERIOD"],
+        continent:        row_data["CONTINENT"],
+        diet:             row_data["DIET"],
+        weight_in_lbs:    row_data["WEIGHT_IN_LBS"].to_i,
+        size:             row_data["SIZE"],
```
jmtai: Does the `dinodex.csv` file have a `SIZE` column?

dino_catalog/lib/dino_catalog/formatters/pirate_bay_formatter.rb
```diff
@@ -0,0 +1,14 @@
+module DinoCatalog
+  class PirateBayFormatter
+    def self.convert(row_data)
+      {
+        name:             row_data["Genus"],
+        period:           row_data["Period"],
+        continent:        "Africa",
+        weight_in_lbs:    row_data["Weight"],
+        walking:          row_data["Walking"],
+        diet:             "Carnivore" == "Yes" ? "Carnivore" : "Herbivore",
```
jmtai: Hmmm... are you sure this is doing what you think? Does this
```ruby
"Carnivore" == "Yes"
```
ever return anything besides `false`?

dino_catalog/lib/dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,28 @@
...
+      @dinosaurs = dinosaurs
+    end
+
+    def carnivores
+      self.class.new(dinosaurs.select(&:carnivore?))
+    end
+
+    def bipeds
+      self.class.new(dinosaurs.select(&:biped?))
+    end
```
jmtai: Is there any reason you only include convenience functions for `carnivores` and `bipeds`? Why not `big` and `small` as well, since those are also specifically called out in the list of requirements?

dino_catalog/lib/dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    def filter(attribute:, value:)
+      dinos = dinosaurs.select do |dinosaur|
+        # raise "Invalid attribute type. Use a String." unless attribute.is_a?(String)
```
jmtai: Delete commented code.

dino_catalog/lib/dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,28 @@
...
+        val_of_attribute = dinosaur.send(attribute).to_s
+        val_of_attribute.downcase == value.downcase unless val_of_attribute.nil?
```
jmtai: `nil.to_s` returns an empty string, so you could also do this:
```ruby
val_of_attribute.to_s.downcase == value.downcase
```

dino_catalog/lib/dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,77 @@
...
+        send(attribute)
+      end
+
+      attributes_with_values.map do |attribute|
+        "#{attribute}: #{send(attribute)}"
+      end.join("\n")
+    end
+
+    def calculated_size(weight_in_lbs)
+      if weight_in_lbs >= 2000
```
jmtai: Hmmm... how much is a ton in pounds?

dino_catalog/lib/dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,77 @@
...
+      walking:       nil,
+      size:          nil,
+      description:   nil)
+
+      @name	          = name
+      @period	        = period
+      @continent = continent
+      @diet = diet
+      @weight_in_lbs	= weight_in_lbs
+      @size = calculated_size(weight_in_lbs.to_i)
```
jmtai: One of these things is not like the others. Not a big deal but perhaps it would make sense to create predicate methods `big?` and `small?` instead? That way all your instance variables correspond directly to the data read from the files and you'd maintain internal consistency in your implementation.

dino_catalog/lib/dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,77 @@
...
+      @walking	      = walking
+      @description	  = description
+    end
+
+    def carnivore?
```
jmtai: :+1: Nice use of predicate methods.

dino_catalog/lib/dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,77 @@
...
+    def to_s
+      formatted_facts
+    end
+
+    def to_h
+      dino_hash = {}
+      instance_variables.each do |variable|
+        dino_hash[variable] = instance_variable_get(variable)
+      end
+      dino_hash
```
jmtai: Whenever you see the pattern:
```
variable = some empty object
some_collection.each { do stuff, updating variable }
return variable
```
You can compact it using `each_with_object`: http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-each_with_object

dino_catalog/lib/dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    def formatted_facts
+      attributes_with_values = ATTRIBUTES.select do |attribute|
+        send(attribute)
+      end
+
+      attributes_with_values.map do |attribute|
+        "#{attribute}: #{send(attribute)}"
+      end.join("\n")
+    end
```
jmtai: Here you end up repeating yourself a bit, since you call `send_attribute` for each attribute to get your `attributes_with_values` and then you call it again when creating the output string. How about something like this instead:
```ruby
ATTRIBUTES.map do |attribute|
  value = send(attribute)
  "#{attribute}: #{value}" if value
end.compact.join("\n")
```

dino_catalog/lib/dino_catalog/formatters/pirate_bay_formatter.rb
```diff
@@ -0,0 +1,14 @@
+module DinoCatalog
+  class PirateBayFormatter
+    def self.convert(row_data)
+      {
+        name:             row_data["Genus"],
+        period:           row_data["Period"],
+        continent:        "Africa",
+        weight_in_lbs:    row_data["Weight"],
```
jmtai: For the `dinodex.csv` data, you call `to_i` on the weight but here you are not. This is probably not what you want, as now the converted data is not consistent.

dino_catalog/lib/dino_catalog/formatters/dinodex_formatter.rb
```diff
@@ -0,0 +1,16 @@
+module DinoCatalog
+  class DinodexFormatter
+    def self.convert(row_data)
+      {
+        name:             row_data["NAME"],
+        period:           row_data["PERIOD"],
+        continent:        row_data["CONTINENT"],
+        diet:             row_data["DIET"],
+        weight_in_lbs:    row_data["WEIGHT_IN_LBS"].to_i,
```
jmtai: Are you sure this is what you want? `nil.to_s` returns 0, so ff a dinosaur entry does not have a weight in the source data, do you want to display 0 pounds as the weight? This is a valid design choice but semantically it doesn't make much sense to have a dinosaur that weighs 0 pounds.

dino_catalog/spec/unit/dino_importer_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    @importer = DinoCatalog::DinoImporter.new('lib/dinodex.csv')
+  end
+
+  describe '#import_from_csv' do
+    it 'returns a collection of Dinosaur objects' do
+      @importer.import_from_csv('lib/dinodex.csv')
+      expect(@importer.dinosaur_list).to all(be_a DinoCatalog::Dinosaur)
+    end
+
+    it 'changes data format for pirate bay format'
```
jmtai: Are you still working on these tests? Empty `it` blocks don't do much for you. :smile:

dino_catalog/spec/unit/dinodex_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+        expect(dinosaur.diet.downcase).to satisfy { "carnivore" || "insectivore" || "piscivore" }
+      end
+    end
+  end
+
+  describe '#bipeds' do
+    it 'returns only biped Dinosaurs' do
+      bipeds = @dinodex.bipeds.dinosaurs
+      bipeds.each do |dinosaur|
+        expect(dinosaur.walking.downcase).to eq("biped")
```
jmtai: Since you've got a predicate method in the dinosaur class, you can use Rspec's predicate matchers (https://www.relishapp.com/rspec/rspec-expectations/v/3-1/docs/built-in-matchers/predicate-matchers) to make this a bit cleaner:
```
expect(dinosaur).to be_biped
```

dino_catalog/lib/dino_catalog/dinosaur.rb
djkotowski: Make sure your text editor is configured to replace tabs with 2 spaces.  This happens when you combine tabs and spaces.

dino_catalog/lib/dino_catalog/dinosaur.rb
djkotowski: :+1:

jmtai:
Nice work! The code is nice and clean and you've done a good job of following the single responsibility principle. From a design standpoint, my only question is why did you choose to implement all your classes using class methods? From a usage standpoint, this makes actually using the gem a bit awkward. For example, to serialize the a collection to JSON, you currently do this:

```ruby
small_dinos = dinodex.filter(attribute: "size", value: "big").dinosaurs
DinoCatalog::DinoSerializer.serialize_collection(small_dinos)
```
but wouldn't it be nicer to do something like this?:
```ruby
json = dinodex.filter(attribute: "size", value: "big").dinosaurs.serialize
```

One of the goals of this exercise is to build an interactive command line interface, so for your next pass you can actually use your gem to build the interface and see if you still think your design works well. Feel free to come grab me if you want to talk about anything here.

paulghaddad:
@cstavitsky Nice work on this iteration. Please take a look at Josh's comments and submit a last iteration.

PR #1074, opened 2015-11-30
Dinodex Implementation Part 2

jrade-enova:


dino_catalog/csv_utility.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def read(path)
+    raise "You need to specify a path and file" unless valid_path?(path)
+    csv_options = { encoding: "UTF-8", skip_blanks: true, headers: true }
+    process_csv_rows(path, csv_options)
+  end
+
+  private
+
+  def valid_path?(path)
```
schepedw: You call this method `valid_path`, but this will break if you pass a [`Pathname`](http://ruby-doc.org/stdlib-1.9.3/libdoc/pathname/rdoc/Pathname.html) object.

How can you extend this to accept both strings and Pathnames? Can you think of other data types that might be convenient to accept?

dino_catalog/csv_utility.rb
```diff
@@ -0,0 +1,47 @@
...
+      data_objects << process_row(row)
+    end
+
+    data_objects
+  end
+
+  def process_row(row)
+    data = {}
+
+    row.headers.each do |header|
```
schepedw: The pattern of 
1. Initialize some data structure, call it `foo`
2. Populate `foo`

Is encapsulated in the [`inject`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-inject) method

dino_catalog/dinodex.thor
```diff
@@ -0,0 +1,40 @@
...
+require './dinodex_app'
+require './csv_utility'
+require './filter_utility'
+require './merge_utility'
+
+class Dinodex < Thor
+  desc "read_file FILE", "This task will read a file (You can search " \
+    "based on given criteria. Also, provide an optional 'columns' option " \
+    "'file_system' option or 'to_json' for JSON format)"
+  method_option :criteria, :type => :hash, :required => false
```
schepedw: nitpicking: prefer `{foo: bar}` to `{:foo => bar}`

dino_catalog/dinodex.thor
```diff
@@ -0,0 +1,40 @@
...
+  method_option :columns, :type => :array, :required => false
+  method_option :to_json, :type => :boolean, :required => false
```
schepedw: nitpicking: put a space in between calls to class method calls / dsl calls & instance method definitions

dino_catalog/dinodex.thor
```diff
@@ -0,0 +1,40 @@
...
+  def read_file(file)
+    search_options = options[:criteria]
+    columns = options[:columns]
+    to_json = options[:to_json] || false
+    app = DinodexApp.new(CsvUtility.new, FilterUtility.new, MergeUtility.new)
+    data = app.read(file)
+    data = app.filter(data, search_options) if search_options
+    data = app.filter_columns(data, columns) if columns
+    output = to_json ? JSON.generate(data) : app.write(data)
+    puts output
```
schepedw: If I were to call a method called `read_file`, I wouldn't expect a bunch of console output and a nil return. I'd probably expect a string representation of the file, that I can later print to the console if I so choose

dino_catalog/dinodex.thor
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  desc "merge_files FILE1 FILE2", "This task will merge two data files. (You " \
+    " can also provide 'to_json' option for JSON format)"
+  method_option :mapping, :type => :hash, :required => true
+  method_option :to_json, :type => :boolean, :required => false
+  def merge_files(file1, file2)
+    hash_object = options[:mapping]
+    to_json = options[:to_json] || false
```
schepedw: Unless you're explicitly testing against `false`, you can leave nil as the value.
```
[12] pry(main)> nil ? 'nil is truthy' : 'nil is falsey'
=> "nil is falsey"
```

dino_catalog/dinodex.thor
```diff
@@ -0,0 +1,40 @@
...
+    data = app.filter_columns(data, columns) if columns
+    output = to_json ? JSON.generate(data) : app.write(data)
+    puts output
+  end
+
+  desc "merge_files FILE1 FILE2", "This task will merge two data files. (You " \
+    " can also provide 'to_json' option for JSON format)"
+  method_option :mapping, :type => :hash, :required => true
+  method_option :to_json, :type => :boolean, :required => false
+  def merge_files(file1, file2)
```
schepedw: Same point as above: `merge_files`, from an outsider perspective, doesn't seem like it should return nil.
For public methods, always think about the external interface you're providing for your object

dino_catalog/csv_utility.rb
```diff
@@ -0,0 +1,47 @@
...
+    CSV.generate do |csv_data|
+      csv_data << data_objects.first.keys
+      data_objects.each do |data_object|
+        csv_data << data_object.values
+      end
+    end
+  end
+
+  def read(path)
+    raise "You need to specify a path and file" unless valid_path?(path)
```
schepedw: :+1: for error checking

dino_catalog/dinodex_app.rb
```diff
@@ -0,0 +1,31 @@
+class DinodexApp
+  attr_accessor :io
```
schepedw: You can inline these, i.e. `attr_accessor :io, :filter, :merger`

Just an FYI. I'm not sure there's any convention here.

dino_catalog/dinodex_app.rb
```diff
@@ -0,0 +1,31 @@
...
+  attr_accessor :filter
+  attr_accessor :merger
+
+  def initialize(io, filter, merger)
```
schepedw: afaict, none of these objects has a state when they are passed. There's no need to initialize them, then pass them in. Just initialize them here.

In a related train of thought, what's the difference between a stateless class and a module? Which do you think you need?

dino_catalog/filter_utility.rb
```diff
@@ -0,0 +1,83 @@
+class FilterUtility
+  def filter_data(data, search_options)
+    search_criteria = build_search_params(search_options)
+    data.delete_if do |data_object|
```
schepedw: `delete_if` both modifies the data structure, which is pretty weird. You're using the return value, [here](https://github.com/jmmastey/level_up_exercises/pull/1074/files#diff-50e82a25fd60441685ef4475ddcf6af3R21), so you could use `reject`. 

In your case, the outcome is the same since you're doing `data = filter_data(data)`. `reject` might be more handy if you were going to separate the data, i.e. `filtered_data` = filter_data(data)`

dino_catalog/filter_utility.rb
```diff
@@ -0,0 +1,83 @@
...
+      !data_matches?(data_object, search_criteria)
+    end
+  end
+
+  def data_matches?(data_object, search_criteria)
+    data_object.each do |obj_key, obj_value|
+      return true if matches_search?(obj_key, obj_value, search_criteria)
+    end
+    false
```
schepedw: :+1: for explicit `false` rather than `nil` from a predicate function

dino_catalog/filter_utility.rb
```diff
@@ -0,0 +1,83 @@
+class FilterUtility
+  def filter_data(data, search_options)
+    search_criteria = build_search_params(search_options)
+    data.delete_if do |data_object|
+      !data_matches?(data_object, search_criteria)
+    end
+  end
+
+  def data_matches?(data_object, search_criteria)
+    data_object.each do |obj_key, obj_value|
```
schepedw: Check out the [`any`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-any-3F) method

dino_catalog/filter_utility.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  def matches_search?(obj_key, obj_value, search_criteria)
+    criteria_includes?(obj_key, search_criteria) &&
+      !obj_value.nil? &&
+      criteria_match?(obj_key, obj_value, search_criteria)
+  end
+
+  def build_search_params(search_options)
+    search_criteria = []
```
schepedw: `inject`

dino_catalog/merge_utility.rb
```diff
@@ -0,0 +1,27 @@
...
+  def merge_data(data1, data2, name_mapping)
+    common_columns = data1.first.keys & data2.first.keys
+    current_data = data1 + data2
+    generate_data(current_data, common_columns, name_mapping)
+  end
+
+  private
+
+  def generate_data(current_data, common_columns, name_mapping)
+    result_data = []
```
schepedw: `inject`

dino_catalog/merge_utility.rb
```diff
@@ -0,0 +1,27 @@
...
+    current_data.each do |data|
+      data_object = data.select { |key, _| common_columns.include?(key) }
+      data_object.merge!(create_hash_from_mapping(data, name_mapping))
+      result_data << data_object if data_object.length > 0
+    end
+    result_data
+  end
+
+  def create_hash_from_mapping(hash, column_mapping)
+    result_hash = {}
```
schepedw: `inject` some more :D

dino_catalog/dinodex_app.rb
jrade-enova: @schepedw  In my first iteration everything I wrote was in a module, but I didn't take advantage of dependency injection.  With this PR, I am trying to take a better OOD approach.  But to answer your question, I am not sure.  My guess is to use a class since I can then create objects from them and pass them along to other objects.  Modules are for Namespacing and cross-cutting concerns across an entire application.

jrade-enova:
@schepedw Thank you Dan for reviewing.  I will work on revisions first thing tomorrow morning and comment under your feedback accordingly.

dino_catalog/filter_utility.rb
jrade-enova: @schepedw I see the [difference](http://ruby-doc.org/core-2.2.0/Array.html#class-Array-label-Non-destructive+Selection) and can see when `reject` might be better.  I'll leave as `delete_if`for now.

dino_catalog/merge_utility.rb
jrade-enova: Loving `inject` now that I know about it :)

dino_catalog/dinodex.thor
jrade-enova: @schepedw I agree.  I was recommended to use Thor and started with [this example](https://github.com/erikhuda/thor/wiki/Getting-Started#a-simple-example) that just `puts` the output as you mentioned.  That is why I use pipe `|` and `tee` on the command line to redirect to a file.  This definitely turned more into a command line utility as opposed to something interactive for an everyday user.

jrade-enova:
@schepedw I have committed all suggested changes.  Anything not changed, I have left a reply stating reasons, but please let me know if there is anything I should address.  Thanks!

schepedw:
:+1:

paulghaddad:
@jrade-enova Nice work on this! You're good to move on!

PR #1075, opened 2015-12-02
Arrowhead exercise v2

mostlybadfly:
Hello! I wanted to submit this to get some feedback after the initial suggestions.  I can see how much cleaner the code looks now, any other suggestions would be appreciated :smile:

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -15,18 +15,22 @@ class Arrowhead
     },
   }
 
-  # FIXME: I don't have time to deal with this.
   def self.classify(region, shape)
-    if CLASSIFICATIONS.include? region
-      shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
+    shapes = CLASSIFICATIONS[region] unless check_region(region)
+    arrowhead = shapes[shape] unless check_shape(shapes, shape)
+    "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
+  end
+
+  private
+  def self.check_region(region)
+    unless CLASSIFICATIONS.include? region
+      raise "Uknown region, please provide a valid region."
+    end
+  end
+
+  def self.check_shape(shapes, shape)
+    unless shapes.include? shape
+      raise "Uknown shape value. Are you sure you know what you're talking about?"
     end
   end
 end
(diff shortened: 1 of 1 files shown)


arrowhead/arrowhead.rb
```diff
@@ -15,19 +15,20 @@ class Arrowhead
...
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+    shapes = check_region(region)
+    arrowhead = check_shape(shapes, shape)
+    "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
+  end
+
+  def self.check_region(region)
+    raise "Uknown region, please provide a valid region." unless CLASSIFICATIONS.include? region
```
ochagata: You can create custom errors that have the messages in them, that way you don't have to write the same string over and over again.

arrowhead/arrowhead.rb
```diff
@@ -15,19 +15,20 @@ class Arrowhead
...
+    CLASSIFICATIONS[region]
+  end
+
+  def self.check_shape(shapes, shape)
```
ochagata: This is a method that returns a boolean value, thus it should have a question mark at the end.

ochagata: Also, consider renaming to something more descriptive: `valid_shape?`

arrowhead/arrowhead.rb
```diff
@@ -15,19 +15,20 @@ class Arrowhead
...
-      end
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+    shapes = check_region(region)
+    arrowhead = check_shape(shapes, shape)
+    "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
+  end
+
+  def self.check_region(region)
```
ochagata: A bit confusing with the name of this method, it sounds like it should be checking, but returns a hash as an output. Consider renaming

arrowhead/arrowhead.rb
```diff
@@ -15,19 +15,20 @@ class Arrowhead
...
+    raise "Uknown region, please provide a valid region." unless CLASSIFICATIONS.include? region
+    CLASSIFICATIONS[region]
+  end
+
+  def self.check_shape(shapes, shape)
+    raise "Uknown shape value. Are you sure you know what you're talking about?" unless shapes.include? shape
```
ochagata: Same here for the error message

arrowhead/arrowhead.rb
ochagata: Consider refactoring this method so it returns a bool value

ochagata:
I left a few comments, otherwise it looks fine

mostlybadfly:
Thanks for the suggestions @ochagata :) . I moved the error messages to their own methods and realized I didn't need to return from them so the names changed.  Going to move on to the next exercise.

paulghaddad:
@mostlybadfly Nice work! As Ozzy said, you are free to move on.

PR #1078, opened 2015-12-03
Data science - Iteration 1

cstavitsky:
Haven't wrapped my head around what the actual data science means yet. That may be why I feel I didn't understand how to do real TDD for those functions. Definitely looking for specific feedback on how to better approach TDD for the behavior that uses the ABAnalyzer gem.


zsyed1:
@cstavitsky Hi Chris, Would you be able to clean up these hound errors? Most of them have to do with long-lines. It will make the review comments more readable. Sorry for the delay, I'll review this once you re-submit.

cstavitsky:
Sure! Sorry, didn't see this until just now. Will clean the hound errors up
today.

PR #1082, opened 2015-12-08
refactored for shorter methods. took out internal use of call and ano…

elialbert:
…n fns

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -8,21 +8,33 @@ class Robot
   def initialize(args = {})
     @@registry ||= []
     @name_generator = args[:name_generator]
+    @name = @name_generator ? @name_generator.call : generate_new_name
+    check_name
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
-
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
     @@registry << @name
+  end
+
+  def generate_char 
+    ('A'..'Z').to_a.sample
+  end
+
+  def generate_num
+    rand(10)
+  end
+
+  def generate_new_name
+      "#{generate_char}#{generate_char}#{generate_num}#{generate_num}#{generate_num}"
+  end
+
+  def check_name
+    if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
+      raise NameCollisionError, 'There was a problem generating the robot name!' 
+    end
   end
 end
 
+# generator = -> {'bb121'}
+# robot = Robot.new(name_generator: generator)
 robot = Robot.new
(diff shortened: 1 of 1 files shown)


robot_name/robot_name.rb
```diff
@@ -8,21 +8,33 @@ class Robot
   def initialize(args = {})
     @@registry ||= []
     @name_generator = args[:name_generator]
+    @name = @name_generator ? @name_generator.call : generate_new_name
```
paulghaddad: Perhaps extract this as a method?

robot_name/robot_name.rb
```diff
@@ -8,21 +8,33 @@ class Robot
   def initialize(args = {})
```
paulghaddad: As an overall comment, consider extracting the registry as a separate class.

robot_name/robot_name.rb
```diff
@@ -8,21 +8,33 @@ class Robot
...
     @@registry ||= []
```
paulghaddad: A `Set` instead of an array would ensure its values are unique.

robot_name/robot_name.rb
```diff
@@ -8,21 +8,33 @@ class Robot
...
+  def generate_num
+    rand(10)
+  end
+
+  def generate_new_name
+      "#{generate_char}#{generate_char}#{generate_num}#{generate_num}#{generate_num}"
+  end
+
+  def check_name
+    if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
```
paulghaddad: Consider creating two separate errors for each condition: invalid name and a non-unique name.

robot_name/robot_name.rb
```diff
@@ -8,21 +8,33 @@ class Robot
...
+      raise NameCollisionError, 'There was a problem generating the robot name!'
```
paulghaddad: To continue the comment above, consider extracting the error(s) into a separate class.

robot_name/robot_name.rb
```diff
@@ -8,21 +8,33 @@ class Robot
...
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+    @@registry << @name
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
+  def generate_char 
+    ('A'..'Z').to_a.sample
```
paulghaddad: `[*'A'..'Z'].sample` is nifty to generate the array and sample it.

robot_name/robot_name.rb
paulghaddad: Consider extracting this regex to a well-named constant.

robot_name/robot_name.rb
```diff
@@ -8,21 +8,33 @@ class Robot
...
-    @@registry << @name
+  def generate_num
+    rand(10)
+  end
+
+  def generate_new_name
+      "#{generate_char}#{generate_char}#{generate_num}#{generate_num}#{generate_num}"
+  end
+
+  def check_name
```
paulghaddad: Consider checking all of the method's of this class and asking whether they should be public or not. In this case, `check_name` is probably a internal, `private` method.

paulghaddad:
@elialbert Nice work on this. Take a look at my comments and submit a second iteration. I'd recommend extracting a class or two. Also, see if there is a better way to handle the class variable.

PR #1083, opened 2015-12-09
Dino catalog

elialbert:


dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,136 @@
...
+    end
+    self
+  end
+
+  def merge_data(dinodex, african)
+    merged = []
+    dinodex.each { |entry| merged << normalize_dinodex_entry(entry) }
+    african.each { |entry| merged << normalize_african_entry(entry) }
+    merged
+  end
```
sshao: this method can be reduced down to one line: 

* `dinodex.map { |entry| normalize_dinodex_entry(entry) } + african.map { |entry| normalize_african_entry(entry) }`
* or, if you can abstract a `normalize` method that accepts both "dinodex" and "african"-type entries: `(dinodex + african).map { |entry| normalize(entry) }` 
* or, if you convert your "entries" to Dinosaur objects and create a `Dinosaur#normalize` method: `(dinodex + african).map(&:normalize)`

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,136 @@
...
+      :name => entry[:genus],
+      :period => entry[:period],
+      :continent => nil,
+      :diet => nil,
+      :carnivore => (entry[:diet] == "Yes"),
+      :weight => entry[:weight],
+      :walking => entry[:walking],
+      :description => nil,
+    }
+  end
```
sshao: as touched upon in the previous comment: the two methods `normalize_african_entry` and `normalize_dinodex_entry` are both `normalize_{some object type}`, and as such might be better structured as a method *upon* that object.

so instead of the method call `normalize_dinodex_entry(entry)`, you could define a `DinodexEntry` object with a `normalize` method and use `entry.normalize` instead.

and taking that further: you'd then have two objects: a `DinodexEntry` and an `AfricanEntry`. those objects are representing two variations upon the same thing (an Entry, or in reality, a Dinosaur). you could have a `Dinosaur` object that covers both African- and Dinodex-type entries.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,136 @@
...
+
+  def get_carnivore_for_dinodex(diet)
+    %w(Carnivore Pescivore Insectivore).include? diet
+  end
+end
+
+module CSVHelper
+  CSV::Converters[:blank_to_nil] = lambda do |field|
+    field && field.empty? ? nil : field
+  end
```
sshao: or you could take those `normalize_*` methods and make them into lambdas, add them as CSV Converters, and do the "normalization" immediately, possibly skipping the need to create various flavors of a Dinosaur object at all

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,136 @@
...
+
+  def to_json
+    JSON.generate(results)
+  end
+
+  def print
+    JSON.pretty_generate(results)
+  end
+
+  def grab_bipeds
```
sshao: no need to include the "grab" verb in these method names, just make them #bipeds, #carnivores, etc. the call `catalog.carnivores` already reads as "get carnivores from the catalog", without the `grab_` prefix.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,136 @@
...
+  def start_filter
+    @filtered ||= @data.dup
+  end
+
+  def results
+    start_filter
+    results = @filtered.dup
+    @filtered = nil
+    results
+  end
```
sshao: instead of creating a copy of `@data` and modifying that copy, try creating and returning an entirely new DinoCatalog object instead. your method calls will still be chainable, while reducing the number of "catalog" representations you have (you currently have three: `@data`, `@filtered`, and `DinoCatalog#results`, which is a lot of things to keep track of)

for example, `DinoCatalog#select` would become:

```ruby
def select(&block)
  DinoCatalog.new(@data.select(&block))
end
```

this would also require rethinking and refactoring your `initialize` method, as it's currently hardcoded to only initialize from files.

sshao:
code's very clean, i like it. there's higher-level refactors and abstractions that you can make (as suggested in previous comments) and i'd definitely suggest taking at shot at those and see if it cleans up your code and reduces duplications further. :+1:

RedWizzrobe:
Be sure to separate each class and module into its own file, following Ruby naming conventions.

djkotowski:
Be sure to separate each class and module into its own file.

dino_catalog/dino_catalog.rb
djkotowski: I also have a slight concern with naming this `#select`, as another developer may expect it to have the same output type as `Enumerable#select`.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,136 @@
...
+  def length
+    results.length
+  end
+
+  def to_s
+    results.to_s
+  end
+
+  def to_json
+    JSON.generate(results)
```
djkotowski: You could also delegate this as Hound suggests with the others, as the JSON module adds `#to_json` to arrays.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  def print
+    JSON.pretty_generate(results)
+  end
+
+  def grab_bipeds
+    search(:walking => "Biped")
```
djkotowski: DinoCatalog has a lot of responsibility in this solution.  It's responsible for categorizing the dinosaurs, filtering them if needed, understanding where the CSV files are located and calling CSVHelper on them, and formatting the dinosaurs.

I recommend extracting some of this logic out into a Dinosaur model class.  Then that model is in charge of knowing itself (Am I big?  Am I a carnivore?), and DinoCatalog can refer to this logic when filtering.

djkotowski:
Good first submission.  Take a look at the comments, refactor and resubmit.  Let me know if you have any questions.

dino_catalog/dino_catalog.rb
elialbert: @sshao oh! duh! thanks for that. Knew there had to be a better way.

@djkotowski totally.

dino_catalog/dino_catalog.rb
elialbert: I did factor out the normalize logic into Dinosaur, but can you help me understand something - If I put the "knowing itself" logic into Dinosaur, and I want to filter out (for instance) "all bipeds" in the DinoCatalog, I would still need either:
- a method for all bipeds in the dino catalog that refers to the method for biped in the dinosaur class (seems redundant)
- a dynamic method lookup in the dino class (seems overcomplicated)
- no methods for all bipeds in the dino catalog, pass a block directly into DinoCatalog (not useful enough)

See my current reopened PR, curious if you have some advice about how to put the logic into Dinosaur without the above pitfalls.
thanks!

sshao: I think the first option (method for all bipeds in the dino catalog) is sufficient; in this case we want to "know about all bipeds". A single Dinosaur object only knows if it itself is a biped -- it shouldn't know if other Dinosaurs are bipeds, so there's not much of a way to put "all bipeds" logic into Dinosaur. A DinoCatalog, on the other hand, can know if multiple dinosaurs are bipeds.

I also don't think there's anything wrong with the third option (no specific #bipeds method). It's true that it's not as nice as `catalog.bipeds`, but on the other hand you're likely already expecting that users can query arbitrarily for some attributes (for example `catalog.search(:continent => 'Africa')`), so it's not a far stretch to let them query for bipeds (`catalog.search(:walking => 'bipeds')`) either.

PR #1084, opened 2015-12-09
Data science

elialbert:


data_science/data_science.rb
```diff
@@ -0,0 +1,65 @@
+require 'json'
+require 'abanalyzer'
+
+class Optimizer
+  attr_accessor :data
```
djkotowski: Add space between accessor and initialize.

data_science/data_science.rb
```diff
@@ -0,0 +1,65 @@
+require 'json'
+require 'abanalyzer'
+
+class Optimizer
```
djkotowski: Extract each class into its own file.

data_science/data_science.rb
```diff
@@ -0,0 +1,65 @@
...
+  attr_accessor :data
+  def initialize(initial_data)
+    @data = DataLoader.load_data initial_data
```
djkotowski: Wrap arguments in parentheses

data_science/data_science.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def simple_counts
+    result = {}
+    [:A, :B].each do |cohort|
+      result[cohort] = { :sample_size => @data[cohort].values.inject(:+),
+                        :conversions => @data[cohort][:successes] }
+    end
+    result
```
djkotowski:
```ruby
obj = obj_value
enumerable.each do |element|
  # do stuff with obj
end
obj
```
can be simplified to
```ruby
enumerable.each_with_object(obj_value) do |element, obj|
  # do stuff with obj
end
```

data_science/data_science.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def conversion_rates
+    counts = simple_counts
+    result = {}
+    [:A, :B].each do |cohort|
+      result[cohort] = ABAnalyzer.confidence_interval(
+        counts[cohort][:conversions], counts[cohort][:sample_size], 0.95)
+    end
+    result
```
djkotowski: see above comment about `each_with_object`

data_science/data_science.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def result_confidence
+    tester = ABAnalyzer::ABTest.new @data
+    tester.chisquare_p
+  end
+end
+
+class DataLoader
+  def self.load_data(filename_or_array)
+    return parse filename_or_array if filename_or_array.class == Array
```
djkotowski: This is an example of when wrapping args with parentheses really helps with readability.

data_science/data_science.rb
```diff
@@ -0,0 +1,65 @@
...
+    self.parse(
+      self.load_json(filename_or_array)
+    )
+  end
+
+  def self.load_json(json_data_filename)
+    f = File.open(json_data_filename, "r")
+    data = f.read
```
djkotowski: If this raises an error, the file will not close.  Either `f.close` should be in an `ensure` block, or (and more preferably) you should use the block syntax.

```ruby
data = File.open(json_data_filename, "r") do |f|
  f.read
end
```
which can be simplified to
```ruby
data = File.open(json_data_filename, "r", &:read)
```

data_science/data_science.rb
```diff
@@ -0,0 +1,65 @@
...
+    f.close
+    JSON.parse(data)
+  end
+
+  def self.parse(data)
+    result = Hash.new { |hash, key| hash[key] = { :successes => 0, :failures => 0 } }
```
djkotowski: Nice use of Hash's default value syntax.

data_science/data_science.rb
```diff
@@ -0,0 +1,65 @@
...
+    data.each do |entry|
+      result_key = self.key_or_sym(entry, "result") == 1 ? :successes : :failures
+      result[self.key_or_sym(entry, "cohort").to_sym][result_key] += 1
+    end
+    result
+  end
+
+  def self.key_or_sym(hash, value)
+    hash[value.to_sym] || hash[value.to_s]
```
djkotowski: This will get easier in Rails when you have access to `HashWithIndifferentAccess`

data_science/data_science.rb
elialbert: cool I knew there had to be a more ruby way to do that kind of thing

data_science/data_science_spec.rb
```diff
@@ -0,0 +1,85 @@
+require_relative 'data_science'
+
+fake_data = [
```
djkotowski: This should be loaded from a testing JSON file.

elialbert:
I just made a new commit fixing all mentioned changes so far

data_science/data_science_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  {"cohort": "B", "result": 1},
+  {"cohort": "A", "result": 0}, 
+  {"cohort": "A", "result": 0}, 
+  {"cohort": "B", "result": 0},
+]
+
+
+describe Optimizer do
+  before :each do
+    @optimizer = Optimizer.new fake_data
```
djkotowski: Avoid instance variables in tests.  Use `let` and `subject`s instead.

data_science/data_science_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  it "counts total sample size per cohort" do
+    expect(@optimizer.simple_counts[:A][:sample_size]).to eq(28)
+    expect(@optimizer.simple_counts[:B][:sample_size]).to eq(14)
+  end
+
+  it "counds number of conversions per cohort" do
+    expect(@optimizer.simple_counts[:A][:conversions]).to eq(12)
+  end
```
djkotowski: L67-74 should be within a `describe "#simple_counts"` block.

data_science/data_science_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  it "should calculate conversion rates" do
+    result = @optimizer.conversion_rates
+    expect(result[:A].class).to eq(Array)
+    expect(result[:A][0]).to be < result[:A][1]
+  end
+
+  it "should do chisquared calculations" do
+    expect(@optimizer.result_confidence).to be < 1
+  end
```
djkotowski: Put these two in their own `describe` statements.

data_science/data_science_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  {"cohort": "B", "result": 0},
+  {"cohort": "A", "result": 1}, 
+  {"cohort": "A", "result": 0}, 
+  {"cohort": "B", "result": 1},
+  {"cohort": "A", "result": 0}, 
+  {"cohort": "A", "result": 0}, 
+  {"cohort": "B", "result": 0},
+]
+
+
```
djkotowski: You should also have a spec for `DataLoader`.

data_science/data_science_spec.rb
```diff
@@ -0,0 +1,85 @@
+require_relative 'data_science'
+
```
djkotowski: This file should be `spec/optimizer_spec.rb`.

data_science/data_science.rb
```diff
@@ -0,0 +1,65 @@
...
+  def simple_counts
+    result = {}
+    [:A, :B].each do |cohort|
+      result[cohort] = { :sample_size => @data[cohort].values.inject(:+),
+                        :conversions => @data[cohort][:successes] }
+    end
+    result
+  end
+
+  def conversion_rates
```
djkotowski: Given you're mapping calculations to each cohort hash, it would be good to extract them out to a Cohort object.  Then, rather than having to eager-calculate the conversion rates and corresponding error bars, you can have methods on Cohort to do that for you.

djkotowski:
Nice first attempt.  Take a look at the comments, refactor and resubmit when ready.

Also, I recommend avoiding submitting new PRs until the outstanding one is closed.  Often, reviews are done throughout the day and suffer interruptions.

PR #1085, opened 2015-12-09
Overlord

elialbert:
went for the focus on capybara since that's where I needed a refresher, didn't bother prettying up the page; supervillain can hire a designer.


overlord/features/bomb.feature
```diff
@@ -0,0 +1,55 @@
+Feature: Bomb activation and deactivation
```
djkotowski: This should be two features.

overlord/features/bomb.feature
```diff
@@ -0,0 +1,55 @@
...
+    Then the bomb is now active
+    And the number of attempts remaining is 2
+
+  Scenario: bomb deactivated incorrectly 3 times
+    Given the bomb page is loaded
+    And I input the default activation code
+    And the bomb is now active
+    When I input an incorrect deactivation code
+    And I input an incorrect deactivation code
+    And I input an incorrect deactivation code
```
djkotowski: This could be simplified into one or two steps.

One option would be to combine the first two deactivations to a single `Given` step (e.g. `And I have entered an incorrect deactivation code 2 times`), having only one `When` to indicate the action triggering the explosion.  The other would be to combine all three into a single step (`When I input an incorrect activation code 3 times`).  With either of these options, you'd be able to take advantage of Cucumber's Regex parsing.

```ruby
When(/^I input an incorrect deactivation code (\d+) times$/) do |count|
  count.times do
    # input actions here
  end
end
```

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,83 @@
...
+Capybara.register_driver :selenium do |app|
+  Capybara::Selenium::Driver.new(app, :browser => :chrome)
+end
+
+Capybara.javascript_driver = :chrome
+Capybara.configure do |config|
+  config.run_server = false
+  config.app_host = "localhost:4567"
+end
+
```
djkotowski: Everything in L1-15 should be in a separate `features/support/env.rb` file.

overlord/overlord.rb
```diff
@@ -1,14 +1,64 @@
...
 
 require 'sinatra'
+require_relative 'lib/bomb'
 
 enable :sessions
 
-get '/' do
-  "Time to build an app around here. Start time: " + start_time
+get '/bomb/:activation_code/:deactivation_code' do
+  @bomb = get_bomb params
```
djkotowski: Wrap args/params in parentheses.

overlord/overlord.rb
```diff
@@ -1,14 +1,64 @@
 # run `ruby overlord.rb` to run a webserver for this app
 
 require 'sinatra'
+require_relative 'lib/bomb'
 
 enable :sessions
 
-get '/' do
```
djkotowski: It'd be good practice to have a root route.  Maybe just redirect to `/bomb`?

overlord/overlord.rb
```diff
@@ -1,14 +1,64 @@
...
+  @bomb.enter_code(params[:deactivation_code])
+  set_bomb @bomb
+  redirect to('/bomb')
+end
+
+post '/newbomb' do
+  session[:bomb] = nil
+  redirect to('/bomb')
+end
+
```
djkotowski: All methods below this line should be `private`.

overlord/overlord.rb
```diff
@@ -1,14 +1,64 @@
...
+def new_bomb params
+  raise "need a number for code" unless is_number?(params[:activation_code]) and 
+                                        is_number?(params[:deactivation_code])
+  Bomb.new(params[:activation_code] || "1234", 
+           params[:deactivation_code] || "0000")
```
djkotowski: This is a fun issue that pops up occasionally in Hound.  Hound expects the second line to be indented by two spaces, not to the same position as the first arg.  I like your formatting better.  Ignore this.

overlord/views/bomb.erb
```diff
@@ -0,0 +1,25 @@
+They set us up the bomb.
```
djkotowski: We get signal.

overlord/views/bomb.erb
```diff
@@ -0,0 +1,25 @@
...
+<br/>
```
djkotowski: Not an issue pre-UI development, but avoid `<br />` tags.  CSS should handle spacing functionality.

Again, don't worry about fixing this now.  It's just something to be aware of.

overlord/views/bomb.erb
```diff
@@ -0,0 +1,25 @@
...
+  <form action="/deactivate" method="post">
+    <label>Deactivation Code</label>
+    <input id="deactivation_input" type="text" name="deactivation_code"/>
+    <input type="submit" value="submit"/>
+  </form>
+<% end %>
+
+<form action="/newbomb" method="post">
+  <input id="newbomb" type="submit" value="New Bomb"/>
+</form>
```
djkotowski: Since this isn't being rendered with a layout, this should be a full HTML file, complete with `<html>`, `<head>`, and `<body>` tags.

elialbert: I think it is tho, isn't it getting yielded into application.html.erb or somesuch?

overlord/overlord.rb
```diff
@@ -1,14 +1,64 @@
...
+end
+
+def get_bomb(params)
+  puts "found session bomb #{session[:bomb]}"
+  session[:bomb] || new_bomb(params)
+end
+
+def is_number? string
+  return true if string.nil? || string == ''
+  true if Integer(string) rescue false
```
djkotowski: Should really be:
```ruby
begin
  Integer(string)
  true
rescue
  false
end
```

overlord/views/bomb.erb
elialbert: for sure

overlord/overlord.rb
```diff
@@ -1,14 +1,64 @@
...
+  session[:bomb] = bomb
+end
+
+def get_bomb(params)
+  puts "found session bomb #{session[:bomb]}"
+  session[:bomb] || new_bomb(params)
+end
+
+def is_number? string
+  return true if string.nil? || string == ''
```
djkotowski: `''` and `nil` aren't numbers, so this method is a bit misnamed.  Maybe `valid_code?` would be a better name?

overlord/features/bomb.feature
```diff
@@ -0,0 +1,55 @@
...
+
+  Scenario: bomb deactivated incorrectly
+    Given the bomb page is loaded
+    And I input the default activation code
+    And the bomb is now active
+    When I input an incorrect deactivation code
+    Then the bomb is now active
+    And the number of attempts remaining is 2
+
+  Scenario: bomb deactivated incorrectly 3 times
```
djkotowski: Exploding a bomb should probably be its own feature (on the border with this, since it does involve attempted deactivation).

overlord/features/bomb.feature
```diff
@@ -0,0 +1,55 @@
...
+  Scenario: bomb activated with default code
+    Given the bomb page is loaded
+    When I input the default activation code
+    Then the bomb is active
+
+  Scenario: bomb activated with specified code
+    Given the bomb page is loaded with specified
+    When I input the specified activation code
+    Then the bomb is active
+
```
djkotowski: You should also consider bad paths.  What if the user enters "fail" or "012" (too few digits) for the code?

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,83 @@
...
+  expect(page).to have_content("inactive")
+  click_button "newbomb"
+end
+
+When(/^I input the default activation code$/) do
+  fill_in "activation_input", :with => "1234"
+  click_button 'submit'
+end
+
+And(/^the bomb is now active$/) do
```
djkotowski: Don't use `And`.  Use `Given`, `When`, or `Then`.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,83 @@
...
+  Capybara::Selenium::Driver.new(app, :browser => :chrome)
+end
+
+Capybara.javascript_driver = :chrome
+Capybara.configure do |config|
+  config.run_server = false
+  config.app_host = "localhost:4567"
+end
+
+Given(/^the bomb page is loaded$/) do
```
djkotowski: For readability, it's helpful to sort your steps by their conjunction, usually Givens first, Whens second, and Thens last.

overlord/features/step_definitions/bomb_steps.rb
djkotowski: Also, it looks like this step is used as a `Given`.  Only `Then`s should have expectations.  `Given` and `When` steps should just be actions.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,83 @@
...
+  fill_in "activation_input", :with => "1234"
+  click_button 'submit'
+end
+
+And(/^the bomb is now active$/) do
+  expect(page).to(have_content("active")) && expect(page).not_to(have_content("inactive"))
+end
+
+Then(/^the bomb is active$/) do
+  expect(page).to(have_content("active")) && expect(page).not_to(have_content("inactive"))
```
djkotowski: Don't use `&&` between expectations.  Just have each one on a separate line.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,83 @@
...
+  config.app_host = "localhost:4567"
+end
+
+Given(/^the bomb page is loaded$/) do
+  visit('/bomb')
+end
+
+Then(/^it is not activated$/) do
+  expect(page).to have_content("inactive")
+  click_button "newbomb"
```
djkotowski: Actions should not be in `Then` steps.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,83 @@
...
+  expect(page).to have_content("error")
+end
+
+When(/^I input the default deactivation code$/) do
+  fill_in "deactivation_input", :with => "0000"
+  click_button 'submit'
+end
+
+When(/^I input the specified deactivation code$/) do
+  fill_in "deactivation_input", :with => "1111"
```
djkotowski: Since you're referencing the same sets of digits throughout the feature, it may be useful to extract them to constants (e.g. `DEFAULT_ACTIVATION_CODE`, `INCORRECT_DEACTIVATION_CODE`, etc.)

djkotowski:
This is looking good.  Take a look at the comments, refactor and resubmit.  I imagine just one refactor should do it.

elialbert:
thanks, will do

overlord/views/bomb.erb
elialbert: oh duh not rails

overlord/features/bomb.feature
elialbert: I already had that at the bottom of the file I think

PR #1087, opened 2015-12-10
Robot name exercise

mostlybadfly:
I moved a lot of what was going in in `initialize` into separate methods, I think it makes it easier to read. The registry didn't seem to be something that belonged to the robot, so I moved it up out of the method, but perhaps it should be its own class? The exercise mentioned needing tests so I attempted some specs that I checked when I made changes.

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -9,7 +9,7 @@ class Robot
     @name_generator = args[:name_generator]
     @name = make_name
 
-    check_name  
+    check_name
   end
 
   def check_name
@@ -28,7 +28,7 @@ class Robot
     if @name_generator
       @name = @name_generator.call
     else
-     @name = generate_name
+      @name = generate_name
     end
   end
 
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -17,19 +17,18 @@ class Robot
     @@registry << @name
   end
 
-  def generate_char
-    ('A'..'Z').to_a.sample
-  end
-
-  def generate_num
-    rand(10)
+  def generate_name
+    output = ""
+    2.times { output << ('A'..'Z').to_a.sample }
+    3.times { output << rand(10).to_s }
+    output
   end
 
   def make_name
     if @name_generator
       @name = @name_generator.call
     else
-      @name = "#{generate_char}#{generate_char}#{generate_num}#{generate_num}#{generate_num}"
+     @name = generate_name
     end
   end
 
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -2,24 +2,39 @@ class NameCollisionError < RuntimeError; end
 
 class Robot
   attr_accessor :name
-
-  @@registry
+  VALID_NAME = /[[:alpha:]]{2}[[:digit:]]{3}/
+  @@registry ||= []
 
   def initialize(args = {})
-    @@registry ||= []
     @name_generator = args[:name_generator]
+    @name = make_name
 
+    check_name  
+  end
+
+  def check_name
+    raise NameCollisionError, 'There was a problem generating the robot name!' if invalid_name?
+    @@registry << @name
+  end
+
+  def generate_char
+    ('A'..'Z').to_a.sample
+  end
+
+  def generate_num
+    rand(10)
+  end
+
+  def make_name
     if @name_generator
       @name = @name_generator.call
     else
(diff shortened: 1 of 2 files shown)


robot_name/robot_name.rb
```diff
@@ -2,24 +2,38 @@ class NameCollisionError < RuntimeError; end
...
     @name_generator = args[:name_generator]
+    @name = make_name
+
+    check_name
+  end
+
+  def check_name
+    raise NameCollisionError, 'There was a problem generating the robot name!' if invalid_name?
+    @@registry << @name
+  end
```
jmmastey: Consider that it would be much easier to test individual methods if `invalid_name?` and `check_name` both took arguments. They're both pretty simple at this second, but it's likely that they'd grow over time, and it'd be nice to be able to say:

`expect { robot.check_name("abc123") }.not_to raise_error`

robot_name/robot_name.rb
```diff
@@ -2,24 +2,38 @@ class NameCollisionError < RuntimeError; end
...
 
+  def generate_name
+    output = ""
+    2.times { output << ('A'..'Z').to_a.sample }
+    3.times { output << rand(10).to_s }
+    output
+  end
```
jmmastey: Nice and clean :D

robot_name/robot_name.rb
```diff
@@ -2,24 +2,38 @@ class NameCollisionError < RuntimeError; end
...
     if @name_generator
       @name = @name_generator.call
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+      @name = generate_name
     end
+  end
```
jmmastey: The only other thing I might change here is to coalesce the usage of the default name generator versus the arg one. So:

```
def initialize
  @name_generator = args.fetch(:name_generator, method(:generate_name))
end

def make_name
  @name = @name_generator.call
end
```

robot_name/robot_name.rb
```diff
@@ -2,24 +2,38 @@ class NameCollisionError < RuntimeError; end
...
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def invalid_name?
+    !(name =~ VALID_NAME) || @@registry.include?(name)
   end
```
jmmastey: Noice.

robot_name/robot_name_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rspec'
+require_relative 'robot_name'
+
+describe Robot do
+  it "should create a new Robot" do
+    expect(Robot.new).to be_a Robot
+  end
```
jmmastey: This may be a low value test. What are the odds that `Robot.new` returns anything other than a `Robot` instance? If you were TDDing this, you might generate this test and then delete it later. My personal process is to write only tests that fit two of these three criteria:
1. Is it easy to write and maintain this test (yes)
2. Is the functionality complicated or likely to break (not at all)
3. Are the consequences of the functionality breaking significant (probably not)

robot_name/robot_name_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rspec'
```
jmmastey: Yay tests!

robot_name/robot_name_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  it "should have a name of two characters and 3 numbers" do
+    expect(Robot.new.name).to match(/[[:alpha:]]{2}[[:digit:]]{3}/)
+  end
+
+  it "should raise an error if two names are the same" do
+    generator = -> { 'AA111' }
+    Robot.new(name_generator: generator)
+    expect { Robot.new(name_generator: generator) }.to raise_error(NameCollisionError) 
+  end
+end
```
jmmastey: Remember to test the generator itself (that it assigns the name based on the generator). Also, it's possible that you'd want to test the registry. I'm not entirely sure.

jmmastey:
Hey @mostlybadfly, good to hear from you! :D 

Solid code. I added a few nitpicks, but they're not big deals. Read, grok, and move on.

mostlybadfly:
Thanks @jmmastey , sounds good! :100: :sunglasses:

PR #1091, opened 2015-12-12
Overlord - Iteration 1

creativisionist:


overlord/features/bomb_activation.feature
```diff
@@ -0,0 +1,40 @@
...
+  In order to activate the bomb
+  As the evil Vault Boy
+  I want to input the correct activation code
+
+  Scenario: Set valid activation and deactivation codes
+    Given I am on the bomb page
+    When I set valid activation and deactivation codes
+    Then I should be directed to the inactive_bomb page
+
+  Scenario: Set invalid activation and deactivation codes
```
ochagata: When it comes to testing happy/sad path, try to test most of the happy path in one scenario(like the one before this one), and try to have an individual test for each sad path. 

The way the test is written now, it only tests the validity of the first code that the application processes

overlord/features/bomb_activation.feature
```diff
@@ -0,0 +1,40 @@
...
+    Given I am on the bomb page
+    When I set invalid activation or deactivation codes
+    Then I should be directed to the bomb page
+      And the page should contain Codes can only be numeric.
```
ochagata: Indent everything the same in Gherkin.

overlord/features/bomb_activation.feature
```diff
@@ -0,0 +1,40 @@
...
+    Given I am on the bomb page
+    When I set invalid duplicate activation or deactivation codes
+    Then I should be directed to the bomb page
+      And the page should contain Codes must be different.
+
+  Scenario: Set blank duplicate activation and deactivation codes
+    Given I am on the bomb page
+    When I set blank activation or deactivation codes
+    Then I should be directed to the bomb page
+      And the page should contain Codes can only be numeric.
```
ochagata: Wouldn't the expectation be that you either have a javascript check in place to make sure that the field is being populated. Or, have a page that says there must be an input.

The page that says the input should be numeric should come up only if the input contains non-numeric characters

overlord/features/bomb_activation.feature
```diff
@@ -0,0 +1,40 @@
+Feature: Bomb activation
+  In order to activate the bomb
+  As the evil Vault Boy
+  I want to input the correct activation code
+
+  Scenario: Set valid activation and deactivation codes
```
ochagata: You could move this and the next few scenarios(ones before activation) into a configuration.feature file, since they don't deal with activation, but with configuration instead.

overlord/features/bomb_deactivation.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Bomb deactivation
+  In order to deactivate the bomb
+  As the good Vault Boy
+  I want to input the correct deactivation code
+
+  Scenario: Set activation and deactivation codes
```
ochagata: Only test each scenario once, this was already tested in the `activation.feature`

overlord/features/step_definitions/error_page_steps.rb
```diff
@@ -0,0 +1,7 @@
+When(/^I go to a page that does not exist$/) do
+  visit "/asdf"
+end
+
+Then(/^I will get a (\d+) page$/) do |arg1|
```
ochagata: Agreed with hound. You can omit arguments you don't end up using.

overlord/features/step_definitions/index_steps.rb
```diff
@@ -0,0 +1,11 @@
+Given(/^I am on the welcome page$/) do
+  visit "/"
+end
+
+When(/^I enter Megaton$/) do
+  click_link "Enter Megaton"
+end
+
+Then(/^I should be directed to the bomb page$/) do
+  expect(current_path).to eq "/bomb"
```
ochagata: Paths can be the same for different pages depending on the state of the user looking at the page. Prefer using css selectors instead

overlord/features/step_definitions/overlord_steps.rb
```diff
@@ -0,0 +1,73 @@
+Given(/^I am on the bomb page$/) do
+  visit '/bomb'
+end
+
+When(/^I set valid activation and deactivation codes$/) do
+  ACTIVATION_CODE = 1234
```
ochagata: You can declare these at the top of the file. No need to do so inside a step definition

overlord/features/step_definitions/overlord_steps.rb
```diff
@@ -0,0 +1,73 @@
...
+
+Then(/^I should be directed to the active_bomb page$/) do
+  expect(current_path).to eq '/active_bomb'
+end
+
+When(/^I input the incorrect activation code$/) do
+  validate_code('activation_code', INCORRECT_CODE)
+end
+
+Then /the page should contain (.*)/ do |input|
```
ochagata: Hound suggesting this: `Then(/the page should contain (.*)/) do`

overlord/features/step_definitions/overlord_steps.rb
```diff
@@ -0,0 +1,73 @@
...
+end
+
+When(/^I input the incorrect deactivation code max times$/) do
+  validate_code('deactivation_code', INCORRECT_CODE)
+  validate_code('deactivation_code', INCORRECT_CODE)
+  validate_code('deactivation_code', INCORRECT_CODE)
+end
+
+Then(/^I should be directed to the explosion page$/) do
+  expect(current_path).to eq '/explosion'
```
ochagata: Again, prefer css selectors. Or the native selectors in capybara. It can find based on text

overlord/features/support/helpers.rb
```diff
@@ -0,0 +1,14 @@
+module Helpers
+  def initialize_bomb(code_1, code_2)
```
ochagata: Already discussed with you about using pageObject models. You can call this method at any time and the application will try to run the scripts no matter what page it's on. With pageObject models, the application will know that it should be on the page that the pageobject model is associated with before it runs the method.

overlord/overlord.rb
```diff
@@ -1,14 +1,111 @@
...
 # we can shove stuff into the session cookie YAY!
 def start_time
   session[:start_time] ||= (Time.now).to_s
 end
+
+def bomb
+  session[:bomb]
+end
+
+def create_bomb(activation_code, deactivation_code, max_failed_deactivations)
```
ochagata: consider using the variable name `max_attempts`

ochagata:
Good iteration. Review comments, refactor and resubmit :goat:

overlord/features/bomb_activation.feature
creativisionist: Can you explain in further detail what you mean by this?

overlord/features/step_definitions/overlord_steps.rb
creativisionist: If I'm using pageObject models shouldn't I expect to be directed to a certain page instead of looking for elements b/c those elements will be contained with the pageObject model?

overlord/features/bomb_activation.feature
ochagata: You want to test each sad path by itself so that one failure doesn't mask a subsequent one.

"The reason that individual test cases cover invalid cases is that certain
erroneous-input checks mask or supersede other erroneous-input checks.
For instance, if the specification states ‘‘enter book type (HARDCOVER,
SOFTCOVER, or LOOSE) and amount (1–999),’’ the test case, (XYZ 0),
expressing two error conditions (invalid book type and amount) will probably
not exercise the check for the amount, since the program may say
‘‘XYZ IS UNKNOWN BOOK TYPE’’ and not bother to examine the remainder
of the input."

From http://it-ebooks.info/book/2899/ -
p. 52

overlord/features/step_definitions/overlord_steps.rb
ochagata: If you decide to use site_prism, you can do something like 
```
@explosion_page = Explosion.new
expect(@explosion_page).to be_displayed
```
Given that you have an `Explosion` page object defined

PR #1093, opened 2015-12-14
Overlord Exercise - Iteration 1

jrade-enova:
web_steps.rb was generated from gem, therefore ignored rubocop on this file


magicalbanana:
Quick tip on the Gemfile, when you're committing just gems, no need to make a single commit on each. Just squash them together (look into git rebase) and you can summarize your title like 'Add gems for testing' and you can add a description (add a line underneath it) so it'll be like this:

```
Add gems for testing # no dots

Feature testing gems:
cucumber
watir ..etc
```

Here's a resource on writing git commit messages:
http://chris.beams.io/posts/git-commit/

overlord/features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,241 @@
+# Taken from the cucumber-rails project.
```
jmtai: @djkotowski will tell you not to use web_steps.rb, as it encourages [poor scenario writing](http://aslakhellesoy.com/post/11055981222/the-training-wheels-came-off).

overlord/features/manage_bomb.feature
```diff
@@ -0,0 +1,36 @@
+Feature: A Working Bomb
```
jmtai: With TDD, we test units of software (which usually correspond to a single class in Ruby) but with BDD we are testing the behavior of the software from a user perspective. That means our tests should be organized according to related behaviors. In this case, you've got all your tests grouped under a single `manage_bomb` feature but if we look at the behavior of the site, the user can do a number of different things:

Arm the bomb
Disarm the bomb
Explode the bomb
etc.

overlord/overlord.rb
```diff
@@ -1,14 +1,76 @@
...
 
 require 'sinatra'
+require 'haml'
+require 'json'
+require './lib/bomb'
+require './lib/timer'
 
 enable :sessions
 
+ERROR = { "error" => "Codes must be numeric" }
```
jmtai: `ERROR` is a bit generic - might want to name it something a bit more descriptive. Maybe `CODE_FORMAT_ERROR`?

overlord/overlord.rb
```diff
@@ -1,14 +1,76 @@
...
+
 get '/' do
-  "Time to build an app around here. Start time: " + start_time
+  haml :index
```
jmtai: From an operational standpoint, what happens to an activated bomb if I just reload the home page? Since `/boot` always creates a new bomb, the old one would just get discarded. Does this make sense?

overlord/overlord.rb
```diff
@@ -1,14 +1,76 @@
...
+end
+
+get '/boot', provides: :json do
+  content_type :json
+  bomb = initialize_bomb(params['activation_code'], params['deactivation_code'])
```
jmtai: Semantically, this is a little bit sketchy since your `initialize_bomb` method might actually return an error hash, and not a bomb object.

overlord/overlord.rb
```diff
@@ -1,14 +1,76 @@
...
+  bomb.to_json
+end
+
+get '/activate', provides: :json do
+  content_type :json
+  bomb = get_from_session(:bomb)
+  return bomb.to_json if bomb.active?
```
jmtai: Since you're always returning `bomb.to_json` in your routes, why not just call `.to_json` within the helper method itself?

overlord/overlord.rb
```diff
@@ -1,14 +1,76 @@
...
+get '/deactivate', provides: :json do
+  content_type :json
+  bomb = deactivate_bomb(params['code'])
+  bomb.to_json
+end
+
+get '/state', provides: :json do
+  content_type :json
+  bomb = get_from_session(:bomb)
+  bomb.state
```
jmtai: Are you using this for anything?

overlord/overlord.rb
```diff
@@ -1,14 +1,76 @@
...
+  bomb.to_json
+end
+
+def initialize_bomb(activation_code, deactivation_code)
+  return ERROR unless valid_codes(activation_code, deactivation_code)
```
jmtai: For consistency, you might want to use the predicate method naming convention here (i.e. `valid_codes?` instead of `valid_codes`)

overlord/overlord.rb
```diff
@@ -1,14 +1,76 @@
...
+  bomb = Bomb.new(activation_code, deactivation_code, 3)
```
jmtai: Definitely avoid sprinkling magic number into the code - assign this number to a descriptively named constant.

overlord/features/step_definitions/web_steps.rb
jeffrade: I believe this file was generated from the cucumber-sinatra gem that was suggested.

overlord/features/manage_bomb.feature
jeffrade: @jmtai Makes sense.  I will break these out in my next PR

overlord/overlord.rb
jeffrade: @jmtai Yes it does.  I'll look back at any requirements and test locally to make sure UX is reasonable

overlord/overlord.rb
jeffrade: I'll definitely refactor this to do so

overlord/overlord.rb
jeffrade: To get the time left.  Using Ajax polling on front-end to get the status.

overlord/overlord.rb
jeffrade: @jmtai  Will do

overlord/overlord.rb
jeffrade: @jmtai Will add as a constant at top of the class

jeffrade:
@jmtai Thanks for reviewing!  I will make changes and submit a second iteration tomorrow.

overlord/overlord.rb
```diff
@@ -1,14 +1,76 @@
...
+  save_to_session(:bomb, bomb)
+  bomb
+end
+
+def activate_bomb(bomb, code, time)
+  bomb = bomb.enter_code(code)
+  bomb.timer = Timer.new(time) if bomb.active?
```
jmtai: Codes are getting validated when the bomb is initialized but what about the time?

overlord/overlord.rb
```diff
@@ -1,14 +1,76 @@
...
+def deactivate_bomb(code)
+  bomb = get_from_session(:bomb)
+  bomb = bomb.enter_code(code)
+  bomb.timer = nil if bomb.inactive?
+  save_to_session(:bomb, bomb)
+  bomb
+end
+
+def valid_codes(activation_code, deactivation_code)
+  activation_code.match(/^\d+$/) && deactivation_code.match(/^\d+$/)
```
jmtai: Are you sure you want to allow arbitrary length codes? It would be pretty easy to guess a single digit activation/deactivation code.

overlord/public/main.js
```diff
@@ -0,0 +1,83 @@
+$(document).ready(function() {
+  $("#boot-bomb").click(function bootBomb(){
```
jmtai: Keep your whitespace consistent (e.g. put a space between `bootBomb()` and the curly brace).

Also, for readability, it might be a bit easier to read if you declare the functions outside of the click handler so you can just have three lines of initialization code inside `$(document).ready()`.

overlord/public/main.js
```diff
@@ -0,0 +1,83 @@
...
+    $("#error").html('');
```
jmtai: Since this is being used within each callback, you probably want to move this into a separate function so if you end up renaming the `#error` element you'll only have to change it in one place.

overlord/public/main.js
```diff
@@ -0,0 +1,83 @@
...
+    $.ajax({
+      url: "/boot",
+      data: {
+        activation_code : $("#activation-code").val(),
+        deactivation_code : $("#deactivation-code").val() },
+      dataType: "json"
+    }).done(function callback(response){
+      console.log('response=' + JSON.stringify(response));
```
jmtai: Remove debugging statements.

overlord/public/main.js
```diff
@@ -0,0 +1,83 @@
...
+      data: {
+        code : $("#code").val(),
+        time: $("#time").val() },
+      dataType: "json"
+    }).done(function callback(response){
+      console.log('response=' + JSON.stringify(response));
+      if(response.error != null) {
+        $("#error").html(response.error);
+      }
+      if(response.state == ':active') {
```
jmtai: What if you get an unexpected state here... what would you want to display to the user?

overlord/public/main.js
```diff
@@ -0,0 +1,83 @@
...
+};
+
+function addExplodeBackground() {
+  $body = $("body");
+  $body.empty();
+  $body.css("background-color", "black");
+  $body.css("background-image", "url('img/boom.gif')");
+  $body.css("background-repeat", "no-repeat");
+  $body.css("background-attachment", "fixed");
+  $body.css("background-position", "center");
```
jmtai: Not a big deal, but these can all be combined into a single `background` property setting.

overlord/overlord.rb
jmtai: Ah, sorry, that was unclear. The actual line of code `bomb.state` doesn't really do anything since you're returning bomb.to_json.

jmtai:
Looking good! Definitely work on the tests a bit by breaking the features up. This will allow you to add sad/error path tests to the appropriate feature. Feel free to swing by if you want clarification on any of the comments.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,69 @@
+When(/^I submit an activation and deactivation code$/) do
+  step 'I fill in "activation-code" with "4321"'
```
jmtai: These codes are being used in multiple steps, so define them as constants.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,69 @@
...
+  step 'I press "boot-bomb"'
+end
+
+Then(/^the bomb should be booted$/) do
+  step 'I should see "Time:"'
+  step 'the "time" field should contain ""'
+end
+
+Then(/^the bomb should be activated$/) do
+  sleep(2)
```
jmtai: In general, try to avoid using sleeps within tests - they can lead to sporadic behavior and unreliable tests are almost as bad as no tests at all.

overlord/overlord.rb
jeffrade: Looks like we do need this.  `bomb.state` will call methods `check_wires` and `check_timer` in `bomb.rb`.

jmtai: Ah... interesting. I'd argue that the method is poorly named, since it gives no indication that it produces side effects. But that's not your fault. :smile:

jeffrade:
@jmtai @paulghaddad Made all major changes/suggestions.  Submitted PR [here](https://github.com/enova/level_up_exercises/pull/6)

PR #1096, opened 2015-12-14
Cyoa congress

elialbert:
Notes:
* there will be hound errors. I satisfied rubocop on every file that I touched. I didn't want to go through and sanitize the worst of the autogenerated files... my apologies.
* This is a combination rails + angular app. So it runs as a SPA on the frontend. The cucumber tests are therefore all run in database truncation mode.
* run cucumber and rspec the normal ways (just `cucumber` and `rspec` suffice on my machine).
* includes full omniauth login support.


cyoa/goodbills/app/assets/javascripts/welcome.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
```
paulghaddad: Be sure to delete any unnecessary files.

cyoa/goodbills/app/assets/templates/bill.html.erb
```diff
@@ -0,0 +1,27 @@
+<div class="row">
+  <b>Title: </b>{{bill.official_title}}
```
paulghaddad: I don't believe this is valid HTML. The text should be in a header or paragraph element.

cyoa/goodbills/app/assets/templates/homepage.html.erb
```diff
@@ -0,0 +1,20 @@
+
+<ul ng-repeat="bill in bills">
+<li><i>{{bill.official_title}}</i></li>
```
paulghaddad: For your view templates, I recommend indenting your HTML elements for readability.

cyoa/goodbills/app/assets/templates/homepage.html.erb
```diff
@@ -0,0 +1,20 @@
...
+<li><b>Score:</b> {{bill.score}}</li>
+<li><a href="#/bills/{{bill.id}}">Bill Detail</a></li>
```
paulghaddad: You'll want to use link helpers instead of hardcoding links...but maybe that's not how Angular works. In Rails you'll want to use helpers.

cyoa/goodbills/app/controllers/api/v1/bills_controller.rb
```diff
@@ -0,0 +1,27 @@
+class Api::V1::BillsController < ApplicationController
+  def index
+    @bills = Bill.order("id ASC").paginate(
```
paulghaddad: An alternative for the parameter to `order` is: `order(id: :asc)`

cyoa/goodbills/app/controllers/api/v1/bills_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+      page: params[:page] || 1,
+      per_page: 10)
+    render json: @bills
+  end
+
+  def show
+    @bill = Bill.find(params[:id])
```
paulghaddad: You can extract this into a private method since it is used more than once in this controller.

cyoa/goodbills/app/controllers/api/v1/bills_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    Bill.find(params[:id]).tap do |bill|
+      bill.update_score(bill_update_params[:vote])
+      @bill = bill
+    end
+    render json: @bill
+  end
+
+  private
+
+  def bill_update_params
```
paulghaddad: The Rails convention for this method is `bill_params`.

cyoa/goodbills/app/controllers/welcome_controller.rb
```diff
@@ -0,0 +1,7 @@
+require 'bills_api'
+
+class WelcomeController < ApplicationController
+  def index
+    BillsAPI.repull_bills if (Date.today - Bill.latest_timestamp) > 6
```
paulghaddad: Consider extracting the conditional logic into it's own, well-named method. It is not clear to me the reason for the time element, and the number 6 is a magic number.

cyoa/cyoa-plan.txt
```diff
@@ -0,0 +1,25 @@
...
+High level business requirements:
+
+  Grab each bill that congress passes.
+  Store bill data in a local db.
+  Pull new bill data on a schedule.
+  Display last N bills on the homepage.
+  Allow users to log in with facebook and github.
+  Allow users to vote on whether or not a bill represents a good deed. (positive or negative feedback)
+  Allow users to tweet about a bill.
+  Republish bill data as a JSON feed with voting results included.
```
jmtai: These requirements don't seem to match what is actually implemented in the app. Is this still a work in progress?

cyoa/cyoa-plan.txt
```diff
@@ -0,0 +1,25 @@
...
+
+Test plan:
+  Attributes: Fast, Simple, Persistent
+  Components: Main bill page, individual bill view, voting system, API in, API out
+  Capabilities:
+    Main bill page: view and paginate bills, show voting results at a glance
+    Individual bill: view details about the bill, allow voting via click, 
+      show voting results details
+    Voting system: Allow voting once per user (TODO!!), allow vote change (?!),
```
jmtai: Voting is allowed even if a user is not logged in. Is this by design? If so, does this make sense?

cyoa/goodbills/Gemfile
```diff
@@ -0,0 +1,48 @@
+source 'https://rubygems.org'
+
+# Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
+gem 'rails', '4.2.5'
+# Use sqlite3 as the database for Active Record
+gem 'sqlite3'
```
jmtai: Postgres should be the default database for all Rails apps - we do not use sqlite for any real apps here.

cyoa/goodbills/README.rdoc
```diff
@@ -0,0 +1,28 @@
+== README
```
jmtai: At a minimum, this README should contain instructions on how to get your app up and running. I had to dig through the code to figure out the dependencies and get everything installed, which is not great.

cyoa/goodbills/app/assets/templates/homepage.html.erb
jmtai: Definitely agree with @paulghaddad here. These templates are difficult to read.

cyoa/goodbills/app/controllers/api/v1/bills_controller.rb
```diff
@@ -0,0 +1,27 @@
+class Api::V1::BillsController < ApplicationController
```
jmtai: :+1: for versioning the API.

cyoa/goodbills/app/controllers/welcome_controller.rb
jmtai: Agreed. Replace magic numbers with descriptively named constants, instead.

cyoa/goodbills/app/models/bill.rb
```diff
@@ -0,0 +1,23 @@
+class Bill < ActiveRecord::Base
+  def update_score(vote)
+    self.score += 1 if vote == "like"
+    self.score -= 1 if vote == "dislike"
+    self.num_voted += 1
+    self.save!
+  end
```
jmtai: So any user, authenticated or not, can vote an arbitrary number of times? Does that make sense?

cyoa/goodbills/app/models/bill.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def self.latest_timestamp
+    order('last_action_at DESC').first.last_action_at ||
```
jmtai: Did you test this with a clean install? This is what you'll see:
```
NoMethodError in WelcomeController#index
undefined method `last_action_at' for nil:NilClass
```
You can use `try` to avoid nil errors:
```
 order('last_action_at DESC').first.try(:last_action_at)
```

jmtai: Are you sure `last_action_at` is the value you want to use for the latest_timestamp calculation? If the most recent record has a `last_action_at` value that is greater than 6 days old, the app is going to repull all the bills on every time the home page is loaded. Perhaps `created_at` or `updated_at` would be more appropriate?

cyoa/goodbills/app/models/bill.rb
```diff
@@ -0,0 +1,23 @@
...
+      Date.civil(2000, 1, 1)
+  end
+
+  def self.update_or_create(attributes)
+    assign_or_new(attributes).save
```
jmtai: Hmmm... what's the point of the extra abstraction here?

cyoa/goodbills/app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,38 @@
...
+        <span class="navbar-text">Logged in as: <%= current_user.email %></span>
+        <span class="navbar-text"><%= link_to "Sign out", destroy_user_session_path %></span>
+      <% else %>
+        <span class="navbar-text">
+          <%= link_to "Sign in with Github", user_omniauth_authorize_path(:github) %>
+        </span>
+      <% end %>
+    </div>
+  </div>
+</nav>
```
jmtai: Keep your indentation consistent.

cyoa/goodbills/config/secrets.yml
```diff
@@ -0,0 +1,22 @@
...
+development:
+  secret_key_base: b1485a15cb03fad5e1d845ae20496cef599def38a9041b876d06fc87374ebb4380e6328abdeeff0188b49ec2846f8511ca877712bec2fa576ca44700678c7080
+
+test:
+  secret_key_base: 0d3d3b6c626af17200c0934619f98d9af0bf49e356f780189ac873ed58539c47387997a80453cb8eeff59de2b30240d897da0eda4aef52194111d2be4866b2b0
+
+# Do not keep production secrets in the repository,
+# instead read values from the environment.
+production:
+  secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
```
jmtai: :+1: for using `secrets.yml` and not hardcoding keys into the source. :smile:

cyoa/goodbills/app/assets/templates/bill.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  ng-click="vote('dislike')">
+  Dislike
+</button>
+<button type="button" 
+  id="like"
+  class="btn btn-default" 
+  ng-click="vote('like')">
+  Like
+</button>
+</div>
```
jmtai: You're getting a whole bunch of other information on each bill via the API, so why is this so limited? At a minimum, I'd expect a link to the text of the bill.

jmtai: Also, from a UX point of view, why isn't there a link or button to return me to the page I clicked the bill details from? Currently, I have to use the browser back button and then it puts me back on page 1.

cyoa/goodbills/app/assets/templates/homepage.html.erb
```diff
@@ -0,0 +1,20 @@
+
+<ul ng-repeat="bill in bills">
```
jmtai: Visually, the bulleted list of items is a bit hard on the eyes. All the bills are kind of smashed together without anything to separate them.

cyoa/goodbills/db/migrate/20151210184033_scoring_for_bills.rb
```diff
@@ -0,0 +1,8 @@
+class ScoringForBills < ActiveRecord::Migration
+  def change
+    add_column :bills, :score, :number
+    add_column :bills, :num_voted, :number
+    add_index :bills, :score
+    add_index :bills, :num_voted
```
jmtai: Do you really need these indexes? Are you querying anywhere by `score` or `num_voted`?

cyoa/goodbills/features/api/v1/bills.feature
```diff
@@ -0,0 +1,20 @@
+Feature: GoodBills API
+  In order to show bills and voting status
+  as an API client
+  I want to be able to request bills and vote on them
+
+  Scenario: Get bills
+    Given some bills
+    When I ask for bills from the API
+    Then I should receive them with voting information
```
jmtai: This is pretty vague. The description here implies that any number of bills will be returned by the API, but the actual test looks for a specific number of records.

cyoa/goodbills/features/api/v1/bills.feature
```diff
@@ -0,0 +1,20 @@
...
+
+  Scenario: Vote on a bill
+    Given some bills
+    When I vote that I like one
+    And I ask for bills from the API
+    Then my vote should be counted
```
jmtai: A vote can be `Like` or `Dislike` but you aren't testing the `Dislike` action.

cyoa/goodbills/features/api/v1/bills.feature
```diff
@@ -0,0 +1,20 @@
...
+
+  Scenario: Get a single bill
+    Given some bills
+    When I ask for a bill from the API
```
jmtai: How are you asking for the bill from the API? By name? By ID?

cyoa/goodbills/features/billpage.feature
```diff
@@ -0,0 +1,21 @@
...
+  And I click an individual bill link
+  Then I should be on the individual bill page
+  And there should be information about the bill
+  And there should be voting controls
+
+Scenario: Vote Positive
+  Given some bills
+  When I am on an individual bill page
+  And I click the like button
+  Then the bill score should increase
```
jmtai: What about the number of votes? That's present on the bill page, as well.

cyoa/goodbills/features/billpage.feature
```diff
@@ -0,0 +1,21 @@
...
+
+Scenario: Vote Negative
+  Given some bills
+  When I am on a different individual bill page
+  And I click the dislike button
+  Then the bill score should decrease
```
jmtai: So you've got the happy path covered, but what about the sad path and error path?

cyoa/goodbills/features/homepage.feature
```diff
@@ -0,0 +1,22 @@
...
+    And I should not see a previous page button
+
+  Scenario: Paginate bills on home page
+    Given some more bills
+    And I am on the home page
+    When I click the next page button
+    Then I should see different bills
+    And I should see a next page button
+    And I should see a previous page button
+
```
jmtai: You're testing the next page button, but what about the previous page button?

cyoa/goodbills/features/homepage.feature
```diff
@@ -0,0 +1,22 @@
...
+  Scenario: show voting results
+    Given some more bills
+    And I am on the home page
+    Then bills should include voting results
```
jmtai: For consistency, make sure each clause reads like a sentence - i.e. `Then the bills should include voting results`

cyoa/goodbills/features/step_definitions/api/v1/bills_steps.rb
```diff
@@ -0,0 +1,37 @@
...
+When(/^I ask for bills from the API$/) do
+  header 'Accept', 'application/json'
+  get '/api/v1/bills'
+end
+
+Then(/^I should receive them with voting information$/) do
+  bills_json = JSON last_response.body
+  expect(bills_json.length).to eq(10)
+  expect(bills_json[0]["official_title"]).to eq("title-1")
+  expect(bills_json[0]["score"]).to eq(250)
```
jmtai: What if the API returned a valid first record but corrupt or empty remaining records?

cyoa/goodbills/features/step_definitions/api/v1/bills_steps.rb
```diff
@@ -0,0 +1,37 @@
...
+When(/^I vote that I like one$/) do
+  header 'Accept', 'application/json'
+  put '/api/v1/bills/1', bill: { vote: "like" }
+end
+
+Then(/^my vote should be counted$/) do
+  bills_json = JSON last_response.body
+  expect(bills_json.length).to eq(10)
+  expect(bills_json[0]["score"]).to eq(251)
+end
```
jmtai: This is not a great test. It only works because you're hardcoding the test to vote on the first record. A much better test would be to randomly select and verify one of the records. It would be very easy to break the code but still have this test pass.

cyoa/goodbills/features/step_definitions/api/v1/bills_steps.rb
```diff
@@ -0,0 +1,37 @@
...
+
+When(/^I ask for a bill from the API$/) do
+  header 'Accept', 'application/json'
+  get '/api/v1/bills/2'
+end
+
+Then(/^I should receive only that bill, with all information$/) do
+  bill_json = JSON last_response.body
+  expect(bill_json["id"]).to eq(2)
+  expect(bill_json["score"]).to eq(250)
```
jmtai: The bill page also displays number of votes, but it's not being checked here.

cyoa/goodbills/features/step_definitions/billpage_steps.rb
```diff
@@ -0,0 +1,44 @@
+When(/^I click an individual bill link$/) do
+  first(:link, "Bill Detail").click
```
jmtai: Again, here you should not hardcode the test to pick just the first link on the page.

cyoa/goodbills/features/step_definitions/billpage_steps.rb
```diff
@@ -0,0 +1,44 @@
...
+When(/^I am on the homepage$/) do
+  visit('/')
+end
+
+Then(/^I should be on the individual bill page$/) do
+  expect(current_url.include? 'bills').to be(true)
+end
+
+Then(/^there should be information about the bill$/) do
+  expect(page).to have_content("summary")
```
jmtai: This doesn't actually verify that clicking the `Bill Detail` link for a particular bill displays the correct content for that bill. This test will pass even if the info for an entirely different bill were displayed.

cyoa/goodbills/features/step_definitions/billpage_steps.rb
```diff
@@ -0,0 +1,44 @@
...
+end
+
+Then(/^there should be voting controls$/) do
+  expect(page).to have_content("Like")
+  expect(page).to have_content("Dislike")
+end
+
+When(/^I am on an individual bill page$/) do
+  visit('/#/bills/1')
```
jmtai: Ditto here on the hardcoded ID.

cyoa/goodbills/lib/bills_api.rb
```diff
@@ -0,0 +1,40 @@
+require 'congress'
+
+class BillsAPI
+  def self.repull_bills
+    client = Congress::Client.new("1e1ef54e51a840d7a20dee5768e2459e")
```
jmtai: Magic numbers are bad. Assign this to a descriptively named constant, instead.

cyoa/goodbills/lib/bills_api.rb
```diff
@@ -0,0 +1,40 @@
...
+    (1...5).each { |page| repull_bill_page(client, page) }
```
jmtai: The number of pages seems like a good candidate for a configuration option. Definitely shouldn't be hardcoded.

cyoa/goodbills/lib/bills_api.rb
```diff
@@ -0,0 +1,40 @@
...
+  def self.transform(bill)
+    result = add_simple_properties(bill)
+    result[:enacted] = bill['history']['enacted']
+    result[:congress_url] = bill['urls']['congress']
+    result[:summary] = bill['summary_short']
+    result
+  end
+
+  def self.add_simple_properties(bill)
+    [:official_title, :popular_title, :short_title, :bill_id, :last_vote_at,
```
jmtai: My preference would be to assign this to a constant so it reads a little cleaner:
```ruby
SIMPLE_PROPERTIES.each_with_object({}) do |simple_property, result|
  result[simple_property] = bill[simple_property]
end
```

cyoa/goodbills/lib/bills_api.rb
```diff
@@ -0,0 +1,40 @@
...
+     :bill_type, :chamber, :congress, :introduced_on, :last_action_at,
+     :last_version_on].each_with_object({}) do |simple_property, result|
+      result[simple_property] = bill[simple_property]
+    end
+  end
+
+  FIELDS = "summary_short,bill_id,official_title,popular_title," \
```
jmtai: Typically, constants are located at the top of the class definition.

cyoa/goodbills/public/500.html
```diff
@@ -0,0 +1,66 @@
...
+    border-right-color: #999;
+    border-left-color: #999;
+    border-bottom-color: #999;
+    border-bottom-left-radius: 4px;
+    border-bottom-right-radius: 4px;
+    border-top-color: #DADADA;
+    color: #666;
+    box-shadow: 0 3px 8px rgba(50, 50, 50, 0.17);
+  }
+  </style>
```
jmtai: These styles are duplicated in all the error pages. Why not stick them in a separate stylesheet?

cyoa/goodbills/app/controllers/welcome_controller.rb
jmtai: Are you sure you want to seed the DB via a home page view? In development mode, I have to sit and watch a blank screen for almost 30 seconds while the initial data is being fetched and saved. I can't think of too many users that would be willing to sit around that long waiting for a site to load.

Perhaps it would be better to initially seed the DB via an external rake task?

jmtai:
Good start! The code is clean and easy to follow. Is this still a work in progress? There seems to be a lot of missing functionality (both from the plan document `cyoa-plan.txt` as well as the original description of this particular CYOA - https://github.com/jmmastey/level_up_exercises/tree/master/opposite_of_progress).

One question I have is what made you go with Angular for this particular app? The actual implementation could just as easily have been made using straight Rails and a little bit of Ajax without imposing the additional dependencies of Angular, Bower, Restangular, etc. Using the right tools for the job is a big part of the design process and I'm not sure Angular is adding anything here.

The other question is why are there no access controls on voting? Users can log in but there is no difference in app functionality if the user is logged in or out. I should be able to do *something* more if I login to the site.

Finally, make sure you go back and flesh out the testing for the app. Most of the happy path is covered but there's zero sad or error path testing.

elialbert:
hey @jmtai - thanks for the really helpful comments. The story is that I started this, got to a "finishing" point, then switched over to find me a sandwich, the replacement project that dan and blake have been working on. So you're right that there could be more functionality.

I chose angular because 1) I really like it and 2) I wanted more practice integrating angular and rails in sustainable, maintainable ways. This was helpful for figuring that out.

I will go back and follow your suggestions if time permits.

PR #1097, opened 2015-12-14
using lots of activesupport

elialbert:


supportive/supportive.rb
```diff
@@ -9,94 +9,56 @@ class BlagPost
...
+    @categories = args[:categories].reject do |category|
+      DISALLOWED_CATEGORIES.include? category
     end
-
-    @comments = args[:comments] || []
+    @comments = args[:comments]
     @body = args[:body].gsub(/\s{2,}|\n/, ' ').gsub(/^\s+/, '')
-    @publish_date = (args[:publish_date] && Date.parse(args[:publish_date])) || Date.today
+    @publish_date = Date.parse(args[:publish_date])
   end
```
magicalbanana: How would your application handle non-recognized keys?

magicalbanana:
No tests? And add a description to your PR.

supportive/supportive.rb
```diff
@@ -9,94 +9,56 @@ class BlagPost
...
-      hash[key.to_sym] = value
-      hash
-    end
+    args = args.transform_keys(&:to_sym)
+    args = args.reverse_merge(categories: [],
+                              comments: [],
+                              publish_date: Date.today.to_s)
 
-    if args[:author] != '' && args[:author_url] != ''
+    if args[:author].present? && args[:author_url].present?
```
magicalbanana: This is not a bad approach seeing this is a tiny app, but I want you to show me how you would handle this using a setter method for ```@author```.

supportive/supportive.rb
```diff
@@ -9,94 +9,56 @@ class BlagPost
...
-    publish_date + (365 * 3) > Date.today
+    publish_date.years_since(3) > Date.today
   end
 
   def abstract
-    if body.length < 200
-      body
-    else
-      body[0..200] + "..."
-    end
```
magicalbanana: This wasn't entirely terrible. It's good to see that you know in terms of raw ruby how to do this and not be dependent on the built-in methods.

But truncate is a good choice as well.

supportive/supportive.rb
elialbert: Not sure what you mean by that - do you mean an unparseable date?

elialbert:
@magicalbanana the exercise was to replace as much regular ruby with activesupport methods as possible. So I didn't write tests. I could see an argument that this would have been helpful for making sure underlying functionality hadn't changed, but since the file itself had a test pretty much built in at the bottom, I used that. Considering the size of the application, I think that's sufficient. It would be trivial to wrap that in an rspec, but I didn't because it was not in the scope of the exercise.

supportive/supportive.rb
elialbert: do you mean to factor this into a separate "author" method like
def author(args)
  if args[:author].present? && args[:author_url].present?
    @author = Author.new(args[:author], args[:author_url])

or am I missing something

magicalbanana:
@elialbert Ah, is that what this exercise is? Haha. My bad. But yeah, if that's the case it's good. Well, since we are TDD shop here a test would always be not out of scope so writing one ensuring that the functionality hasn't changed actually ensures that your code is working the way it's supposed to work.

supportive/supportive.rb
magicalbanana: A setter method so it would start of like this:

```
def author=(author)
  #logics here..
end
```

elialbert:
yah true. I was just running the file and visually identifying that the output hadn't changed, but if the file got any more complicated I would definitely wrap that in rspec.

magicalbanana:
Yep. But still, a test would've been good at the get go. But I do realize this is not part of the exercise although it should be seeing that we're very TDD (in a way - lots of folks write their test AFTER they write their code) and we are enforcing coverage.

elialbert:
sure thing. just wrapped that stuff at the bottom in an rspec test. I don't think hound will be super pleased...

paulghaddad:
@elialbert Nice work on adding the ActiveSupport methods. Also, good extra credit on adding specs. You're good to move on!

elialbert:
cool thanks

PR #1098, opened 2015-12-14
Shopping Cart - Iteration 1

jrade-enova:


shopping_cart/cart.feature
```diff
@@ -0,0 +1,47 @@
+Feature: A working shopping cart
```
magicalbanana: You're feature title should be the actual feature in which this case is a shopping cart and you can further explain on your scenarios what the states of those shopping carts are and how your application handles those scenarios.

shopping_cart/cart.feature
```diff
@@ -0,0 +1,47 @@
...
+  As a consumer
+  I want a shopping cart
+  So that I can manage my ordering
+
+  Scenario: Add item
```
magicalbanana: Be more expressive in your scenarios. Explore or state the different possibilities that may occur and how your application handles those situation.

Here are some resources:

1. https://semaphoreci.com/community/tutorials/introduction-to-writing-acceptance-tests-with-cucumber
2. http://mixandgo.com/blog/7-steps-to-better-cucumber-tests

And I believe we have some books in the library here in Enova as well.

magicalbanana:
Review my comments and read up more on writing feature tests before attempting resubmission.

jrade-enova:
@magicalbanana Will do, thanks!

PR #1099, opened 2015-12-16
redux duo

Dwin357:


dino_catalog/first_solution/dino_catalog.rb
```diff
@@ -0,0 +1,65 @@
...
+                  converters: :numeric, 
+                  headers: true,
+                  :header_converters => lambda {|h| h.downcase}
+                  ) do |row|
+        args = DinoSaur.construct_arguments(row)
+        rtn << DinoSaur.new(args)
+      end
+    end
+    rtn
+  end
```
jmmastey: We're into minor nitpicks at this point. If converting a single file had a method (say, `build_dinos_from_file`) this method would be:

```
def build_collection
  Dir['../dino_fact_csvs/*'].inject do |arr, path|
    arr + build_dinos_from_file(path)
  end
end

File: dino_catalog/first_solution/dino_catalog.rb [hunk 2]
```diff
@@ -0,0 +1,65 @@
...
+        collection = collection.select{ |d| d.has_attribute_value?(katagory, value)}
+      elsif DinoSaur::RECOGNIZED_CSV_COLUMN_HEADERS.include?(katagory)
+        attribute = DinoSaur.look_up_attribute_by_csv_header(katagory)
+        collection = collection.select{ |d| d.has_attribute_value?(attribute, value)}
+      else
+        collection = collection.select{ |d| d.has_additional_info_value?(katagory, value)}
+      end
+    end
+    collection
+  end
```
jmmastey: Notice that this class has two purposes right now. 1) building dinosaurs from CSVs and 2) filtering and searching those dinosaurs.

As the file is only 65 lines right now, it's probably good enough right now. Later on, if you added more methods to load data (say, JSON, or a webservice), the right answer would be to separate build strategies into a different file. At which point this one follows SRP.

If you read POODR, Sandi goes into the idea that you only refactor when there's pain, or something that drives you to that refactor. Actually, I think the All the Little Things talk also goes into this. Consider that talk as well.

dino_catalog/first_solution/dino_command_line_interface.rb
```diff
@@ -0,0 +1,93 @@
...
+          end
+          if filter_criteria.include?("-b") || filter_criteria.include?("big")
+            collection = dino_catalog.big_dinosaurs(collection)
+          end
+          if filter_criteria.include?("-2") || filter_criteria.include?("biped")
+            collection = dino_catalog.biped_dinosaurs(collection)
+          end
+        end
+        display_collection(collection)
+
```
jmmastey: Generally `when` blocks shouldn't have large chunks of code. This pretty clearly should be refactored into a method. I think there's probably a way to get at that triple if statement above, but it'd be easier to see if it wasn't buried in a case block.

dino_catalog/first_solution/dino_command_line_interface.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  def parse_search_terms(search_terms)
+    rtn = {}
+    search_terms.each do |term|
+      search_pair = term.split(":")
+      rtn[search_pair[0].downcase] = search_pair[1].downcase
+    end
+    rtn
+  end
```
jmmastey: At the least, `each_with_object` here. This block is likely to stay a little ugly, tho.

dino_catalog/first_solution/dino_saur.rb
```diff
@@ -0,0 +1,141 @@
...
+        additional_info.each do |k, v|
+          if v
+            output << "#{k}: #{v}"
+          end
+        end
+      else
+        output << "#{attribute.to_s[1..-1]}: #{self.instance_variable_get(attribute)}"
+      end
+    end.compact.join(", ") + "\n\n"
+  end
```
jmmastey: This still uses the same algorithm as `hash_dino` (which should still be `to_h`). Use that instead and this method will mostly disappear.

dino_catalog/first_solution/dino_saur.rb
```diff
@@ -0,0 +1,141 @@
...
+
+  def self.hash_dino(dino_instance)
+    dino_instance.instance_variables.each_with_object({}) do |attribute, hash|
+      next if !dino_instance.instance_variable_get(attribute)
+      hash[attribute.to_s[1..-1]] = dino_instance.instance_variable_get(attribute)
+    end
+  end
+
+end
```
jmmastey: Same as with the original driver class, eventually you're going to get another CSV and this file will grow, and it shouldn't. In the medium term, this stuff would need to be refactored into CSV strategies. Then this class would control only the regularized dinosaur attributes.

dino_catalog/first_solution/spec/dino_saur_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+# so what do I care about?
+# I case that all the various boolean tests work
+# I care that the object instantiates correctly
+# I care that the various outputs (to_s, to_h) work correctly
+# I care... that the class methods work
+
+describe DinoSaur do
+  it "" do
+  end
+end
```
jmmastey: :)

dino_catalog/second_solution/dinosaur.rb
```diff
@@ -0,0 +1,128 @@
+class Dinosaur
```
jmmastey: ACH, god, just realized there are two sets in here. Watch while all my notes were already followed. This already looks tons better.

dino_catalog/second_solution/dinosaur.rb
```diff
@@ -0,0 +1,128 @@
...
+    when "weight", "weight_in_lbs"
+      :weight
+    when "locomotion", "walking"
+      :locomotion
+    when "period"
+      :period
+    when "diet"
+      :diet
+    end
+  end
```
jmmastey: I think we lost continent somewhere here. Did we lose continent?

dino_catalog/second_solution/dinosaurs.rb
```diff
@@ -0,0 +1,80 @@
...
+    Dinosaurs.new( collection.select(&:biped?) )
+  end
+
+  def carnivorus_dinosaurs
+    Dinosaurs.new( collection.select(&:carnivore?) )
+  end
+
+  def dinosaurs_from(period)
+    Dinosaurs.new( collection.select{|d| d.from?(period)} )
+  end
```
jmmastey: Noice. :+1:

dino_catalog/second_solution/dinosaurs.rb
```diff
@@ -0,0 +1,80 @@
...
+    file_paths.each_with_object(Dinosaurs.new) do |file_path, dinosaurs|
+      CSV.foreach(file_path, 
+                  converters: :numeric, 
+                  headers: true,
+                  :header_converters => lambda {|h| h.downcase}
+                  ) do |row|
+        dinosaurs.add_dino( Dinosaur.from_csv_row(row) )
+      end
+    end
+  end
```
jmmastey: See? You already did the thing :D 

If you use `each_with_object`, you probably don't need to return `self` from `add_dino`. `inject` requires you to return the right object, but `each_with_object` ignores return values entirely.

dino_catalog/second_solution/spec/dino_runner_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+FORMAT_ONE_FILE_DATA = "NAME,PERIOD,CONTINENT,DIET,WEIGHT_IN_LBS,WALKING,DESCRIPTION\nAlbertosaurus,Late Cretaceous,North America,Carnivore,2000,Biped,Like a T-Rex but smaller.\nAlbertonykus,Early Cretaceous,North America,Insectivore,,Biped,Earliest known Alvarezsaurid.\nBaryonyx,Early Cretaceous,Europe,Piscivore,6000,Biped,One of the only known dinosaurs with a fish-only diet.\nDeinonychus,Early Cretaceous,North America,Carnivore,150,Biped,\nDiplocaulus,Late Permian,North America,Carnivore,,Quadruped,They actually had fins on the side of their body."
+
+FORMAT_TWO_FILE_DATA = "Genus,Period,Carnivore,Weight,Walking\nAbrictosaurus,Jurassic,No,100,Biped\nAfrovenator,Jurassic,Yes,,Biped\nCarcharodontosaurus,Albian,Yes,3000,Biped\nParalititan,Cretaceous,No,120000,Quadruped\nSuchomimus,Cretaceous,Yes,10400,Biped"
+
+def stage_csv_directory
+  Dir.mkdir(TEST_DIR_PATH) unless File.exists?(TEST_DIR_PATH)
+  File.open("#{TEST_DIR_PATH}/file_one.csv", "w") { |file| file.write(FORMAT_ONE_FILE_DATA)}
+  File.open("#{TEST_DIR_PATH}/file_two.csv", "w") { |file| file.write(FORMAT_TWO_FILE_DATA)}
+  [TEST_DIR_PATH]
+end
```
jmmastey: So, test questions!

If at all possible, it's a huge benefit to not need to change the filesystem to run your tests. If you do need to have file references, there is a concept of a temp file (that lives in memory). Basically, files are slow.

So, preferably, you'd have some tests that deal with whether files are getting accessed, but every other test would just start with a prepared set of data.

dino_catalog/second_solution/spec/dino_runner_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  let(:runner) { DinoRunner.new(stage_csv_directory)}
+  let(:dinosaurs) {Dinosaurs.from_directory(stage_csv_directory.first)}
+  let(:default_runner) {DinoRunner.new}
+  let(:default_dinosaurs) {Dinosaurs.from_directory}
+
+  describe "#initialize" do
+    it "creates a DinoRunner obj" do
+      expect(runner).to be_an_instance_of DinoRunner
+    end
```
jmmastey: I'm not sure it's possible for `DinoRunner.new` to be anything other than a `DinoRunner` instance. Remember that tests should fulfill probably at least two of the following criteria:

1) The test is easy to write and maintain.
2) The error it catches is likely to happen.
3) The consequences of the error are significant.

I think this test is maybe only #1.

dino_catalog/second_solution/spec/dino_runner_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+    it "populates a dino_catalog w/ a Dinosaurs obj" do
+      expect(runner.dino_catalog.class).to eq(dinosaurs.class)
+      expect(runner.dino_catalog.collection.count).to eq(dinosaurs.collection.count)
+    end
+    it "defaults to csv directory in dinosaurs if no arg given" #do
+    #   expect(default_runner.dino_catalog).to eq(default_dinosaurs)
+    # end
```
jmmastey: The way you'd test this is to look at what the runner invokes.

```
allow(File).to receive(:open).and_call_original
expect(File).to have_received(:open).with("/my/default/path")
```

dino_catalog/second_solution/spec/dino_runner_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  describe "#run" do
+    # no idea how to test any of this
+    it "calls #greeting_display at opening" 
+    it "calls #options_display at opening"
+    it "loops until it receives the break command"
+    it "pulls in user's command line requests"
+    it "calls the correct method w/ correct args based on user command"
+    it "displays user requested info to terminal"
+  end
```
jmmastey: Most of these are the same as the above (testing whether a message gets called).

The loop stuff is the hardest. Long and short is that I probably wouldn't actually test the loop itself. Hopefully the loop is in its own method and just runs a `do_command(str)` method. Then, I would test:

```
allow(runner).to receive(:stop_execution) { nil }
do_command("break")
expect(runner).to have_received(:stop_execution)
```
do_command

jmmastey: For testing output, there are basically three strategies:
1. Have your runner class return strings and some other wrapper actually output those strings. Test that.
2. Use Dependency Injection to pass in which object gets puts'd:
```
def initialize(outputter = Kernel)
  @outputter = outputter
  ...
  @outputter.puts("Foo")
end

class FakeOutputter
  def puts(str)
    @output_strings << str
  end
end

it "does output shit" do
  fake_outputter = FakeOutputter.new
  runner = Runner.new(fake_outputter)
  expect(fake_outputter.output_data).to include("foo")
end
```
3. Fuck it, don't bother testing output strings. Maybe they don't pass 2 of the 3 criteria at the top.

dino_catalog/second_solution/spec/dino_runner_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    it "returns a Dinosaurs collection" do
+      expect(runner.filter_catalog(filter_args_shorthand)).to be_an_instance_of Dinosaurs
+    end
+    it "accepts an array of arg filters" #no idea
+    it "correctly parses filter terms & applies methods" do
+      expect(runner.filter_catalog(filter_args_b).collection.all?{ |d| d.big?}).to be true
+      expect(runner.filter_catalog(filter_args_2).collection.all?{ |d| d.biped? }).to be true
+      expect(runner.filter_catalog(filter_args_c).collection.all?{ |d| d.carnivore? }).to be true
+    end
```
jmmastey: Use `be_truthy`. Also, this _may_ be a lot of expectations loaded into one test.

jmmastey: Also, `collection.all?(&:biped?)`

dino_catalog/second_solution/spec/dino_runner_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    it "applies a method for each arg" do
+      expect(runner.filter_catalog(filter_args_verbose).collection.all?{ |d| d.carnivore? && d.biped? && d.big?}).to be true
+      expect(runner.filter_catalog(filter_args_shorthand).collection.all?{ |d| d.carnivore? && d.biped? && d.big?}).to be true
+      expect(runner.filter_catalog(filter_args_bc).collection.all?{ |d| d.carnivore? && d.big?}).to be true
+      expect(runner.filter_catalog(filter_args_2b).collection.all?{ |d| d.biped? && d.big?}).to be true
+      expect(runner.filter_catalog(filter_args_2c).collection.all?{ |d| d.carnivore? && d.biped? }).to be true
+    end
```
jmmastey: This is definitely a lot of expectations in one test method. I'm not sure I'd test every single permutation here. It might be sufficient to test just a few combinations.

dino_catalog/second_solution/spec/dinosaur_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+
+describe Dinosaur do
+
+  let(:bob) { Dinosaur.new({name: "Bob", 
+                            weight: 280, 
+                            diet: "coffee & doughnuts", 
+                            locomotion: "rolling chair", 
+                            period: "haven't gone since highschool", 
+                            additional_info: {fun_fact: "I only drink when I smoke", 
+                                              best_quote: "there aint no sunshine when shes gone"
```
jmmastey: only darkness every day :(

(bum, bum, buum)

dino_catalog/second_solution/spec/dinosaur_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+  let(:african_carnivore_csv_row){ CSV::Row.new ['genus','period','carnivore','weight','walking'], ['Suchomimus','Cretaceous','Yes',10400,'Biped']}
+  let(:dex_insectivore_csv_row){ CSV::Row.new ['name','period','continent','diet','walking','description'], ['Albertonykus','Early Cretaceous','North America','Insectivore','Biped','Earliest known Alvarezsaurid.']}
+  let(:dex_carnivore_csv_row){ CSV::Row.new ['name','period','continent','diet','weight_in_lbs','walking','description'],['Albertosaurus','Late Cretaceous','North America','Carnivore',2000,'Biped','Like a T-Rex but smaller.']}
+  let(:dex_piscivore_csv_row){CSV::Row.new ['name','period','continent','diet','weight_in_lbs','walking','description'], ["Baryonyx", "Early Cretaceous","Europe","Piscivore",6000,"Biped","One of the only known dinosaurs with a fish-only diet."]}
+  let(:dex_herbivore_csv_row){CSV::Row.new ['name','period','continent','diet','walking','description'], ["Dracopelta","Early Cretaceous or Late Jurassic","South America","Herbivore","Quadruped","One of the most primitive known Ankylosauria"]}
```
jmmastey: I _think_ (but don't necessarily totally know) that CSV::Row conforms to either the Array or Hash interface. I think the latter. In which case, use an Array or a Hash instead.

dino_catalog/second_solution/spec/dinosaur_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+  let(:dex_insectivore_dino) {Dinosaur.from_csv_row(dex_insectivore_csv_row)}
+  let(:dex_carnivore_dino) {Dinosaur.from_csv_row(dex_carnivore_csv_row)}
+  let(:dex_piscivore_dino) {Dinosaur.from_csv_row(dex_piscivore_csv_row)}
+  let(:dex_herbivore_dino) {Dinosaur.from_csv_row(dex_herbivore_csv_row)}
+
+
+  describe "#initialize and accesors" do
+    it "creates a dinosaur object" do
+      expect(bob).to be_an_instance_of Dinosaur
+    end
```
jmmastey: Low value test

dino_catalog/second_solution/spec/dinosaur_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+    it "has a #name" do
+      expect(bob.name).to eq("Bob")
+    end
```
jmmastey: Ehhhh, assignment tests are possibly low value as well. If I were to test this, I might throw them all into one assertion because I don't care very much about them.

dino_catalog/second_solution/spec/dinosaur_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+  end
+
+  describe "#big?" do
+    it "responds true if weight is >= 4000" do
+      expect(african_carnivore_dino.big?).to be true
+    end
+    it "responds false if weight is < 4000" do
+      expect(dex_carnivore_dino.big?).to be false
+    end
+  end
```
jmmastey: Depending on naming, you should be able to say `expect(african_carnivore_dino).to be_big`. and `not_to be_big`

dino_catalog/second_solution/spec/dinosaur_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+
+  describe "#biped?" do
+    it "responds true if locomotion is =='biped'" do
+      expect(dex_piscivore_dino.biped?).to be true
+    end
+    it "responds false if locomotion is !='biped'" do
+      expect(dex_herbivore_dino.biped?).to be false
+    end
+  end
```
jmmastey: I'm worried that the `dex_piscivore_dino` isn't really necessarily a biped, which means that you might change it and break an unrelated test. Consider this possible structure:

```
def dinosaur(attrs = {})
  attrs = { name: "Charizard", size: 2000, ... }.merge(attrs)
  Dinosaur.new(attrs)
end

it "detects bipeds directly" do
  expect(dinosaur(locomotion: "biped")).to be_biped
end
```

Also, while I'm at it, try to make your test descriptions more english explanations. On most codebases, the tests *are* the documentation, and documentation is written for humans.

dino_catalog/second_solution/spec/dinosaur_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+      expect(african_herbivore_dino.to_h).to be_an_instance_of Hash
+    end
+    it "has all the model's attributes" do
+      expect(african_herbivore_dino.to_h["name"]).to eq("Abrictosaurus")
+      expect(african_herbivore_dino.to_h["weight"]).to eq(100)
+      expect(african_herbivore_dino.to_h["diet"]).to eq("herbivore")
+      expect(african_herbivore_dino.to_h["period"]).to eq("Jurassic")
+      expect(african_herbivore_dino.to_h["locomotion"]).to eq("Biped")
+    end
+  end
```
jmmastey: Include (and document) a null attr. Skip repetition of `african_herbivore_dino.to_h`:

```
it "has all the model's attributes" do
  dino_hash = african_herbivore_dino.to_h
  expect(dino_hash["name"]).to eq("Abrictosaurus")
end
```

dino_catalog/second_solution/spec/dinosaur_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+      expect(dex_insectivore_dino).to be_an_instance_of Dinosaur
+      expect(african_herbivore_dino.name).to eq("Abrictosaurus")
+      expect(african_herbivore_dino.weight).to eq(100)
+      expect(dex_carnivore_dino.weight).to eq(2000)
+      expect(african_herbivore_dino.diet).to eq("herbivore")
+      expect(african_carnivore_dino.diet).to eq("carnivore")      
+      expect(african_herbivore_dino.locomotion).to eq("Biped")
+      expect(african_herbivore_dino.period).to eq("Jurassic")
+      expect(dex_insectivore_dino.additional_info).to eq({"continent"=>"North America", "description"=>"Earliest known Alvarezsaurid."})
+    end
```
jmmastey: I cannot tell how to verify that this method is doing what's expected, so all I can do is believe you. If you included the hashes more locally and did targeted tests, I could understand better what the method does:

it "returns a fully populated dinosaur obj" do
  csv_data = { name: "Abrictosaurus", carnivore: "false" ... }
  
  dino = Dinosaur.from_csv_row(csv_data)

  expect(dino.diet).to eq("carnivore")
```

jmmastey:
Code is in a much better place. I'm nitpicking now :D 

Overall tests need to surface information about the intention of the software better. This is the really hard part about testing. Three things to read / watch if you haven't already:
1. Sandi Metz on Unit Testing: https://www.youtube.com/watch?v=URSWYvyc42M
2. Justin Searls on How Not to Hate Your Test Suite: https://www.youtube.com/watch?v=VD51AkG8EZw
3. Ruby Rspec Guidelines / Best Practices: http://betterspecs.org/

It's in one of the comments, but remember these criteria for writing tests:
1. Is it easy to write and maintain this test.
2. Is the functionality likely to break.
3. Is the breakage likely to have significant consequences.

Only test if you can hit 2 out of 3. Keep at it!

PR #1101, opened 2015-12-24
Overlord - Iteration 2

creativisionist:
• Edited according to suggestions/input.
• How deep/wide typically does one go when utilizing/testing with/for pO models?  Being that it is another level of abstraction is there also a point (in bigger applications, I suspect) where it creates efficiency in testing if/when code gets changed?  (Can discuss this more in person.)
• Aside from utilizing only "if's" is there a more elegant way to test for a double conditional?  I recognize I could have simply edited the instructions but thought it might be a good exercise for me in testing the sad path(s) for at least one scenario (though as a user I might like to have every double scenario covered as well despite that those were already tested, or more likely just have it all in the instructions upfront to begin with, unless it was my deliberate intent as a developer to have it otherwise).


overlord/features/bomb_activation.feature
```diff
@@ -0,0 +1,21 @@
+Feature: Bomb activation
+  In order to destroy Megaton
+  As the evil Vault Boy
+  I want to activate the bomb
```
jmtai: :+1: for having a feature description.

overlord/features/bomb_activation.feature
```diff
@@ -0,0 +1,21 @@
...
+
+  Scenario: Visit the inactive bomb page
+    When I am on the inactive_bomb page
```
jmtai: `inactive_bomb` looks like a variable name. No need to use anything other than plain english in the scenarios since using specific variable names makes the test a bit too tied to a particular implementation.

overlord/features/bomb_activation.feature
```diff
@@ -0,0 +1,21 @@
...
+    Then I should see instructions and the form to submit the activation code
```
jmtai: Make the `and` condition a separate line, just like you've done with the other scenarios.

overlord/features/bomb_activation.feature
```diff
@@ -0,0 +1,21 @@
...
+
+  Scenario: Successful bomb activation
+    Given I am on the inactive_bomb page
+    When I input the correct activation code
+    Then I should be directed to the active_bomb page
+    And the page should contain Bomb Activated
```
jmtai: Once you have been successfully directed to the active bomb page, is it possible for the page to NOT display `Bomb Activated`? If not, this condition doesn't really need to be tested.

overlord/features/bomb_activation.feature
```diff
@@ -0,0 +1,21 @@
...
+
+  Scenario: Unsuccessful bomb activation
+    Given I am on the inactive_bomb page
+    When I input the incorrect activation code
```
jmtai: Minor nitpick - the wording here should probably be `When I put *an* incorrect activation code`, since `the` implies there is only one possible incorrect code.

overlord/features/bomb_deactivation.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Bomb deactivation
+  In order to deactivate the bomb
+  As the good Vault Boy
+  I want to input the correct deactivation code
```
jmtai: Since you're testing more than just the correct deactivation code, you probably want to write this description more like you did for bomb activation. e.g.

```
In order to save Megaton
As the good Vault Boy
I want to deactivate the bomb
```

overlord/features/bomb_deactivation.feature
```diff
@@ -0,0 +1,25 @@
...
+
+  Scenario: Visit the active_bomb page
+    When I am on the active_bomb page
+    Then I should see instructions and the form to submit the deactivation code
```
jmtai: Again, make the `and` condition a separate line here.

overlord/features/bomb_deactivation.feature
```diff
@@ -0,0 +1,25 @@
...
+
+  Scenario: Successful bomb deactivation
+    Given I have successfully activated the bomb
+    When I input the correct deactivation code
+    Then I should be directed to the inactive_bomb page
+    And the page should contain You've Saved Megaton! For now...
+
+  Scenario: Unsuccessful bomb deactivation
+    Given I have successfully activated the bomb
+    When I input the incorrect deactivation code
```
jmtai: s/the/an

overlord/features/bomb_deactivation.feature
```diff
@@ -0,0 +1,25 @@
...
+    Then I should be directed to the active_bomb page
+    And the page should contain The Bomb Is Still Active!
+
+  Scenario: Unsuccessful deactivation leading to explosion
+    Given I have successfully activated the bomb
+    When I input the incorrect deactivation code max times
```
jmtai: Nice job keeping the test description generic here and not hardcoding a number.

overlord/features/bomb_deactivation.feature
```diff
@@ -0,0 +1,25 @@
...
+
+  Scenario: Visit the active_bomb page
+    When I am on the active_bomb page
+    Then I should see instructions and the form to submit the deactivation code
+
+  Scenario: Successful bomb deactivation
+    Given I have successfully activated the bomb
+    When I input the correct deactivation code
+    Then I should be directed to the inactive_bomb page
+    And the page should contain You've Saved Megaton! For now...
```
jmtai: One thing you might want to do instead of checking for exact string matches is to use a generic descriptor, instead (like you do with `max times` below). If you do something like

```
And the page should contain a success message
```

Then if you decide to modify the text of the success message, this cucumber test is still valid and you only have to change the matcher in the step definitions.

overlord/features/configuration.feature
```diff
@@ -0,0 +1,56 @@
+Feature: Code configurations
+  In order to activate the bomb
+  As the evil Vault Boy
+  I want to input the correct activation code
```
jmtai: This page doesn't actually activate the bomb, so this description is slightly misleading. Perhaps something like this would be better?

```
In order to configure the bomb
As the evil Vault Boy
I want to input activation and deactivation codes
```

overlord/features/error_page.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Error page
+  As a Vault Boy
+  When I go to a page that does not exist
+  I will get a 404 page
```
jmtai: For this feature, the description basically mirrors the only scenario so you could probably just omit it.

overlord/features/error_page.feature
```diff
@@ -0,0 +1,8 @@
...
+
+  Scenario: Get a 404 page
+    When I go to a page that does not exist
+    Then I will get a 404 page
```
jmtai: These are expectations, so stick with the `should` wording instead of `will`.

overlord/features/step_definitions/error_page_steps.rb
```diff
@@ -0,0 +1,7 @@
+When(/^I go to a page that does not exist$/) do
+  visit "/asdf"
+end
+
+Then(/^I will get a 404 page$/) do
+  expect(page).to have_title "#fail"
```
jmtai: Here your test specifically describes a 404 page but is indirectly testing for that response by examining the page title. This is ok since selenium doesn't have a good way to check response codes (although capybara-webkit and poltergeist do). Since the important thing is the user sees your "Please Stand By" page, I'd probably just change the wording of the test to that, instead (i.e. `Then I should see the Please Stand By page`).

overlord/features/step_definitions/overlord_steps.rb
```diff
@@ -0,0 +1,120 @@
+ACTIVATION_CODE = 1234
+DEACTIVATION_CODE = 0000
+DUPLICATE_CODE = 1234
+INCORRECT_CODE = "asdf1234"
+BLANK_CODE = ""
```
jmtai: Nice use of constants here.

overlord/features/step_definitions/overlord_steps.rb
```diff
@@ -0,0 +1,120 @@
...
+  expect(@bomb_page).to have_deactivation_code_field
+  expect(@bomb_page).to have_submit_button
+end
+
+When(/^I set valid activation and deactivation codes$/) do
+  initialize_bomb(ACTIVATION_CODE, DEACTIVATION_CODE)
+end
+
+When(/^I set an invalid activation code with a valid deactivation code$/) do
+  initialize_bomb(INCORRECT_CODE, DEACTIVATION_CODE)
```
jmtai: Minor nit:  Your test uses the descriptor `invalid` but your constant uses `INCORRECT`.

overlord/features/configuration.feature
```diff
@@ -0,0 +1,56 @@
...
+    When I set invalid duplicate activation and deactivation codes
+    Then I should be directed to the bomb page
+    And the page should contain Codes must be different.
+    And the page should contain Codes can only be numeric.
+
+  Scenario: Set valid duplicate activation and deactivation codes
+    Given I am on the bomb page
+    When I set valid duplicate activation or deactivation codes
+    Then I should be directed to the bomb page
+    And the page should contain Codes must be different.
```
jmtai: Nice job testing all the possible combinations.

overlord/features/support/helpers.rb
```diff
@@ -0,0 +1,14 @@
+module Helpers
+  def initialize_bomb(code_1, code_2)
```
jmtai: Why `code_1` and `code_2`? Why not `activation_code` and `deactivation_code`?

overlord/features/step_definitions/overlord_steps.rb
```diff
@@ -0,0 +1,120 @@
...
+
+Then(/^I should see instructions and the form to submit the deactivation code$/) do
+  expect(@active_bomb_page).to have_instructions
+  expect(@active_bomb_page).to have_deactivation_code_field
+  expect(@active_bomb_page).to have_submit_button
+end
+
+When(/^I input the incorrect activation code$/) do
+  initialize_bomb(ACTIVATION_CODE, DEACTIVATION_CODE)
+  validate_code('activation_code', INCORRECT_CODE)
```
jmtai: Since your `validate_code` method is submitting the code AND validating it, you might want to change the name to something like `submit_and_validate_code`. Method names should make it clear what they are actually doing.

overlord/overlord.rb
```diff
@@ -1,14 +1,121 @@
...
 
 enable :sessions
 
 get '/' do
-  "Time to build an app around here. Start time: " + start_time
+  erb :index
+end
+
+get '/bomb' do
+  session[:bomb] = nil if session[:bomb]
```
jmtai: From a UX perspective, do you think it makes sense to discard an existing bomb when this page is loaded? What if I go through the trouble to configure and activate the bomb and then accidentally go to the `/bomb` page? The activated bomb gets discarded, which means I've completely bypassed the need to deactivate the bomb.

overlord/overlord.rb
```diff
@@ -1,14 +1,121 @@
...
+  erb :bomb
+end
+
+post '/boot' do
+  activation_code = params[:activation_code]
+  deactivation_code = params[:deactivation_code]
+  max_failed_deactivations = 3
+  errors = collect_errors(activation_code, deactivation_code)
```
jmtai: Naming is tricky but we definitely want to make our method names as descriptive as possible. You're actually validating the codes here, so something like `validate_codes` might be more appropriate than `collect_errors`.

overlord/overlord.rb
```diff
@@ -1,14 +1,121 @@
...
+
+  if errors.empty?
+    create_bomb(activation_code, deactivation_code, max_failed_deactivations)
+    redirect '/inactive_bomb'
+  else
+    errors.map do |k, v|
+      flash[k] = v
+    end
```
jmtai: For one line blocks, it's usually preferred to use the curly brace notation:
```ruby
errors.map { |k, v| flash[k] = v }
```

overlord/overlord.rb
```diff
@@ -1,14 +1,121 @@
...
+    redirect '/bomb'
+  end
+end
+
+get '/inactive_bomb' do
+  return erb :inactive_bomb if bomb && bomb.inactive?
```
jmtai: FYI, when you start working with Rails, there is a [try](http://apidock.com/rails/Object/try) method that allows you to eliminate the initial nil check:
```ruby
return erb :inactive_bomb if bomb.try(:inactive?)
```

overlord/overlord.rb
```diff
@@ -1,14 +1,121 @@
...
+
+def numeric?(code)
+  (/[^[:digit:]]+/).match(code).to_s.empty?
+end
+
+def empty?(code)
+  code.empty?
+end
+
+get '/reroute' do
```
jmtai: Having this be a route instead of a method is a little funky since you end up redirecting twice.

overlord/overlord.rb
```diff
@@ -1,14 +1,121 @@
...
+  session[:bomb]
+end
+
+def create_bomb(activation_code, deactivation_code, max_failed_deactivations)
+  new_bomb = Bomb.new(activation_code, deactivation_code, max_failed_deactivations)
+  session[:bomb] = new_bomb
+end
+
+def numeric?(code)
+  (/[^[:digit:]]+/).match(code).to_s.empty?
```
jmtai: This is a little more complicated than it needs to be. Instead of checking for a negative condition, you could write it like this:
```ruby
>> /^\d+$/.match("123")
=> #<MatchData "123">
```
However, this won't return a boolean result - just a MatchData object if there's a match or nil if there's no match. To convert it to a boolean value, we can use a little trick: `!!` (double negate)
```ruby
>> !! /^\d+$/.match("123")
=> true
```

overlord/public/stylesheets/overlord.css
```diff
@@ -0,0 +1,115 @@
...
+body {
+  -moz-background-size: cover;
+  -o-background-size: cover;
+  -webkit-background-size: cover;
+  background-size: cover;
+  background-repeat: no-repeat;
+  background-attachment: fixed;
+  background-position: center;
+  font-family: 'Share Tech Mono';
+}
```
jmtai: :+1: for listing properties in alphabetical order.

overlord/views/active_bomb.erb
```diff
@@ -0,0 +1,33 @@
...
+        <input type="text" placeholder="Deactivation Code" autocomplete="off" name="deactivation_code">
+        <div>
+          <button class="whitebutton" type="submit" name="submit_button">Save Megaton</button>
+        </div>
+      </form>
+
+      <div class="flash">
+        <% if flash[:uhoh] %>
+          <%= flash[:uhoh] %>
+        <% end %>
```
jmtai: Since fetching a nonexistent key from a hash will just return nil, there's no need for the extra `if`. You can just use
```ruby
<%= flash[:uhoh] %>
```

overlord/views/inactive_bomb.erb
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>Nuked.</title>
```
jmtai: Was this title intended to be something else?

overlord/views/inactive_bomb.erb
```diff
@@ -0,0 +1,34 @@
...
+        <input type="text" placeholder="Activation Code" autocomplete="off" name="activation_code">
+        <div>
+          <button class="redbutton" type="submit" name="submit_button">Activate</button>
+        </div>
+      </form>
+
+      <div class="flash">
+          <% if flash[:error] %>
+            <%= flash[:error] %>
+          <% else flash[:safe] %>
```
jmtai: I think you intended this to be `elsif`?

jmtai:
Nice work! The code is clean and easy to follow and the test coverage is excellent. The only issue I have with the implementation is the disappearing bomb problem, although that is a UX/design issue and not necessarily within the scope of the exercise.

Read over the comments and let me know if you have any questions - I don't think you need to resubmit this one, though.

overlord/features/bomb_activation.feature
creativisionist: According to the pO model descriptor it seems to follow the one line convention.  I think it's because it's referring to all the elements on the page but I am uncertain.  In this case would it just be a stylistic preference?

Thoughts?

overlord/features/support/helpers.rb
creativisionist: I used code_1 and code_2 b/c this method is a helper to test for different scenarios, so the user might not actually be inputting a real activation or deactivation code, but an invalid code instead.

overlord/overlord.rb
creativisionist: So cool! :+1:

overlord/views/inactive_bomb.erb
creativisionist: :innocent:

overlord/features/bomb_deactivation.feature
creativisionist: I had it set the reverse way: one general cucumber test to test for all the different string matchers (cf overlord_steps.rb).

overlord/features/support/helpers.rb
jmtai: Since code_1 is always being used to fill in the activation_code, and code_2 is always being used to fill in the deactivation_code, it would make more sense to just use those names. The actual content of the code (whether or not it's valid) doesn't really matter for naming purposes.

PR #1103, opened 2015-12-24
Overlord - Iteration 1

cstavitsky:
Didn't take on the optional stuff; I'd like to get feedback on how this looks so far. Thanks!!


magicalbanana:
First off, please address all of the hound errors you got. Thanks!

overlord/features/active_deactivate.feature
```diff
@@ -0,0 +1,31 @@
+Feature: Disarm active bomb
+	As a supervillain
+	I want to be able to disarm the bomb when activated
+	So that I can stay alive if I accidentally activate it
+
+	Scenario: User views disarm interface for an active bomb
+		Given The bomb has been initialized
```
magicalbanana: This is just formatting but try to limit your  tab spacing to two.

overlord/features/step_definitions/general_steps.rb
```diff
@@ -0,0 +1,71 @@
...
+
+Then(/^I should see a[n]? "(.*?)" submit button$/) do |button_id|
+  expect(page).to have_css("input#{button_id}")
+end
+
+When(/^I press the "(.*?)" button$/) do |button_name|
+  click_button(button_name)
+end
+
+Then(/^I fill in the following:$/) do |_object, value|
```
magicalbanana: What's the purpose of the first attribute? If you have no need for this simple replace is with an underscore.

overlord/features/step_definitions/general_steps.rb
```diff
@@ -0,0 +1,71 @@
...
+  object = value
+end
+
+Then(/^I should be on the homepage$/) do
+  visit("/")
+end
+
+CSS_MATCHERS = {
```
magicalbanana: I wouldn't call this CSS matchers since this is mostly html attributes, attributes with id's specifically.

magicalbanana:
Where are your unit tests? Also, add test coverage (simplecov) to your submission. I want to see the full coverage of your app.

overlord/overlord.rb
```diff
@@ -1,14 +1,66 @@
...
+    @bomb.enter_code(params["actcode"])
+    if @bomb.state == :active
+      redirect '/'
+    else
+      @error = "You failed to activate the bomb."
+      erb :inactive
+    end
+  end
+
+  post '/deactivate' do
```
magicalbanana: Do you think you'll be able to make a class to handle this instead of redirecting each time it is incorrect and doing the count?

But I think it'll also involve some JS which if you're not comfortable it's ok. But I would love to see this logic be in a class or something the very least.

overlord/features/active_deactivate.feature
```diff
@@ -0,0 +1,31 @@
...
+		Then I should see "The bomb is initialized with activation and deactivation codes, but is not active. Activate it with the correct code:"
+
+	Scenario: User enters incorrect deactivation code for an active bomb
+		Given the bomb has been initialized with deactivation code "4444"
+		And the bomb has been activated
+		And I go to the homepage
+		When I fill in "1111" as the deactivation code
+		And I click the "Deactivate" link
+		Then I should see "The bomb is active"
+		And I should see "You have used 1 of 3 attempts at deactivating the bomb."
```
magicalbanana: Actually, for this test instead of manually specifying that you have used 1 of 3 steps try to make this more "dynamic".

overlord/features/explode.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Bomb explosion
+	As a supervillain
+	I want the bomb to explode if someone tries to disarm it incorrectly 3 times
+	So that only people who know the code can disarm it
+
+	Scenario: User enters incorrect disarm code for the third time.
```
magicalbanana: And just leave this scenario to say something like:
```
Scenario: User has one attempt left to disarm the bomb
  Given a user has already used two attempts to disarm the bomb
  When a user enters the incorrect code to disarm the bomb
  Then the user is expected to see the denotation page
```

Note I use the word expected vs should because it commands a stricter assertion.

magicalbanana:
Please address the comments and feel free to reach out to me anytime if you wish. Another note is to try and reuse some of your steps to the extend of your capability.

paulghaddad:
@cstavitsky Also, try to write your feature tests at a higher level of abstraction. This will be easier to read and make your tests more concise:

For example: 
```
Given I am on the homepage
And I activate the bomb
When I enter the deactivation code
Then the bomb should be deactivated
```

And here is a good article for reference:

Let me know if you have any questions.

PR #1109, opened 2016-01-08
Rubcop has style, who knew?

ericbrooke:
Fix the code example in the Triangle Facts exercise so that it conforms to the Ruby style guide. Which style guide who knows?

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -1,58 +1,72 @@
 # Killer facts about triangles AWW YEAH
 class Triangle
-	attr_accessor :side1,:side2,:side3
+  # attr_accessor :side1, :side2, :side3
 
-	def initialize(side1,side2,side3)
-		@side1,@side2,@side3 = side1,side2,side3
-	end
+  def initialize(side1, side2, side3)
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+  end
 
-	def equilateral()
-		return side1 == side2 && side2 == side3
-	end
+  #  equilateral - all 3 sides are equal
+  def equilateral?
+    @side1 == @side2 && @side2 == @side3
+  end
 
-	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
-	end
+  #  isosceles - 2 sides have equal lengths
+  def isosceles?
+    [@side1, @side2, @side3].uniq.length == 2
+  end
 
-	def scalene()
-		unless ! (equilateral || isosceles)
-	    false
-	  else
-	    true
-	  end
-	end
+  #  scalene - all 3 sides have different lengths
+  def scalene?
+    if equilateral? || isosceles?
+      true
+    else
+      false
+    end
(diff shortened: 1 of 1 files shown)


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,72 @@
...
+    else
+      false
+    end
+  end
+
+  def triangle_sides
+    puts 'This triangle is equilateral!' if equilateral?
+
+    puts
+    'This triangle is isosceles! Also, that word is hard to type.' if isosceles?
```
oharace: Not sure if this was done in order to make rubocop happy, but making this its own line isn't the best format. (Sorry, have to be really picky since you did so well in this exercise! haha)

oharace:
You did an excellent job in this exercise! You're ready for the next one.

triangle_facts/triangle_facts.rb
ericbrooke: @oharace You are exactly right, what is your preferred style for breaking up multi-lines?

oharace: Good question, it really depends on the app or project. (For these exercises, I think rubocop is a bit more strict). Perhaps in this case you could wrap it in an if statement, rather than having an if statement at the end. However, I wouldn't worry about it for this exercise.
One quick side note is that your method name there isn't the clearest. The original method name 'recite_facts' is more descriptive of what the method does than 'triangle_sides'. Maybe an even better name would be 'print_triangle_description' or something along that nature. Just some extra food for thought!

ericbrooke: Good points @oharace thanks

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,72 @@
...
 
+    puts 'This triangle is scalene and mathematically boring.' if scalene?
+  end
+
+  def triangle_angles
+    angles = calculate_angles(@side1, @side2, @side3)
+    puts 'The angles of this triangle are ' + angles.join(',')
+
+    puts 'This triangle is also a right triangle!' if angles.include? 90
```
paulghaddad: While omitting parentheses around the `90`, I'd recommend being consistent throughout the file.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,72 @@
...
+
+    puts
+    'This triangle is isosceles! Also, that word is hard to type.' if isosceles?
 
+    puts 'This triangle is scalene and mathematically boring.' if scalene?
+  end
+
+  def triangle_angles
+    angles = calculate_angles(@side1, @side2, @side3)
+    puts 'The angles of this triangle are ' + angles.join(',')
```
paulghaddad: You could also use string interpolation here.

paulghaddad:
@ericbrooke Nice work on this exercise! Very minor comments. For future exercises, I'd recommend making methods `private` if they aren't calling from outside the class.

PR #1110, opened 2016-01-08
Arrowheads are bad for you

ericbrooke:
Look at the code in the Arrowhead exercise. It currently fails to put error conditions near their conditions. Refactor it so that it isn't terrible anymore.

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -1,4 +1,8 @@
 class Arrowhead
+  # Error Messages
+  BAD_REGION = "Unknown region, please provide a valid region."
+  BAD_SHAPE = "Unknown shape value. please provide a valid shape"
+
   # This seriously belongs in a database.
   CLASSIFICATIONS = {
     far_west: {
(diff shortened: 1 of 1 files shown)


arrowhead/arrowhead.rb
```diff
@@ -1,4 +1,8 @@
 class Arrowhead
+  # Error Messages
+  BAD_REGION = "Unknown region, please provide a valid region."
```
oharace: Good idea setting these error messages aside.

arrowhead/arrowhead.rb
```diff
@@ -15,19 +19,20 @@ class Arrowhead
     },
   }
 
-  # FIXME: I don't have time to deal with this.
+  def self.valid_region?(region)
```
oharace: These methods are simple and to-the-point. Awesome!

arrowhead/arrowhead.rb
```diff
@@ -15,19 +19,20 @@ class Arrowhead
...
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+    return raise BAD_REGION unless valid_region?(region)
```
oharace: I like the way you simplified this method. Much more readable.

oharace:
Excellent work! You're ready for the next exercise.

ericbrooke:
@oharace Thanks :-)

arrowhead/arrowhead.rb
paulghaddad: You don't need the `return` if you're raising an exception.

arrowhead/arrowhead.rb
```diff
@@ -15,19 +19,20 @@ class Arrowhead
...
+    shapes = CLASSIFICATIONS[region]
```
paulghaddad: Consider extracting this as a method.

paulghaddad:
@ericbrooke Nice work on this! Feel free to move on!

PR #1111, opened 2016-01-11
added string sub example file

jcurlee:


paulghaddad:
@jcurlee You're good!

PR #1112, opened 2016-01-11
code clean up

jcurlee:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,70 @@
...
+  def equilateral?
+    side1 == side2 && side2 == side3
+  end
+
+  def isosceles?
+    [side1, side2, side3].uniq.length == 2
+  end
+
+  def scalene?
+    if !(equilateral? || isosceles?)
```
oharace: What if, instead we simplified it to `!(equilateral? || isosceles?)`? That would make it even better!

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,70 @@
...
+        'This triangle is isosceles! Also, that word is hard to type.'
+      when scalene?
+        'This triangle is scalene and mathematically boring.'
+    end
+  end
+
+  def recite_angle_facts
+    angles = calculate_angles(side1, side2, side3)
 
+    result = "\nThe angles of this triangle are " + angles.join(',')
```
oharace: Using `puts` statements seems better here in my opinion. First, you can just call `puts` at a few places where you need to print a string out. Second, `puts` takes care of newline characters, so you don't need to worry about including them in your strings.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,70 @@
...
 
-triangles = [
-	[5,5,5],
-	[5,12,13],
-]
-triangles.each { |sides|
-	tri = Triangle.new(*sides)
-	tri.recite_facts
-}
+    if angles.include? 90
```
oharace: In most circumstances, you should put parameters in parentheses. There are exceptions (like `puts`), but most of the time you should include the parentheses.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,70 @@
...
+  def radians_to_degrees(rads)
+    (rads * 180 / Math::PI).round
+  end
+end
+
+[
+  [5, 5, 5],
+  [5, 12, 13],
+].each do |sides|
+  tri = Triangle.new(*sides)
```
oharace: That's neat, I never though of using the splat character here to split the sides into separate arguments. Cool.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,70 @@
...
+
+  def scalene?
+    if !(equilateral? || isosceles?)
+      true
+    else
+      false
+    end
+  end
+
+  def recite_facts
```
oharace: In the method below this one, you're specific, but here is pretty generic. Maybe a better name would be `recite_triangle_type`.

oharace:
Nice work! You did a great job simplifying this code.

paulghaddad:
@jcurlee Nice work! You're good to move on.

PR #1113, opened 2016-01-11
refactored arrowhead code

jcurlee:


arrowhead/arrowhead.rb
```diff
@@ -15,19 +15,20 @@ class Arrowhead
...
       raise "Unknown region, please provide a valid region."
     end
+
+    CLASSIFICATIONS[region]
+  end
+
+  def self.classify(region, shape)
+    shapes = get_regional_shapes(region)
+
+    raise "Unknown shape value." unless shapes.include? shape
```
oharace: Perhaps this should be extracted to another method, since this doesn't actually classify the shape. It can be argued either way. Maybe to a method like `valid_shape?` or something.

oharace:
Your simplification looks great! Nice work. :+1:

paulghaddad:
@jcurlee Nice work! Feel free to move on.

PR #1114, opened 2016-01-11
Robot Name - 1

ericbrooke:
removed the Procs to make simpler
moved to instance variables from local
added more examples to make testing easier

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -39,7 +39,7 @@ class Robot
 
   def incorrect_name_format?
     fail NameCollisionError,
-      'NO: 2 CHARs and 3 Digits' if !(@name =~ /[[:alpha:]]{2}[[:digit:]]{3}/)
+      'NO: 2 CHARs and 3 Digits' unless @name =~ /[[:alpha:]]{2}[[:digit:]]{3}/
   end
 
   def generate_char
@@ -51,7 +51,8 @@ class Robot
   end
 
   def create_new_name
-    @name = "#{generate_char}#{generate_char}#{generate_num}#{generate_num}#{generate_num}"
+    @name = "#{generate_char}#{generate_char}#{generate_num}#{generate_num}" \
+      "#{generate_num}"
   end
 end
 
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -6,10 +6,13 @@ end
 class Robot
   attr_accessor :name
 
-  @@registry
+  @registry ||= []
+
+  class << self
+    attr_accessor :registry
+  end
 
   def initialize(args = {})
-    @@registry ||= []
     @name_generator = args[:name_generator]
     create_or_allocate_name
     check_and_register_name
@@ -27,14 +30,14 @@ class Robot
     if name_alreadly_used? || incorrect_name_format?
       return
     else
-      @@registry << @name
+      Robot.registry << @name
       puts @name + " Registered"
     end
   end
 
   def name_alreadly_used?
     fail NameCollisionError,
-      'NO: Name Alreadly Used' if @@registry.include?(@name)
+      'NO: Name Alreadly Used' if Robot.registry.include?(@name)
   end
 
   def incorrect_name_format?
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -1,5 +1,8 @@
-class NameCollisionError < RuntimeError; end
+class NameCollisionError < RuntimeError
+end
 
+# A Random name will be created if you give no gudiance
+# To allocate a name give an arg (name_generator: 'ted')
 class Robot
   attr_accessor :name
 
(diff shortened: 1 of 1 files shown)


robot_name/robot_name.rb
```diff
@@ -1,32 +1,75 @@
...
   attr_accessor :name
 
-  @@registry
+  @registry ||= []
+
+  class << self
+    attr_accessor :registry
+  end
 
   def initialize(args = {})
```
oharace: Since we know there will be a max of one argument, maybe we can be more specific, like `def initialize(name_generator = nil)`, which will set it to nil if we don't pass anything.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,75 @@
...
-    @@registry ||= []
     @name_generator = args[:name_generator]
+    create_or_allocate_name
+    check_and_register_name
+  end
 
+  def create_or_allocate_name
     if @name_generator
```
oharace: This can be simplified to `@name = @name_generator || create_new_name` (and then you'd have to change `create_new_name` to pass back a name instead)

robot_name/robot_name.rb
```diff
@@ -1,32 +1,75 @@
...
+      @name = @name_generator
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+      create_new_name
+    end
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def check_and_register_name
```
oharace: Even though this is a simple program, it would be better to split this into separate methods. A method should only perform one task. So let's have one method to check the name (`valid_name?` or something), and one method to register the name if it's valid (`register_name`).

oharace:
Good job! :+1: Keep these comments in mind as you progress through the courses.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,75 @@
...
 
+# A Random name will be created if you give no gudiance
+# To allocate a name give an arg (name_generator: 'ted')
 class Robot
   attr_accessor :name
 
-  @@registry
+  @registry ||= []
+
+  class << self
```
paulghaddad: Prefer `self.method_name` for class methods.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,75 @@
...
+      'NO: Name Alreadly Used' if Robot.registry.include?(@name)
+  end
+
+  def incorrect_name_format?
+    fail NameCollisionError,
+      'NO: 2 CHARs and 3 Digits' unless @name =~ /[[:alpha:]]{2}[[:digit:]]{3}/
+  end
+
+  def generate_char
+    ('A'..'Z').to_a.sample
```
paulghaddad: You can also use `[*'A'..'Z'].sample`.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,75 @@
...
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def name_alreadly_used?
+    fail NameCollisionError,
+      'NO: Name Alreadly Used' if Robot.registry.include?(@name)
+  end
+
+  def incorrect_name_format?
+    fail NameCollisionError,
+      'NO: 2 CHARs and 3 Digits' unless @name =~ /[[:alpha:]]{2}[[:digit:]]{3}/
```
paulghaddad: Consider extracting this regex as a well-named constant.

paulghaddad:
@ericbrooke Nice first iteration. Refactor and you should be good to go on the second iteration.

robot_name/robot_name.rb
ericbrooke: Can you give an example?

robot_name/robot_name.rb
ericbrooke: nice

robot_name/robot_name.rb
ericbrooke: Good point

robot_name/robot_name.rb
paulghaddad: In general, for class methods, `self.method` is preferred over using the `class << self` syntax.

ericbrooke: I get that its just that ```self.attr_accessor :registry``` does not function

paulghaddad: Gotcha! Learned something new myself! :)

robot_name/robot_name.rb
```diff
@@ -1,32 +1,75 @@
...
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def name_alreadly_used?
+    fail NameCollisionError,
+      'NO: Name Alreadly Used' if Robot.registry.include?(@name)
+  end
+
+  def incorrect_name_format?
```
ericbrooke: Thanks

robot_name/robot_name.rb
djkotowski: It'd be nice if there were a `class_attr_accessor` method to skip around this.  Maybe Ruby 2.4.

PR #1115, opened 2016-01-11
Robot name

jcurlee:


jcurlee:
need to fix some things pointed out by hound... closing until that is complete.

PR #1116, opened 2016-01-11
Robot name

jcurlee:


robot_name/robot_name.rb
```diff
@@ -7,26 +7,44 @@ class Robot
 
   def initialize(args = {})
     @@registry ||= []
-    @name_generator = args[:name_generator]
+    @name = generate_name(args[:name])
```
djkotowski: What if I want to add a custom name generator?

robot_name/robot_name.rb
```diff
@@ -7,26 +7,44 @@ class Robot
...
+    @@registry.include?(robot_name) ? false : true
+  end
+
+  def generate_char
+    ('A'..'Z').to_a.sample(2).join
+  end
+
+  def generate_num
+    prng = Random.new
+    prng.rand(100..999)
```
djkotowski: This doesn't have the same effect, as it doesn't support numbers 000-099.

robot_name/robot_name.rb
```diff
@@ -7,26 +7,44 @@ class Robot
...
+      robot_name
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+      generate_name
+    end
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+  def valid_name?(robot_name)
```
djkotowski: Predicate methods shouldn't raise errors.  They should just respond `true` or `false`.

robot_name/robot_name.rb
```diff
@@ -7,26 +7,44 @@ class Robot
 
   def initialize(args = {})
     @@registry ||= []
```
djkotowski: Avoid the use of class variables.  Prefer [class instance variables](http://stackoverflow.com/questions/3802540/difference-between-class-variables-and-class-instance-variables) (which are separate from object instance variables).

robot_name/robot_name.rb
```diff
@@ -7,26 +7,44 @@ class Robot
...
 
-    if @name_generator
-      @name = @name_generator.call
+    if valid_name?(robot_name)
+      @@registry << robot_name
+      robot_name
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+      generate_name
```
djkotowski: Do we want to generate a new name even if we provide a custom, invalid one?

djkotowski:
Nice iteration.  Take a look at the comments and refactor where necessary.  Let me know if you have any questions.

robot_name/robot_name.rb
jcurlee: @djkotowski Won't this negate the purpose of registry (each robot instance would have its own unique registry and checking against it would fail)? Would it be better to handle registry as a separate class?

djkotowski: You could do a separate class, but a class instance variable would be class instance-scoped, not object-scoped.  Take a look at the link above to see what I mean.

PR #1120, opened 2016-01-11
Arrowhead exercise - Iteration 1

locker1776:
Look at the code in the Arrowhead exercise. It currently fails to put error conditions near their conditions. Refactor it so that it isn't terrible anymore.

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -15,20 +15,29 @@ class Arrowhead
     },
   }
 
-  # FIXME: I don't have time to deal with this.
-  def self.classify(region, shape)
-    if CLASSIFICATIONS.include? region
-      shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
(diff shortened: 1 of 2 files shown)


arrowhead/arrowhead.rb
```diff
@@ -15,20 +15,29 @@ class Arrowhead
...
-    else
+  def self.region_lookup(region)
+    unless CLASSIFICATIONS.include? region
       raise "Unknown region, please provide a valid region."
     end
+    CLASSIFICATIONS[region]
+  end
+
+  def self.shape_lookup(region_info, shape)
+    unless region_info.include? shape
```
oharace: Can't remember the character limit for rubocop for these exercises.. but maybe this would be more simplified:
```
return region_info[shape] if region_info.include? shape
raise "Unknown shape value. Are you sure you know what you're talking about?"
```
Not super important, but does cut down the line count from 6 to 2.

oharace:
Awesome job! You made the code much easier to read. Good work. :+1:

paulghaddad:
@locker1776 Nice work on this iteration. Take a look at Chris's feedback and submit a second iteration.

PR #1121, opened 2016-01-12
Robot name exercise - Iteration 1

locker1776:


robot_name/robot_name.rb
```diff
@@ -1,32 +1,50 @@
...
 class Robot
+  @created_names = []
   attr_accessor :name
 
-  @@registry
+  class << self
+    attr_accessor :created_names
+  end
 
   def initialize(args = {})
```
oharace: Since we know there will be a maximum of one argument, and we know what the argument would be, we can make this more specific, like `def initialize(name_generator = nil)`. Then, you can change the method `create_name`.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,50 @@
...
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+      @name = name_generator.call
     end
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def random_name
+    generated_name = ""
+    2.times { generated_name << ('A'..'Z').to_a.sample }
```
oharace: Nice use of `2.times` and `3.times`!

robot_name/robot_name.rb
```diff
@@ -1,32 +1,50 @@
...
 end
 
-robot = Robot.new
-puts "My pet robot's name is #{robot.name}, but we usually call him sparky."
+robot_1 = Robot.new
+puts "My pet robot's name is #{robot_1.name}, but we usually call him sparky."
+robot_2 = Robot.new
+puts "My pet robot's name is #{robot_2.name}, but we usually call him riley."
+
+generator = -> { 'AA111' }
```
oharace: What if instead we just pass in a name, rather than a whole generator?

file_for_git_exercise.txt
```diff
@@ -0,0 +1 @@
+This is a file for the Git exercise in the just_git branch
```
oharace: Oops! Looks like you branched off your previous exercise.

oharace:
Nice work! Your solution looks good. Keep the feedback in mind as you continue progressing through the exercises!

robot_name/robot_name.rb
paulghaddad: You can try: `[*'A'..'Z'].sample`

robot_name/robot_name.rb
```diff
@@ -1,32 +1,50 @@
...
+  class << self
+    attr_accessor :created_names
+  end
 
   def initialize(args = {})
-    @@registry ||= []
-    @name_generator = args[:name_generator]
+    create_name(args[:name_generator])
+    check_for_duplicate_name
+    Robot.created_names << @name
```
paulghaddad: I'd extract `Robot.created_names` since you use it in more than one spot.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,50 @@
...
+  end
 
-    if @name_generator
-      @name = @name_generator.call
+  def create_name(name_generator = nil)
+    if name_generator.nil?
```
paulghaddad: You don't need to use the nil check as the following would work:

```
if name_generator
  @name = name_generator.call
else
  @name = random_name
end
```

robot_name/robot_name.rb
```diff
@@ -1,32 +1,50 @@
...
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def random_name
+    generated_name = ""
+    2.times { generated_name << ('A'..'Z').to_a.sample }
+    3.times { generated_name << rand(10).to_s }
+    generated_name
+  end
+
+  def check_for_duplicate_name
```
paulghaddad: I'd recommend making methods `private` if they aren't called outside the class.

paulghaddad:
@locker1776 Nice work on this iteration. Take a look at our comments and submit a second iteration.

PR #1127, opened 2016-01-14
Triangle facts exercise - Iteration 3

locker1776:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,74 @@
...
-
-	def equilateral()
-		return side1 == side2 && side2 == side3
-	end
-
-	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
-	end
-
-	def scalene()
```
mcdaniel67: Where did this method go? I noticed that in the previous PR for this branch that you had refactored it into 
```
if equilateral || isosceles
  false
else
  true
end
```

Consider utilizing triangle_type for this method or utilizing the expression used in the if statement.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,74 @@
...
+    end
+  end
 
+  def angle_facts
+    angles = calculate_angles
+    puts 'The angles of this triangle are ' << angles.join(',')
+    puts 'This triangle is also a right triangle!' if angles.include? 90
+  end
+
+  private def calculate_angles
```
mcdaniel67: I would structure the private methods differently. Look here https://github.com/bbatsov/ruby-style-guide under Classes & Modules towards the bottom of the first section for a good example of what I'm suggesting.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,74 @@
...
+    denominator = (2.0 * other_side_1 * other_side_2)
+    radians_to_degrees(Math.acos(numerator / denominator))
+  end
+
+  private def radians_to_degrees(rads)
+    (rads * 180 / Math::PI).round
+  end
+end
 
 triangles = [
```
mcdaniel67: Consider pulling everything below this out into a separate file and including `triangle_facts.rb`

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,74 @@
...
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+  end
+
+  def facts
+    puts "\nFacts of a triangle with sides #{side1}, #{side2}, #{side3}"
+    triangle_type_facts
+    angle_facts
+    puts "\n"
```
mcdaniel67: Consider putting the newline in the `angle_facts` method. This would clean up `facts` a bit more. You could probably even combine it with one of the puts statements in the method to make it really clean.

mcdaniel67:
Great Job! This submission definitely looks better than what we started with!

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,74 @@
...
+      when :isosceles then puts 'This triangle is isosceles!
+                               Also, that word is hard to type.'
+      when :scalene then puts 'This triangle is scalene and mathematically
+                               boring.'
+    end
+  end
 
+  def angle_facts
+    angles = calculate_angles
+    puts 'The angles of this triangle are ' << angles.join(',')
```
paulghaddad: Consider using `puts "The angles of this triangle are  #{angles.join(',')}"`

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,74 @@
...
+  end
+
+  def triangle_type
+    return :equilateral if equilateral?
+    return :isosceles if isosceles?
+    :scalene
+  end
+
+  def triangle_type_facts
+    case triangle_type
```
mcdaniel67: Look here https://github.com/bbatsov/ruby-style-guide under the part that says "Indent when as deep as case". This could benefit from structuring similarly to that section of this article.

triangle_facts/triangle_facts.rb
paulghaddad: Definitely need a new line after the `private` call.

paulghaddad:
@locker1776 Nice work on this iteration. Please take a look at our comments and resubmit.

PR #1128, opened 2016-01-14
Dinosaur catalog - iteration 1

jcurlee:
run from cli with `ruby main.rb`


jcurlee:
premature pull req -- cleaning up a few things

jcurlee:
I mad another change to the ui file -- removed usage of lambdas. I think this simplifies the code. The lambda usage was vestigial code. Originally I was creating menus and passing methods from UI as actions that the menus would call, later I reduced this to actions the UI would call on itself... there's no need to do this with lambda of course, but originally I liked the fact that I could have a smaller method (fewer lines). However favoring fewer lines for more complexity seemed like a bad idea.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,81 @@
...
+  def find_large
+    find_by_weight(2000, 999_999_999)
+  end
+
+  def find_small
+    find_by_weight(1, 2000)
+  end
+
+  def find_by_weight(min_weight, max_weight)
+    @dinodex.select { |row| row[:weight].to_i.between?(min_weight, max_weight) }
```
oharace: If a dinosaur weighs 2000, this will return that dinosaur if you search for large dinosaurs or if you search for small dinosaurs.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,81 @@
...
+  def dinodex_periods
+    @dinodex.map { |row| row[:period] }.uniq
+  end
+
+  def find_large
+    find_by_weight(2000, 999_999_999)
+  end
+
+  def find_small
+    find_by_weight(1, 2000)
```
oharace: What if instead of calling another method, you do the logic here? `@dinodex.select { |row| row[:weight].to_i < ONE_TON }`
You should make 2000 a constant, rather than a magic number.
Also, this change can also be made to `find_large` (so >= ONE_TON)

dino_catalog/dinodexui.rb
```diff
@@ -0,0 +1,131 @@
...
+        "(eg: walking:biped,period:jurassic)\n"),
+    }
+  end
+
+  def self.run
+    @running = true
+
+    while @running
+      show_menu_options
+      user_input = gets.chomp.downcase
```
oharace: Nice job handling upcase/downcase user input.

dino_catalog/dinodexui.rb
```diff
@@ -0,0 +1,131 @@
...
+      when :periods
+        show_dinos_by_period(user_input)
+      when :sizes
+        show_dinos_by_size(user_input)
+      when :search
+        handle_search_input(user_input)
+    end
+  end
+
+  def self.handle_home_menu(user_input)
```
oharace: Some of these methods should probably be private.

oharace:
This was a really neat solution to Dinodex. I'm really impressed with what you did. Your code is clean and easy to read. Excellent work! :+1:

jcurlee:
Thanks for the feedback. Made the suggested changes and opening a new PR

PR #1129, opened 2016-01-14
Robot name - Iteration 3

jcurlee:
modifications according to feedback in iteration 2 -- can now pass in a lambda to generate a robot name.

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -26,9 +26,7 @@ class Robot
       raise NameFormatError, "Robot name \"#{robot_name}\" is wrong format!"
     end
 
-    return true unless @@registry.include?(robot_name)
-
-    false
+    @@registry.include?(robot_name) ? false : true
   end
 
   def generate_char
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -4,7 +4,12 @@ class NameCollisionError < RuntimeError; end
 class Robot
   attr_accessor :name
 
+  def name_generator
+    @name_generator ||= -> { name_builder }
+  end
+
   def initialize(args = {})
+    @name_generator = args[:name_generator]
     @name = args[:name] ? register_name(args[:name]) : generate_name
   end
 
@@ -15,11 +20,17 @@ class Robot
   end
 
   def generate_name
-    robot_name = "#{generate_char}#{generate_num_string}"
-
-    return generate_name if name_exists?(robot_name)
+    robot_name = name_generator.call
 
     register_name(robot_name)
+  end
+
+  def name_builder
+    robot_name = "#{generate_char}#{generate_num_string}"
+
+    return name_generator.call if name_exists?(robot_name)
+
+    robot_name
   end
 
   def register_name(robot_name)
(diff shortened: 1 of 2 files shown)


robot_name/robot_name.rb
```diff
@@ -1,32 +1,70 @@
...
 
-  @@registry
+  def name_generator
+    @name_generator ||= -> { name_builder }
+  end
 
   def initialize(args = {})
-    @@registry ||= []
     @name_generator = args[:name_generator]
+    @name = args[:name] ? register_name(args[:name]) : generate_name
```
magicalbanana: This is not a bad way to initialize a value, but I wan to see (no need to make a resubmission, just reply in the comment and show me) how you would do this via setter method.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,70 @@
...
+  def name_builder
+    robot_name = "#{generate_char}#{generate_num_string}"
+
+    return name_generator.call if name_exists?(robot_name)
+
+    robot_name
+  end
+
+  def register_name(robot_name)
+    raise NameCollisionError, "Robot name \"#{robot_name}\" "\
```
magicalbanana: Instead of inheriting from `RuntimeError`, inherit from StandardError. See why it's better to inherit from [StandardError]. And here's a neat way of creating an exception class than if you want, can even automatically pass in to a logger or STDOUT (helpful when building web apps such as ours).

[StandardError]: http://blog.honeybadger.io/ruby-exception-vs-standarderror-whats-the-difference/

```
class NameCollisionError < StandardError
  def initialize(message)
    # maybe log it?
    # Logger.log(message)
    # The good thing with what I have written above is in a web app, often times the logs get polluted
    # and you want to be able to log your exceptions nicely.
    super(message) # this will still raise the exception and print it via STDOUT
  end
end
```

robot_name/robot_name.rb
```diff
@@ -1,32 +1,70 @@
...
-# Robot.new(name_generator: generator)
+    robot_name
+  end
+
+  def generate_char
+    ('A'..'Z').to_a.sample(2).join
+  end
+
+  def generate_num_string
+    nums = ""
```
magicalbanana: Try to rewrite using [reduce] which is an alias for [inject]. Just do it here on the comments.

[inject]: http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-inject
[reduce]: http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-reduce

magicalbanana:
And this maybe a little late, but I would like to see some tests on this.

paulghaddad:
@jcurlee Nice job adding specs! You are good to go on this exercise.

robot_name/robot_name.rb
jcurlee: went ahead and switched to setter attribute in code:

def name=(name)
  @name = name ? register_name(name) : generate_name
end

robot_name/robot_name.rb
jcurlee: switched to reduce (in code as well)

def generate_num_string
  (1..3).reduce("") { |nums| nums << rand(10).to_s }
end

PR #1131, opened 2016-01-19
Dino Catalog - Iteration 1

locker1776:


dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
...
+  def create_hash_from_dinos
+    array_of_hashes = []
+    @collection_of_dinos.each do |dino|
+      array_of_hashes << dino.hash_of_attributes
+    end
+    array_of_hashes
+  end
+end
+
+class Dino
```
DejaAugustine: Best practice: put each class in a separate file with a class name that corresponds to the class name (e.g. Dino would be dino.rb and DinoCollection would be in dino_collection.rb)

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
...
+    @continent = attrs[:continent]
+    @diet = attrs[:diet]
+    @weight = attrs[:weight]
+    @walking = attrs[:walking]
+    @description = attrs[:description]
+  end
+
+  def carnivore_or_herbivore
+    return "Herbivore" if @diet == "Herbivore"
+    "Carnivore"
```
DejaAugustine: This could be simplified to a single line `@diet == "Herbivore" ? @diet : "Carnivore"`

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def fancy_display
```
DejaAugustine: Consider renaming `fancy_display` to `to_s`

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
...
+    puts
+    DINO_ATTRIBUTES.each do |attribute|
+      next if send(attribute).nil?
+      puts "#{attribute}...#{send(attribute)}"
+    end
+  end
+
+  def hash_of_attributes
```
DejaAugustine: Consider renaming `hash_of_attributes` to 'to_hash`

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def carnivore_or_herbivore
+    return "Herbivore" if @diet == "Herbivore"
+    "Carnivore"
+  end
+
+  def fancy_display
+    puts
+    DINO_ATTRIBUTES.each do |attribute|
+      next if send(attribute).nil?
```
DejaAugustine: Consider stashing the results of `send(attribute)` in a variable:

```
value = send(attribute)
next if value.nil?
puts "#{attribute}...#{value}"
```

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
+DINO_ATTRIBUTES = [:name, :period, :continent, :diet, :weight, :walking,
```
DejaAugustine: Global constants can be dangerous.  Consider moving this into the `Dino` class and accessing it via `Dino::DINO_ATTRIBUTES`

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def add_many_dinos(many_dino_objects)
+    many_dino_objects.each { |dino| add_dino(dino) }
+  end
+
+  def fancy_display
+    @collection_of_dinos.each(&:fancy_display)
+  end
+
+  def create_hash_from_dinos
```
DejaAugustine: Unless there's another reason that you'd want a hash of dinos, consider moving the call to `JSON.pretty_generate` into this method and rename it `to_json`

Otherwise, consider renaming to `to_hash`

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,52 @@
...
+Use search categories :weight_greater_than and :weight_less_than for weight
+ranges.
+
+ - Output is always printed to the screen. You can choose a path for a
+ file that outputs your answers to json using --output_file
+
+- Usage:
+      dinodex.rb [options]
+where [options] are:
+EOS
```
DejaAugustine: Add a blank line after a large text block like this

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
...
+    @description = attrs[:description]
+  end
+
+  def carnivore_or_herbivore
+    return "Herbivore" if @diet == "Herbivore"
+    "Carnivore"
+  end
+
+  def fancy_display
+    puts
```
DejaAugustine: As a best practice, you should separate models (objects for storing and manipulating data) from views (objects for formatting and displaying data)

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
...
+class DinoCollection
+  attr_accessor :collection_of_dinos
+
+  def initialize(load_from_csv = false)
+    @collection_of_dinos = []
+    load_from_csv_files if load_from_csv
+  end
+
+  def load_from_csv_files
+    data_to_load = ImportData.load_and_transform
```
DejaAugustine: This line couples `DinoCollection` and `ImportData`. What about changing the initializer for `DinoCollection` to take an `ImportData` object as an argument rather than a boolean `load_from_csv`?

This will make it easier to ingest other data sources in the future (such as an XML feed from a web API)

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
...
+    @name = attrs[:name]
+    @period = attrs[:period]
+    @continent = attrs[:continent]
+    @diet = attrs[:diet]
+    @weight = attrs[:weight]
+    @walking = attrs[:walking]
+    @description = attrs[:description]
+  end
+
+  def carnivore_or_herbivore
```
DejaAugustine: Consider renaming this method to `carnivore?` or `is_carnivore?` and returning a boolean value rather than returning a string

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
...
+    return "Herbivore" if @diet == "Herbivore"
```
DejaAugustine: Since you've declared `attr_reader :diet`, you should use `diet` rather than `@diet`

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
+DINO_ATTRIBUTES = [:name, :period, :continent, :diet, :weight, :walking,
+                   :description]
+
+class DinoCollection
+  attr_accessor :collection_of_dinos
```
DejaAugustine: What is the benefit of calling this `collection_of_dinos` rather than simply `dinos`?

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def create_dino(attrs)
+    @collection_of_dinos << Dino.new(attrs)
+  end
+
+  def add_dino(dino)
+    @collection_of_dinos << dino
+  end
+
+  def add_many_dinos(many_dino_objects)
```
DejaAugustine: `add_many_dinos` is redundant.  `add_dinos` already indicates multiple

DejaAugustine: Same thing with `many_dino_objects`.  You could shorten it to `dino_objects` or `dinos` without sacrificing clarity

DejaAugustine: On an architectural note, is this method really necessary?  In order to power this method, at some point you're going to have to iterate through a number of dino objects, is it really better to aggregate them all to the object at once or to simply call `add_dino()` during the iteration?

dino_catalog/filter_dinos.rb
```diff
@@ -0,0 +1,34 @@
...
+  def self.search(dino_collection, search_attrs)
+    working_collection = dino_collection
+    search_attrs.each do |category, value|
+      category = :carnivore_or_herbivore if category == :diet
+      working_collection = find_matches(working_collection, category, value)
+    end
+    working_collection
+  end
+
+  def self.find_matches(dino_collection, category, value)
```
DejaAugustine: The way this method is written, it iterates through the collection twice (once to aggregate the `matching_dinos` and once during the `add_many_dinos` call inside `create_filtered_dino_collection`).

This could be simplified as follows:

```
matching_dinos = DinoCollection.new
dino_collection.collection_of_dinos.each do |dino|
  matching_dinos.add_dino(dino) if dino_match?(dino, category, value)
end
matching_dinos
```

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
...
+  def load_from_csv_files
+    data_to_load = ImportData.load_and_transform
+    data_to_load.each { |attrs| create_dino(attrs) }
+  end
+
+  def create_dino(attrs)
+    @collection_of_dinos << Dino.new(attrs)
+  end
+
+  def add_dino(dino)
```
DejaAugustine: Consider aliasing `add_dino` as `<<` so that you can do:

```
collection = DinoCollection.new
collection << dino
```

dino_catalog/dino_data.rb
```diff
@@ -0,0 +1,82 @@
...
+    load_from_csv_files if load_from_csv
+  end
+
+  def load_from_csv_files
+    data_to_load = ImportData.load_and_transform
+    data_to_load.each { |attrs| create_dino(attrs) }
+  end
+
+  def create_dino(attrs)
+    @collection_of_dinos << Dino.new(attrs)
```
DejaAugustine: This line couples `Dino` and `DinoCollection`.  Consider removing the entire `create_dino` method and have `ImportData` create the Dino objects directly.

dino_catalog/import.rb
```diff
@@ -0,0 +1,62 @@
...
+    attr = {}
+    DINO_ATTRIBUTES.each do |attribute|
+      method_to_use = "#{attribute}_transform"
+      attr[attribute] = send(method_to_use, dino_info)
+    end
+    attr
+  end
+
+  def self.name_transform(dino_info)
+    dino_info["NAME"] || dino_info["Genus"]
```
DejaAugustine: What about `Name` or `name` or `GENUS`?

dino_catalog/import.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def self.period_transform(dino_info)
+    dino_info["PERIOD"] || dino_info["Period"]
+  end
+
+  def self.continent_transform(dino_info)
+    dino_info["CONTINENT"] || "Africa"
```
DejaAugustine: Is "Africa" a reasonable assumption?

DejaAugustine:
On the whole, this is a good first draft.

My biggest overarching concern is flexibility. While this approach will get the immediate job done, it makes a lot of assumptions that will make it more difficult to expand or update in the future.

As you work on the next iteration, think about what you'd have to do to use multiple sources (4 or 5 CSVs) or to be able to parse lists of dinos stored in an alternative format (JSON or XML instead of CSV).  Think about what would be involved in adding support for additional fields or dealing with a source that's missing one or more fields.

paulghaddad:
@locker1776 Please take a look at Deja's comments and submit a second iteration. Nice work!

PR #1132, opened 2016-01-19
Triangle facts - Iteration 4

locker1776:


triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,67 @@
...
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+  end
+
+  def facts
+    40.times { print "-" }
+    puts "\nFacts of a triangle with sides #{side1}, #{side2}, #{side3}\n\n"
+    puts "#{triangle_type_facts}\n"
+    puts "#{angle_facts}\n"
```
mcdaniel67: It's kind of weird that the code is puts-ing functions whose only purpose is to puts stuff. I'd try to make the functions either return strings (so you can print the output here) or just have them print inside the function. See my comment in `angle_facts` about pulling the newline into that function.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,67 @@
...
-	[5,5,5],
-	[5,12,13],
-]
-triangles.each { |sides|
-	tri = Triangle.new(*sides)
-	tri.recite_facts
-}
+  def angle_facts
+    angles = calculate_angles
+    puts "The angles of this triangle are #{angles.join(',')}"
```
mcdaniel67: Consider refactoring this (and subsequent lines) to include the newline from `facts`. It would make this line look something like:
`puts "The angles of this triangle are #{angles.join(',')} \n"`
and totally clean up `facts`

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,67 @@
...
+
+  def triangle_type
+    return :equilateral if equilateral?
+    return :isosceles if isosceles?
+    :scalene
+  end
 
+  def triangle_type_facts
+    case triangle_type
+      when :equilateral then puts 'This triangle is equilateral!'
```
mcdaniel67: These lines could benefit from newlines too. The whens are also not indented as deep as the case statements, which is the preferred method in the ruby style guide.  (pulled from said guide)
  
Indent `when` as deep as `case`. This is the style established in both
  "The Ruby Programming Language" and "Programming Ruby".

  ```Ruby
  # bad
  case
    when song.name == 'Misty'
      puts 'Not again!'
    when song.duration > 120
      puts 'Too long!'
    when Time.now.hour > 21
      puts "It's too late"
    else
      song.play
  end

  # good
  case
  when song.name == 'Misty'
    puts 'Not again!'
  when song.duration > 120
    puts 'Too long!'
  when Time.now.hour > 21
    puts "It's too late"
  else
    song.play
  end
  ```

triangle_facts/triangles.rb
```diff
@@ -0,0 +1,10 @@
+require './triangle_facts.rb'
```
mcdaniel67: This looks so much better as it's own file!

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,67 @@
...
-	tri = Triangle.new(*sides)
-	tri.recite_facts
-}
+  def angle_facts
+    angles = calculate_angles
+    puts "The angles of this triangle are #{angles.join(',')}"
+    puts 'This triangle is also a right triangle!' if angles.include? 90
+  end
+
+  private
```
mcdaniel67: Private methods look great!

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,67 @@
...
-
-	def equilateral()
-		return side1 == side2 && side2 == side3
-	end
-
-	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
-	end
-
-	def scalene()
```
mcdaniel67: Still don't see a `scalene?` method.

mcdaniel67:
This looks good!

paulghaddad:
@locker1776 Nice work on this iteration. Please take a look at Kyle's comments. You don't need to resubmit, and can move on.

PR #1134, opened 2016-01-19
Arrowhead exercise - Iteration 3

locker1776:


arrowhead/arrowhead.rb
```diff
@@ -15,20 +17,24 @@ class Arrowhead
...
-      raise "Unknown region, please provide a valid region."
-    end
+  def classify(region, shape)
+    region_info = region_lookup(region)
+    arrowhead_info = shape_lookup(region_info, shape)
+    "You have a(n) '#{arrowhead_info}' arrowhead. Probably priceless."
+  end
+
+  def region_lookup(region)
+    return CLASSIFICATIONS[region] if CLASSIFICATIONS.include? region
```
oharace: Use parentheses to wrap parameters.

arrowhead/arrowhead.rb
```diff
@@ -15,20 +17,24 @@ class Arrowhead
...
+    raise "Unknown region, please provide a valid region."
+  end
+
+  def shape_lookup(region_info, shape)
+    return region_info[shape] if region_info.include? shape
```
oharace: Use parentheses to wrap parameters.

arrowhead/arrowhead.rb
```diff
@@ -1,4 +1,6 @@
-class Arrowhead
+module Arrowhead
+  module_function
```
oharace: Why this, instead of `self.classify`, etc.?

oharace:
Looks good. :+1:

arrowhead/arrowhead.rb
locker1776: Some style guides say that if you are not going to instantiate an instance of a class, then you should make it a module instead.

paulghaddad:
@locker1776 Nice work! You're good to move on.

PR #1135, opened 2016-01-19
Testing mindset written exercises

locker1776:


paulghaddad:
@locker1776 Looks good! If you could submit these as Github Gists in the future, that would be preferred. Thanks!

PR #1138, opened 2016-01-19
Dinosaur catalog - iteration 2

jcurlee:


dino_catalog/dinodexui.rb
```diff
@@ -0,0 +1,133 @@
...
+    @action_menu = LetteredMenu.new(["JSON Save", "Back", "Quit"])
+
+    clear_screen
+
+    File.open('dinobanner.txt', 'r') { |f| puts f.read }
+
+    run
+  end
+
+  private
```
jcurlee: Ah -- I knew there was a reason I removed the private modifier... these are all class methods and are automatically private... removing again.

dino_catalog/dinodexui.rb
```diff
@@ -0,0 +1,131 @@
+require_relative "menus"
+require_relative "dinosaur_catalog"
+
+class DinodexUI
```
jmtai: Note that when you start working with Rails, there is a convention that class names are written in CamelCase and filenames are written in snake_case. Rails has a class autoloader so if the filename deviates from the convention, the autoloader gets confused. In this case, `DinodexUI` would be expected to be defined in a file named `dinodex_ui.rb`.

dino_catalog/dinodexui.rb
```diff
@@ -0,0 +1,131 @@
...
+  @menu_index = :home
+  @sub_menus = {}
```
jmtai: Why not initialize these variables in the `start` method?

dino_catalog/dinodexui.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    File.open('dinobanner.txt', 'r') { |f| puts f.read }
+
+    run
+  end
+
+  def self.load_sub_menus
+    @sub_menus = {
+      home: Menu.new(%w(Bipeds Carnivores Periods Sizes Search)),
+      periods: Menu.new(@catalog.dinodex_periods),
```
jmtai: Careful when using instance variables as global variables. Although convenient, it can obscure the dependencies between methods. In this case `load_sub_menus` requires that a valid `@catalog` variable exists, although from the method signature it looks like a standalone method.

It might be clearer to have it accept a `periods` argument to make the external dependency explicit:
```ruby
def self.load_sub_menus(periods)
```
and then call it using
```ruby
load_sub_menus(@catalog.dinodex_periods)
```

We always want to try to adhere to the (Principle of Least Astonishment)[https://en.wikipedia.org/wiki/Principle_of_least_astonishment]

dino_catalog/dinodexui.rb
```diff
@@ -0,0 +1,131 @@
...
+    while @running
+      show_menu_options
+      user_input = gets.chomp.downcase
+      clear_screen
+      handle_user_input(user_input)
+    end
+  end
+
+  def self.clear_screen
+    system "clear" || "cls"
```
jmtai: This is equivalent to
```ruby
system("clear") || "cls"
```
Are you sure that's doing what you expect?

jmtai: This is equivalent to
```ruby
system("clear" || "cls")
```
Are you sure this is doing what you want? Will the statement
```ruby
"clear" || "cls"
```
ever return "cls"?

dino_catalog/dinodexui.rb
```diff
@@ -0,0 +1,131 @@
...
+  end
+
+  def self.show_menu_options
+    @sub_menus[@menu_index].show
+    @action_menu.show
+  end
+
+  def self.handle_user_input(user_input)
+    if %w(j b q).include?(user_input)
```
jmtai: You probably don't want to hardcode these options, since the list depends on the entries used to create the action_menu. Perhaps it would be better to define a constant with the menu items and their associated menu letters so you only have to maintain the menu data in a single spot?

dino_catalog/dinodexui.rb
```diff
@@ -0,0 +1,131 @@
...
+      handle_action_menu(user_input)
+    else
+      handle_sub_menu_actions(user_input)
+    end
+  end
+
+  def self.switch_menu(user_input)
+    @menu_index = :periods if user_input == 3
+    @menu_index = :sizes if user_input == 4
+    @menu_index = :search if user_input == 5
```
jmtai: Again here, you've hardcoded these indices but what if you want to change the order of the menu items? Now you have multiple spots in the code that need to be modified.

jmtai: If you defined your menu options as an array constant, you could do something like this:
```ruby
HOME_MENU_OPTS = ["Bipeds", "Carnivores", "Periods", "Sizes", "Search"].freeze

def self.switch_menu(index)
  @menu_index = HOME_MENU_OPTS[index - 1].downcase.to_sym
end
```

jcurlee:
thanks -- I made modifications based on your feedback. I especially like your home menu constant solution.

jmtai:
@jcurlee Whoops, wasn't quite done yet - I've got a few more comments coming.

jcurlee:
@jmtai cool no worries

dino_catalog/dinodex_ui.rb
```diff
@@ -0,0 +1,129 @@
...
+  end
+
+  def self.switch_menu(user_input)
+    @menu_index = HOME_MENU_OPTIONS[user_input - 1].downcase.to_sym
+  end
+
+  def self.handle_action_menu(user_input)
+    save_to_json if user_input == 'j'
+    @menu_index = :home if user_input == 'b'
+    @running = false if user_input == 'q'
```
jmtai: This is another case of global variables obscuring the operation of your app. At line 38, calling `handle_user_input` gives me no indication that there are side effects to calling the method (namely, updating internal state).

dino_catalog/dinodex_ui.rb
```diff
@@ -0,0 +1,129 @@
...
+        show_dinos_by_size(user_input)
+      when :search
+        handle_search_input(user_input)
+    end
+  end
+
+  def self.handle_home_menu(user_input)
+    user_input = user_input.to_i
+
+    if user_input > 2
```
jmtai: This is fragile code, since rearranging the menu options will break this logic.

dino_catalog/dinodex_ui.rb
```diff
@@ -0,0 +1,129 @@
...
+      switch_menu(user_input)
+    else
+      show_dinosaur_facts(@catalog.find_bipeds) if user_input == 1
+      show_dinosaur_facts(@catalog.find_carnivores) if user_input == 2
+    end
+  end
+
+  def self.show_dinos_by_period(user_input)
+    show_dinosaur_facts @catalog.find_dinos(
```
jmtai: Always wrap your method arguments in parentheses. Additionally, it might make sense to break this into several statements to make it read better, even though it adds additional lines of code:
```ruby
selected_period = @catalog.dinodex_periods[user_input.to_i - 1].downcase
dinos = @catalog.find_dinos(selected_period)
show_dinosaur_facts(dinos)
```
Compact code is nice, but readable code is better.

dino_catalog/dinodex_ui.rb
```diff
@@ -0,0 +1,129 @@
...
+      :period, @catalog.dinodex_periods[user_input.to_i - 1].downcase
+    )
+  end
+
+  def self.show_dinos_by_size(user_input)
+    dinos = user_input == "1" ? @catalog.find_large : @catalog.find_small
+
```
jmtai: When methods are small, the Ruby convention is to not add additional blank lines between statements unless there's a specific need to do so.

dino_catalog/dinodex_ui.rb
```diff
@@ -0,0 +1,129 @@
...
+    show_dinosaur_facts(dinos)
+  end
+
+  def self.handle_search_input(user_input)
+    index = nil
```
jmtai: The choice of `index` for this variable is a bit confusing, since it's pretty conventional for `index` to refer to an integer value. Naming is tricky (and sometimes subjective) but this is a case where a naming choice makes it hard to understand what the code is actually doing. A better choice might be `dinos` or something similar.

dino_catalog/dinodex_ui.rb
```diff
@@ -0,0 +1,129 @@
...
+    user_input.split(',').each do |search|
+      facet = search.split(':')
+      index = @catalog.find_dinos(facet[0].to_sym, facet[1], index)
```
jmtai: Again here, it might be clearer to use temp variables to make it clear what `facet` contains:
```
key = facet[0].to_sym,
value = facet[1]
filtered_dinos = @catalog.find_dinos(key, value, filtered_dinos)
```
Although it adds a couple of extra lines of code, it makes it very obvious what is being passed to the find method.

dino_catalog/dinodex_ui.rb
```diff
@@ -0,0 +1,129 @@
...
+    end
+    show_dinosaur_facts(index)
+  end
+
+  def self.show_dinosaur_facts(results)
```
jmtai: We want our Ruby classes to follow the Single Responsibility Principle. That is, each class should be responsible for one thing, and one thing only. Here your UI class is responsible for a number of different things: user input, JSON export, and console output.

dino_catalog/dinodex_ui.rb
```diff
@@ -0,0 +1,129 @@
...
+  end
+
+  def self.show_dinos_by_period(user_input)
+    show_dinosaur_facts @catalog.find_dinos(
+      :period, @catalog.dinodex_periods[user_input.to_i - 1].downcase
+    )
+  end
+
+  def self.show_dinos_by_size(user_input)
+    dinos = user_input == "1" ? @catalog.find_large : @catalog.find_small
```
jmtai: Use parentheses here to visually separate the ternary operator from the variable assignment:
`dinos = (user_input == "1") ? @catalog.find_large : @catalog.find_small`

dino_catalog/dinodex_ui.rb
```diff
@@ -0,0 +1,129 @@
...
+      facet = search.split(':')
+      index = @catalog.find_dinos(facet[0].to_sym, facet[1], index)
+    end
+    show_dinosaur_facts(index)
+  end
+
+  def self.show_dinosaur_facts(results)
+    results.each do |dino|
+      @catalog.default_keys.each do |key|
+        sym_key = key.to_sym
```
jmtai: It looks like `default_keys` are only used here, so why not define them as symbols in dinosaur_catalog and avoid the conversion here?

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,87 @@
+require "CSV"
+require "JSON"
+
+class DinosaurCatalog
+  ONE_TON = 2000
```
jmtai: Good use of a constant here.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  attr_accessor :json_file_name
+  attr_accessor :default_keys
```
jmtai: Do you expect default keys to be modified outside of this class? If not, there's not need to define an accessor for it.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def default_keys
+    @default_keys = %w(name period continent diet walking weight description)
```
jmtai: Default implies non-modifiable, so perhaps this should be declared as a constant, instead?

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  def initialize(attrs = {})
+    @dinodex = []
+
```
jmtai: No need for these extra blank lines.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,87 @@
...
+    @json_file_name = 'dinodex.json'
+
+    import_csv_files(attrs[:csv_files])
+
+    parse_keys
+  end
+
+  def dinodex_periods
```
jmtai: Here you're using `dinodex`, but the class name is `DinosaurCatalog`. Try to keep your naming consistent.

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def find_small
+    find_by_weight('small')
+  end
+
+  def find_by_weight(size)
+    @dinodex.select do |row|
+      weight = row[:weight].to_i
+
+      size == 'large' ? weight >= ONE_TON : weight < ONE_TON
```
jmtai: Make sure you get the requirements right when implementing functionality:
From the Requirements - `Grab only big (> 2 tons) or small dinosaurs.`

jmtai: What if instead of working with raw row data, you had a `Dinosaur` class. A dinosaur object could then tell you if it was large or small and your find function could be something like this:
```ruby
def find_large
  @dinodex.select { |dino| dino.large? }
end
```
or the even more compact form
```ruby
def find_large
  @dinodex.select(&:large?)
end
```

dino_catalog/dinosaur_catalog.rb
```diff
@@ -0,0 +1,87 @@
...
+    find_dinos(:period, time_period)
+  end
+
+  def find_dinos(key, value, catalog = nil)
+    current_catalog = catalog ? catalog : @dinodex
+    current_catalog.select { |row| row[key] && row[key].downcase == value }
+  end
+
+  def export_to_json
+    File.open(@json_file_name, "w") { |f| f.write JSON.dump @dinodex }
```
jmtai: Wrap your method parameters in parentheses:
```ruby
File.open(@json_file_name, "w") { |f| f.write(JSON.dump(@dinodex)) }
```
Although in this case, it would probably be more readable to use the do-end format.
```ruby
File.open(@json_file_name, "w") do |f|
  json = JSON.dump(@dinodex)
  f.write(json)
end
```

dino_catalog/menus.rb
```diff
@@ -0,0 +1,44 @@
...
+  def show_options
+    menu_text = ""
+    @menu_options.each_with_index do |option, index|
+      menu_text += "#{index + 1}) #{option} "
+    end
+    puts menu_text
+  end
+end
+
+class LetteredMenu < Menu
```
jmtai: Put each class in a separate file. Even though these classes are very small, sticking to the one class-per-file guideline makes it easier to search for classes when the codebase gets very large.

dino_catalog/menus.rb
```diff
@@ -0,0 +1,44 @@
...
+  def initialize(attrs = {})
+    super(attrs)
+    hashed_menu_options = {}
+    @menu_options.each do |option|
+      hashed_menu_options[option[0].upcase] = option
+    end
+    @menu_options = hashed_menu_options
```
jmtai: Whenever you see the pattern:
```
init output_object
iterate over some array, updating output_object
use output_object
```
It means you can use `each_with_object` to make the code more compact.

jmtai:
Nice job! The implementation is clean and easy to understand. For your next iteration, I'd suggest implementing a `Dinosaur` class and see if that makes your implementation even cleaner and more readable. Additionally, try refactoring the `DinodexUI` class to adhere more closely to the Single Responsibility Principle.

Feel free to come talk to me if you have any questions.

paulghaddad:
@jcurlee Nice work! Please take a look at Josh's comments, and submit one more iteration.

PR #1140, opened 2016-01-19
Dino Catalog

ericbrooke:


dino_catalog/dino_store.rb
```diff
@@ -0,0 +1,34 @@
+$LOAD_PATH << '.'
+
+require 'dinosaur'
+require 'import'
+require 'catalog'
+require 'export'
+
+class Dinodex
+  @dinosaurs_catalog = Catalog.new
```
zsyed1: Why not name this `@catalog`

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,45 @@
...
+    attr.each { |k, v| instance_variable_set("@#{k}", v) }
+  end
+
+  def carnivore?
+    carnivorous = %w(Carnivore Insectivore Piscivore)
+    carnivorous.include?(@diet)
+  end
+
+  def what_size
+    return "Unknown" if @weight.nil?
```
zsyed1: `return "Unknown" unless @weight`

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,45 @@
...
+    @weight >= 2000 ? "Big - Over a ton" : "Small - Under a ton"
```
zsyed1: Not sure if ternary expressions are preferred to:
```
return "Big - Over a ton" if @weight >= 2000
"Small - Under a ton"
```

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def to_hash
```
zsyed1: Use `each_with_object` or `inject` to do this:

```
def to_hash
  instance_variables.each_with_object({}) do |hash, var|
  hash[var.to_s.delete("@")] = instance_variable_get(var)
end
```

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,45 @@
...
+    record << "Name:".ljust(pad) + "#{@name}\n" unless @name == ""
+    record << "Period:".ljust(pad) + "#{@period}\n" unless @period == ""
+    record << "Continent:".ljust(pad) + "#{@continent}\n" \
+      unless @continent == ""
+    record << "Diet:".ljust(pad) + "#{@diet}\n" unless @diet == ""
+    record << "Weight:".ljust(pad) + "#{@weight}\n" unless @weight == 0
+    record << "Walking:".ljust(pad) + "#{@walking}\n" unless @walking == ""
+    record << "Description:".ljust(pad) + "#{@description}\n" \
+      unless @description == "" || @description.nil?
+    record << "-" * 50 + "\n"
```
zsyed1: Lines 33 - 41: You can DRY this up by creating a method `def display(label, feature)`
```
def display(label, feature)
  return '' if feature == "" || feature == 0
  "#{label}:".ljust(15) + "#{feature}"
end
```

Then this method becomes:
```
def to_s
  [display("Name", @name),
   display("Period", @period),
   ...
  ]..reject(&:empty?).join("\n")
```

dino_catalog/export.rb
```diff
@@ -0,0 +1,20 @@
...
+    File.open(EXPORT_FILENAME, "w") do |f|
+      f.puts JSON.generate(temp)
+      puts "Dino Stored exported to file "
+    end
+  end
+
+  def convert_to_hash(catalog)
+    temp = []
+    catalog.each { |row| temp << row.to_hash }
+    temp
```
zsyed1: Lines 16-18: Whenever you see this pattern in Ruby where you initialize something (`temp = []`) then iteratively build it, then return it, there is probably an `Enumerable` method that will allow you to remove the first and last lines.

```
def convert_to_hash(catalog)
  catalog.map { |row| row.to_hash }
end
```

or even more concisely:

```
def convert_to_hash(catalog)
  catalog.map(&:to_hash)
end
```

dino_catalog/import.rb
```diff
@@ -0,0 +1,49 @@
...
+    puts "file two complete"
+  end
+
+  def convert_carnivore_to_diet(row)
+    if row["Carnivore"] == "Yes"
+      "Carnivore"
+    else
+      ""
+    end
+  end
```
zsyed1: Lines 43-47:

```
return "Carnivore" if row["Carnivore"] == "Yes"
""
```

dino_catalog/catalog.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+    self
+  end
+
+  def search_with_hash(h)
+    p "hash search"
+    @results = nil
+    carry = []
+    h.each_pair do |key, value|
+      carry = filter(key, value)
```
zsyed1: What is the purpose of `carry` here?

dino_catalog/import.rb
```diff
@@ -0,0 +1,49 @@
+require 'csv'
+
+class Import
+  def initialize(catalog)
+    @dinosaurs = catalog.items
```
zsyed1: Mutating the member `items` of `Catalog` kind of crosses boundaries of the `Catalog` object. Instead you should probably have a method on `Catalog` say `def add(dinosaur)` instead of making the array `@items` and `attr_accessor`.

dino_catalog/catalog.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    @results = current_items.select do |x|
+      x.instance_variable_get("@#{category}").include?(keyword)
+    end
+    self
+  end
+
+  def search_with_hash(h)
+    p "hash search"
+    @results = nil
```
zsyed1: Interesting approach - using `@results` to store the "current" filter.

dino_catalog/import.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def convert_carnivore_to_diet(row)
+    if row["Carnivore"] == "Yes"
+      "Carnivore"
+    else
+      ""
+    end
+  end
+end
```
zsyed1: Nice work implementing Dinodex! I would recommend making the suggested changes and resubmitting since this task is early on in Level-Up.

dino_catalog/dino_store.rb
ericbrooke: Fair point, but I wanted it to be clearly different to separate from the class

dino_catalog/dinosaur.rb
ericbrooke: @zsyed1  Thanks for showing actual examples of what you mean :-)

dino_catalog/dinosaur.rb
ericbrooke: Nice. Thanks again for showing your examples

dino_catalog/export.rb
ericbrooke: Nice. I like you top example as it reads more clearly

dino_catalog/import.rb
ericbrooke: Fair point

dino_catalog/import.rb
ericbrooke: Really clear guidance thanks @zsyed1

paulghaddad:
@ericbrooke Nice job. Take a look at Zamir's comments, and submit a second iteration.

PR #1144, opened 2016-01-21
Shopping cart test plan - iteration 1

jcurlee:


shopping_cart/features/cart.features
```diff
@@ -0,0 +1,76 @@
+Feature: Shopping cart
```
magicalbanana: The file name is just ```*.feature```

shopping_cart/features/cart.features
```diff
@@ -0,0 +1,76 @@
...
+  In order to purchase items from a page
+  As an online shopper
+  I want the ability to add and remove items to a cart, as well as change quantity.
+
+  Scenario: Ability to add items to cart
+    Given an item has an add to cart button
+    And a user clicks the add to cart button
```
magicalbanana: You need to have ```When``` steps in your feature tests. ```When``` steps are your action steps or the key action that your test is doing.

In a good test there are three things I always want to see:

1. Your arrangement (in this case your Givens)
2. Your action (in this case your Whens)
3. Your assertions (in this case your Thens).

magicalbanana:
Couple of things I want to see when this is resubmitted.

1. Split the features into individual features.
2. Elaborate more on your tests (maybe add some examples).
3. Look into drying up some of your steps.

Here's a [reading material - Instant Cucumber BDD]  I recommend you read to get you started.
[reading material - Instant Cucumber BDD]: https://www.packtpub.com/web-development/instant-cucumber-bdd-how-instant

jcurlee:
Thanks for the feedback. Made the suggested changes to the cart features -- will read the suggested book as well.

paulghaddad:
@jcurlee This looks good! Nice work on using high-level step definitions at a consistent level of abstraction. You're good on this exercise!

PR #1145, opened 2016-01-21
Tautological test example - iteration 1

jcurlee:


paulghaddad:
@jcurlee Nice job on this. Please submit future write ups as a Gist, though. :)

PR #1146, opened 2016-01-25
Data Science TTD

ericbrooke:
I tried to avoid testing the actual gem but not sure if I got the balance right?


data_science/spec/understand_spec.rb
```diff
@@ -0,0 +1,67 @@
+
+require 'spec_helper'
+require_relative '../understand.rb'
+
+RSpec.describe "test" do
+  before :all do
+    file = File.read(FILETOIMPORT)
+    @data_hash = JSON.parse(file)
```
mcdaniel67: You might want to consider refactoring these instance variables into let statements. 
http://betterspecs.org/#let

data_science/spec/understand_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    DataScience.raw_results
+    DataScience.significant_difference?
+  end
+
+  it "should have a class import" do
+    expect(DataScience).to be_kind_of(Class)
+  end
+
+  it "should have correct filename" do
+    expect(FILETOIMPORT).to eq "data_export_2014_06_20_15_59_02.json"
```
mcdaniel67: I would put parentheses around what the expected value is as well. This would keep consistency among `be_kind_of`, `include`, `match`, `eq`, etc.

dino_catalog/catalog.rb
```diff
@@ -0,0 +1,17 @@
+class Catalog
```
mcdaniel67: Everything under this is a leftover from another exercise, right? Probably shouldn't be here...

data_science/spec/understand_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  it "should have a data structure that follows" do
+    expect(@hash_keys).to include("date")
+    expect(@hash_keys).to include("cohort")
+    expect(@hash_keys).to include("result")
+  end
+
+  it "should have cohorts that are only A or B" do
+    expect(@random_sample1["cohort"]).to match(/(A|B)/)
```
mcdaniel67: You might be able to DRY this out a bit more by making a random_sample function instead of three distinct instance variables. I.e. you could do something like this:

```ruby
3.times { expect(random_sample["cohort"]).to match(/(A|B)/) }
```

data_science/spec/understand_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    expect(@random_sample2["cohort"]).to match(/(A|B)/)
+    expect(@random_sample3["cohort"]).to match(/(A|B)/)
+  end
+
+  it "should have results that are numbers" do
+    expect(@random_sample1["result"]).to be_a(Integer)
```
mcdaniel67: Same here as L36

data_science/spec/understand_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    expect(@random_sample2["result"]).to be_a(Integer)
+    expect(@random_sample3["result"]).to be_a(Integer)
+  end
+
+  it "produces a summary of raw results" do
+    expect(DataScience.raw_results).to be_kind_of(Object)
```
mcdaniel67: I'm uncertain of the benefit of asserting the `raw_results` are a kind of object. Maybe a more restrictive check would be beneficial?

data_science/understand.rb
```diff
@@ -0,0 +1,79 @@
+# parse this {"date":"2014-03-20","cohort":"B","result":0}
+FILETOIMPORT = "data_export_2014_06_20_15_59_02.json"
```
mcdaniel67: I agree with hound that this should be frozen. Also a minor thing, but any reason you chose to avoid underscores here? Makes it a little harder to read.

data_science/understand.rb
```diff
@@ -0,0 +1,79 @@
...
+    values[:bsum] = { converted: @b_converts, not_converted: @b_nonconverts }
+    tester = ABAnalyzer::ABTest.new values
+    puts "Chi leader"
+    puts "Chi-Square Test if less then 0.05, we cannot accept null hypothesis"
+
+    puts tester.chisquare_p
+  end
+end
+
+DataScience.import
```
mcdaniel67: Consider splitting everything below this line into it's own file.

data_science/understand.rb
```diff
@@ -0,0 +1,79 @@
...
+class DataScience
+  def self.import
+    file = File.read(FILETOIMPORT)
+    @data_hash = JSON.parse(file)
+    @sample = @data_hash.length
+    puts "loaded #{@sample} sample\n\n"
+  end
+
+  def self.raw_results
+    @a_converts = 0
```
mcdaniel67: Using `attr_accessor`, or `attr_reader` might make this a little more testable as well as a little cleaner/legible.

data_science/understand.rb
```diff
@@ -0,0 +1,79 @@
+# parse this {"date":"2014-03-20","cohort":"B","result":0}
+FILETOIMPORT = "data_export_2014_06_20_15_59_02.json"
+require 'json'
+require 'abanalyzer'
+
+class DataScience
+  def self.import
+    file = File.read(FILETOIMPORT)
```
mcdaniel67: Might be beneficial to make this function take the file to import as an argument. This would allow you to create a specific dataset to run your tests on. It would also give you the opportunity to remove the constant when you split this file. 

This would also generate some more tests you could do (i.e. file does not exist, file of a bad format, etc)

data_science/understand.rb
```diff
@@ -0,0 +1,79 @@
...
+        @a_count += 1
+      else
+        @b_converts += k["result"].to_i
+        @b_count += 1
+      end
+      @a_nonconverts = @a_count - @a_converts
+      @b_nonconverts = @b_count - @b_converts
+    end
+
+    puts "A conversion: #{@a_converts} non converts: " +
```
mcdaniel67: Consider factoring the printing out to it's own function. That would simplify this function greatly.

data_science/understand.rb
```diff
@@ -0,0 +1,79 @@
...
+    puts "B conversion: #{@b_converts} non converts: " +
+      "#{@b_nonconverts} over #{@b_count} days"
+    puts "Total Sample: #{@sample}\n\n"
+  end
+
+  def self.significant_difference?
+    values = {}
+    values[:asum] = { converted: @a_converts, not_converted: @a_nonconverts }
+    values[:bsum] = { converted: @b_converts, not_converted: @b_nonconverts }
+    @significant_difference = ABAnalyzer::ABTest.new values
```
mcdaniel67: You should wrap function arguments in parentheses.

data_science/understand.rb
```diff
@@ -0,0 +1,79 @@
...
+    puts "Is there a difference between the groups?"
+    puts "#{@significant_difference.different?} \n\n"
+  end
+
+  def self.confidence_level
+    p = 0.95
```
mcdaniel67: Might be worth refactoring this into a constant, or make it a default function argument.

data_science/understand.rb
```diff
@@ -0,0 +1,79 @@
...
+      @b_converts, @b_nonconverts, p)
+
+    puts "Conversion range from #{@b_conf_rang[0]} to #{@b_conf_rang[1]}\n\n"
+  end
+
+  def self.chi_leader
+    values = {}
+    values[:asum] = { converted: @a_converts, not_converted: @a_nonconverts }
+    values[:bsum] = { converted: @b_converts, not_converted: @b_nonconverts }
+    tester = ABAnalyzer::ABTest.new values
```
mcdaniel67: Parentheses here too.

mcdaniel67:
Overall this is a great first submission!! One thing I would strongly suggest is creating a sample dataset to run the tests on. This would allow you to do some more clever assertions on data that you have absolute control over.

data_science/understand.rb
```diff
@@ -0,0 +1,79 @@
...
+    @significant_difference = ABAnalyzer::ABTest.new values
+    puts "Is there a difference between the groups?"
+    puts "#{@significant_difference.different?} \n\n"
+  end
+
+  def self.confidence_level
+    p = 0.95
+    puts "Conversion rates with #{p} certain"
+    puts "Group A"
+    @a_conf_rang = ABAnalyzer.confidence_interval(
```
mcdaniel67: There might be a cleaner way to break this potentially long line up.

ericbrooke:
@mcdaniel67 Thanks for your time and comments :-)

PR #1148, opened 2016-01-25
Overlord - Iteration 2

cstavitsky:


oharace:
Some of these rubocop comments are valid.

overlord/features/active_deactivate.feature
```diff
@@ -0,0 +1,31 @@
+Feature: Disarm active bomb
+	As a supervillain
+	I want to be able to disarm the bomb when activated
+	So that I can stay alive if I accidentally activate it
+
+	Scenario: User views disarm interface for an active bomb
+		Given The bomb has been initialized
+		And the bomb has been activated
```
oharace: What's the difference between being initialized and being activated? These two lines could probably be combined into one line `Given the bomb has been activated <with code if needed>`, and then the step definition will take care it.

overlord/features/active_deactivate.feature
```diff
@@ -0,0 +1,31 @@
...
+		And I go to the homepage
+		Then I should see "The bomb is active"
+		And I should see a "Deactivate" link
+		And I should see a "#bomb-deactivation-code" submit button
```
oharace: This is too specific. Simplify it to `And I should see a submit button` or something like that, and then let the corresponding step definition be specific. The step definition can check for the element with that ID.

overlord/features/active_deactivate.feature
```diff
@@ -0,0 +1,31 @@
...
+
+	Scenario: User enters correct deactivation code for an active bomb
+		Given the bomb has been initialized w deactivation code "4444"
+		And the bomb has been activated
+		And I go to the homepage
+		When I fill in "4444" as the deactivation code
```
oharace: Being specific about the code is OK in this circumstance.

overlord/features/active_deactivate.feature
```diff
@@ -0,0 +1,31 @@
...
+		And I click the "Deactivate" link
+		Then I should see "The bomb is initialized with activation and deactivation codes, but is not active. Activate it with the correct code:"
```
oharace: This is OK too. At least for NetCredit, we sometimes do this to test error/flash messages. That way we can use a single step definition to test what messages we see (the step definition takes the message as a parameter to look for).

overlord/features/active_deactivate.feature
```diff
@@ -0,0 +1,31 @@
+Feature: Disarm active bomb
+	As a supervillain
```
oharace: I'm pretty sure I brought this up in a previous pull request, but you'll need to fix your indentation. In your text editor, change the number of spaces per tab/indent from 4 to 2.

overlord/features/initialize.feature
```diff
@@ -0,0 +1,31 @@
...
+		And I should see "Mwahahaha!"
+		And I should see a "Bomb activation code" input box
+		And I should see a "Bomb deactivation code" input box
+		And I should see an "#initialize" submit button
+
+	@javascript
+	Scenario Outline: User initializes (but does not activate) bomb
+		Given The bomb has not been initialized
+		And I go to the homepage
+		When I fill in "<activation>" as the activation code
```
oharace: Do you need quotes around `<activation>`?

overlord/features/initialize.feature
```diff
@@ -0,0 +1,31 @@
...
+		And I fill in "<deactivation>" as the deactivation code
+		And I click the "Initialize" link
+		And I confirm the pop-up dialog
```
oharace: This should be combined into the step definition for clicking the "Initialize" link. No need for a feature line to do this.

overlord/features/initialize.feature
```diff
@@ -0,0 +1,31 @@
...
+		Then I should see "The bomb is initialized with activation and deactivation codes, but is not active. Activate it with the correct code:"
+		And I should see a "Bomb activation code" input box
+
+		Scenarios:
+			| activation      | deactivation        |
+			| 1234            |  N/A   |
```
oharace: Instead of `N/A`, you could just leave it blank, right? Also, make sure you space out the pipes nicely so the table lines all match up.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,40 @@
+Given(/^The bomb has not been initialized$/) do
+  @bomb = nil
```
oharace: Interacting with objects directly like this is usually a code smell for feature tests. Unit tests (rspec) are meant to do this kind of work. Feature tests (cucumber) are integration tests, so they should only interact with the webpage. You can check the values of objects and such in step definitions, but you shouldn't modify them. These tests are meant to verify that everything on the front-end works as expected. If we set/modify things in here, we can't be sure that everything is working as expected.

overlord/features/step_definitions/bomb_steps.rb
```diff
@@ -0,0 +1,40 @@
...
+end
+
+Given(/^I enter an incorrect code three times$/) do
+  visit('/')
+  fill_in('bomb-deactivation-code', with: "wrong")
+  page.find('#deactivate').click
+  fill_in('bomb-deactivation-code', with: "wrong")
+  page.find('#deactivate').click
+  fill_in('bomb-deactivation-code', with: "wrong")
+  page.find('#deactivate').click
```
oharace: Good! You could also use ruby to simplify it:
```
3.times do
  fill_in('bomb-deactivation-code', with: "wrong")
  page.find('#deactivate').click
end
```

overlord/features/step_definitions/general_steps.rb
```diff
@@ -0,0 +1,71 @@
...
+Given(/^I go to the bomb active page$/) do
+  visit("/active_bomb")
+end
+
+When(/^I click the "(.*?)" link$/) do |link|
+  click_on(link)
+end
+
+When(/^I fill in "(.*?)" as the activation code$/) do |value|
+  page.fill_in 'bomb-activation-code', with: value unless value == "N/A"
```
oharace: hmm, instead of 'N/A', I think this would perform as you expect if you just pass in an empty string (see my comment in the other file)

overlord/features/step_definitions/general_steps.rb
```diff
@@ -0,0 +1,71 @@
...
+
+Then(/^I should see a[n]? "(.*?)" input box$/) do |box_name|
+  require 'pry'
+  if CSS_MATCHERS[box_name]
+    expect(page).to have_css("input#{CSS_MATCHERS[box_name]}")
+  elsif CSS_MATCHERS.keys.include?(box_name)
+    raise StandardError, "This element does not exist on the page"
+  else
+    raise StandardError, "Your tests don't contain a match for '#{box_name}'"
+  end
```
oharace: I'm unsure what all you're trying to do here. I don't think you need to raise errors, as Capybara/Cucumber will do that for you when you run your tests. In addition, I'm sure there is a simpler way to find what you're looking for. If you're looking for an input box with a certain ID, then just search for an element with that ID. You can also be specific, like `input#ID` or `find('input[name=foo]')`

overlord/views/inactive.html.erb
```diff
@@ -0,0 +1,7 @@
+<h1><%= @error %></h1>
+
+	<form action="/activate" method="POST">
```
oharace: You could probably use a rails `form_tag` instead of writing a html form manually. Might be easier

oharace:
Good progress on your tests! Take a look at my comments, and ask any questions if you have them!

paulghaddad:
@cstavitsky Nice work on this iteration. For the last iteration, really strive for writing your feature tests at a higher level of abstraction. Reach out to me if you'd like to pair on this together.

PR #1149, opened 2016-01-25
Shopping Cart - Iteration 2

cstavitsky:


shopping_cart/add_another_of_item_in_cart.feature
```diff
@@ -0,0 +1,12 @@
+Feature: Add another item of the same SKU that's already in user's cart updates qty in cart
```
oharace: You should change your text editor to make tabs indent to 2 spaces. Looks like it is set to 4 or 6.

oharace: I think this scenario fits right in with our `adjust_cart_quantities

shopping_cart/adjust_cart_quantities.feature
```diff
@@ -0,0 +1,57 @@
...
+		Given that I have 1 Broomstick in my cart
+		And I am on the shopping cart page
+		When I fill in the item quantity for the first item with "3"
+		And I click the "update quantity" button
+		Then I should see 3 Broomsticks in my cart
+
+	Scenario: Updating item quantity in cart on shopping cart page doesn't affect other items in cart
+		Given that I have 1 Broomstick and 1 Wand in my cart
+		And I am on the shopping cart page
+		When I fill in the item quantity for the first item with "3"
```
oharace: Be a little more specific here - rather than 'first item', say 'broomstick'

shopping_cart/coupon_codes.feature
```diff
@@ -0,0 +1,51 @@
+Feature: add coupon code for discount
+	As a user
+	I want to add a coupon code
+	So I can save money and get the discount I was promised
+
+	Scenario: User enters a valid coupon code to save a static amount
+		Given that I have 1 Textbook in my cart
+		And the price of a Textbook is $100
+		And the "LOCKHART_FAN" coupon code is active
+		And I am on the shopping cart page
```
oharace: Because this step is a `Given` step for every scenario, you can extract it to the beginning, so it would look like:
```
Background:
  Given I am on the shopping cart page

Scenario: User enters a valid....
```

shopping_cart/coupon_codes.feature
```diff
@@ -0,0 +1,51 @@
...
+		Given that I have 1 Butterbeer in my cart
+		And the price of a Butterbeer is $15
+		And the "SAVE_10_PERCENT" coupon code is active
+		And I am on the shopping cart page
+		When I fill in the coupon code form with "SAVE_10_PERCENT"
+		And I click the "Apply coupon code" button
+		Then I should see "-1.50"
+		And I should see "You saved 10%"
+
+	Scenario: User enters an expired or invalid coupon code
```
oharace: This should probably be split into two scenarios. We would want a different message for a valid, but expired, coupon, right?

shopping_cart/shipping_calculator.feature
```diff
@@ -0,0 +1,92 @@
...
+
+	Scenario: User enters more than 30 characters in the shipping city field
+		Given that I have 1 Broomstick in my cart
+		And I am on the checkout page
+		And I fill in the shipping address field with "175 W. Jackson Blvd"
+		And I fill in the shipping city field with "Chicagooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo"
+		And I fill in the shipping state field with "IL"
+		And I fill in the shipping zipcode field with "60604"
+		When I click the "Estimate shipping" button
+		Then I should see "Please enter a shipping address under 70 characters"
```
oharace: Why would they see this message? They've satisfied the condition. (I'm guessing this was a copy-paste mistake)

shopping_cart/shipping_calculator.feature
```diff
@@ -0,0 +1,92 @@
...
+
+	Scenario: User enters zipcode that is not exactly 5 integers
+		Given that I have 1 Broomstick in my cart
+		And I am on the checkout page
+		And I fill in the shipping address field with "175 W. Jackson Blvd"
+		And I fill in the shipping city field with "Chicago"
+		And I fill in the shipping state field with "IL"
+		And I fill in the shipping zipcode field with "6060442242424hello"
+		When I click the "Estimate shipping" button
+		Then I should see "Please enter a shipping address under 70 characters"
```
oharace: Same deal here, this is the wrong message.

shopping_cart/shipping_calculator.feature
```diff
@@ -0,0 +1,92 @@
+Feature: weight-agnostic shipping calculator
+	As a user
+	I want to know the cost to ship products to my address
+	So I can expect how much my grand order total will be
+
+	Scenarios: Enter valid address to calculate shipping cost
+		Given that I have 1 Broomstick in my cart
```
oharace: When every scenario starts with the same step(s), you can extract them to the beginning of the file before all the scenarios like so:
```
Background:
  Given that I have 1 Broomstick in my cart

Scenario: Enter a valid address to calculate...
```
Then delete that line from each scenario.

shopping_cart/shipping_calculator.feature
```diff
@@ -0,0 +1,92 @@
+Feature: weight-agnostic shipping calculator
+	As a user
+	I want to know the cost to ship products to my address
+	So I can expect how much my grand order total will be
+
+	Scenarios: Enter valid address to calculate shipping cost
```
oharace: Oops: `Scenario` rather than `Scenarios`

oharace:
Nice work on this exercise. You did well thinking through the different scenarios/paths! :+1:

paulghaddad:
@cstavitsky Nice job on this iteration. One recommendation is to strive for step definitions at a higher level of abstraction. So instead of "fill in the input"; "click this button", etc, you can say "when I submit the form". The low-level implementation details can then be hidden in the step definition. 

Here are some good examples of high-level step definitions:

https://github.com/jmmastey/level_up/blob/master/features/users/sign_in.feature

Furthermore, strive to stay on the same level of abstraction. It makes it easier to read since you'll have fewer step definitions. Here is a good article that dives into this:

http://robots.thoughtbot.com/acceptance-tests-at-a-single-level-of-abstraction

For your next iteration, really focus on refactoring your existing feature tests with this philosophy.

Nice work! :)

PR #1150, opened 2016-01-25
Stylish Ruby Exercises

bap13:
My code for the 'Stylish (and Sane) Ruby' section of 'Basic Ruby'.

Thanks for your feedback!

### DIFF
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -1,33 +1,19 @@
-class Arrowhead
-  # This seriously belongs in a database.
-  CLASSIFICATIONS = {
-    far_west: {
-      notched: "Archaic Side Notch",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Agate Basin",
-      bifurcated: "Cody",
-    },
-    northern_plains: {
-      notched: "Besant",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Humboldt Constricted Base",
-      bifurcated: "Oxbow",
-    },
-  }
+require 'json'
 
-  # FIXME: I don't have time to deal with this.
+data = File.read("arrowhead_data.json")
+CLASSIFICATIONS = JSON.parse(data, :symbolize_names => true)
+
+class Arrowhead
   def self.classify(region, shape)
-    if CLASSIFICATIONS.include? region
-      shapes = CLASSIFICATIONS[region]
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
+    unless shapes = CLASSIFICATIONS[region]
       raise "Unknown region, please provide a valid region."
     end
+
+    unless arrowhead = shapes[shape]
+      raise "Unknown shape value. Are you sure you know what you're talking about?"
+    end
+
+    "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
   end
 end
 
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -1,17 +1,16 @@
 require 'json'
 
 data = File.read("arrowhead_data.json")
-CLASSIFICATIONS = JSON.parse(data, :symbolize_names => true)
+CLASSIFICATIONS = JSON.parse(data, symbolize_names: true)
 
 class Arrowhead
-  def self.classify(region, shape)
-    unless shapes = CLASSIFICATIONS[region]
-      raise "Unknown region, please provide a valid region."
-    end
+  REGION_ERROR = "Unknown region, please provide a valid region."
+  SHAPE_ERROR  = "Unknown shape value. "\
+                 "Are you sure you know what you're talking about?"
 
-    unless arrowhead = shapes[shape]
-      raise "Unknown shape value. Are you sure you know what you're talking about?"
-    end
+  def self.classify(region, shape)
+    raise REGION_ERROR unless (shapes = CLASSIFICATIONS[region])
+    raise SHAPE_ERROR  unless (arrowhead = shapes[shape])
 
     "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
   end
--- a/arrowhead/arrowhead.rb
+++ b/arrowhead/arrowhead.rb
@@ -1,12 +1,12 @@
 require 'json'
 
-data = File.read("arrowhead_data.json")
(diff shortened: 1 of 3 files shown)


arrowhead/arrowhead.rb
```diff
@@ -1,33 +1,18 @@
...
-      bifurcated: "Cody",
-    },
-    northern_plains: {
-      notched: "Besant",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Humboldt Constricted Base",
-      bifurcated: "Oxbow",
-    },
-  }
+  REGION_ERROR = "Unknown region, please provide a valid region."
```
oharace: I'll be honest, I'm not sure why this rubocop message is happening. I haven't seen it before.
Besides that, write strings with single quotes if there won't be any interpolation. 

`REGION_ERROR = 'Unknown region, please provide a valid region.'`

arrowhead/arrowhead.rb
```diff
@@ -1,33 +1,18 @@
...
+  SHAPE_ERROR  = "Unknown shape value. "\
```
oharace: Single quotes

arrowhead/arrowhead.rb
```diff
@@ -1,33 +1,18 @@
...
-      if shapes.include? shape
-        arrowhead = shapes[shape]
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+    raise REGION_ERROR unless (shapes = CLASSIFICATIONS[region])
```
oharace: This is a good, concise way to do this!

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
 # Killer facts about triangles AWW YEAH
```
oharace: For future exercises, please split each exercise into it's own pull request, and name the pull <Exercise, Version>, so like 'Triangle Facts, v1'

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
+
+  def recite_facts
+    puts side_facts
+    puts angle_facts
+    puts 
+  end
+
+  def side_facts
+    return 'This triangle is equilateral!' if equilateral?
+    return "This triangle is isosceles! "\
```
oharace: Single quotes

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
+           "Also, that word is hard to type." if isosceles?
+    return 'This triangle is scalene and mathematically boring.' if scalene?
+  end
 
+  def angle_facts
+    angles = calculate_angles(side1, side2, side3)
+    message = "The angles of this triangle are #{angles.join(',')}"
+    message << "\nThis triangle is also a right triangle!" if angles.include? 90
```
oharace: Wrap arguments in parentheses

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
+    angle_a = calculate_angle(a, b, c)
+    angle_b = calculate_angle(b, a, c)
+    angle_c = calculate_angle(c, a, b)
+
+    [angle_a, angle_b, angle_c]
+  end
+
+  # Uses the law of cosines to calculate angle opposite of 'side'
+  # given three side lengths
+  def calculate_angle(side, other1, other2)
```
oharace: I think a little more descriptive parameter would be beneficial. Rather than `other1`, maybe use `first_opposite_side`, and same for `other2`

arrowhead/arrowhead.rb
```diff
@@ -1,33 +1,18 @@
...
+    raise SHAPE_ERROR  unless (arrowhead = shapes[shape])
+
+    "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
```
oharace: This is a good example where double quotes should be used

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,71 @@
...
-end
+  attr_reader :side1, :side2, :side3
+
+  def initialize(side1, side2, side3)
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+  end
+
+  def equilateral?
```
oharace: Some of these methods should be private.

oharace:
Nice work with these! Take a look at my comments and keep them in mind for your future exercises. :+1:

triangle_facts/triangle_facts.rb
bap13: Thanks. I wasn't sure how I should be doing that.

arrowhead/arrowhead.rb
bap13: I was hoping this line would clearly show my intentions, but seeing it again myself after a couple days, it wasn't immediately clear. Even though it takes up more space, I am thinking this would be more transparent:
```ruby
raise REGION_ERROR unless CLASSIFICATIONS.include?(region)
```
And do the same thing for the next line where the shape lookup happens.

paulghaddad: And you can refactor this further by extracting this condition (`CLASSIFICATIONS.include?(region)`) as a well-named method.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,63 @@
...
-	def radians_to_degrees(rads)
-		return (rads * 180 / Math::PI).round
-	end
+  attr_reader :sides
+
+  def initialize(*sides)
+    @sides = sides[0, 3]
+  end
+
+  def recite_facts
```
paulghaddad: I'd personally recommend not using`puts` inside a method like this. I'd recommend calling `puts` on the method itself, from wherever you call it, i.e. `puts recite_facts`.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,63 @@
...
+    puts side_facts
+    puts angle_facts + "\n\n"
```
paulghaddad: Consider using string interpolation, and then you can condense this into one line.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,63 @@
...
+  end
+
+  private
```
paulghaddad: Good job making non-public methods `private`.

paulghaddad:
@bap13 Nice work! You understand what fundamentals of each exercise, so it's your call whether to submit another iteration. But as Chris said, split each exercise into its own PR.

PR #1154, opened 2016-01-29
Supportive - iteration 1 (fixed)

jcurlee:


blag_post.rb
```diff
@@ -0,0 +1,81 @@
...
+    begin
+      pub_date = pub_date.to_date
+    rescue ArgumentError
+      pub_date = nil
+    end
+
+    @publish_date = pub_date ? pub_date : Date.current
+  end
+
+  def initialize(args = {})
```
ssapra: I believe the original data used strings for the keys in args which was meant to encourage you to look for a cleaner way to convert a hash with strings as keys to be accessed with symbols

blag_post.rb
```diff
@@ -0,0 +1,81 @@
...
+    unless args[:author].blank? && args[:author_url].blank?
+      @author = Author.new(args[:author], args[:author_url])
+    end
+
+    @comments = args[:comments].presence || []
```
ssapra: Check out `Array.wrap` to handle nil

blag_post.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def publish_date=(pub_date)
+    pub_date = pub_date.presence || ''
+
+    begin
+      pub_date = pub_date.to_date
+    rescue ArgumentError
+      pub_date = nil
+    end
```
ssapra: This is fine, but I would check out the method `try` as a way of handling something that could be nil.

blag_post.rb
```diff
@@ -0,0 +1,81 @@
...
+    by_line += ", at #{author.url}" if author.url
+
+    "By #{by_line}"
+  end
+
+  def category_list
+    return nil if categories.empty?
+
+    label = "Category".pluralize(categories.length) + ': '
+    label + categories.map { |cat| cat.to_s.humanize.titleize }.to_sentence
```
ssapra: This is fine but to make the string concatenation a bit easier to grasp, I would do:

```
label = "Category".pluralize(categories.length)
sentences = categories.map { |cat| cat.to_s.humanize.titleize }.to_sentence
"#{label}: #{sentences}"
```

Personal preference really more than anything

ssapra:
A few comments, but otherwise your code looks great. Great job adding tests!

blag_post.rb
jcurlee: I like that a lot better actually -- wasn't a fan of my solution

paulghaddad:
@jcurlee You are good to move on! Nice job adding specs as well!

PR #1155, opened 2016-01-30
Zach's Code Challenge for Fulton Works

zpfled:


shopping_cart/adding_coupons.md
```diff
@@ -0,0 +1,63 @@
...
+
+####Scenario: Adding a valid coupon code for one item (Happy)
+* Given that I am on my shopping cart page
+* And I have an item in my cart
+* And I have a coupon code
+* And it is not expired
+* And it is applicable to the item in my cart
+* And I enter the coupon code
+* When I click 'Apply discount'
+* Then I should see my order total reduced by the value of the coupon
```
jmmastey: The structure of this test is generally fine, but more verbose than it needs to be. While it'll still pass, it means that over time your test suite will become a huge pain in your ass and you won't want to update your tests. A few things to help here.

1. Move invariants into a `Background` block. Usually, user authentication lives there. Also, stuff like `I am on my shopping cart page`.
2. Combine steps that aren't relevant to the test. Three or four steps here could be combined as `I have a valid coupon.` This causes your step definitions to be a little more verbose, but that's *okay*. Tests are documentation as well as verification, and the documentation role is more important in this case. If you find yourself writing the same parts of step definitions over and over, move them into `def` statements within the step definition file and call them from there.

Actually, those are the two most important. Skipping my other advice. :)

shopping_cart/adding_coupons.md
```diff
@@ -0,0 +1,63 @@
...
+
+####Scenario: Adding a coupon code for an ineligible item (Bad)
+* Given that I am on my shopping cart page
+* And I have an item in my cart
+* And I have a coupon code
+* And it is not expired
+* And it is not applicable to the item in my cart
+* And I enter the coupon code
+* When I click 'Apply discount'
+* Then I should see an error message that informs my item is ineligible for discount
```
jmmastey: What about a case where I add an item and get the discount, then remove the item from my cart? I always liked that case.

shopping_cart/adding_items.md
```diff
@@ -0,0 +1,38 @@
...
+* Given that I am on an item page
+* And the item is in stock
+* When I click 'Add to cart'
+* Then I should see a confirmation message that the item was successfully added to my cart
+
+####Scenario: Adding an in-stock item and viewing it in my cart (Happy)
+* Given that I am on an item page
+* And the item is in stock
+* And I have NOT already added the item to my cart
+* And I click 'Add to cart'
```
jmmastey: This is probably a `When`

shopping_cart/adding_items.md
```diff
@@ -0,0 +1,38 @@
...
+####Scenario: Adding an in-stock item to my cart (Happy)
+* Given that I am on an item page
+* And the item is in stock
+* When I click 'Add to cart'
+* Then I should see a confirmation message that the item was successfully added to my cart
+
+####Scenario: Adding an in-stock item and viewing it in my cart (Happy)
+* Given that I am on an item page
+* And the item is in stock
+* And I have NOT already added the item to my cart
```
jmmastey: Negative tests can be of low value overall, but I like that you're using it to provide contrast to the definition itself. Good call.

shopping_cart/adding_items.md
```diff
@@ -0,0 +1,38 @@
...
+* And I click 'Add to cart'
+* When I visit the shopping cart page
+* Then I should see my item listed
+
+####Scenario: Adding another in-stock item and viewing it in my cart (Happy)
+* Given that I am on an item page
+* And the item is in stock
+* And I have already added the item to my cart n times
```
jmmastey: This isn't usually how shopping carts work, but I get what you're saying :)

shopping_cart/changing_quantity.md
```diff
@@ -0,0 +1,60 @@
+###Feature: Changing quantity of shopping cart items
+* In order to buy as much of the merchandise I want
+* I need to be able to change the quantity of items my shopping cart
+
+####Scenario: Increasing quantity (Happy)
+* Given that I am on my shopping cart page
+* And I have 3 items in my cart
+* And I want to change the quantity of that item to 4
```
jmmastey: Your desires are probably not relevant to the test.

shopping_cart/changing_quantity.md
```diff
@@ -0,0 +1,60 @@
...
+* And there are at least 4 items of that kind in stock
+* And I click 'Change Quantity'
+* And I enter 4
+* When I click 'submit'
```
jmmastey: This kind of step definition is really reusable, but ultimately leads to the same verbose tests. It's okay to just say "I submit a new quantity of 4"

shopping_cart/changing_quantity.md
```diff
@@ -0,0 +1,60 @@
...
+* And I fail to enter any number
+* When I click 'submit'
+* Then I should see an error message that prompts me to enter a number
+
+####Scenario: Attempting to change quantity to a non-number (Bad)
+* Given that I am on my shopping cart page
+* And I have 3 items in my cart
+* And I want to change the quantity of that item
+* And I click 'Change Quantity'
+* And I enter a non-number (such as 'a')
```
jmmastey: The parenthetical here is probably unnecessary. Actually, if you already have a step for `I submit a new quantity of :x` you can just reuse it here.

shopping_cart/changing_quantity.md
```diff
@@ -0,0 +1,60 @@
...
+####Scenario: Attempting to increase quantity when there is not enough stock (Bad)
+* Given that I am on my shopping cart page
+* And I have 3 items in my cart
+* And I want to change the quantity of that item
+* And I click 'Change Quantity'
+* And there are only 4 items in stock
+* And I enter 5
+* When I click 'submit'
+* Then I should see an error message that tells me there are only 4 items in stock
+* Then I should not see my item quantity change
```
jmmastey: What about a negative number?

shopping_cart/getting_shipping_estimates.md
```diff
@@ -0,0 +1,72 @@
...
+####Scenario: Invalidating shipping info when items are added (Bad)
+* Given that I am on my shopping cart page
+* And I have an item in my cart
+* And I enter my zip code
+* And I select a shipping carrier
+* And I click 'Estimate shipping costs'
+* And I see a shipping estimate
+* When I add an item to my cart that is ineligible to be shipped by my carrier
+* Then I should see my shipping estimate reset automatically
+* Then I should see an error message prompting me to select a shipping carrier
```
jmmastey: I like this scenario overall. Not sure it should generate an error-message per se, but it's good to reset.

shopping_cart/logging_in.md
```diff
@@ -0,0 +1,51 @@
...
+* In order to have as few barriers to purchasing merchandise as possible
+* I need to be able to save my cart and come back to it later
+
+####Scenario: Logging in with a saved in-stock item (Happy)
+* Given that I am not logged in
+* And I am viewing my empty cart
+* And I added an item to my cart previously when I was logged in
+* And the item is still in stock
+* When I log in with valid credentials
+* Then I should see the item in my cart from my last session
```
jmmastey: The more I read these, the more I wish the item had a name other than "an item".

shopping_cart/logging_in.md
```diff
@@ -0,0 +1,51 @@
...
+* When I log in with valid credentials
+* Then I should see both items in my cart
+
+####Scenario: Attempting to log in without password (Sad)
+* Given that I am not logged in
+* And I have added an item to my cart as a guest user
+* And I added an item to my cart previously when I was logged in
+* When I attempt to log in without entering my password
+* Then I should see an error message that prompts me to enter a password
+* Then I should see no change in my cart
```
jmmastey: This may be a hard assertion to write, fwiw.

jmmastey: Also, this scenario is more about authentication than carts. They happen at the same time, but they are separate features. They really should be in different tests entirely.

shopping_cart/navigating_to_item_pages.md
```diff
@@ -0,0 +1,18 @@
+###Feature: Navigating to item pages
+* In order to make informed decisions about the merchandise I buy
+* I need to be able to navigate to merchandise detail pages
+
+####Scenario: Navigating to valid item page (Happy)
+* Given that I am on my shopping cart page
+* And I have an item in my cart
+* And I want to review that item in more detail
```
jmmastey: Wants don't figure into tests themselves. What would this step look like?

shopping_cart/navigating_to_item_pages.md
```diff
@@ -0,0 +1,18 @@
...
+* When I click on the item name
+* Then the page title should be {{ item_name }}
+
+####Scenario: Navigating to missing item page (Bad)
+* Given that I am on my shopping cart page
+* And I have an item in my cart
+* And I want to review that item in more detail
+* And some reckless developer has deleted the item page
+* When I click on the item name
+* Then I should see a 404 page
```
jmmastey: Hah. I like this.

shopping_cart/removing_items.md
```diff
@@ -0,0 +1,25 @@
...
+* And I want to remove the item
+* When I click 'Remove Item'
+* Then I should no longer see the item in my cart
+
+####Scenario: Removing an item from my cart when there are multiple of the same item (Happy)
+* Given that I am on my shopping cart page
+* And it contains multiple of the same item
+* And I want to remove one of the items
+* When I click 'Remove Item'
+* Then I should see the item's quantity decrease by one
```
jmmastey: This seems incompatible with your quantity tests.

shopping_cart/removing_items.md
```diff
@@ -0,0 +1,25 @@
...
+
+####Scenario: Removing a non-existing item from my cart (Bad)
+* Given that I have my shopping cart page open in two windows
+* And the cart contains one item
+* And I want to remove the item
+* And I click 'Remove Item' in one of the windows
+* When I click 'Remove Item' in the other window
+* Then I should see an error message that my cart is already empty
```
jmmastey: This error message may not be accurate. This test seems oddly formulated (and hard to test).

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,65 @@
 # Killer facts about triangles AWW YEAH
```
jmmastey: Oh no! Polluted commits!

Make sure your master branch doesn't have any extraneous commits. If it does, you'll want to delete your local master branch and set up a new one to track the upstream master.

jmmastey:
Let me know if you want me to leave this open for Juan or anyone. Generally good tests. A little refactoring will make them much tighter and easier to work with, though.

PR #1156, opened 2016-02-01
Dinodex Iteration 1:

balajisubr:
To run,

ruby dino_runner.rb

If you want to print info for a particular dino, call 
print_info_for_species("Megalosaurus") in the above file

### DIFF
--- a/dino_catalog/afro_dino_parser.rb
+++ b/dino_catalog/afro_dino_parser.rb
@@ -1,0 +1,23 @@
+require 'csv'
+require 'pry'
+load 'dino_parser.rb'
+class AfroDinoParser < DinoParser
+  def initialize(file_name)
+    super(file_name)
+  end
+
+  def find_carnivores
+    @csv_data.each do |line|
+      if((line[@food_index]) == 'Yes')
+        @carnivores << line[@map_indices[@fields_mapping[:name].to_sym]]
+      end
+    end
+    print_dinosaurs("Carnivores Dinosaurs", @carnivores)
+  end
+end
+
+afro = AfroDinoParser.new('/Users/bsubramanian/level_up_exercises/dino_catalog/african_dinosaur_export.csv')
+afro.find_carnivores
+afro.find_bipeds
+afro.dino_classified_by_periods
+afro.classify_by_weight
--- a/dino_catalog/afro_dino_parser.rb
+++ b/dino_catalog/afro_dino_parser.rb
@@ -1,23 +1,18 @@
 require 'csv'
-require 'pry'
 load 'dino_parser.rb'
 class AfroDinoParser < DinoParser
   def initialize(file_name)
     super(file_name)
   end
 
-  def find_carnivores
+  def classify_by_carnivores
     @csv_data.each do |line|
       if((line[@food_index]) == 'Yes')
         @carnivores << line[@map_indices[@fields_mapping[:name].to_sym]]
       end
     end
-    print_dinosaurs("Carnivores Dinosaurs", @carnivores)
+    return @carnivores
   end
 end
 
-afro = AfroDinoParser.new('/Users/bsubramanian/level_up_exercises/dino_catalog/african_dinosaur_export.csv')
-afro.find_carnivores
-afro.find_bipeds
-afro.dino_classified_by_periods
-afro.classify_by_weight
+#afro = AfroDinoParser.new('/Users/bsubramanian/level_up_exercises/dino_catalog/african_dinosaur_export.csv')
--- a/dino_catalog/dino_parser.rb
+++ b/dino_catalog/dino_parser.rb
@@ -1,5 +1,5 @@
-require 'pry'
 class DinoParser
+  attr_accessor :csv_data, :data_fields
   @@fields =  
     [:food, :name, :period, :weight,
      :walking, :continent, :description]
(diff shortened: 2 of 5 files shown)


oharace:
All these rubocop messages are valid, and should be addressed/fixed.

dino_catalog/afro_dino_parser.rb
```diff
@@ -0,0 +1,18 @@
...
+    @csv_data.each do |line|
+      if((line[@food_index]) == 'Yes')
+        @carnivores << line[@map_indices[@fields_mapping[:name].to_sym]]
+      end
+    end
+    return @carnivores
+  end
+end
+
+#afro = AfroDinoParser.new('/Users/bsubramanian/level_up_exercises/dino_catalog/african_dinosaur_export.csv')
```
oharace: Remove comments

dino_catalog/afro_dino_parser.rb
```diff
@@ -0,0 +1,18 @@
+require 'csv'
+load 'dino_parser.rb'
+class AfroDinoParser < DinoParser
+  def initialize(file_name)
+    super(file_name)
+  end
+
+  def classify_by_carnivores
+    @csv_data.each do |line|
+      if((line[@food_index]) == 'Yes')
```
oharace: Generally, you don't need to wrap conditions in parentheses, unless there are multiple conditions using `&&` or `||`. So, this would be:
`if line[@food_index] == 'Yes'`
Makes it easier to read.

dino_catalog/dino_parser.rb
```diff
@@ -0,0 +1,116 @@
+class DinoParser
+  attr_accessor :csv_data, :data_fields
+  @@fields =
```
oharace: Hmm. Rather than a class-level variable, I think it would be better to use a constant.

dino_catalog/dino_parser.rb
```diff
@@ -0,0 +1,116 @@
...
+    @map_indices = {}
+    @data_fields.each_with_index do |data_field, i|
+      @map_indices[data_field.to_sym] = i
+    end 
+  end
+
+  def classify_by_bipeds
+    bipeds = []
+    @csv_data.each do |line|
+      if(line[walking_index].downcase == 'biped')
```
oharace: Remove parentheses around condition

dino_catalog/dino_parser.rb
```diff
@@ -0,0 +1,116 @@
...
+    return bipeds
+  end
+
+  def classify_by_periods
+    dinos_in_periods = {}
+    @csv_data.each  do |line|
+      key = line[period_index]
+      key = 'Cretaceous' if(line[period_index].match(/Cretaceous/))
+   
+      if(dinos_in_periods[key]).nil?
```
oharace: Remove parentheses around condition --> `if dino_in_periods[key].nil?`

dino_catalog/dino_parser.rb
```diff
@@ -0,0 +1,116 @@
...
+        dinos_in_periods[key] = [line[name_index]]
+      else
+        dinos_in_periods[key] << line[name_index]
+      end
+    end   
+    return dinos_in_periods
+  end
+
+  def find_carnivores
```
oharace: Empty method?

dino_catalog/dino_parser.rb
```diff
@@ -0,0 +1,116 @@
...
+    
+  end
+
+  def classify_by_weight
+    dinos_by_weight = {:small => [], :big => []}
```
oharace: This is Ruby's old hash syntax. The new syntax is like: `{ small: [], big: [] }` (so put the colon after the key instead)

dino_catalog/dino_parser.rb
```diff
@@ -0,0 +1,116 @@
...
+    @csv_data.each do |line|
+      if(line[weight_index] && line[weight_index].to_i >= 2000)
```
oharace: Rather than these magic numbers (even though we know what they are in this case), and make them constants
`ONE_TON = 2000`

dino_catalog/dino_parser.rb
```diff
@@ -0,0 +1,116 @@
...
+    @food_index = @map_indices[@fields_mapping[:food].to_sym]
+    @species = []
+    @@fields.each do |field|
+      self.class.send(:define_method,"#{field}_index") do 
+        @map_indices[@fields_mapping[field.to_sym].to_sym]
+      end
+    end
+  end
+
+  def map_fields
```
oharace: Some of these methods should probably be `private`

dino_catalog/dino_runner.rb
```diff
@@ -0,0 +1,76 @@
...
+
+joes_parser = JoesCsvParser.new(File.join(File.dirname(__FILE__),'dinodex.csv'))
+afro_parser = AfroDinoParser.new(File.join(File.dirname(__FILE__),'african_dinosaur_export.csv'))
+
+@parsers = [joes_parser, afro_parser]
+@parsed_data = {"weight" => {"small" => [], "big" => []}, "bipeds" => [], 
+  "periods" => {}, "carnivores" => []}
+
+
+def parse_and_print_info(criteria)
```
oharace: Methods should only perform one task, so this method is doing too much. Have one method parse the data, and one method print the data. That will split the logic and make your methods simpler and easier to read.

oharace:
I _think_ that this exercise wants you to chain calls. So for example, if I did:
`dino_dex = DinoParser(<file>)`
Then I would want to chain calls together to filter the dinosaurs like so:
`dinodex.carnivores.large.print_facts`
This would get all the carnivores that are large in weight and then print the facts. Let myself or another mentor know if you need clarification or additional help.

oharace:
Nice work on your first iteration! :+1: Please take a look at the comments, make some fixes, and submit another iteration. As always, feel free to ask for help!

PR #1161, opened 2016-02-01
robot_name, v1

bap13:
More practice with Ruby style.

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -1,25 +1,34 @@
-class NameCollisionError < RuntimeError; end
-
 class Robot
-  attr_accessor :name
+  NameCollisionError = Class.new(RuntimeError)
 
-  @@registry
+  NAME_ERROR_MSG = 'There was a problem generating the robot name!'.freeze
+
+  attr_reader :name
+
+  class << self
+    attr_accessor :registry
+  end
+  @registry ||= []
 
   def initialize(args = {})
-    @@registry ||= []
-    @name_generator = args[:name_generator]
+    name_generator = args.fetch(:name_generator, false)
+    @name = name_generator ? name_generator.call : generate_new_name
+    raise NameCollisionError, NAME_ERROR_MSG if invalid_name?
+    Robot.registry << @name
+  end
 
-    if @name_generator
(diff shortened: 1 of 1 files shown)


robot_name/robot_name.rb
```diff
@@ -1,25 +1,34 @@
...
+  class << self
+    attr_accessor :registry
+  end
+  @registry ||= []
 
   def initialize(args = {})
-    @@registry ||= []
-    @name_generator = args[:name_generator]
+    name_generator = args.fetch(:name_generator, false)
+    @name = name_generator ? name_generator.call : generate_new_name
```
jmmastey: Use `method(:generate_new_name)` and you can get a callable reference in either case. That'll clear up these few lines.

robot_name/robot_name.rb
```diff
@@ -1,25 +1,34 @@
...
+  private
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  def generate_new_name
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
```
jmmastey: If we're doing short methods, these really are begging to be moved out into methods.

robot_name/robot_name.rb
```diff
@@ -1,25 +1,34 @@
...
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
+    [Array.new(2) { generate_char.call },
+     Array.new(3) { generate_num.call }].join
```
jmmastey: Ehhhh, not sure I like this form of Array initialization. It's clever, but harder to understand overall. Prefer clarity to brevity.

jmmastey:
A few awkward constructions still, but overall looks fine. Nice work.

robot_name/robot_name.rb
bap13: Nice! I didn't know you could do that.

bap13:
Thanks for your comments Joe!

PR #1162, opened 2016-02-02
Dino catalog exercise - Iteration 2

locker1776:
Notes:
1. I changed and created a new transform data class. This will take in a hash of dino data, and will normalize and add default values. So now any way we load dino data can be transformed using this class, and it is not dependent on anything. You would have to update the transform for any new transformations needed, but it is obvious where to do it.

2. Also cleaned up the Dino and Dino Collection classes. Dino Collection is still dependent on Dino, but that is on purpose.

3. FilterDinos is much cleaner per Deja's suggestions.

4. There is only 1 default value missing in one of the files (missing continent for Africa). Without knowing use cases, hard to determine how adding default values should look going forward.


dino_catalog/dino.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def to_hash
+    attribute_hash = {}
+    DINO_ATTRIBUTES.each do |attribute|
+      value = send(attribute)
+      attribute_hash[attribute] = send(attribute) unless value.nil?
+    end
+    attribute_hash
+  end
```
jmmastey: Nitpick: idiomatically in ruby, this should be `to_h`.

Generally, if you see this pattern in a method (initialize a value, shove stuff into it, return that value), there is a ruby enumerable method that will express it better. In fact, when it works well, you end up not needing to keep an extra variable (`attribute_hash` here) at all. Compare:

```
def to_h
  DINO_ATTRIBUTES.each_with_object({}) do |attribute, hash|
    value = send(attribute)
    hash[attribute] = value if value
  end
end

File: dino_catalog/dino.rb [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+    @continent = attrs[:continent]
+    @diet = attrs[:diet]
+    @weight = attrs[:weight]
+    @walking = attrs[:walking]
+    @description = attrs[:description]
+  end
+
+  def carnivore_or_herbivore
+    diet == "Herbivore" ? diet : "Carnivore"
+  end
```
jmmastey: Ehhh, unless there's data coercion elsewhere, this method could return, say `Piscivore`, which is not `carnivore_or_herbivore`. So, the name is deceptive at the very least.

dino_catalog/dino_collection.rb
```diff
@@ -0,0 +1,36 @@
+class DinoCollection
+  attr_accessor :dinos
+
+  def initialize(array_of_dinos = nil)
```
jmmastey: Try not to encode data type into a variable name. Which is to say, call this `dinos`, not `array_of_dinos`.

dino_catalog/dino_collection.rb
```diff
@@ -0,0 +1,36 @@
...
+    @dinos = []
+    load(array_of_dinos) unless array_of_dinos.nil?
+  end
+
+  def load(array_of_dinos)
+    array_of_dinos.each { |attrs| create_dino(attrs) }
+  end
```
jmmastey: If we give `array_of_dinos` a default of an array, we can make a only-slightly-shaky assumption that some enumerable will be passed. In which case these whole two methods become:

```
def initialize(dinos = [])
  @dinos = load(dinos)
end

def load(dinos)
  dinos.map { |attrs| dino_from(attrs) }
end
```

As I was writing this, I realized that you have `create_dino` pushing into `@dinos` directly. I'll address that below ;)

dino_catalog/dino_collection.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def create_dino(attrs)
+    dinos << Dino.new(attrs)
+  end
```
jmmastey: So there's an entire wing of programming called Functional Programming. One of the central ideas of FP is that "side effects" are hard to keep track of. That is, if I can call a function over and over, and it doesn't change anything about the world (by outputting data, or shoveling into `@dinos`), then I can reason about it more easily, because I don't depend on, say, the prior state of the `@dinos` hash.

We don't _have_ to use FP in ruby, but once you start writing tests, you'll see that it's massively easier than not doing so in most places.

dino_catalog/dino_collection.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def <<(dino)
+    add_dino(dino)
+  end
+
+  def add_dino(dino)
+    dinos << dino
+  end
```
jmmastey: `create_dino`, `<<` and `add_dino` are _really_ similar. They almost certainly can be combined. I like use of shovel `<<`. What about this?

```
def <<(record)
  record = Dino.new(record) unless record.is_a? Dino
  dinos << record
end
alias_method :<<, :push
```

I think that would encapsulate your set of multiple methods (including an extra name for shovel if we want it instead), and also be easier to work with. For more info on why I structured it to cast to a Dino, look at: https://medium.com/@jmmastey/ruby-and-postels-law-ee94269f322

dino_catalog/dino_collection.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def to_s
+    to_json
+  end
```
jmmastey: Ehhhh, I'm not sold that the natural string representation of a series of dinos is necessarily JSON.

dino_catalog/dino_collection.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def to_json
+    array_of_hashes = []
+    dinos.each do |dino|
+      array_of_hashes << dino.to_hash
+    end
+    JSON.pretty_generate(array_of_hashes)
+  end
```
jmmastey: Same pattern as in dino: initialize, loop, return. Try `.map` here:

`JSON.pretty_generate(dinos.map(&:to_h))`

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env ruby
```
jmmastey: I like the separation of the command line script. Woo!

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,56 @@
...
+
+require './dino_collection'
+require './dino'
+require './filter_dinos'
+require './import_data'
+require './transform_data'
+require 'trollop'
```
jmmastey: O.o?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,56 @@
...
+
+- Usage:
+      dinodex.rb [options]
+where [options] are:
+EOS
+
+  opt :search_file, "Path of the search YAML file", type: String
+  opt :output_file, "Path of the file for JSON output", type: String
+end
+Trollop.die :search_file, "must exist" unless command_opts[:search_file]
```
jmmastey: What an odd library name.

dino_catalog/filter_dinos.rb
```diff
@@ -0,0 +1,28 @@
+class DinoFilter
+  def self.search(dino_collection, search_attrs)
+    working_collection = dino_collection
+    search_attrs.each do |category, value|
+      category = :carnivore_or_herbivore if category == :diet
```
jmmastey: This exception should be an indication that there's a refactoring to be had here.

Interesting, other than maybe two references to dinosaurs, there's no reason why this has to be a Dinosaur filter. Which makes me think we could just make it a filter. But then I feel like we could just collapse it to use built in ruby methods. So, maybe not. I'd want to play with it.

I will point out that this method and `find_matches` both use the initialize/loop/return pattern. Try to find a version that doesn't.

dino_catalog/filter_dinos.rb
```diff
@@ -0,0 +1,28 @@
...
+      working_collection = find_matches(working_collection, category, value)
+    end
+    working_collection
+  end
+
+  def self.find_matches(dino_collection, category, value)
```
jmmastey: So, generally prefer non-static methods, for reasons similar to the FP explanation above. It's a little easier to work with instances.

dino_catalog/filter_dinos.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def self.dino_match?(dino, category, value)
+    case category
+      when :weight_greater_than
+        return !dino.weight.nil? && dino.weight > value
+      when :weight_less_than
+        return !dino.weight.nil? && dino.weight < value
+    end
+    dino.send(category).match(value)
```
jmmastey: I had to jump into IRB to see if this worked :D 

It does for string, but be careful that it doesn't work for integers. Also, as a larger concern, this filter needs to know about the implementation details of dinosaur weights. As our system grows more behavior, that will become a bigger and bigger problem.

Solving it, though, can take a little cleverness. Many folks add a `big?` and `small?` predicate method. Since you're letting users select a weight threshold, that clearly doesn't work here. Having some way for your dinosaur to respond to, say, `dino.bigger_than(value)` might work. Then I'd work to use `#respond_to?` to call that method automatically.

Let me know if that was clear as mud. ;)

dino_catalog/import_data.rb
```diff
@@ -0,0 +1,22 @@
+class ImportData
+  CSVFILES = ['dinodex.csv', 'african_dinosaur_export.csv'].freeze
```
jmmastey: `CSV_FILES`.

Also, honestly, load these from command line. Or at least let me _try_ to load other files. I've got so many files.

dino_catalog/import_data.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def self.load_and_transform
+    processed_data = []
+    CSVFILES.each do |file_path|
+      read_csv_file(file_path).each do |dino_info|
+        processed_data << TransformData.transform(dino_info)
+      end
+    end
+    processed_data
+  end
```
jmmastey: initialize/loop/return

dino_catalog/transform_data.rb
```diff
@@ -0,0 +1,36 @@
...
+  def self.add_default_values(normalized_info)
+    normalized_info[:continent] = normalized_info.fetch(:continent, "Africa")
+    normalized_info
+  end
+
+  def self.normalize(key, value)
+    return { key => value } unless NORMALIZE_MAPPINGS.key?(key)
+    method_to_use = "#{NORMALIZE_MAPPINGS[key]}_transform"
+    send(method_to_use, value)
+  end
```
jmmastey: Good old dynamic dispatch ;)

I think this is probably fine. This is the technique I'd have used in `DinoFilter`, basically. You can only use it sparingly, though, before your code becomes a mess.

dino_catalog/transform_data.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def self.carnivore_transform(value)
+    diet = value == "Yes" ? "Carnivore" : "Herbivore"
+    { diet: diet }
+  end
+
+  def self.weight_transform(value)
+    { weight: value }
+  end
```
jmmastey: I'm not super stoked on the hashes that get returned from these three methods, considering they're just returning a single value. I get that you want both the key and value... Hrm. I'd want to think on this, but overall it smells a bit.

Also, cast your weights to integers ;)

dino_catalog/transform_data.rb
```diff
@@ -0,0 +1,36 @@
...
+    method_to_use = "#{NORMALIZE_MAPPINGS[key]}_transform"
+    send(method_to_use, value)
+  end
+
+  def self.name_transform(value)
+    { name: value }
+  end
+
+  def self.carnivore_transform(value)
+    diet = value == "Yes" ? "Carnivore" : "Herbivore"
```
jmmastey: Throw some parens around the ternary here. It's hard to read otherwise.

dino_catalog/transform_data.rb
```diff
@@ -0,0 +1,36 @@
...
+  def self.transform(dino_info)
+    normalized_info = {}
+    dino_info.each do |k, v|
+      normalized_info.merge!(normalize(k, v))
+    end
+    add_default_values(normalized_info)
+  end
+
+  def self.add_default_values(normalized_info)
+    normalized_info[:continent] = normalized_info.fetch(:continent, "Africa")
```
jmmastey: This is only true for the current CSV. If I were even to add a record to the dinodex CSV that had no continent, this would break.

Overall, it's important to be careful not to specialize too much into the data you have right now. If it changes later, you get into trouble. What should probably happen here is that each CSV format (and later, JSON formats or whatnot), get a configured set of transformations. So if I add an `asian_dinos.csv`, you can specify separate transformations for that one than you have for `african_dinos.csv`.

jmmastey:
This is way cleaner than the first. Good job. And you're right about transforms being better isolated now, which is awesome. I added a bunch of suggestions, but you don't need to run through and fix everything. Make sure you understand the reasons, try out some of the ideas (especially around those initialize/loop/return methods) and move on.

Good work.

dino_catalog/dino.rb
locker1776: Thank You! I have been trying to figure out how to do this in Ruby, with little success.
This is a pattern I use all the time, and now you have made it much better!!!

dino_catalog/dinodex.rb
locker1776: Some of us are fans of Victorian Literature.

PR #1164, opened 2016-02-03
Dino catalog exercise v1

mostlybadfly:
Hello, This is a first attempt at the dino_catalog exercise.  I want to do more with it but I was hoping to get some feedback/clarity if I'm heading in the right direction.  I have tried some specs which I hope give some of my intent with this.

Thanks!

### DIFF
--- a/dino_catalog/african_dino.rb
+++ b/dino_catalog/african_dino.rb
@@ -1,0 +1,12 @@
+class AfricanDino < Dino
+  attr_accessor :name, :period, :diet, :weight, :walking
+
+  def initialize(name, period, diet, weight, walking)
+    @name = name
+    @period = period
+    @diet = diet
+    @weight_in_lbs = weight * 2.2 if weight
+
+    @walking = walking
+  end
+end
--- a/dino_catalog/dino.rb
+++ b/dino_catalog/dino.rb
@@ -1,5 +1,6 @@
 class Dino
-  attr_accessor :name, :period, :continent, :diet, :weight_in_lbs, :walking, :description
+  attr_accessor :name, :period, :continent, :diet, :weight_in_lbs,
+                :walking, :description
 
   def initialize(name, period, continent, diet, weight, walking, description)
     @name = name
--- a/dino_catalog/dino.rb
+++ b/dino_catalog/dino.rb
@@ -1,13 +1,19 @@
 class Dino
-  attr_accessor :name, :period, :continent, :diet, :weight, :walking, :description
+  attr_accessor :name, :period, :continent, :diet, :weight_in_lbs, :walking, :description
 
   def initialize(name, period, continent, diet, weight, walking, description)
     @name = name
     @period = period
     @continent = continent
     @diet = diet
-    @weight = weight
+    @weight_in_lbs = weight
     @walking = walking
     @description = description
+  end
+
+  def to_h
+    instance_variables.each_with_object({}) do |fact, hash|
+      hash[fact.to_s.delete("@").to_sym] = instance_variable_get(fact)
+    end
   end
 end
--- a/dino_catalog/dino.rb
+++ b/dino_catalog/dino.rb
@@ -1,0 +1,13 @@
+class Dino
+  attr_accessor :name, :period, :continent, :diet, :weight, :walking, :description
+
+  def initialize(name, period, continent, diet, weight, walking, description)
+    @name = name
+    @period = period
+    @continent = continent
+    @diet = diet
+    @weight = weight
+    @walking = walking
+    @description = description
+  end
+end
--- a/dino_catalog/dino_catalog.rb
+++ b/dino_catalog/dino_catalog.rb
@@ -13,8 +13,8 @@ class DinoCatalog
   end
 
   def parse(csv_file)
-    listing = CSV.open(csv_file, :converters => :all,
-                          :headers => true, :header_converters => :symbol)
+    listing = CSV.open(csv_file, converters: :all,
+                          headers: true, header_converters: :symbol)
 
     if csv_file.to_s.split("_").include? "african"
       add_dinos(african_dinos(listing))
@@ -42,14 +42,18 @@ class DinoCatalog
     end
   end
 
-  def dino_search(options={})
+  def dino_search(queries)
     dino_results = []
-      options.each {|option,query| dino_results.concat send(option, query)}
-    dino_results.uniq!.each {|dino| puts "----------"; dino_facts(dino)}
+    queries.each { |search, query| dino_results.concat send(search, query) }
+    
+    dino_results.uniq!.each do |dino|
+      puts "----------"
+      dino_facts(dino)
+    end
   end
 
   def name_search(*dino_names)
-    @dino_dex.select { |dino| dino_names.include? dino.name}
+    @dino_dex.select { |dino| dino_names.include? dino.name }
   end
 
   def walking_search(*walk_styles)
@@ -84,7 +88,7 @@ class DinoCatalog
 
   def carnivores
     @dino_dex.select do |dino|
-      ["Carnivore","Insectivore","Piscivore", "Yes"].include? dino.diet
+      %w(Carnivore Insectivore Piscivore Yes).include? dino.diet
     end
   end
 
--- a/dino_catalog/dino_catalog.rb
+++ b/dino_catalog/dino_catalog.rb
@@ -4,7 +4,6 @@ require_relative 'dino'
 require_relative 'african_dino'
 
 class DinoCatalog
-
   attr_accessor :dino_dex
 
   def initialize
@@ -45,7 +44,7 @@ class DinoCatalog
   def dino_search(queries)
     dino_results = []
     queries.each { |search, query| dino_results.concat send(search, query) }
-    
+
     dino_results.uniq!.each do |dino|
       puts "----------"
       dino_facts(dino)
--- a/dino_catalog/dino_catalog.rb
+++ b/dino_catalog/dino_catalog.rb
@@ -1,31 +1,95 @@
 require 'csv'
 require 'pry'
 require_relative 'dino'
+require_relative 'african_dino'
 
 class DinoCatalog
-  attr_accessor :csv_array, :dino_array
+  
+  attr_accessor :dino_dex
 
   def initialize
-    @csv_array = []
-    @dino_array = [] 
+    @dino_dex = [] 
+    @dino_results = []
   end
 
   def parse(csv_file)
-    @csv_array = CSV.open(csv_file, 
-                          :headers => true, 
-                          :header_converters => :symbol,
-                          :converters => :all)
+    dino_array = CSV.open(csv_file, :converters => :all,
+                          :headers => true, :header_converters => :symbol)
+
+    if csv_file.to_s.split("_").include? "african"
+      add_african_dinos(dino_array)
+    else
+      add_dinos(dino_array)
+    end
   end
 
-  def make_dinos
-    @csv_array.each do |dino|
-      @dino_array << Dino.new(dino[:name],
-                              dino[:period],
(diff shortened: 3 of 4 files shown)


jmmastey:
Hound is a cruel mistress, 'eh? Good to see you, Daniel. I'll gladly take a look.

dino_catalog/african_dino.rb
```diff
@@ -0,0 +1,12 @@
+class AfricanDino < Dino
+  attr_accessor :name, :period, :diet, :weight, :walking
+
+  def initialize(name, period, diet, weight, walking)
+    @name = name
+    @period = period
+    @diet = diet
+    @weight_in_lbs = weight * 2.2 if weight
```
jmmastey: How does the dinosaur class know that weight is necessarily in kilos? That's a detail of the datafile you received. Best to handle this logic elsewhere (specifically, where the dino data is imported).

dino_catalog/african_dino.rb
```diff
@@ -0,0 +1,12 @@
...
+
+    @walking = walking
```
jmmastey: Overall, I'm not sure about using subclasses here. None of the behavior is really different, just a few fields. Generally different data sources don't really deserve part in the class hierarchy.

dino_catalog/dino.rb
```diff
@@ -0,0 +1,20 @@
...
+    @walking = walking
+    @description = description
+  end
+
+  def to_h
+    instance_variables.each_with_object({}) do |fact, hash|
+      hash[fact.to_s.delete("@").to_sym] = instance_variable_get(fact)
+    end
+  end
+end
```
jmmastey: Nice and clean, right on.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,103 @@
...
+require 'pry'
+require_relative 'dino'
+require_relative 'african_dino'
+
+class DinoCatalog
+  attr_accessor :dino_dex
+
+  def initialize
+    @dino_dex = []
+    @dino_results = []
```
jmmastey: Usually when the same prefix (or suffix) shows up over and over, you have two options:

1. Create a new abstraction to represent that prefix. In most cases (say, a `Company` class that contains fields for `owner_name`, `owner_address`, `owner_email`) that means pulling the fields out into their own class. You've mostly already done that with the DinoCatalog itself. But the naming remains, so...
2. Delete the prefix most of the places. I think here that `results` is just as clear as `dino_results`.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,100 @@
...
+      Dino.new(dino[:name], dino[:period], dino[:continent],
+               dino[:diet], dino[:weight_in_lbs],
+               dino[:walking], dino[:description])
+    end
+  end
+
+  def african_dinos(listing)
+    listing.map do |dino|
+      AfricanDino.new(dino[:genus], dino[:period], dino[:carnivore],
+                      dino[:weight], dino[:walking])
```
jmmastey: I think this is where I'd put the logic for different fields for African dinos. By using the other `Dino` class (w/ its longer constructor), you'd also be faced with the question of "what should I send for continent?", which leads to the natural answer "oh, just send the string `Africa`". Here, I think the tailor-made constructor hid that decision and made the entire application a little more complicated.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,103 @@
...
+      add_dinos(dinos(listing))
+    end
+  end
+
+  def add_dinos(dinos)
+    @dino_dex.concat dinos
+  end
+
+  def dinos(listing)
+    listing.map do |dino|
```
jmmastey: I like that you didn't modify `@dino_dex` from within this method (or `african_dinos`). That makes for much easier testing.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,103 @@
...
+  def parse(csv_file)
+    listing = CSV.open(csv_file, converters: :all,
+                          headers: true, header_converters: :symbol)
+
+    if csv_file.to_s.split("_").include? "african"
+      add_dinos(african_dinos(listing))
+    else
+      add_dinos(dinos(listing))
+    end
+  end
```
jmmastey: This code is pretty well separated, nicely done. Possible future-looking issue: have you heard of the Single Responsibility Principle? This class is currently responsible for two really big chunks of functionality:
1. Parsing and instantiating Dinos, and
2. Searching

It would be much cleaner to separate those two pieces of functionality. Awesomely, since your code is pretty well factored, that really only requires that you move three or four methods into a class named something like `DinoParser`, and then you can even use that class here if you like.

I won't say that you _have_ to make that refactor right away. I will say that as soon as you start feeling the pain of maintaining this class (as it gets bigger and becomes the place where all new features start to live), that will probably be a good first answer.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  def african_dinos(listing)
+    listing.map do |dino|
+      AfricanDino.new(dino[:genus], dino[:period], dino[:carnivore],
+                      dino[:weight], dino[:walking])
+    end
+  end
+
+  def dino_search(queries)
```
jmmastey: This naming thing is getting smurfy. Just call it `search`. :)

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,103 @@
...
+    dino_results = []
+    queries.each { |search, query| dino_results.concat send(search, query) }
```
jmmastey: This use of `send` might be a little too vague. While reading this method, I have no idea how to *operate* the method. Hopefully it'll be somewhere else, but I think you could get this method to the point where its operation was clear from reading the code.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    dino_results.uniq!.each do |dino|
+      puts "----------"
+      dino_facts(dino)
+    end
+  end
+
+  def name_search(*dino_names)
+    @dino_dex.select { |dino| dino_names.include? dino.name }
+  end
```
jmmastey: These splats are almost certainly unnecessary.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  def walking_search(*walk_styles)
+    @dino_dex.select { |dino| walk_styles.include? dino.walking }
+  end
+
+  def period_search(*periods)
+    @dino_dex.select do |dino|
+      dino.period.split.any? { |period| periods.include? period }
+    end
+  end
```
jmmastey: This particular implementation detail (what constitutes being "from" the Jurassic?) is probably better handled within the `Dino` class. I don't think the `Dex` itself is the best place to contain that knowledge, mostly because dinosaurs otherwise manage their own information.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,103 @@
...
+    elsif diet.eql? "Herbivore"
+      herbivores
+    end
+  end
+
+  def big_dinos
+    @dino_dex.select do |dino|
+      dino.weight_in_lbs > 4000 if dino.weight_in_lbs
+    end
+  end
```
jmmastey: Ditto all these things and where this knowledge lives. Probably in the dino class.

dino_catalog/dino_catalog.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  def small_dinos
+    @dino_dex.select do |dino|
+      dino.weight_in_lbs < 4000 if dino.weight_in_lbs
+    end
+  end
+
+  def carnivores
+    @dino_dex.select do |dino|
+      %w(Carnivore Insectivore Piscivore Yes).include? dino.diet
```
jmmastey: This is an interesting place to illustrate where the parser should be doing more work. At face value, we can see there's something weird when we say "A dinosaur is a carnivore if it's diet is Carnivore, Insectivore, Piscivore, or Yes". Another way to think about it is that I could provide you with another CSV that contains a "Herbivore" column, and now your design would be in trouble.

Instead, the parser (which should be the only place that understands the details of what's in those CSVs) should clean up the data to conform to a sensible standard (in this case, rolling Yes/No into Carnivore/Herbivore). Then your application logic becomes *way* simpler.

dino_catalog/dino_catalog_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'rspec'
+require_relative 'dino_catalog'
+
+describe DinoCatalog do
+  before(:each) do
+    @dino_log = DinoCatalog.new
+    @dino_log.parse('dinodex.csv')
+    @dino_log.parse('african_dinosaur_export.csv')
```
jmmastey: Be careful about using production data as test data. This way leads to brittle, slow tests. For instance, if the production data changed so that there was one extra bipedal dinosaur, your tests would fail, telling you that it did not successfully search by walking style. Did it really fail? No, of course not, but the test depends on a really specific outcome.

This setup makes some assumptions about the idea that `.parse` will actually correctly parse. And finally, it means that I can't actually verify that the tests do what I think they do unless I go open more CSV files. In this case, I have to mentally coalesce two CSVs to make sense of them.

A better way to do this would be to create Dinosaurs on the fly for tests in a way that reveals the intent of the test:

```
it "should search by Dino period" do
  match = create_dino(period: "Cretaceous")
  partial_match = create_dino(period: "Late Cretaceous")
  no_match = create_dino(period: "3rd")
  @dino_log.add_dinos([match, partial_match, no_match])

  results = @dino_log.period_search("Cretaceous")

  expect(results.length).to eq(2)
  expect(results).not_to include(no_match)
end

Comment (jmmastey) [inline dino_catalog/dino_catalog_spec.rb]: Note that this test is technically more verbose, but doesn't depend on data from other places.

File: dino_catalog/dino_catalog_spec.rb [hunk 15]
```diff
@@ -0,0 +1,67 @@
+require 'rspec'
+require_relative 'dino_catalog'
+
+describe DinoCatalog do
+  before(:each) do
+    @dino_log = DinoCatalog.new
```
jmmastey: Use `let` for this. ivars (`@whatever`) are an anti-pattern in spec tests. (see http://betterspecs.org/)

dino_catalog/dino_catalog_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  it "should return Dinos that are carnivores " do
+    dino_results = @dino_log.carnivores
+    expect(dino_results).to include(a_kind_of(Dino))
+    expect(dino_results.count).to be(12)
+  end
+
+  it "should search by Dino period" do
+    dino_results = @dino_log.period_search("Cretaceous")
+    expect(dino_results).to include(a_kind_of(Dino))
```
jmmastey: I'm not sure these `a_kind_of(Dino)` expectations add much. What else would come back? If there is a risk here, test it once at the top and be done with it.

Also, if you follow something like the pattern I gave above, you can assert the _specific_ Dino, which implicitly includes its class.

jmmastey:
Alright, there we go. I gave a lot of higher level advice because the code itself is actually really solid. Well done on that. Many of the things that I'd do differently are really concerned with what the application will look like months down the road.

Also, great job adding tests to this exercise. They're pretty good tests too.

Feel free to submit again if you feel like it. Overall, make sure you understand the feedback, try out some of the changes if they seem confusing, and feel free to move on if you see fit.

Keep it up!

mostlybadfly:
@jmmastey good to know I was actually going somewhere with this! I totally am feeling you on the separation, I was not super keen on all those parsing actions under the Catalog class, I think I definitely have some more insight as to how to proceed. Thanks as usual :)!

PR #1165, opened 2016-02-05
Dinodex, v1

bap13:
Normalizing the data and filtering the dinosaurs got a little messy. I hopefully will be able to think up simpler ways to do those two things.

Thanks for any feedback!

### DIFF
--- a/dino_catalog/dinodex.rb
+++ b/dino_catalog/dinodex.rb
@@ -1,11 +1,23 @@
-# `DinoDex` creates an array of Dinosaur objects from csv data.
 class DinoDex
   require 'csv'
   require_relative 'dinosaur'
-  attr_accessor :dinosaurs
 
   def initialize
     @dinosaurs = create_dinosaurs
+  end
+
+  def all_dinosaurs
+    @dinosaurs
+  end
+
+  def find_by_name(search_string)
+    @dinosaurs.select { |dino| dino.name =~ /^#{search_string}$/i }
+  end
+
+  def filter_by_hash(search_hash)
+    @dinosaurs.select do |dino|
+      search_hash.each { |k,v| break if !(dino.send(k) =~ /#{v}/i) }
+    end
   end
 
   private
@@ -25,16 +37,45 @@ class DinoDex
   def formatted_data(dino)
     {
       name:        dino[:name] || dino[:genus],
-      period:      dino[:period],
+      period:      correct_period(dino[:period]),
       continent:   dino.fetch(:continent, "Africa"),
-      diet:        dino.fetch(:diet) { correct_diet(dino) },
+      diet:        correct_diet(dino),
       weight:      dino[:weight_in_lbs] || dino[:weight],
+      size:        small_or_large(dino),
       walking:     dino[:walking],
       description: dino[:description]
     }
   end
 
   def correct_diet(dino)
-    dino[:carnivore].downcase == 'yes' ? 'Carnivore' : 'Non-carnivore'
+    if dino[:carnivore]
+      fix_carnivore(dino)
+    elsif dino[:diet] =~ /(Piscivore)|(Insectivore)/
+      "Carnivore > #{dino[:diet]}"
+    else
+      dino[:diet]
+    end
+  end
+
+  def fix_carnivore(dino)
+    dino[:carnivore].downcase == 'yes' ? 'Carnivore' : nil
+  end
+
+  def correct_period(dino_period)
+    case dino_period
+    when /Oxfordian/
+      "Late Jurassic > Oxfordian"
+    when /Albian/
+      "Early Cretaceous > Albian"
+    else
+      dino_period
+    end
+  end
+
+  def small_or_large(dino)
+    weight = dino[:weight_in_lbs] || dino[:weight]
+    unless weight.nil?
+      weight > 2000 ? "Large" : "Small"
+    end
   end
 end
--- a/dino_catalog/dinodex.rb
+++ b/dino_catalog/dinodex.rb
@@ -1,0 +1,40 @@
+# `DinoDex` creates an array of Dinosaur objects from csv data.
+class DinoDex
+  require 'csv'
+  require_relative 'dinosaur'
+  attr_accessor :dinosaurs
+
+  def initialize
+    @dinosaurs = create_dinosaurs
+  end
+
+  private
+
+  def create_dinosaurs
+    dino_array = []
+    Dir.glob('*.csv').each do |file|
+      table = CSV.read(file, headers: true, header_converters: :symbol,
+                       converters: :all)
+      table.each do |dino|
+        dino_array << Dinosaur.new(formatted_data(dino))
+      end
+    end
+    dino_array
+  end
+
+  def formatted_data(dino)
+    {
+      name:        dino[:name] || dino[:genus],
+      period:      dino[:period],
+      continent:   dino.fetch(:continent, "Africa"),
+      diet:        dino.fetch(:diet) { correct_diet(dino) },
+      weight:      dino[:weight_in_lbs] || dino[:weight],
+      walking:     dino[:walking],
+      description: dino[:description]
+    }
+  end
+
+  def correct_diet(dino)
+    dino[:carnivore].downcase == 'yes' ? 'Carnivore' : 'Non-carnivore'
+  end
+end
--- a/dino_catalog/dinodexui.rb
+++ b/dino_catalog/dinodexui.rb
@@ -1,0 +1,123 @@
+class DinoDexUI
+  require_relative 'dinodex'
+
+  FILTER_OPTIONS = {
+    period: %w(Cretaceous Jurassic Triassic Permian),
+    continent: %w(North\ America South\ America Europe Africa Asia),
+    diet: %w(Herbivore Carnivore Insectivore Piscivore),
+    size: %w(Small Large),
+    walking: %w(Biped Quadruped)
+  }
+
+  def initialize
+    @dex = DinoDex.new
+    @search_results = []
+    @search_hash = {}
+  end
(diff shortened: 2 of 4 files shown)


dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,81 @@
...
+  require 'csv'
+  require_relative 'dinosaur'
+
+  def initialize
+    @dinosaurs = create_dinosaurs
+  end
+
+  def all_dinosaurs
+    @dinosaurs
+  end
```
jmmastey: Is there a reason to do this, rather than just having an `attr_reader` named `dinosaurs`?

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  def find_by_name(search_string)
+    @dinosaurs.select { |dino| dino.name =~ /^#{search_string}$/i }
+  end
```
jmmastey: Since this search includes `^$`, I think it's equivalent to `==` aside from the case sensitivity. If so, regex is pretty slow and it may be clearer (and more efficient) to use something like `casecmp`).

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  def filter_by_hash(search_hash)
+    @dinosaurs.select do |dino|
+      search_hash.each { |k,v| break if !(dino.send(k) =~ /#{v}/i) }
+    end
+  end
```
jmmastey: I might just call this `filter`. Or `search`. Truth be told, there's no reason why it needs to be a `Hash` object, and generally encoding datatype into names is an antipattern in OO ruby.

jmmastey: Since you mentioned search getting weird, I think it's also plausible that this logic should live in Dinosaur anyway. There's a lot of logic below to handle special cases (by mutating the data for `period` / `diet` etc) that I think would simplify down a bit if you allowed yourself to do special-cases for searching those fields.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,81 @@
...
+    dino_array = []
+    Dir.glob('*.csv').each do |file|
+      table = CSV.read(file, headers: true, header_converters: :symbol,
+                       converters: :all)
+      table.each do |dino|
+        dino_array << Dinosaur.new(formatted_data(dino))
+      end
+    end
+    dino_array
+  end
```
jmmastey: I wish there were a name for this pattern in ruby. Something like "initialize / loop / return". Whenever you see a method in ruby that follows that general structure, see if you can eliminate the "initialize" and "return" steps, using some of ruby's (fantastic amazeballs awesome) iteration methods. So here:

```
def create_dinosaurs
  csv_files = Dir.glob('*.csv')
  csv_files.inject { |array, file| array += parse(file) }
end

def parse
  # all the parse-y stuff
end

Comment (jmmastey) [inline dino_catalog/dinodex.rb]: Also, I think that this method pretends to be very general (create ALL the dinosaurs!) while being overly specific to just the two CSVs that we currently have access to. We almost certainly will end up with other data formats, and while we don't want to over-build for the future, we should at least consider that future.

At the minimum, I would probably move the parsing code *out* of the class where searching code lives. That also makes our code more closely conform to the Single Responsibility Principle. (comment continues below...)

File: dino_catalog/dinodex.rb [hunk 5]
```diff
@@ -0,0 +1,81 @@
...
+
+  def formatted_data(dino)
+    {
+      name:        dino[:name] || dino[:genus],
+      period:      correct_period(dino[:period]),
+      continent:   dino.fetch(:continent, "Africa"),
```
jmmastey: (...continued!)

And this sort of overly-specific fix should be quarantined to one specific location if possible. That way, when I add a third CSV (`asian_dinos.csv`, with an `is_herbivore` field, of course), the logic is easier to extend and keep clean. And even if it *does* get dirty, it lives in the parsing code and never makes its way up to the rest of the program.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+  end
+
+  def small_or_large(dino)
+    weight = dino[:weight_in_lbs] || dino[:weight]
+    unless weight.nil?
+      weight > 2000 ? "Large" : "Small"
+    end
+  end
+end
```
jmmastey: These few methods are a good attempt at getting search to be a simple regex match. I think, though, that you gave up more than you gained. I now cannot search by specific weights for dinosaurs. If I export my data by JSON (see the README) I get different data than what I started with. And in fact, if I run the program _multiple_ times that way I now start to get Periods that look like `Late Jurassic > Late Jurassic > Oxfordian`.

In this case, I think the verbosity of having a longer search method is probably worth the clarity.

dino_catalog/dinodex.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def fix_carnivore(dino)
+    dino[:carnivore].downcase == 'yes' ? 'Carnivore' : nil
+  end
+
+  def correct_period(dino_period)
+    case dino_period
+    when /Oxfordian/
+      "Late Jurassic > Oxfordian"
```
jmmastey: I had to look this up. Thanks for teaching me something about Dinosaurs :D

dino_catalog/dinodexui.rb
```diff
@@ -0,0 +1,123 @@
+class DinoDexUI
+  require_relative 'dinodex'
+
+  FILTER_OPTIONS = {
+    period: %w(Cretaceous Jurassic Triassic Permian),
```
jmmastey: I am a little sad, though, about losing the ability to find only Albian dinosaurs.

dino_catalog/dinodexui.rb
```diff
@@ -0,0 +1,123 @@
+class DinoDexUI
```
jmmastey: I'm not going to put in depth comments on this file. Overall: smart idea separating the UI from the rest of the entire library. Most people don't do that at first. :)

This seems to be a perfectly fine way to handle the UI. We could argue about what a more "custom search" would look like, but I think this satisfies the requirements just fine.

dino_catalog/dinodexui.rb
```diff
@@ -0,0 +1,123 @@
...
+  FILTER_OPTIONS = {
+    period: %w(Cretaceous Jurassic Triassic Permian),
+    continent: %w(North\ America South\ America Europe Africa Asia),
+    diet: %w(Herbivore Carnivore Insectivore Piscivore),
+    size: %w(Small Large),
+    walking: %w(Biped Quadruped)
+  }
+
+  def initialize
+    @dex = DinoDex.new
```
jmmastey: It *is* a little mysterious and surprising that the DinoDex comes with initialized data from specific CSVs on the filesystem. I'd have expected either calling `create_dinosaurs` or initializing with filenames.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,15 @@
+class Dinosaur
+  attr_reader :name, :period, :continent, :diet, :weight, :walking,
+              :description, :size, :to_h
```
jmmastey: Fix thy hound sins :)

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  def initialize(args = {})
+    @to_h = args
```
jmmastey: Ehhhhhhhhhh. `to_h` should almost certainly be a method. I've never really seen this construction before, and if you ever added methods that change data internally (a likelihood), you'd introduce a subtle bug.

dino_catalog/dinosaur.rb
```diff
@@ -0,0 +1,15 @@
...
+    args.each { |k,v| instance_variable_set("@#{k}", v) }
+  end
+
+  def to_s
+    Array(to_h).reject { |pair|
```
jmmastey: `Array()` probably unnecessary here. Hashes are already Enumerable. Also, this is a surprising amount of chained logic in a to_s method.

dino_catalog/executable_dinodex_script.rb
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env ruby
+require_relative 'dinodexui'
+
+DinoDexUI.new.run
```
jmmastey: Even a proper runner. Yay!

jmmastey:
Hey Brett, I think you already see which parts work and which don't. I offered a few ideas of ways to reconfigure the code that should help with the weirdness in parsing and searching. Overall, very solid code, and I like the separation of concerns that you've created for most of it.

Try moving (and cleaning up) the parsing and see if the rest doesn't end up much simpler...

dino_catalog/dinosaur.rb
kotp: `to_h` is a method.  See the `attr_reader` on line 2.

dino_catalog/dinodex.rb
bap13: I wanted to have the public interface of `DinoDex` limited to three methods: `#all_dinosaurs`, `#find_by_name()`, and `filter_by_hash()`. I realize now I could have put `attr_reader :dinosaurs` under the `private` keyword, but I'm not sure if that's proper convention.

dino_catalog/dinosaur.rb
bap13: Yeah... I'm getting rid of this in the next iteration. I did this out of laziness so I wouldn't have to rebuild the same exact hash from the instance variables. And I mainly just wanted `to_h` so it would make the `to_s` method easier to write. Good point, Joe, that if I add methods that change the data, the `to_h` method would give you the wrong info.

PR #1167, opened 2016-02-10
Data science exercise - Iteration 2

locker1776:


data_science/data_science.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def conversion_range(cohort)
+    num_conversion = @stats.num_of_conversions(cohort)
+    total = @stats.size_of_cohort(cohort)
+    ABAnalyzer.confidence_interval(num_conversion, total, 0.95)
+  end
+
+  def analyze
+    ab_test = ABAnalyzer::ABTest.new @stats.prep_data
```
zsyed1: Put parentheses around function arguments: `ab_test = ABAnalyzer::ABTest.new(@stats.prep_data)`

data_science/spec/integration_spec.rb
```diff
@@ -0,0 +1,23 @@
+require './data_science'
+
+describe "Testing the integration with the abanalyzer gem" do
+  context "When the data set for two cohorts are the same" do
+    it "The abanalyzer gem should return false" do
+      test_array =  visits(20, "B", 0)
+      test_array += visits(20, "A", 0)
+      test_array += visits(5, "B", 1)
+      test_array += visits(5, "A", 1)
```
zsyed1: You should move the construction of `test_array` into a `let`

data_science/spec/integration_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+  end
+
+  context "When the data set for two cohorts are very different" do
+    it "The abanalyzer gem should return true" do
+      test_array =  visits(20, "B", 0)
+      test_array += visits(20, "A", 1)
+      test_array += visits(5, "B", 1)
+      test_array += visits(5, "A", 0)
+      expect(DataScience.new(test_array).analyze).to be_truthy
```
zsyed1: Interesting test cases.

data_science/data_science.rb
```diff
@@ -0,0 +1,58 @@
...
+  def display_meta_data
+    puts "Running an analysis on the following data:"
+    @stats.prep_data.each { |k, v| puts "Cohort #{k}: #{v}" }
+  end
+
+  def conversion_ranges
+    %w(A B).each do |cohort|
+      range = conversion_range(cohort)
+      puts "95% Conversion range for #{cohort} is #{range}"
+    end
```
zsyed1: The name of this method should be something like `def display_conversion_ranges` just like you did in the method above. Otherwise it reads as though the method will return the ranges.

data_science/data_science.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def test_hypothesis(p_value)
+    if p_value >= 0.05
+      puts "No correlated evidence from the data."
+      false
+    else
+      puts "Data shows that the evidence is correlated!"
+      true
+    end
```
zsyed1: The responsibility of `test_hypothesis` should be to run the test. Your app could then use the result to display the appropriate message. In your tests below you validate the true/false value so that seems to be the primary function.

data_science/spec/data_from_file_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      allow(File).to receive(:read).and_return(json_data)
+    end
+
+    it "returns an array of hashes" do
+      expect(loaded_data).to be_an(Array)
+    end
+
+    it "includes hashes from the loaded file" do
+      expect(loaded_data).to include(file_data.first)
+      expect(loaded_data).to include(file_data.last)
```
zsyed1: Since the data set is only 2 items, you can just test that the entire data-set (and nothing else) is loaded.

data_science/website_stats.rb
```diff
@@ -0,0 +1,31 @@
...
+  def cohort_values(cohort)
+    { not_converted: num_no_conversions(cohort),
+      converted: num_of_conversions(cohort) }
+  end
+
+  def prep_data
+    # Data is prepared according to API https://github.com/bmuller/abanalyzer
+    { A: cohort_values("A"), B: cohort_values("B") }
+  end
+end
```
zsyed1: Very clean. Nice work.

paulghaddad:
@locker1776 Really nice, clean work on this iteration! No need to resubmit.

PR #1169, opened 2016-02-13
Dinodex, v2

bap13:
I decided to have some fun and make this into a Sinatra app. It lives here: [DinoDex App](https://pacific-bastion-15005.herokuapp.com).

Some major changes since v1:
  - Parsing logic is now separated from searching logic.
  - Specific data-correcting logic lives in subclasses of `DinoDataParser`. This way if another data file is added later, all that needs to be done is create a new subclass of `DinoDataParser` and implement its `#format_data` and `#data_file` methods.
  - `Dinosaur` is now just an `OpenStruct` with a fancier `#to_s` method. I realized a `Dinosaur` is just a bucket of attributes with getters and setters, i.e. an `OpenStruct`, so why reinvent the wheel?
  - Filtering logic inside `DinoDex` is slightly better. There is some unsightly cleaning up of the `params` hash that comes from the requests, but I found no better place to put that code than inside `DinoDex`.
  - `JSON` export is now a thing.

Some things I still would like to improve (but probably won't because I've already spent a lot of time on this):
  - I used the old hash syntax so I could use symbols with spaces, like this `:'period subdivision'`. I did this to make printing and searching easier, without having to `gsub` an underscore a bunch of times. I'd like to find a better way to do this.
  - I don't really like having `FILTERS` hardcoded in DinoDex. I can't think of a better way to do that, but it just feels wrong and difficult to manage in the event that more data is added.

Overall, I had fun making this. Hopefully you enjoy it as much as I did. I appreciate any feedback!


dino_catalog/lib/dino_data_parser.rb
```diff
@@ -0,0 +1,19 @@
...
+  require 'csv'
+  require_relative 'dinosaur'
+
+  def parse
+    table = CSV.read(data_file, headers: true, header_converters: :symbol,
+                     converters: :all)
+    table.reduce([]) { |array, dino| array << Dinosaur.new(format_data(dino)) }
+  end
+
+  def format_data(dino)
```
bap13: I can see why hound gives this note, but I think having `def format_data(dino)` gives you a hint as to how this template method should be implemented.

jmmastey:
Oh man, finally web interface on the Dinodex. :dancers:

dino_catalog/lib/african_data_parser.rb
```diff
@@ -0,0 +1,31 @@
...
+      %w(Early\ Cretaceous Albian)
+    else
+      [period, nil]
+    end
+  end
+
+  def carnivore_or_not(diet)
+    diet.casecmp('yes') == 0 ? 'Carnivore' : nil
+  end
+end
```
jmmastey: This form of massaging data is pretty clean. Good call. Also quite extensible if new CSVs or other data sources appear later.

Small side note: there were vanishingly few omnivorous dinosaurs (it wasn't really a thing yet, evolutionarily), so you could pretty safely call non-carnivores herbivorous.

dino_catalog/lib/dino_data_parser.rb
jmmastey: I think you're more right than hound in this case.

dino_catalog/lib/dinodex.rb
```diff
@@ -0,0 +1,51 @@
...
+  require 'json'
+
+  FILTERS = {
+    :period => %w(jurassic cretaceous triassic permian),
+    :'period subdivision' => %w(oxfordian albian),
+    :continent => %w(north\ america south\ america asia africa europe),
+    :diet => %w(carnivore herbivore),
+    :'diet subtype' => %w(piscivore insectivore),
+    :walking => %w(biped quadruped),
+  }
```
jmmastey: This method of declaring symbols (`:'diet subtype'`) has always been super weird to me. How does it seem to you?

dino_catalog/lib/dinodex.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def find_match(matchable_params)
+    return [] if matchable_params.empty?
+    all_dinosaurs.select do |dino|
+      matchable_params.each do |k, v|
+        break unless dino.send(k) =~ /#{v}/i
+      end
+    end
+  end
```
jmmastey: I think `.all?` may be more appropriate here, given the structure of the method.

dino_catalog/lib/dinodex.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def find_weight(min, max)
+    return [] if min.nil? && max.nil?
+    min = (min &&= min.to_i) || 0
```
jmmastey: This is a tad complicated. Given zero is truthy in ruby, you could probably just do:

`min = min.to_i || 0`

dino_catalog/lib/dinodex.rb
```diff
@@ -0,0 +1,51 @@
...
+    max = (max &&= max.to_i) || Float::INFINITY
+    all_dinosaurs.select do |dino|
+      next if dino.weight.nil?
+      dino.weight.between?(min, max)
+    end
```
jmmastey: Using `next` with `select` can be hard to reason about. I might delegate this to a method on dino that handled this logic. Probably not a big deal.

dino_catalog/lib/dinodex.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def export_json
+    export = all_dinosaurs.reduce([]) { |a, dino| a << dino.to_h }
```
jmmastey: If all_dinosaurs is already an array (it appears that it is), just use `all_dinosaurs.to_h`

dino_catalog/lib/dinosaur.rb
```diff
@@ -0,0 +1,11 @@
...
+class Dinosaur < OpenStruct
+  def to_s
+    valid_fields = to_h.reject { |_k, v| v.nil? }
+    valid_fields.each_with_object("") do |(k, v), output|
+      output << "#{k.capitalize}: #{v}"
+      output << " lbs" if k == :weight
+      output << "\n"
+    end
+  end
+end
```
jmmastey: Be careful with OpenStruct. Its performance characteristics are garbage. Also, if you find yourself defining methods on an OpenStruct, it probably deserves a real class instead.

dino_catalog/public/js/script.js
```diff
@@ -0,0 +1,6 @@
+$(document).ready(function() {
```
jmmastey: Can't comment on an image, so I'll just say:
1) I know that picture!
b) Good choice.

jmmastey:
This looks super clean. Well done.

PR #1172, opened 2016-03-28
Data Science v1

brettapeters:
I'd like a better way of dealing with sample data than what I did. I loaded a test_data.json file for each test case, which feels like the wrong approach. I only wrote tests for the public methods of my classes.

To run: `$ ./bin/data_science <json filename>`

Any feedback is always greatly appreciated!


data_science/bin/data_science
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env ruby
+
+require_relative '../lib/data_science'
+
+raise ArgumentError, "No file given" unless (file = ARGV[0])
```
jmmastey: This style is pretty normal in C but less so in Ruby. By putting the real work of the line on the right hand side, it makes it harder to understand the code when scanning. See also this article: therealadam.com/2016/03/23/one-idea-per-line/

data_science/lib/data_science/analyzer.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    def p_value
+      ABAnalyzer::ABTest.new(data.format_for_analyzer).chisquare_p
+    end
+
+    def different?
+      ABAnalyzer::ABTest.new(data.format_for_analyzer).different?
+    end
+  end
+end
```
jmmastey: I really like that you wrapped the ABAnalyzer dependency in here. Good call.

data_science/lib/data_science/stats.rb
```diff
@@ -0,0 +1,48 @@
...
+    def to_s
+      @output.values.join("\n")
+    end
+
+    private
+
+    def sample_size
+      cohorts.inject([]) do |memo, cohort|
+        memo << "Cohort #{cohort} Sample Size: #{@data.sample_size(cohort)}"
+      end
```
jmmastey: I might use `map` here instead of inject since we're already coming from something iterable.

data_science/lib/data_science/stats.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def conversion_rate
+      interval = 0.95
```
jmmastey: This magic integer belongs somewhere in configuration, or at least at the top of the class.

data_science/lib/data_science/stats.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    def different
+      "The results are" \
+      "#{@analyzer.different? ? '' : ' not'} statistically " \
+      "significantly different."
+    end
+
+    def cohorts
+      @data.data.map(&:cohort).uniq.sort
+    end
```
jmmastey: I do this too ;) This line is very compact, but ultimately difficult to read. If you leave it as-is (which is probably fine), I'd suggest adding a short comment demonstrating the transformation it performs.

Fwiw, as I look at it, it may actually be too coupled to the particular data format already provided as well.

data_science/lib/data_science/website_data.rb
```diff
@@ -0,0 +1,47 @@
...
+      values
+    end
+
+    private
+
+    Entry = Struct.new(:date, :cohort, :converted)
+
+    def convert_data
+      parse_data.inject([]) do |a, e|
+        a << Entry.new(e["date"], e["cohort"], e["result"] == 1 ? true : false)
```
jmmastey: If you have a ternary that ends with `true : false`, just include the predicate (`e["result"] == 1`) and drop the ternary.

data_science/lib/data_science/stats.rb
jmmastey: Actually, I take it back. I see below where you're standardizing the data.

data_science/lib/data_science/website_data.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    def format_for_analyzer
+      values = {}
+      values[:a] = { converted: conversion_count('A'),
+                     not_converted: non_conversion_count('A') }
+      values[:b] = { converted: conversion_count('B'),
+                     not_converted: non_conversion_count('B') }
+      values
+    end
```
jmmastey: This is not my very favorite part, since it's pretty specific. I'd prefer to see it able to handle more than the letters A and B. On the other hand, we know specifically where to go to change this, so at least any issues there are quarantined.

Without more requirements around data with other cohort names, this may be just fine. If you did want to change it, you could grab the cohorts w/ `@cohorts = @data.select(&:cohort).uniq` in the constructor and then iterate over it here.

data_science/spec/analyzer_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'spec_helper'
+
+describe DataScience::Analyzer do
+  before(:all) do
+    data = DataScience::WebsiteData.new("test_data.json")
+    @analyzer = DataScience::Analyzer.new(data)
+    @cohort = 'B'
```
jmmastey: I'm not positive that setting the cohort here is productive, versus hardcoding it in the test. The indirection doesn't really buy you anything, and now you have to keep it in mind.

data_science/spec/analyzer_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'spec_helper'
+
+describe DataScience::Analyzer do
+  before(:all) do
+    data = DataScience::WebsiteData.new("test_data.json")
```
jmmastey: Use `let` for this sort of thing.

Also, I'm glad you didn't use the exported data file provided ultimately. Still, having the data itself abstracted to a different file is another thing that I need to look up if the tests are failing or I want to understand _why_ the output is whatever it is. It'd be better to have data that is generated or specified within the test file itself.

data_science/spec/analyzer_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  describe '#p_value' do
+    it 'gives the p-value for the website data' do
+      expect(@analyzer.p_value).to be_within(0.0001).of(0.0001)
+    end
+  end
+
+  describe '#different?' do
+    it 'tells if results are statistically significantly different' do
+      expect(@analyzer.different?).to be true
```
jmmastey: `be_truthy`, though `be_different` may work as well.

data_science/spec/analyzer_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      interval = 0.95
+      error_bar = @analyzer.conversion_rate_error_bar(@cohort, interval)
+      expect(error_bar[0]).to be_within(0.0001).of(0.0062)
+      expect(error_bar[1]).to be_within(0.0001).of(0.0208)
+    end
+  end
+
+  describe '#p_value' do
+    it 'gives the p-value for the website data' do
+      expect(@analyzer.p_value).to be_within(0.0001).of(0.0001)
```
jmmastey: So this test is correct, and I like the description. However, it's hard to understand what exactly is happening here. Truthfully, when dealing with a dependency under the hood, the better approach might simply be to specify that ABAnalyzer gets called correctly.

If you were to continue testing for correct values, I think you'd need to have enough test cases to cover lower p-values and not-different results, etc etc.

data_science/spec/data_science_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'spec_helper'
+
+describe DataScience do
+  it 'has a version number' do
+    expect(DataScience::VERSION).not_to be nil
+  end
+end
```
jmmastey: I'm not sure this test adds anything at all :)

data_science/spec/stats_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'spec_helper'
+
+describe DataScience::Stats do
+  before do
+    data = DataScience::WebsiteData.new("test_data.json")
```
jmmastey: This is in danger of coupling too hard to WebsiteData. If you can, mocking it out might be better. Especially since you now can't test the stats generator without parsing / loading / computing on an entire test data file. That's a good sign of too much coupling.

data_science/lib/data_science/website_data.rb
```diff
@@ -0,0 +1,47 @@
...
+    Entry = Struct.new(:date, :cohort, :converted)
+
+    def convert_data
+      parse_data.inject([]) do |a, e|
+        a << Entry.new(e["date"], e["cohort"], e["result"] == 1 ? true : false)
+      end
+    end
+
+    def parse_data
+      JSON.parse(File.read(@filename))
```
jmmastey: It may or may not be okay that this class is now unusable over an API, unless I generate a file to parse. We didn't specify any further that you'd get API data, but it's worth at least being aware of.

data_science/spec/website_data_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+describe DataScience::WebsiteData do
+  before(:all) do
+    @website_data = DataScience::WebsiteData.new("test_data.json")
+    @cohort = 'A'
+  end
+
+  describe '#sample_size' do
+    it 'gives the sample size of a cohort' do
+      expect(@website_data.sample_size(@cohort)).to eq(732)
+    end
```
jmmastey: So same here as above, the tests are too coupled to the actual test data itself. This operates as verification, but not terribly well as documentation.

jmmastey:
Hey Brett, this code is pretty solid. The biggest thing I'd try to do is break the dependencies within the tests themselves so that the classes don't all rely on each other for testing, and then to change the tests to include their own data and not just asserting the particular outcome of the test_data. That way, it'll be easier to reason about the actual function of the underlying code.

Thanks!

PR #1173, opened 2016-05-25
Refactor triangle code. Remove explicite returns. Use standard if sta…

aalcesto23:
…tement formats for readability.

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -3,34 +3,40 @@ class Triangle
 	attr_accessor :side1,:side2,:side3
 
 	def initialize(side1,side2,side3)
-		@side1,@side2,@side3 = side1,side2,side3
+		@side1 = side1
+		@side2 = side2
+		@side3 = side3
 	end
 
-	def equilateral()
-		return side1 == side2 && side2 == side3
+	def equilateral
+		side1 == side2 && side2 == side3
 	end
 
-	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
+	def isosceles
+		[side1,side2,side3].uniq.length == 2
 	end
 
-	def scalene()
-		unless ! (equilateral || isosceles)
-	    false
-	  else
-	    true
-	  end
+	def scalene
+		if !(equilateral || isosceles)
+			false
+		end
 	end
 
 	def recite_facts
-		puts 'This triangle is equilateral!' if equilateral
-		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+		if equilateral
+			puts 'This triangle is equilateral!'
+		elsif isosceles
+			puts 'This triangle is isosceles! Also, that word is hard to type.'
+		else
+			puts 'This triangle is scalene and mathematically boring.'
+		end
 
 		angles = self.calculate_angles(side1,side2,side3)
 		puts 'The angles of this triangle are ' + angles.join(',')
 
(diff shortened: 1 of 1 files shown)


triangle_facts/triangle_facts.rb
```diff
@@ -3,34 +3,40 @@ class Triangle
...
-	def scalene()
-		unless ! (equilateral || isosceles)
-	    false
-	  else
-	    true
-	  end
+	def scalene
+		if !(equilateral || isosceles)
+			false
+		end
```
jmmastey: So I think this changed the semantics of the method. But either way, think about what the output of the conditional is here. Whenever you see

```
if thing
   true
else
  false
end
```

Instead you can just reduce to just the predicate: `thing`. Since we're just doing true/false anyway, this is equivalent.

Also, predicate methods in ruby should be named with a `?`, as `scalene?`

triangle_facts/triangle_facts.rb
```diff
@@ -3,34 +3,40 @@ class Triangle
 	attr_accessor :side1,:side2,:side3
 
 	def initialize(side1,side2,side3)
-		@side1,@side2,@side3 = side1,side2,side3
+		@side1 = side1
+		@side2 = side2
+		@side3 = side3
```
jmmastey: Tabs == booo.


Configure your editor to always use 2 spaces.

triangle_facts/triangle_facts.rb
```diff
@@ -3,34 +3,40 @@ class Triangle
...
-		puts 'This triangle is equilateral!' if equilateral
-		puts 'This triangle is isosceles! Also, that word is hard to type.' if isosceles 
-		puts 'This triangle is scalene and mathematically boring.' if scalene 
+		if equilateral
+			puts 'This triangle is equilateral!'
+		elsif isosceles
+			puts 'This triangle is isosceles! Also, that word is hard to type.'
+		else
+			puts 'This triangle is scalene and mathematically boring.'
+		end
```
jmmastey: This is probably fine. It might be more concise as a case statement.

triangle_facts/triangle_facts.rb
```diff
@@ -3,34 +3,40 @@ class Triangle
...
 
 		angles = self.calculate_angles(side1,side2,side3)
```
jmmastey: Put spaces after arguments.

triangle_facts/triangle_facts.rb
```diff
@@ -3,34 +3,40 @@ class Triangle
...
 		puts 'The angles of this triangle are ' + angles.join(',')
 
-		puts 'This triangle is also a right triangle!' if angles.include? 90
+		if angles.include? 90
```
jmmastey: When a method call includes a parameter, use parentheses.

triangle_facts/triangle_facts.rb
```diff
@@ -39,11 +45,11 @@ def calculate_angles(a,b,c)
 		angleB = radians_to_degrees(Math.acos((a**2 + c**2 - b**2) / (2.0 * a * c)))
 		angleC = radians_to_degrees(Math.acos((a**2 + b**2 - c**2) / (2.0 * a * b)))
 
-		return [angleA, angleB, angleC]
+		[angleA, angleB, angleC]
```
jmmastey: Correct to remove explicit returns, but these variables are named more like JavaScript variables than Ruby ones.

jmmastey:
Added a few more suggestions, but this is mostly correct. In future assignments (it's disabled here), just run rubocop to find all these things.

PR #1174, opened 2016-05-25
Aalcesto23

aalcesto23:


jmmastey:
Looks like you've got commits mixed in from the previous exercise. I'll take a look at this one in a moment, but make sure to keep a clean branch without prior commit history.

aalcesto23:
Hey Joe,

Sorry about that, I’ll make sure to use a new branch moving forward. On a side note, I’m working on the Robot Name challenge and having some trouble. Would it be possible to submit what I’ve got and get some feedback?

Regards,
Alan

jmmastey:
Yeah, absolutely.

Joseph Mastey
@jmmastey <http://twitter.com/jmmastey>| jmmastey@gmail.com | 815.228.6924
| http://josephmastey.com | http://leveluprails.com

arrowhead/runner.rb
```diff
@@ -0,0 +1,5 @@
+require_relative 'arrowhead'
+
+puts Arrowhead.classify_region(:northern_plains, :bifurcated)
+puts Arrowhead.classify_region(:poop, :bifurcated)
+puts Arrowhead.classify_region(:northern_plains, :poop)
```
jmmastey: This seems like a reasonable refactor.

arrowhead/classifications.rb
```diff
@@ -0,0 +1,18 @@
+class Classifications
```
jmmastey: Not sure this is an improvement. Really we'd be putting this in a database, but the focus is more on fixing the other lines of code. It's probably not worse, but it does increase the level of abstraction of the code overall without providing much of a clear benefit.

arrowhead/arrowhead.rb
```diff
@@ -1,34 +1,25 @@
...
+    shapes = Classifications::CLASSIFICATIONS[region]
+
+    if shapes.include? shape
+      arrowhead = shapes[shape]
+      "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
+    else
+      "Unknown shape value. Are you sure you know what you're talking about?"
+    end
+  end
+
```
jmmastey: This can probably go farther in getting to short methods and lines of code. This method as written really has two main responsibilities: It finds a classification and then returns some message based on it. The best approach I could see for this would be to split those two domains. A partial refactor would probably look something like this:

```
def self.classify_type(region, shape)
    arrowhead = classify(region, shape)
    if arrowhead
      "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
    else
      "Unknown shape value. Are you sure you know what you're talking about?"
    end
end
```

This has the advantage of separating display and classification logic. That means that when we change off of those ugly configuration hashes, we wouldn't need to change this method. What I'd like you to try to do is to do the rest of that refactor, and see if you can get the other missing region message consistently into the method as well.

Make sense?

arrowhead/arrowhead.rb
```diff
@@ -1,34 +1,25 @@
...
-        "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
-      else
-        raise "Unknown shape value. Are you sure you know what you're talking about?"
-      end
+  def self.classify_region(region, shape)
+    if Classifications::CLASSIFICATIONS.include? region
+      self.classify_type(region, shape)
     else
-      raise "Unknown region, please provide a valid region."
+      "Unknown region, please provide a valid region."
```
jmmastey: You're changing semantics here, so this isn't strictly speaking a refactor. Still, yeah, it may be a stupid place for a raise.

arrowhead/arrowhead.rb
jmmastey: Another way to think about it, take a look at this talk if you have time: http://confreaks.tv/videos/railsconf2014-all-the-little-things

Around five minutes in, she talks about the squint test. That's a good lens for looking at what needs love in this class.

robot_name/robot_name.rb
```diff
@@ -1,4 +1,27 @@
-class NameCollisionError < RuntimeError; end
```
jmmastey: Okay, you definitely need to use different branches for pulls :)

aalcesto23: Hey Joe

Just want to make sure, should I be merging to my branch and then deleting the branch?

Regards,
Alan

robot_name/robot_name.rb
```diff
@@ -12,14 +35,26 @@ def initialize(args = {})
     if @name_generator
       @name = @name_generator.call
     else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
-
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
+      name_constructor
```
jmmastey: This is a good start (it's called the ExtractMethod pattern). We can go further, though. Look at the `method()` method on Kernel for a way to eliminate this if/else structure.

Also, the original code here had a form of symmetry that helps. `@name` is only assigned within `initialize`. With the change you have here it's now set in two places, which will lead to bugs.

robot_name/robot_name.rb
```diff
@@ -12,14 +35,26 @@ def initialize(args = {})
...
     end
+    add_name
+  end
 
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
+  def add_name
     @@registry << @name
+    p @@registry
+  end
```
jmmastey: Where's my dupe and format check? :(

robot_name/robot_name.rb
```diff
@@ -12,14 +35,26 @@ def initialize(args = {})
...
+
+  def name_constructor
+    generate_char = -> { ('A'..'Z').to_a.sample }
+    generate_num = -> { rand(10) }
+    @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
```
jmmastey: So this is a good start, but again we need to go further. Line 51 is just plain too long, and there's no _good_ reason why `generate_char` and `generate_num` should be structured as lambdas within this method.

robot_name/robot_name.rb
```diff
@@ -12,14 +35,26 @@ def initialize(args = {})
...
+  end
+
+  def duplicate_name?
+    if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
+      raise NameCollisionError, 'There was a problem generating the robot name!'
+    end
```
jmmastey: It's funny, this validation is actually a lie. I wasn't paying attention when I wrote it. Anyway, you don't need to fix that if you don't want to, but notice that it doesn't do what it says it does.

jmmastey:
Cool, commented.  Good start.

Reopen with separate branches for the different exercises. Take a look at `git cherry-pick` for more on how to do this. Basically you need a clean master branch, then make a topic branch and cherry-pick in the relevant commits.

👍

robot_name/robot_name.rb
jmmastey: See other comment. Get a clean master, make a topic branch for `robot-name`, then cherry-pick the robot related commits into it. Otherwise we collect old dead assignments here.

PR #1176, opened 2016-05-28
Refactor methods and fix error check.

aalcesto23:
Hey Joe,

I looked over your notes and made as many revisions as possible. These revisions were made on the same branch as the meaningful names exercise. Let me know what you think!

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -1,4 +1,8 @@
-class NameCollisionError < RuntimeError; end
+class NameCollisionError < RuntimeError
+  def initialize(msg='There was a problem generating the robot name!')
+    super
+  end
+end
 
 class Robot
   attr_accessor :name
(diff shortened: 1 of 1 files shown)


robot_name/robot_name.rb
```diff
@@ -7,26 +11,34 @@ class Robot
...
+    @@registry << @name
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
+  def name_constructor
+    generate_char = ('A'..'Z').to_a.sample(2)
+    generate_num = (1..10).to_a.sample(3)
+    @name = generate_char.concat(generate_num).join
+  end
```
jmmastey: This method is probably the biggest target for refactoring. It's kinda-sorta short, but it still contains four pieces of knowledge:

1. How to construct a single char
2. How to construct a single digit
3. How chars and digits are combined to create a name
4. Where to assign the name variable to

Try breaking this method up more and see how that affects your other code.

jmmastey: (actually the `@name = ` is redundant right now because of line 15, so just delete that one ;P

robot_name/robot_name.rb
```diff
@@ -7,26 +11,34 @@ class Robot
...
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  def register_name
+    raise NameCollisionError if bad_name?
+    @@registry << @name
+  end
```
jmmastey: Some folks have argued about whether the format validation stuff belongs in here. I think it's probably reasonable that it does. However, this method also knows about `@name`, which is unnecessary. If you remove it from here, then this method is less coupled to the existing process for creating a robot. I could register other names that were also in use, without instantiating Robot objects with them. This is a win. So upgrade `@name` to a parameter.

jmmastey:
Notes added. No need to resubmit if you see why the changes are helpful. Good work!

PR #1177, opened 2016-05-30
Dino catalog

aalcesto23:
Hey Joe,

So I think I got an MVP for this exercise, but it's in some desperate need for refactoring and some things I wanted to get some clarification on. I wanted to get your input before moving forward with any revisions. A couple of sticking points:

1. parsing both files at one time and compiling the data into one directory
2. not returning nil data
3. I formatting for the return values outside of using ARGV

Let me know what you think and thanks again!!

### DIFF
--- a/dino_catalog/dino_directory.rb
+++ b/dino_catalog/dino_directory.rb
@@ -1,0 +1,75 @@
+require 'csv'
+
+class Dino
+
+	attr_reader :name, :period, :continent, :diet, :weight_in_lbs, :walking, :description
+
+	def initialize(args)
+		@name = args[:name] || nil
+		@period = args[:period]
+		@continent = args[:continent]
+		@diet = args[:diet] || args[:carnivore]
+		@weight_in_lbs = args[:weight_in_lbs] || args[:weight]
+		@walking = args[:walking]
+		@description = args[:description]
+	end
+
+end
+
+module DinoLoader
+
+	def self.get_dinos_from_csv(filepath)
+		dinos = []
+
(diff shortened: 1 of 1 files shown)


jmmastey:
Cool, it's a good start. The biggest thing to focus on here is separation of concerns. When reading this, mostly in my head I'm wondering whether a given class _should_ own some piece of functionality. Examples:

1. The `Dino` class right now needs to know about the exact structure of the CSVs. So if we added another CSV, or a JSON input, or consumed a webservice, we'd have to change the `Dino` class. Does this make sense to do?
2. Should `Directory` own the responsibility of outputting dinosaurs? It's good that you're returning a string, as opposed to doing `puts`, but even still I'm not sure that a Directory should know this kind of stuff.

One possible refactor is to just add a `to_s` to `Dino`, but I'm not sure `Dino` should own it either! One approach that might be better is to provide a `to_h` method on `Dino`, then have the _consumer_ of those string representations (usually a runner, in this case just some ruby) deal with formatting.

3. Should `Directory` know what the weight threshold for big dinosaurs is? Weight doesn't have an obvious unit of measure, which doubles down on the trouble. It's probable that `Dino` should have a predicate method `#big?`.

On the loading question: It's totally okay that you load the two files sequentially, and I like that the parser is separated out. To load the catalog progressively, probably create a `#add(dinos)` to the `Directory`, then use it for each file. Since it's just an array of `Dino` objects, it shouldn't be too bad.

Also, fix yer tabs. ;)

Looking forward to the next iteration.

PR #1178, opened 2016-06-14
Use rubocop in refactor

iamarmanjon:
I didn't touch the #calculate_method since I can't seem to find out what the formula does. It's the only warning I got.

### DIFF
--- a/triangle_facts/triangle_facts.rb
+++ b/triangle_facts/triangle_facts.rb
@@ -1,58 +1,63 @@
-# Killer facts about triangles AWW YEAH
 class Triangle
-	attr_accessor :side1,:side2,:side3
+  attr_accessor :side1, :side2, :side3
 
-	def initialize(side1,side2,side3)
-		@side1,@side2,@side3 = side1,side2,side3
-	end
+  def initialize(side1, side2, side3)
+    @side1 = side1
+    @side2 = side2
+    @side3 = side3
+  end
 
-	def equilateral()
-		return side1 == side2 && side2 == side3
-	end
+  def equilateral?
+    side1 == side2 && side2 == side3
+  end
 
-	def isosceles()
-		return [side1,side2,side3].uniq.length == 2
-	end
+  def isosceles?
+    [side1, side2, side3].uniq.length == 2
+  end
 
-	def scalene()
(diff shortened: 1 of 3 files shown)


jmmastey:
Overall comment: You've merged some commits into master that don't belong there. You'll need to get a clean `master` branch otherwise all your pulls will have all the other exercises in them, which is bad.

Let me know if you need help on how to do that.

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,63 @@
...
+      puts 'This triangle is isosceles! Also, that word is hard to type.'
+    else
+      puts 'This triangle is scalene and mathematically boring.'
+    end
+  end
+
+  def print_triangle_angles
+    angles = calculate_angles(side1, side2, side3)
+    puts "The angles of this triangle are #{angles.join(', ')}"
+    puts 'This triangle is also a right triangle!' if angles.include? 90
```
jmmastey: Minor nit: Add parentheses to method calls that have parameters: `angles.include?(90)`

triangle_facts/triangle_facts.rb
```diff
@@ -1,58 +1,63 @@
...
+    puts ''
+  end
+
+  def calculate_angles(a, b, c)
+    angle_a = radians_to_degrees(Math.acos((b**2 + c**2 - a**2) / (2.0 * b * c)))
+    angle_b = radians_to_degrees(Math.acos((a**2 + c**2 - b**2) / (2.0 * a * c)))
+    angle_c = radians_to_degrees(Math.acos((a**2 + b**2 - c**2) / (2.0 * a * b)))
+    [angle_a, angle_b, angle_c]
+  end
```
jmmastey: A bunch of folks have tried to refactor this method into something that makes sense. It never seems to be better. There's an interesting realization in that this code -- while dense -- works and you can easily tease apart the similarities between each line if you need to. Good call not changing it :)

jmmastey:
Good work. Keep it up!

triangle_facts/triangle_facts.rb
iamarmanjon: overlook that! Thank you!

PR #1179, opened 2016-06-14
Remove arrowhead anti pattern & load database from json file

iamarmanjon:


arrowhead/arrowhead.rb
```diff
@@ -1,34 +1,20 @@
...
-      notched: "Besant",
-      stemmed: "Archaic Stemmed",
-      lanceolate: "Humboldt Constricted Base",
-      bifurcated: "Oxbow",
-    },
-  }
+  REGION_ERROR = "Unknown region, please provide a valid region.".freeze
+  SHAPE_ERROR =  "Unknown shape value. Are you sure you know what
+  you're talking about?".freeze
+  CLASSIFICATIONS = JSON.parse(File.read("arrowhead/database.json"))
```
jmmastey: This isn't really the point of the exercise, but doing `JSON.parse` / `File.read` and reading it into a constant at parsing time seems like a very strange practice.

arrowhead/arrowhead.rb
```diff
@@ -1,34 +1,20 @@
...
-    else
-      raise "Unknown region, please provide a valid region."
-    end
+    raise REGION_ERROR unless CLASSIFICATIONS.include?(region)
+
+    shapes = CLASSIFICATIONS[region]
+    raise SHAPE_ERROR unless shapes.include?(shape)
+
+    arrowhead = shapes[shape]
+    "You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
```
jmmastey: Good call removing the arrowhead of the method itself. It might make sense to move the logic for regions into its own method, but honestly this is a relatively short method either way. 

Here's another possible pattern:

```
shapes      = CLASSIFICATIONS.fetch(region) { raise REGION_ERROR }
arrowhead = shapes.fetch(shape) { raise SHAPE_ERROR }

"You have a(n) '#{arrowhead}' arrowhead. Probably priceless."
```

The nice thing about this pattern is that it puts the normal case (finding an arrowhead) first and up front. The exception case shifts over to the right, so it's more out of the way when trying to understand the method.

jmmastey:
Good work. Take a look at the feedback and see if you like it better. No need to resubmit. Either way, keep it up.

arrowhead/arrowhead.rb
iamarmanjon: Yes, sorry about that. I was trying to make it load from somewhere else, took the wrong approach it seems 👎

arrowhead/arrowhead.rb
iamarmanjon: Wow. That is a lot better, I was trying to find ways not to use `unless` since sometimes it's confusing. Didn't think of `fetch`. Thanks will look forward on other exercises!

PR #1180, opened 2016-06-15
Refactor Robot Class

iamarmanjon:
It works but I am unable to force it to fail. Rubocop suggested to use  instance variable instead of class variable for `@registry`. I also remove the `NameCollissionError`, I'm not sure if it's a good thing since the class is kinda empty. I also avoided the `#call`, I find it difficult to read instead of just making it as a method. So yeah, I think I need to improve the class I just can't find what it is :D

### DIFF
--- a/robot_name/robot_name.rb
+++ b/robot_name/robot_name.rb
@@ -1,32 +1,34 @@
-class NameCollisionError < RuntimeError; end
-
 class Robot
+  CHARS = ('A'..'Z').to_a
+  NUMS = (0..10).to_a
+
   attr_accessor :name
 
-  @@registry
+  def initialize(options = {})
+    @registry = []
+    add_name_to_registry(options)
+  end
 
-  def initialize(args = {})
-    @@registry ||= []
-    @name_generator = args[:name_generator]
+  def generate_name
+    name = ""
+    2.times { name += CHARS.sample }
+    3.times { name += NUMS.sample.to_s }
+    name
+  end
 
-    if @name_generator
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  def add_name_to_registry(options)
+    @name = options[:name] || generate_name
+    raise "Duplicate Error" if duplicate_name?
+    @registry << @name
+  end
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
-
-    raise NameCollisionError, 'There was a problem generating the robot name!' if !(name =~ /[[:alpha:]]{2}[[:digit:]]{3}/) || @@registry.include?(name)
-    @@registry << @name
+  def duplicate_name?
+    @registry.include?(@name)
   end
 end
 
-robot = Robot.new
+name = 'AA111'
+robot = Robot.new(name: name)
 puts "My pet robot's name is #{robot.name}, but we usually call him sparky."
-
-# Errors!
-# generator = -> { 'AA111' }
-# Robot.new(name_generator: generator)
-# Robot.new(name_generator: generator)
+# Robot.new(name: name)
+# Robot.new(name: name)


jmmastey:
Ehhhhh, removing the ability to call a lambda may be stretching your changes a bit far. Let me review the rest, but I would like to see you try to make the `generate_name` / `generator` stuff work while maintaining short methods and consistency.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,34 @@
...
+  CHARS = ('A'..'Z').to_a
+  NUMS = (0..10).to_a
 
-  @@registry
+  attr_accessor :name
 
-  def initialize(args = {})
-    @@registry ||= []
-    @name_generator = args[:name_generator]
+  def initialize(options = {})
```
jmmastey: What's the advantage of using `options` here if it only has one member? Right now I can't just look at the signature for initialize (`def initialize(options = {})`) and see how to actually use this class. Or, for instance, if I pass `{ robot_name: "abc123" }`, it will silently reject my name and give a default name instead.

If it's to be able to do `Robot.new(name: name)`, you should be able to use a named parameter instead.

robot_name/robot_name.rb
```diff
@@ -1,32 +1,34 @@
...
+    @registry = []
```
jmmastey: To your point, this actually breaks the registry, because each Robot gets its own registry as an instance variable. If you need to share a registry between robots, you really have two options:

1. Go back to class var.
2. Separate it into another object.

The latter is what more people tend to do. I actually think the class var is just fine, but apparently rubocop hates it. :)

robot_name/robot_name.rb
```diff
@@ -1,32 +1,34 @@
...
+    add_name_to_registry(options)
```
jmmastey: I like that this is a separate method, but the name may be wrong. This doesn't just add the name to the registry, it also potentially generates a new name. And it sets the instance variable value for the name.

Can you think of a name that might be more appropriate?

robot_name/robot_name.rb
```diff
@@ -1,32 +1,34 @@
...
-      @name = @name_generator.call
-    else
-      generate_char = -> { ('A'..'Z').to_a.sample }
-      generate_num = -> { rand(10) }
+  def generate_name
+    name = ""
+    2.times { name += CHARS.sample }
+    3.times { name += NUMS.sample.to_s }
+    name
+  end
```
jmmastey: I tend to have a thing against the pattern of initializing a variable, doing a few things, and then returning that variable. Ruby usually has good ways to do better than that. However, in this case, the only really plausible refactor is:

`"#{CHARS.sample}#{CHARS.sample}#{NUMS.sample}#{NUMS.sample}#{NUMS.sample}"`, which is a little silly.

If we didn't mind minorly changing the semantics of how names work, we could do:

`"#{CHARS.sample(2)}#{NUMS.sample(3)}`

But, can you see why that would change the semantics of the names?

robot_name/robot_name.rb
```diff
@@ -1,32 +1,34 @@
...
 
-      @name = "#{generate_char.call}#{generate_char.call}#{generate_num.call}#{generate_num.call}#{generate_num.call}"
-    end
+  def add_name_to_registry(options)
+    @name = options[:name] || generate_name
+    raise "Duplicate Error" if duplicate_name?
+    @registry << @name
+  end
```
jmmastey: So we've also lost the concept of validating that a name has the right format, which we should have in some form. Even if we do lose the custom Error class, we should attempt to retain current behavior.

jmmastey:
I've added some notes on ideas. Overall, you definitely made the methods and lines much shorter.

Try to restore the original semantics of the program while making it shorter. It'd be better to have the old, verbose version of the code working correctly, than to lose functionality in the name of pretty code.

Give it another shot and resubmit. Thanks!

iamarmanjon:
Hi @jmmastey I'll re submit later, you have a lot of good suggestions there. Although this activity is quite big  if you really want to factor reading the two books.

jmmastey:
Hah. Don't feel totally obligated to read the books right now. Overall, I will say that even after a dozen years of programming both of those books showed me new ideas that I thought were helpful.

PR #1181, opened 2016-06-15
Refactor RobotName to short methods

iamarmanjon:
I retain the `NameCollisionError` class and `@@registry`. I want the
`initialize` method to only accept a var instead of hash but ruby
doesn't allow an optional parameter. I THINK this is better than my
previous attempt, all functions are still the same. I'm not sure
about my `not_in_the_registry?` method since I use negation there.

Thank you for your awesome feedback!

### DIFF
--- a/robot_name/robot.rb
+++ b/robot_name/robot.rb
@@ -1,0 +1,51 @@
+require_relative 'robot_name'
+
+class NameCollisionError < RuntimeError; end
+
+class Robot
+  CHARS = ('A'..'Z').to_a
+  NUMS = (1..10).to_a
+  ERROR_MESSAGE = 'There was a problem generating the robot name!'.freeze
+
+  attr_accessor :name
+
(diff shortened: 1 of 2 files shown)


robot_name/robot.rb
```diff
@@ -0,0 +1,51 @@
...
+
+class Robot
+  CHARS = ('A'..'Z').to_a
+  NUMS = (1..10).to_a
+  ERROR_MESSAGE = 'There was a problem generating the robot name!'.freeze
+
+  attr_accessor :name
+
+  def initialize(args = {})
+    @name_generator = args[:name_generator]
```
jmmastey: This is the remaining bit that I think you could tighten up. Take a look at the `Kernel#method` method (unfortunate choice of names) for a way to treat `generate_name` identically to the `name_generator`.

jmmastey:
Much better! Good work.

